PR #1, opened 2025-03-24
Feature/onboarding

opcode81:


myproject.demo.yml
```diff
@@ -0,0 +1,6 @@
+# absolute path to the project
```
MischaPanch, line 1: Is this a config file? Should we rather use .env and .env.example for that?

MischaPanch: Ah, nvm, you use this as input to the script. Let's leave it as is for now

scripts/mcp_server.py
```diff
@@ -0,0 +1,200 @@
...
+        pass
+
+    def _add_user_message(self, msg: str) -> None:
+        self.messages.append(UserMessage(content=msg))
+
+
+@mcp.tool()
+def read_file(ctx: Context, relative_path: str) -> str:
+    """
+    :param ctx: the context object, which will be created and provided automatically
```
MischaPanch, line 127: Let's switch to numpy style for this, ok? Easier to provide examples, which can be very useful for the LM to understand the tools.

Afaik, rst doesn't properly support an examples section

MischaPanch: You don't have to do this now, just for the future. After merging I (meaning Claude) will adjust the docstrings

PR #2, opened 2025-04-01
Hack: simpler types in tools to make agno work with google models

MischaPanch:


src/serena/agno.py
```diff
@@ -17,16 +16,16 @@ def __init__(self, serena_agent: SerenaAgent):
...
     @staticmethod
     def _create_agno_function(tool: Tool) -> Function:
-        def entrypoint(**kwargs: Any) -> str:
-            if "kwargs" in kwargs:
-                # Agno sometimes passes a kwargs argument explicitly, so we merge it
-                kwargs.update(kwargs["kwargs"])
-                del kwargs["kwargs"]
-            log.info(f"Calling tool {tool}")
-            return tool.apply_ex(log_call=True, catch_exceptions=True, **kwargs)
+        # def entrypoint(**kwargs: Any) -> str:
```
MischaPanch, line 19: @opcode81 google actually expects kwargs and breaks if we do this. Let's have a look at this together

opcode81: I don't think that's possible, because google never sees or knows about this function. This is just the function that ultimately gets called - in contrast to the function `apply` from which we get the signature google is told about.

opcode81:
We can't merge this. The Gemini adapter in Agno needs to be fixed instead.
Most likely, we can achieve the same effect by overwriting `_format_function_definitions` or `_convert_schema` in Agno's gemini.py. The Agno devs probably just just messed up.

MischaPanch:
Closing, fixed through monkeypatch

PR #13, opened 2025-04-03
Add sqlalchemy to deps (for agno agent)

mdbenito:
That. Was missing from the deps


MischaPanch (approved):

MischaPanch:
Thanks! I wonder why it worked without this for me..

opcode81:
Most likely, a dev dependency adds it as a transitive dependency.

PR #23, opened 2025-04-04
Add Swift language server support..

BSN4:
image


MischaPanch:
Nice, thanks! Could you please try the find_symbol, find_referencing_symbols and replace_symbol_body tools to make sure symbolic read and write operations actually work?

If you have some time, I'd also appreciate if you checked in some dummy swift files to tests/resources and wrote some simple tests for symbolic operations, like the ones we already have for python. An LLM (and probably serena itself) should be able to assist with this fairly well

MischaPanch:
Claude had a brainfart here and thought that one of the very few non-symbolic tools was actually a symbolic one...

BSN4:
I did basic comparison to add the language server , I think I need to see if claude just lied to me

BSN4:
find_symbol - works 
replace_symbol_body - works
find_referencing_symbols - not working

MischaPanch:
Can you describe in more detail what doesn't work with `find_referencing_symbols`?

This would be far easier to debug if you wrote simple unit tests, like I suggested above

MischaPanch:
A possible reason for the failure is wrong initialization params, happened to me for the pyright server

BSN4:
ok here some of my convos with the mcp 
image
image

find_referencing_symbols always returns empty list

BSN4:
Im taking break now and I will see later if I can do a unit test

MischaPanch:
I strongly suspect the problem to be in initialization params

BSN4:
@MischaPanch I did some testing using claude , made it pass but after I saw the code I've realised claude made it worse basically to pass in find_referencing_symbols test it did some methods overrides and used a lot of regex , in my experience parsing source code in regex won't be good at the end .. so I removed the overrides and we back at not passing find_referencing_symbol test , but thats fine for now I will test it more on my workflow ... until then I'll see if I can improve it

MischaPanch:
Ah, I thought you had given up on this and wanted to continue where you stopped. Glad to hear you're still on this! 

Regarding the test: it's enough to just checkin a single swift module with some things referencing each other and check whether the referencing symbols are found. There should be no need for regexes or anything else like this.

Two things may cause the LS finding symbols and not finding references: 

1. Initialization params (the json) - the LS needs to be correctly initialized for finding references 
2. Problems in column counting. There may be inconsistencies in what the LS returns as location of a symbol and what needs to be passed for finding references. For some LS one needs to pass the column from range, for others from selectionRange. 

You can easily find out if it's 2 by debugging the find references method and trying out different values for the character/column param

BSN4:
@MischaPanch to fix it probably you need to understand how sourcekit-lsp works ... Im getting there , but still even with half of functionality serena works just fine

MischaPanch:
Neither Claude not o3-mini-high find errors in the params or the code.

The error might be in our symbolic code and not in the actual find references of the LS. I have implemented a new tool in #30 that retrieves non-symbolic references. Could you pls try that out (by merging it into your branch)? If it works, the error is in our implementation and you might find it hard to fix it.

I am reopening this PR since work on it is being continued.

@opcode81 FYI - this is the reason for the new tool in my PR, but I generally think it might be a good addition to Serena

BSN4:
awesome .. I will try it soon

BSN4:
I don't think those changes did anything ... I believe its just sourcekit-lsp indexing issue

MischaPanch:
Closing in favor of #397

PR #28, opened 2025-04-05
Don't rely on server_lifespan to start the server

MischaPanch:
@opcode81 since there is no Claude app for ubuntu, I tested through connecting serena in Cline (and paying some api costs). It worked for me. Pls have a quick look and a test with Claude Desktop.

Hopefully this will solve many if not all stability issues.

### DIFF
--- a/src/serena/agent.py
+++ b/src/serena/agent.py
@@ -10,8 +10,7 @@ import sys
 import traceback
 from abc import ABC
 from collections import defaultdict
-from collections.abc import Callable, Generator, Iterable, Iterator
-from contextlib import contextmanager
+from collections.abc import Callable, Generator, Iterable
 from logging import Logger
 from pathlib import Path
 from typing import Any, TypeVar, cast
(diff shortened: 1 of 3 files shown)


src/serena/mcp.py
```diff
@@ -68,30 +64,46 @@ def execute_fn(ctx: Context, **kwargs) -> str:  # type: ignore
...
-    async def server_lifespan(mcp_server: FastMCP) -> AsyncIterator[SerenaMCPRequestContext]:
+    async def server_lifespan(mcp_server: FastMCP) -> AsyncIterator[None]:
         """Manage server startup and shutdown lifecycle."""
-        with agent.language_server_lifecycle_context():
-            yield SerenaMCPRequestContext(agent=agent)
+        try:
+            yield
+        finally:
+            for agent in _SERENA_AGENT_REGISTRY.values():
+                agent.language_server.stop()
```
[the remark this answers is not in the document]
MischaPanch: nah, man

src/serena/agent.py
```diff
@@ -123,6 +122,14 @@ def __init__(self, project_file_path: str, start_language_server: bool = False):
         if self._start_language_server:
             log.info("Starting the language server ...")
             self.language_server.start()
+            if not self.language_server.is_running():
+                raise RuntimeError(f"Failed to start the language server for {self.language} and project {self.project_root}")
+
+    def reset_language_server(self) -> None:
+        self.language_server.stop()
```
opcode81, line 129: I think this will crash if the agent is ever instantiated with `start_language_server=False`, as stopping will raise an exception if it is not running.

https://github.com/oraios/serena/blob/8d5664bf12ffd1c7e11250335c4ec34592bf78f3/src/multilspy/language_server.py#L1841-L1853

Also, it is actually a contradiction to the code that triggers this call in the first place:

https://github.com/oraios/serena/blob/8d5664bf12ffd1c7e11250335c4ec34592bf78f3/src/serena/agent.py#L289-L290

Therefore, I don't think this can work as intended.

MischaPanch: I adjusted stop to no longer raise an exception but to just log on info level. Should be enough, right?

opcode81:
Superseded by #34

PR #30, opened 2025-04-05
New tool: GetReferencingCodeExtractsTool

MischaPanch:


opcode81:
I think I would prefer the term "excerpt" over "extract".
Also, I think we should prefer the verb "find" over "get".

src/serena/agent.py
```diff
@@ -536,6 +536,49 @@ def apply(
...
+        context_lines_before: int = 0,
+        context_lines_after: int = 0,
+        max_answer_chars: int = _DEFAULT_MAX_ANSWER_LENGTH,
+    ) -> str:
+        """
+        Returns short code extracts where the symbol at the given location is referenced.
+
+        Contrary to the `find_referencing_symbols` tool, this tool returns references that are not symbols but instead
+        code extracts that may or may not be contained in a symbol (for example, file-level calls).
+        It may make sense to use this tool if you want to get a quick and dirty overview of the code that references
```
opcode81: I don't really think there is anything "dirty" about this. Depending on the use case, you might not care at all about the full context (symbol retrieval could only give you the symbol's full body) but only about the very concrete usage (e.g. is the function being called with some parameter or not). I think it's actually a very reasonable way of looking at usages (which I constantly make use of in IDEs).
So I'd lose the "and dirty" part.

opcode81:
Apart from the naming preferences, lgtm!

src/serena/agent.py
```diff
@@ -536,6 +536,49 @@ def apply(
         return self._limit_length(result, max_answer_chars)
 
 
+class GetReferencingCodeExtractsTool(Tool):
```
opcode81: I think I would prefer the term "excerpt" or even "snippet" over "extract".
Also, I think we should prefer the verb "find" over "get".

MischaPanch: Done

src/serena/agent.py
MischaPanch: Done

PR #33, opened 2025-04-06
Add Serena version to server startup logs

ShadovvBeast:
This change adds the Serena version number to the server startup log message, making it easier to identify which version is running when reviewing logs.


MischaPanch:
Sure, that's useful, thanks for the PR. We'll have to remember to bump the version regularly

PR #34, opened 2025-04-06
Support project changes from within Serena (via tools)

opcode81:
**Main change: Support project changes from within Serena (via tools)**

* Establish new Serena-wide configuration file serena_config.yml with
   * Setting whether to enable project activation
   * List of project configuration files
   * GUI logger configuration (no longer part of project configuration)

* Add new tools for project activation (activate_project, get_active_project)

* Project configuration file is no longer required for startup (now optional):
  If project activation is enabled, no project must be provided (but at
  least one project must be configured)

**Peripheral changes: Improved logging**


src/serena/agent.py
```diff
@@ -37,6 +37,67 @@
...
+    SERENA_MANAGED_DIR = ".serena"
+
+    def __init__(self, config_dict: dict[str, Any], project_name: str):
+        self.project_name: str = project_name
+        self.language: Language = Language(config_dict["language"])
+        self.project_root: str = str(Path(config_dict["project_root"]).resolve())
+        self.ignored_dirs: list[str] = config_dict.get("ignored_dirs", [])
+        self.excluded_tools: set[str] = set(config_dict.get("excluded_tools", []))
+
+    def _tostring_excludes(self) -> list[str]:
```
MischaPanch: why that? It's rather relevant info, I'd say

src/serena/mcp.py
```diff
@@ -68,29 +65,55 @@ def execute_fn(ctx: Context, **kwargs) -> str:  # type: ignore
...
-    agent = SerenaAgent(project_file_path)
+    if (len(argv) == 1 and argv[0] == "--help") or len(argv) > 1:
+        print("\nUsage: mcp_server [.yml project file]", file=sys.stderr)
+        sys.exit(0)
+
+    mcp: FastMCP | None = None
+
+    def update_tools() -> None:
+        """Update the tools in the MCP server."""
+        # Tools may change as a result of project activation.
```
MischaPanch, line 83: We should document this in the readme (or did I miss it?)

MischaPanch (approved):
Nicely done!

Only minor comments (though for some parts of the code I only did a superficial review).

Pls adjust the changelog and let's bump the version (it is now being shown in the log). Maybe we should have date-based versioning instead of semver?

src/serena/agent.py
opcode81: Yeah, you're probably right. I'll remove it.

src/serena/mcp.py
opcode81: Document what exactly?
I think what you're looking for is documented in `serena_config.template.yml`, because that's where the setting is controlled that affects this behaviour.

https://github.com/oraios/serena/blob/a4a8c19f6cfd6bd2ba0a4ce0879f0e77bbd46211/serena_config.template.yml#L1-L7

I don't want to duplicate this in the README.

MischaPanch: Ah, ok, there is enough

serena_config.template.yml
```diff
@@ -0,0 +1,28 @@
...
+# by its name (see "projects" below).
+# Note that if this is enabled, then per-project tool activation will be handled within Serena,
+# i.e. Serena will provide all tools but will disallow the execution of tools that are not enabled for
+# the current project. This is because clients typically do not respond to a changed set of tools.
+enable_project_activation: True
+
+# Add your list of .yml project files here (which you can switch between using "activate_project").
+# Serena will know the projects by their base filename, i.e. "MyProject.yml" will be project "MyProject" to Serena.
+# Paths should either be relative to the directory this configuration file is in or absolute.
+projects:
```
MischaPanch, line 12: so here now just a path to a repo is also allowed, right?

opcode81: Yes

opcode81: Ah, now I get why you asked this question. I added the updated description to my `serena_config.yml` instead of the template.

PR #35, opened 2025-04-06
Native support for gitignore in LanguageServer and ProjectConfig

MischaPanch:
This fixes several issues, including the slowness of the regex and overview tools, but also potentially the instabilities (they might have been due to OOM)

@shyraptor, @a-bonus - this will fix (or at least improve) several of the problems you reported


src/multilspy/language_server.py
```diff
@@ -196,6 +212,27 @@ def logging_fn(source, target, msg):
...
+    # Helper function to check if a relative path should be ignored
+    def should_ignore_path(self, relative_path: str) -> bool:
+        fn_matcher = self.language.get_source_fn_matcher()
+        if os.path.isfile(relative_path) and not fn_matcher.is_relevant_filename(relative_path):
+            return True
+        
+        # Normalize path separators for consistent comparison
+        rel_path = Path(relative_path)
+        for ignored_path in self.ignored_paths:
+            if rel_path.is_relative_to(ignored_path):
```
opcode81: This is not sufficient, I think.

Note the description in the configuration:
```
# list of directories to ignore: either names (e.g. "temp") or relative paths (e.g. "build/foo")
ignored_dirs: [".git", "temp"]
```

So if you had "baz" in there, it should not only ignore "./baz" but also "./foo/baz" or "./foo/bar/baz".
In other words, "baz" stands for the glob pattern "**/baz".

To clearly differentiate between suffix matching and relative path matching, we should probably do this gitignore-style:
 * `/baz` ignores only `./baz`
 * `baz` ignores `**/baz`
 * `/foo/bar` ignores only `./foo/bar`
 * `foo/bar` ignores `**/foo/bar`
 
 I think this would be the best - albeit a bit more complicated.

opcode81: Maybe we could use this: https://gitmatch.readthedocs.io/en/stable/

MischaPanch: We could also permit to configure globs here, this way it would also be easier for users to copy config from gitignore (later we could give them the option to just use gitignore directly). Wdyt?

opcode81: Yes.

MischaPanch: Done

src/multilspy/language_server.py
```diff
@@ -196,6 +224,64 @@ def logging_fn(source, target, msg):
...
+        # Set up the pathspec matcher for the ignored paths
+        # for all absolute paths in ignored_paths, convert them to relative paths
+        processed_patterns = []
+        for pattern in set(config.ignored_paths):
+            # Normalize separators (pathspec expects forward slashes)
+            pattern = pattern.replace(os.path.sep, '/')
+            processed_patterns.append(pattern)
+        # Combine explicitly passed patterns with
+        if config.gitignore_file_content is not None:
+            processed_patterns.extend(config.gitignore_file_content.splitlines())
```
opcode81: You don't seem to be filtering out comments in the gitignore file.

MischaPanch: No, the matcher can deal with this (this line is taken from the official docs there, but I also checked the source code, it filters out the comments)

src/multilspy/language_server.py
```diff
@@ -82,8 +85,13 @@ class LanguageServer:
     The LanguageServer class provides a language agnostic interface to the Language Server Protocol.
     It is used to communicate with Language Servers of different programming languages.
     """
+
+    # To be overridden and extended by subclasses
+    def should_always_ignore(self, dirname: str) -> bool:
```
opcode81, line 90: Should add docstring docstring

MischaPanch:
@opcode81 I disabled the gui by default in the last commit since it doesn't work with the (unofficial) Linux app.

If you disagree, I can drop this, but I prefer to keep it safe and let the user enable it consciously

src/multilspy/language_server.py
```diff
@@ -196,6 +224,64 @@ def logging_fn(source, target, msg):
...
         self.language = Language(language_id)
+        
+        # Set up the pathspec matcher for the ignored paths
+        # for all absolute paths in ignored_paths, convert them to relative paths
+        processed_patterns = []
+        for pattern in set(config.ignored_paths):
+            # Normalize separators (pathspec expects forward slashes)
+            pattern = pattern.replace(os.path.sep, '/')
+            processed_patterns.append(pattern)
+        # Combine explicitly passed patterns with
```
opcode81: incomplete comment

src/multilspy/language_server.py
MischaPanch: I will still filter out the comments myself here. The matcher will filter them on each match, that's inefficient (although it likely doesn't matter)

src/multilspy/language_server.py
```diff
@@ -749,12 +841,13 @@ def visit_tree_nodes_and_build_tree_repr(node: GenericDocumentSymbol) -> List[mu
...
         Will go through all files in the project and build a tree of symbols. Note: this may be slow the first time it is called.
-        
+
         For each file, a symbol of kind Module (3) will be created. For directories, a symbol of kind Package (4) will be created.
-        All symbols will have a children attribute, thereby representing the tree structure of all symbols in the project 
+        All symbols will have a children attribute, thereby representing the tree structure of all symbols in the project
         that are within the repository.
-        Will ignore all directories that start with a dot (.) and __pycache__ directories.
-        
+        Will ignore directories starting with '.', '__pycache__', language-specific defaults (self.default_ignored_dirs),
```
opcode81: This is no longer accurate. `__pycache__` is not globally ignored afaict.

MischaPanch: Good catch!

opcode81: Actually, I think the defaults could be problematic in general, and it might be a good idea to make them optional (in a future update).

src/multilspy/language_server.py
MischaPanch: Done

src/multilspy/language_server.py
MischaPanch: Done

src/serena/agent.py
```diff
@@ -47,8 +47,12 @@ def __init__(self, config_dict: dict[str, Any], project_name: str, project_root:
         if project_root is None:
             project_root = Path(config_dict["project_root"])
         self.project_root: str = str(project_root.resolve())
-        self.ignored_dirs: list[str] = config_dict.get("ignored_dirs", [])
+        self.ignored_paths: list[str] = config_dict.get("ignored_paths", [])
```
opcode81, line 54: This is problematic. Users who don't update their configs (still have `ignored_dirs`) will have this fail silently, so we need an exception if it's not there.

src/serena/agent.py
```diff
@@ -47,8 +47,12 @@ def __init__(self, config_dict: dict[str, Any], project_name: str, project_root:
...
         self.excluded_tools: set[str] = set(config_dict.get("excluded_tools", []))
+        self.ignore_all_files_in_gitignore = config_dict.get("ignore_all_files_in_gitignore", True)
```
opcode81: Same here. We should not assume that a default is OK.

opcode81: I see that you (perhaps unexpectedly) handled backward compatbility below.

src/serena/agent.py
opcode81: I see that you (perhaps unexpectedly) handled backward compatbility below.

src/serena/agent.py
opcode81: If a default is OK, it is False instead of True, because that's what users had previously.

src/serena/agent.py
```diff
@@ -79,23 +83,31 @@ def __init__(self) -> None:
...
+            try:
+                log.info(f"Loading Serena configuration from {config_file}")
+                config_yaml = yaml.safe_load(f)
+            except Exception as e:
+                raise ValueError(f"Error loading Serena configuration from {config_file}: {e}") from e
 
         # read projects
         self.projects: dict[str, ProjectConfig] = {}
-        for project_config_path in config_yaml["projects"]:
+        for project_config_path in config_yaml.get("projects", []):
```
opcode81: Why?

src/serena/agent.py
```diff
@@ -79,23 +83,31 @@ def __init__(self) -> None:
...
                 project_config_path = project_config_path / ProjectConfig.SERENA_MANAGED_DIR / ProjectConfig.SERENA_DEFAULT_PROJECT_FILE
             if not project_config_path.is_file():
                 raise FileNotFoundError(f"Project file not found: {project_config_path}")
+            log.info(f"Loading project configuration from {project_config_path}")
             project_config = ProjectConfig.from_yml(project_config_path)
-            self.projects[project_config.project_name] = project_config
+            try:
+                self.projects[project_config.project_name] = project_config
+            except Exception as e:
+                raise ValueError(f"Error loading project configuration from {project_config_path}: {e}") from e
```
opcode81: This makes no sense to me. Did you mean to include the `from_yml` call?

src/serena/agent.py
```diff
@@ -79,23 +83,31 @@ def __init__(self) -> None:
...
         self.project_names = list(self.projects.keys())
 
-        self.gui_log_window_enabled = config_yaml.get("gui_log_window", True)
+        self.gui_log_window_enabled = config_yaml.get("gui_log_window", False)
```
opcode81: Why did you change this?

src/serena/agent.py
```diff
@@ -180,7 +192,11 @@ def __init__(self, project_file_path: str | None = None, project_activation_call
         if project_file_path is not None:
             if not os.path.exists(project_file_path):
                 raise FileNotFoundError(f"Project file not found: {project_file_path}")
-            project_config = ProjectConfig.from_yml(Path(project_file_path))
+            log.info(f"Loading project configuration from {project_file_path}")
+            try:
+                project_config = ProjectConfig.from_yml(Path(project_file_path))
+            except Exception as e:
+                raise ValueError(f"Error loading project configuration from {project_file_path}: {e}") from e
```
opcode81: Instead of doing this twice, you could wrap the constructor in a try-except block.

src/serena/agent.py
```diff
@@ -544,7 +565,7 @@ def apply(self, relative_path: str, recursive: bool, max_answer_chars: int = _DE
             os.path.join(self.project_root, relative_path),
             relative_to=self.project_root,
             recursive=recursive,
-            ignored_dirs=self.project_config.ignored_dirs,
+            ignored_dirs=self.project_config.ignored_paths,
```
opcode81, line 582: Shouldn't this be updated to use the full set of ignored paths, i.e. using the pathspec matcher (which also contains the gitignore if enabled)?

src/serena/agent.py
opcode81: I see your comment above about the Linux Claude app not supporting this.
Any idea why? What is the error you get?

MischaPanch: Something something DISPLAY variable not set

src/serena/agent.py
MischaPanch: Yeah, out of place, sry

src/serena/agent.py
MischaPanch: It's fine to have an empty list of projects, right? I find crashes with key errors to be rather bad, so if the key is missing, I'd rather raise a proper error (if empty list is not allowed)

src/serena/agent.py
MischaPanch: I would rather roll it out automatically to those who already setup their configs. It's not like they consciously chose to not include gitignore, and I suspect it's a behavior that most if not all users would want

src/serena/agent.py
MischaPanch: Yes, let me do that in a followup PR, ok?

src/serena/agent.py
opcode81: Also, if there are known limitations, we should mention them in `myproject.template.yml`.

src/serena/agent.py
MischaPanch: Done

src/serena/agent.py
opcode81: The option you chose can cause unexpected behaviour.
The two valid options are:
* False as a default (unchanged behaviour)
* forcing a decision via an exception

I don't truly care, but it's usually the best approach to never make _any_ assumptions about what the user really wants.

src/serena/agent.py
opcode81: It can be empty if the user specifies a project at startup. The correct handling is ensured by the SerenaAgent constructor.

But if it is empty, it should be empty explicitly, i.e. the value should always be there. It not being there is an indication of an error, which should be raised.

src/serena/agent.py
opcode81: > DISPLAY variable not set

Sounds like it should be fixable. We should investigate this.
Can you add an issue with the exact error?

src/serena/agent.py
opcode81: There are advantages to forcing a decision, as it also creates awareness of the new feature.
If we never force users to make decisions on new settings, some of them will likely continue with the original set of settings forever (and even copy their configs instead of the template).

src/serena/agent.py
MischaPanch: Raising dedicated exception now

src/serena/agent.py
MischaPanch: I agree, let's force. I added a callout section to the readme warning about this, and to the changelog about the breaking changes in the config

src/serena/agent.py
MischaPanch: Huh, I can't reproduce now. MCP started up and all tools work.

But I don't see the gui, so that's also weird

MischaPanch: Changed the default back and added explanation on limitations

opcode81 (approved):

Issue #16, opened 2025-04-03
Port recent updates from multilspy - including Kotlin and Dart support


MischaPanch:

opcode81:
We could consider restructuring our code to use multilspy as a dependency - or at least enable a synchronisation (e.g. based on sensAI's repo_dir_sync).

Shabinder:
Thanks for all the work being done here, pretty excited for Kotlin support!

PR #41, opened 2025-04-08
Multilspy update

opcode81:
Updates multilspy to the latest version 
#16


src/multilspy/language_server.py
```diff
@@ -538,9 +560,7 @@ async def request_definition(
             new_item: multilspy_types.Location = {}
             new_item.update(response)
             new_item["absolutePath"] = PathUtils.uri_to_path(new_item["uri"])
-            new_item["relativePath"] = str(
-                PurePath(os.path.relpath(new_item["absolutePath"], self.repository_root_path))
-            )
+            new_item["relativePath"] = PathUtils.get_relative_path(new_item["absolutePath"], self.repository_root_path)
```
MischaPanch, line 552: So does the pathutil do the right job even if the definition is on a different drive?

src/multilspy/language_server.py
```diff
@@ -1539,43 +1559,72 @@ def load_cache(self):
                 # cache often becomes corrupt, so just skip loading it
                 self.logger.log(
                         f"Failed to load document symbols cache from {self._cache_path}: {e}. Possible cause: the cache file is corrupted. " 
-                        "Check for any errors related to saving the cache in the logs.", 
+                        "Check for any errors related to saving the cache in the logs.",
                         logging.ERROR
                     )
 
 
+    async def request_workspace_symbol(self, query: str) -> Union[List[multilspy_types.UnifiedSymbolInformation], None]:
```
MischaPanch, line 1557: Great! We should adjust our find_symbol to use this instead of our workaround

src/multilspy/language_server.py
```diff
@@ -1539,43 +1559,72 @@ def load_cache(self):
...
         self.loop_thread = None
-        
+        self.timeout = timeout
+
         self._server_context = None
 
     @classmethod
     def create(
         cls, config: MultilspyConfig, logger: MultilspyLogger, repository_root_path: str, add_gitignore_content_to_config=True,
+        timeout: Optional[int] = None
```
MischaPanch, line 1606: Docstring missing

MischaPanch (approved):
Looks good, minor comments. 

We can incorporate the new symbol finding endpoint in a later PR if you want to

src/multilspy/language_server.py
opcode81: Depends on what "the right job" is.
It will return None:

https://github.com/oraios/serena/blob/2fa6a95aac1f051ae6c187e44c3b3b7aa159b884/src/multilspy/multilspy_utils.py#L98-L105

src/multilspy/language_server.py
opcode81: The multilspy devs did not add it. Why do we care?

MischaPanch: Just noticed and mentioned it, it's ok to not add it ofc

src/multilspy/language_server.py
MischaPanch: previously we were adding absPath.

It explains why they changed the type for relativePath to optional in the location type-dict. I guess None is good, not sure whether this means we need none-checks in our code now

README.md
```diff
@@ -56,6 +56,8 @@ With Serena, we provide
      * Go (untested)
      * C# (untested)
      * Rust (untested)
+     * Kotlin (untested)
+     * Dart (untested)
```
MischaPanch, line 60:
```suggestion
     * Dart (untested)
     * C++ (untested)
```

opcode81: Pushed two commits, one of which adds "C/C++" to the list (as C is also supported).

src/multilspy/language_server.py
opcode81: I actually hate it when programmers don't add a unit to this type of parameter. I suppose it's in seconds, but you never know without reading the code.

src/multilspy/language_server.py
opcode81: Let's add an issue on that.

src/multilspy/language_server.py
opcode81: This probably needs consolidation. Semantically, it's wrong to use an absolute path. Then again, the original authors basically assumed that one can always use a relative path (which is wrong for Windows though), so in that sense, using an absolute path is the right thing to do.

So shall we change their function to return the absolute path instead of None and use it everywhere, getting rid of our exception handlers?

src/multilspy/language_server.py
opcode81: Added #44

src/multilspy/language_server.py
MischaPanch: Thing is, receiving an abs path when one expects a relative path is also wrong. In many places we join relative paths with the project root. I feel like it's better to return None if relativePath cannot be achieved, bite into the sour apple (is that an expression in English as well?) and do None handling everywhere

opcode81: Agreed.

opcode81: I changed all calls to use the PathUtils function

opcode81: A potential downside of using None is, however, that Serena will be completely unable to read any files for which no relative paths can be constructed, as we never return the absolute paths.

We might consider changing Serena's interface to return an unspecified "path", which could either be relative or absolute. But this is something for another PR.

src/multilspy/language_server.py
opcode81: Added docstrings.

src/multilspy/language_server.py
scottyybt: Today was setting up mcp server on my new linux machine and started getting this error when tried to run the server with vscode:

> Should` ignore all files in gitignore not not .gitignore found at /home/fedora/scottie/serena/myproject.yml/.gitignore. Skipping. Fatal exception: [Errno 20] Not a directory: '/home/fedora/scottie/serena/myproject.yml'

_The constructor of SerenaAgent takes a project_file_path and loads the project configuration from it if it exists. It uses ProjectConfig.from_yml to load the config from the given path. Then it calls activate_project with the loaded project_config.
The activate_project method calls reset_language_server, which creates the language server with the project_root path. The project_root is set in the ProjectConfig constructor, but it may be None or a file path.
  You have to ensure that project_root is always a directory path. Added this at line 79 and it worked. Cant even believe.. hehe 
      79        if project_root.is_file():
      80          project_root = project_root.parent

scottyybt: Sorry for my bad english btw. Hope someone will be able to understand what I tried to say :D

MischaPanch: Pls open a new issue for that and share your MCP server config and your Serena config

PR #43, opened 2025-04-09
eclipse_jdtl launcher on osx intel

ahmetsevki:
runtime dependency brought in from [multilspy repository](https://github.com/microsoft/multilspy/blob/ab588214afb733c8d0084451d34ee6c7d7b6e0f0/src/multilspy/language_servers/eclipse_jdtls/runtime_dependencies.json#L26C9-L35C11) for osx with x64 architecture.


opcode81:
Thanks for the contribution, but this change is already included in #41.
Note that we have synchronisation mechanism to update multilspy from the original repository, so we do not need to manually take care of such things.

Issue #6, opened 2025-04-02
SearchInAllCodeTool is too limited


opcode81:
The tool is limited to the file extensions listed here:

https://github.com/oraios/serena/blob/8a935b7fa1cd094e1f468d13984e795e003136ac/src/multilspy/multilspy_config.py#L40-L59

This is potentially too limiting, as we want to search in all text files, whether they are strictly understood source code or not.
The tool gets its filenames from the full symbol tree tool, which is also suboptimal.

We could think about either
* adding a second tool that searches more globally
* parametrising this one

Open question: How to determine the set of files to include in/exclude from the search?

PR #45, opened 2025-04-09
Feature/tools extension

MischaPanch:
Closes: #6 

Also addresses using Serena while editing in a different tool


src/serena/agent.py
```diff
@@ -1143,32 +1158,56 @@ def apply(
...
+                context_lines_before=context_lines_before,
+                context_lines_after=context_lines_after,
+                paths_include_glob=paths_include_glob,
+                paths_exclude_glob=paths_exclude_glob,
+            )
+        else:
+            # we walk through all files in the project starting from the root
+            files_to_search = []
+            ignore_spec = self.language_server.get_ignore_spec()
+            for root, dirs, files in os.walk(self.project_root):
```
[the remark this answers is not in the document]
MischaPanch: Man you're stupid... 
os is imported, embarassing!

src/serena/agent.py
```diff
@@ -502,6 +503,20 @@ class ToolMarkerDoesNotRequireActiveProject:
...
+
+    def apply(self) -> str:
+        """Use this tool only on explicit user request or after confirmation.
+        It may be necessary to restart the language server if the user performs edits
+        not through Serena, so the language server state becomes outdated and further editing attempts lead to errors.
+
+        If such editing errors happen, you should suggest using this tool.
+        """
+        self.agent.reset_language_server()
+        return "OK"
```
opcode81: Use constant `SUCCESS_RESULT`

src/serena/text_utils.py
```diff
@@ -205,3 +210,48 @@ def search_text(
...
 
     return matches
+
+
+def default_content_reader(file_path: str) -> str:
+    with open(file_path) as f:
+        return f.read()
+
+
+def search_files(
```
opcode81, line 223: Needs docstring

src/serena/text_utils.py
```diff
@@ -205,3 +210,48 @@ def search_text(
...
+    file_paths: list[str],
+    pattern: re.Pattern | str,
+    content_reader: Callable[[str], str] = default_content_reader,
```
opcode81: Name is imprecise; `file_reader` would be more to the point

src/serena/text_utils.py
```diff
@@ -205,3 +210,48 @@ def search_text(
                 matches.append(MatchedConsecutiveLines(lines=context_lines, source_file_path=source_file_path))
 
     return matches
+
+
+def default_content_reader(file_path: str) -> str:
+    with open(file_path) as f:
```
opcode81: While code will mostly be ASCII, it might make sense to explicitly specify utf-8 encoding instead of using Python's rather impractical default (basically latin-1 on Windows).

src/serena/agent.py
```diff
@@ -1143,32 +1158,56 @@ def apply(
...
+            for root, dirs, files in os.walk(self.project_root):
+                # Don't go into directories that are ignored by modifying dirs inplace
+                # Explanation for the  + "/" part:
+                # pathspec can't handle the matching of directories if they don't end with a slash!
+                # see https://github.com/cpburnz/python-pathspec/issues/89
+                dirs[:] = [d for d in dirs if not ignore_spec.match_file(d + "/")]
+                for file in files:
+                    if not ignore_spec.match_file(os.path.join(root, file)):
+                        files_to_search.append(os.path.join(root, file))
+            # TODO (maybe): not super efficient to walk through the files again and filter if glob patterns are provided
```
opcode81, line 1201: What do you mean "go through the files again"? We never went through the files themselves, only the file system's directory listings - or am I missing something?

src/serena/text_utils.py
```diff
@@ -205,3 +210,48 @@ def search_text(
...
+        if paths_include_glob and not fnmatch(path, paths_include_glob):
+            log.debug(f"Skipping {path}: does not match include pattern {paths_include_glob}")
+            continue
+
+        if paths_exclude_glob and fnmatch(path, paths_exclude_glob):
+            log.debug(f"Skipping {path}: matches exclude pattern {paths_exclude_glob}")
+            continue
+        try:
+            file_content = content_reader(path)
+        except Exception as e:
```
opcode81, line 258: This probably triggers for most binary files, as they will contain character combinations that are not valid in utf-8.
Is this intentional? I believe it will spam the log too much.

src/serena/agent.py
MischaPanch: I mean a second loop through the file names. Just ugly to me but probably not a real problem in any sense, unless there is a bazillion of them

src/serena/text_utils.py
MischaPanch: Done

src/serena/text_utils.py
MischaPanch: Added utf8

src/serena/text_utils.py
MischaPanch: I now make a single log with all skipped files on debug level at the end of the execution

MischaPanch:
I addressed the TODO about fnmatch and added a bunch of tests for glob-matching logic

MischaPanch:
Merging, if you have other comments, we can address them post merge

PR #46, opened 2025-04-09
Readme/better high level explanation

MischaPanch:


opcode81:
Please separate the code changes from the README and let's address the README together (in the same room) one of these days.

opcode81:
The very essence can already be communicated in the three leading bullet items. Here's a proposal:

* Serena is a powerful **coding agent toolkit** that enables LLMs to efficiently **work directly on your codebase**.
* Serena provides **essential semantic code retrieval and editing tools** that are akin to an IDE's capabilities, extracting code entities and exploiting relational structure.
* Serena is **free & open-source**, enhancing the capabilities of LLMs you alredy have access to free of charge.

These items answer the questions
* What is it at a very abstract level?
* How does it achieve good performance/what sets it apart?
* What does it cost me?

The existing bullet items are rather misleading.
And I don't think much more than these changes are necessary, as we already do a good job of explaining how the tools can be leveraged in an MCP server or an Agno agent.

The old section on "LLM Integration" very clearly and concisely stated how the tools could be combined with an LLM to form an agent. The verbose sections you are proposing are not adding anything meaningful, so I'd prefer to keep it short and to the point.

And I think statements like "Serena is three things at once" are very confusing. Serena is, at the very core, an agent toolkit - and that toolkit can specifically be made accessible in several ways.

src/serena/agent.py
```diff
@@ -502,6 +502,20 @@ class ToolMarkerDoesNotRequireActiveProject:
     pass
 
 
+class RestartLanguageServerTool(Tool):
+    """Restarts the language server, may be necessary when edits not through Serena happen."""
```
opcode81, line 507: "when edits not through Serena happen" is not proper English. Change to "when edits occur outside of Serena"?
Similar phrase in docstring below.

src/serena/agent.py
```diff
@@ -502,6 +502,20 @@ class ToolMarkerDoesNotRequireActiveProject:
...
+
+    def apply(self) -> str:
+        """Use this tool only on explicit user request or after confirmation.
+        It may be necessary to restart the language server if the user performs edits
+        not through Serena, so the language server state becomes outdated and further editing attempts lead to errors.
+
+        If such editing errors happen, you should suggest using this tool.
+        """
+        self.agent.reset_language_server()
+        return "OK"
```
opcode81, line 517: Use SUCCESS_RESULT

PR #56, opened 2025-04-10
Close #15: Don't try to use the GUI if unavailable

### LINKED ISSUE #15: Assertion `!xcb_xlib_unknown_seq_number' failed
With the latest changes to address agno-ui issues (https://github.com/oraios/serena/commit/e88c1e6fac5003c520e692c38e0278f00f960e4f), if the log gui is enabled, tinker starts on a separate thread and this happens:

```
$ python scripts/agno_agent.py --project-file ~/devel/pydvl/serena.yml 
Starting thread

[xcb] Unknown sequence number while appending request
[xcb] You called XInitThreads, this is not your fault
[xcb] Aborting, sorry about that.
python3: ../../src/xcb_io.c:157: append_pending_request: Assertion `!xcb_xlib_unknown_seq_number' failed.
Aborted (core dumped)
```

(Incidentally, this also happens if starting serena from within pycharm)

Several projects trying to open several X11 windows in a multithreaded app hit this problem. See e.g. [this one](https://github.com/rust-windowing/winit/issues/458). 

Also, possibly because of how agno starts the agent / app (?), one would always get two X11 windows upon startup showing slightly different logs (can't check now because of the issue above)

mdbenito:
This is just a quick fix that avoids creating a gui window when the gui has been deactivated in the config. I think centralized handling of gui / non-gui logging would make more sense, but in the interim this avoids an obscure crash. Closes #15 

PS: Sorry I can't provide a more general solution right now, but I will be offline for a few days and thought it best to at least patch the problem.

### DIFF
--- a/src/serena/agno.py
+++ b/src/serena/agno.py
@@ -15,7 +15,7 @@ from dotenv import load_dotenv
 from sensai.util.logging import LogTime
 
 from serena import serena_root_path
-from serena.agent import SerenaAgent, Tool
+from serena.agent import SerenaAgent, SerenaConfig, Tool
 from serena.gui_log_viewer import show_fatal_exception
 
 log = logging.getLogger(__name__)
@@ -150,7 +150,10 @@ class SerenaAgnoAgentProvider:
                 try:
                     serena_agent = SerenaAgent(project_file)
                 except Exception as e:
-                    show_fatal_exception(e)
+                    if SerenaConfig().gui_log_window_enabled:
+                        show_fatal_exception(e)
+                    else:
+                        log.exception("Failed to start Serena agent.")
                     raise
 
             # Even though we don't want to keep history between sessions,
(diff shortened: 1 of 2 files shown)


mdbenito:
To reproduce the issue: use a malformed project config, while having the gui deactivated in the main config

opcode81:
The point of `show_fatal_exception` was to make the user see the exception, even if he does _not_ have the GUI enabled.
The GUI for showing the fatal exception is opened on the main thread (in contrast to the regular log window), such that it wouldn't be subject to some platform-specific limitations (macOS insists that UIs live on the main thread, always).

Alas, there are conditions that prevent the GUI from working even on the main thread (no working Tk installation on macOS, no DISPLAY on *nix); Windows is really the only platform where this will never fail. So I have now added a function which tries to show the GUI but will also make sure that the error is logged and if there's no logging configured, written to stderr: #60

opcode81:
Closing in favour of #60

PR #61, opened 2025-04-16
Fix handling of ignored paths

### LINKED ISSUE #59: Help Adding More File Extensions for list_dir command
Hi @MischaPanch and Team,

I noticed that when the LLM uses `list_dir`, whether recursive or not, it only lists files and directories that match the extensions relevant to the language server (e.g., for PHP).

I’m currently working on a custom LSP setup for Serena, using the Intelephense LSP for PHP. I’d like to include additional file types—specifically .xml, .json, and .phtml.

Could you help me figure out which files or settings I need to modify to support these extra extensions?

Thanks!

opcode81:
Fix handling of ignored paths in ListDirTool:
Do not limit results to files not ignored by the LanguageServer, but retain ignoring of ignored directories

Improve handling of ignored paths in LanguageServer:
  * Improve method naming: 
      * should_always_ignore -> is_ignored_dirname 
      * should_ignore_path -> is_ignored_path
  * Fix is_ignored_dirname (should_always_ignore) being applied to filenames
  * Add option ignore_unsupported_files to is_ignored_path to improve versatility

Fixes #59

### DIFF
--- a/src/serena/util/file_system.py
+++ b/src/serena/util/file_system.py
@@ -7,12 +7,14 @@ def scan_directory(
     recursive: bool = False,
     relative_to: str | None = None,
     is_ignored_dir: Callable[[str], bool] = lambda x: False,
+    is_ignored_file: Callable[[str], bool] = lambda x: False,
 ) -> tuple[list[str], list[str]]:
     """
     :param path: the path to scan
     :param recursive: whether to recursively scan subdirectories
     :param relative_to: the path to which the results should be relative to; if None, provide absolute paths
     :param is_ignored_dir: a function with which to determine whether the given directory (abs. path) shall be ignored
+    :param is_ignored_file: a function with which to determine whether the given file (abs. path) shall be ignored
     :return: the list of directories and files
     """
     files = []
@@ -31,7 +33,8 @@ def scan_directory(
                 result_path = entry_path
 
             if entry.is_file():
-                files.append(result_path)
+                if not is_ignored_file(entry_path):
+                    files.append(result_path)
             elif entry.is_dir():
                 if not is_ignored_dir(entry_path):
                     directories.append(result_path)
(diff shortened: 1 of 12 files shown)


src/serena/util/file_system.py
```diff
@@ -1,20 +1,18 @@
...
-from collections.abc import Sequence
+from collections.abc import Callable
 
 
 def scan_directory(
     path: str,
     recursive: bool = False,
     relative_to: str | None = None,
-    ignored_dirs: Sequence[str] = (),
-    ignored_files: Sequence[str] = (),
```
MischaPanch, line 10: should we not also have is_ingored_file here? If feel like it would be more convenient than having to filter afterwards

MischaPanch, line 10: should we not also have is_ingored_file here? If feel like it would be more convenient than having to filter afterwards

opcode81: Depends on what we want the tool to do.
The original idea was: List everything except (files in) sub-directories that are explicitly ignored.
I restored this behaviour.

The previous behaviour was: Ignore everything the language server ignores (in particular, all non-source files).

If we want to ignore files, we could think about applying the new option I added and define
```
is_ignored_file = lambda p: language_server.is_ignored_path(p, ignore_unsupported_files=False)
```
which would ignore only files that are defined in `.gitignore` (if enabled by the user) as well as explicitly configured ignored paths but keep non-source files otherwise.

opcode81: The option I added basically makes `get_ignore_spec` obsolete, and I think we should get rid of it, as it is easy to use it incorrectly.

MischaPanch: Yes, I think that's a good default for is_ignored_file

opcode81: Add this in the second commit.

PR #71, opened 2025-04-20
support for dynamic lsp switching (projects with multiple languages)

shyraptor:
Works really well for me. 
Feel free to modify it to your liking or return it back to me for fixing.

### DIFF
--- a/myproject.template.yml
+++ b/myproject.template.yml
@@ -12,7 +12,7 @@ project_root: /path/to/project
 #
 # Special requirements:
 #  * csharp: Requires the presence of a .sln file in the project folder.
-language: python
+default_language: python
 
 # Whether to enable dynamic language server switching within this project.
 # If true, when a tool operates on a file with a supported language extension (e.g., .py, .ts, .java)
--- a/myproject.template.yml
+++ b/myproject.template.yml
@@ -2,10 +2,26 @@
 # This is optional if this file is placed in the project directory under `.serena/project.yml`.
 project_root: /path/to/project
 
-# language of the project (csharp, python, rust, java, typescript, javascript, go, or ruby)
+# Default language of the project (csharp, python, rust, java, typescript, javascript, go, or ruby).
+# This is used to start the initial language server when the project is activated.
+#
+# If a tool operation targets a file with a different supported language (based on its extension),
+# Serena will attempt to restart the language server for that specific language.
+# If the file's language is not supported or cannot be determined, this default language server will
+# be used (if running).
+#
 # Special requirements:
 #  * csharp: Requires the presence of a .sln file in the project folder.
 language: python
+
+# Whether to enable dynamic language server switching within this project.
+# If true, when a tool operates on a file with a supported language extension (e.g., .py, .ts, .java)
+# that differs from the currently running LSP language, Serena will attempt to restart the LSP server
+# specifically for that file's language. The `default_language` server is used otherwise.
+# If false, Serena will only use the LSP server configured for the `default_language`,
+# and tools might fail or provide limited results for files of other languages.
+# Added on 2025-04-20
+enable_dynamic_lsp_switching: true
 
 # whether to use the project's gitignore file to ignore files
 # Added on 2025-04-07
--- a/src/serena/agent.py
+++ b/src/serena/agent.py
@@ -41,6 +41,11 @@ LOG_FORMAT = "%(levelname)-5s %(asctime)-15s %(name)s:%(funcName)s:%(lineno)d - 
 TTool = TypeVar("TTool", bound="Tool")
 SUCCESS_RESULT = "OK"
 
+def get_language_from_file_path(file_path: str | Path, mapping: dict[str, Language]) -> Language | None:
+    """Determines the Language enum based on the file extension using the provided mapping.
+    """
+    _, ext = os.path.splitext(str(file_path).lower())
+    return mapping.get(ext, None)
 
 def show_fatal_exception_safe(e: Exception) -> None:
     """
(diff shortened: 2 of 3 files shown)


shyraptor:
I'd like to add support for files like *.py.new since tools have a tendency to create such files but multispy would have to be updated as well and I won't have a chance to get around to it in the upcoming week.

src/serena/agent.py
```diff
@@ -155,6 +170,24 @@ def __init__(self) -> None:
         self.gui_log_window_enabled = config_yaml.get("gui_log_window", False)
         self.gui_log_window_level = config_yaml.get("gui_log_level", logging.INFO)
         self.enable_project_activation = config_yaml.get("enable_project_activation", True)
+        
+        # --- Load dynamic language mapping ---
+        raw_mapping = config_yaml.get("dynamic_language_mapping", {})
```
MischaPanch, line 175: There already is a mapping in multilspy, we should use that one

src/serena/agent.py
```diff
@@ -41,6 +41,11 @@
 TTool = TypeVar("TTool", bound="Tool")
 SUCCESS_RESULT = "OK"
 
+def get_language_from_file_path(file_path: str | Path, mapping: dict[str, Language]) -> Language | None:
```
MischaPanch, line 44: Pls instead add a classmethod `from_file_path` to `Language`

src/serena/agent.py
```diff
@@ -331,7 +366,71 @@ def reset_language_server(self) -> None:
         )
         self.language_server.start()
         if not self.language_server.is_running():
+            self._current_lsp_language = None
             raise RuntimeError(f"Failed to start the language server for {self.project_config}")
+        else:
+            self._current_lsp_language = self.project_config.language
+
+    def restart_lsp_with_language(self, target_language: Language) -> bool:
```
MischaPanch, line 374: Maybe instead of restarting we should start another one and keep all of them? If a project uses multiple languages, one should be able to switch quickly

src/serena/agent.py
```diff
@@ -884,6 +1068,10 @@ def apply(
         :param body: the new symbol body. Important: Provide the correct level of indentation
             (as the original body). Note that the first line must not be indented (i.e. no leading spaces).
         """
+        if not self.ensure_language_server_for_file(relative_path):
```
MischaPanch, line 1071: see above, instead we could `activate_language_server_for_file` and keep multiple servers alive

src/serena/agent.py
```diff
@@ -73,13 +78,23 @@ class ProjectConfig(ToStringMixin):
...
+             raise SerenaConfigError(f"Missing required configuration key 'default_language' in project '{project_name}'.")
+
+        self.language: Language = Language(config_dict[lang_key])
         if project_root is None:
             project_root = Path(config_dict["project_root"])
         self.project_root: str = str(project_root.resolve())
         self.ignored_paths: list[str] = config_dict.get("ignored_paths", [])
         self.excluded_tools: set[str] = set(config_dict.get("excluded_tools", []))
         self.read_only: bool = config_dict.get("read_only", False)
+        self.enable_dynamic_lsp_switching: bool = config_dict.get("enable_dynamic_lsp_switching", True)
```
MischaPanch, line 97: With multiple LS running, I think this config is not necessary

myproject.template.yml
```diff
@@ -2,10 +2,26 @@
...
+#
+# If a tool operation targets a file with a different supported language (based on its extension),
+# Serena will attempt to restart the language server for that specific language.
+# If the file's language is not supported or cannot be determined, this default language server will
+# be used (if running).
+#
 # Special requirements:
 #  * csharp: Requires the presence of a .sln file in the project folder.
-language: python
+default_language: python
```
MischaPanch, line 15: How about instead we support an entry `languages` that can also be empty. If multiple are provided, then multiple LS instances are spun up.

MischaPanch (changes requested):
Thank you for the PR @shyraptor ! 

@opcode81 will need to discuss it. I left some comments now, overall I think for supporting this it's better to move towards full support of multi-language projects.

If I didn't get it wrong, your current approach is incomplete. There are tools that use the language server but don't pass any file paths, like find_symbol and overview tools. They can't decide which LS to use.

One solution would be to allow them to select one of the configured languages, though that makes the interface harder to use for the LM. Maybe there is a way to fully support multiple languages without making the interface harder for a single-language project. @opcode81 and I will talk about this and get back to you. Or maybe you already have ideas for how to resolve it - like a multi-LS variant of SerenaAgent? Would be great to achieve it without duplicating code

shyraptor:
> If I didn't get it wrong, your current approach is incomplete. There are tools that use the language server but don't pass any file paths, like find_symbol and overview tools. They can't decide which LS to use.

This is a good point and very much true! Haven't thought this through.
Although, with my specific case (language set to typescript, trying to use find_symbol on python file), find_symbol was always successful with my solution and never successful without it. Now that I think about it, it must have been because of referring to a specific path in the prompt ("Using tool find_symbol, find the definition of the function named register_user in the file marketplace-listing-assistant\backend\app\api\auth.py"). 
For whichever reason this might have worked for me, your argument is still valid.

> Or maybe you already have ideas for how to resolve it - like a multi-LS variant of SerenaAgent?

Sadly, I won't have much time in the upcoming days/weeks to look into it. But I selfishly hope you will work on this idea further since projects (like web apps) are often multi-lingual.

src/serena/agent.py
shyraptor: Very good point, although restarting the server completely instead of "caching" it has been pretty swift for me. Just an insight.

MischaPanch:
> Sadly, I won't have much time in the upcoming days/weeks to look into it. But I selfishly hope you will work on this idea further since projects (like web apps) are often multi-lingual.

Multi language support is very much on our radar, but the highest priority now is to evaluate on SWE bench and inprove quality by comparing tools and prompts with codex and Claude Code, so we will likely not start the multi-language topic before finishing those. But your PR goes in a good direction and is a valuable first step :)

Foxfire1st:
Too bad this is still in an open state. My projects use more than 1 language. So would have been good if you could make multiple entries and use the right language at the right time. For me this would be php and typescript.

MischaPanch:
It's coming soon, we need to finish the work on improving editing tools and do at least a small evaluation on a part of SWE Bench first, as these are higher priority issues. 

Many of our projects are also multi-language, so we need ourselves. If you wanna try solving it yourself meanwhile, feel free to give it a go

Foxfire1st:
@MischaPanch alright that sounds very good. Especially since I think that I could combine Serena + VScode + Copilot + Pieces into a pretty strong unit. Github copilot does already use a lot of native vs code features. But sadly does poorly with terminal and debugger outputs. It can only read from it's own terminals. I hope Serena can close that loop for me. I haven't yet really tested into it but read the documentation and then looked for multi language support.

MischaPanch:
Multiple projects are natively supported, I started on native multi-language support here #168 . Closing this as superseded

PR #73, opened 2025-04-21
Parse parameter descriptions from docstring into MCP tool

marcelomendoncasoares:
## Summary
This PR enhances the MCP server by extracting parameter descriptions from Python docstrings - that would appear in the tools description - and passing them to the proper `description` field of each parameter. The change improves quality of tool documentation available to LLMs.

## Testing
The PR includes test coverage to validate the parameter description extraction functionality. Tests ensure that docstrings in various formats are correctly parsed and that the parameter descriptions are properly integrated into MCP tools.

## Impact
This enhancement will improve the parsing of the tools by LLMs and also the tool evaluating experience for users. Below is an example of the tool listing on RooCode before and after the change:

Before | After
:-------------------------:|:-------------------------:
image | image

### DIFF
--- a/src/serena/mcp.py
+++ b/src/serena/mcp.py
@@ -10,6 +10,7 @@ from logging import Formatter, Logger, StreamHandler
 from typing import Any, Literal
 
 import click  # Add click import
+import docstring_parser
 from mcp.server.fastmcp import server
 from mcp.server.fastmcp.server import FastMCP, Settings
 from mcp.server.fastmcp.tools.base import Tool as MCPTool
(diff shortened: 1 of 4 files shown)


src/serena/mcp.py
```diff
@@ -41,6 +41,36 @@ class SerenaMCPRequestContext:
...
+def _process_parameter_descriptions(
+    parameters_properties: dict[str, dict[str, Any]],
+    func_doc: str,
+) -> str:
+    """
+    Move the parameter descriptions from the docstring to the parameters properties
+    and return the docstring without the parameter descriptions, but keeping the
+    return description if present.
+    """
+    split_docs = func_doc.split(":param ")
```
MischaPanch: Is there a library that can do the docstring parsing for us? Like sphinx but maybe less heavy?

Apart from that, some comments:

1. Generally, a function should either return or have a side-effect, not both
2. Since you are importing it in tests, it should not be private. In fact, I suggest just moving the logic to make_tool, we don't need a separate function for that. You can write a block comment on what this is doing. This way there are also no problems with mutation of user input. The tests should then be rewritten to use make_tool (by definining tools with docstrings on apply within tests)

test/serena/test_mcp.py
```diff
@@ -0,0 +1,139 @@
+"""Tests for the mcp.py module in serena."""
+
+from serena.mcp import _process_parameter_descriptions
+
+
+def test_basic_parameter_processing() -> None:
```
MischaPanch: as mentioned above, pls test the logic of param processing by directly using make_tool. If there is a docstring parsing library we can use (I strongly suspect so), we don't need almost any tests - just one would be enough

MischaPanch (changes requested):
Thanks! This is really helpful, some small comments on code organization

src/serena/mcp.py
marcelomendoncasoares: Agree with you that its surely a better approach to use an existing lib. I implemented without one to avoid adding a dependency, but just changed to use the [docstring_parser](https://pypi.org/project/docstring_parser/) library. Tests were  first refactored to target the `make_tool` function, as suggested.

test/serena/test_mcp.py
marcelomendoncasoares: Done.

MischaPanch:
Thanks, merging. I might delete some tests later

PR #76, opened 2025-04-22
fixup: add missing description to ListDir tool

sysradium:
Fixes the following issue:
```
  File "/Users/user/serena/.venv/bin/serena-mcp-server", line 10, in <module>
    sys.exit(start_mcp_server())
             ^^^^^^^^^^^^^^^^^^
  File "/Users/user/serena/.venv/lib/python3.11/site-packages/click/core.py", line 1161, in __call__
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/user/serena/.venv/lib/python3.11/site-packages/click/core.py", line 1082, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/Users/user/serena/.venv/lib/python3.11/site-packages/click/core.py", line 1443, in invoke
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/user/serena/.venv/lib/python3.11/site-packages/click/core.py", line 788, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/user/serena/src/serena/mcp.py", line 187, in start_mcp_server
    mcp_server = create_mcp_server(project_file_path=project_file, host=host, port=port)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/user/serena/src/serena/mcp.py", line 135, in create_mcp_server
    update_tools()
  File "/Users/user/serena/src/serena/mcp.py", line 124, in update_tools
    mcp._tool_manager._tools[tool.get_name()] = make_tool(tool)
                                                ^^^^^^^^^^^^^^^
  File "/Users/user/serena/src/serena/mcp.py", line 65, in make_tool
    func_doc = f"{docstring.description.strip().strip('.')}."
```
Since `ListDir.apply` does not have a description the parsing tool returns `None`. The easiest fix was to correct the docstring itself.

The regression came from https://github.com/oraios/serena/pull/73

### DIFF
--- a/src/serena/agent.py
+++ b/src/serena/agent.py
@@ -644,6 +644,8 @@ class ListDirTool(Tool):
 
     def apply(self, relative_path: str, recursive: bool, max_answer_chars: int = _DEFAULT_MAX_ANSWER_LENGTH) -> str:
         """
+        Lists files and directories in the given directory (optionally with recursion).
+
         :param relative_path: the relative path to the directory to list; pass "." to scan the project root
         :param recursive: whether to scan subdirectories recursively
         :param max_answer_chars: if the output is longer than this number of characters,
@@ -1092,7 +1094,7 @@ class WriteMemoryTool(Tool):
         """
         if len(content) > max_answer_chars:
             raise ValueError(
-                f"Content for {memory_file_name    } is too long. Max length is {max_answer_chars} characters. "
+                f"Content for {memory_file_name} is too long. Max length is {max_answer_chars} characters. "
                 + "Please make the content shorter."
             )
 


MischaPanch:
Thanks! Merging, but that's not enough - an absence of description should not lead to errors. I will fix the parsing and add a test over all tools

sysradium:
@MischaPanch yeah, that would be a proper fix for sure.

Also I think project is mature enough to switch to releases, so that you don't get accidently a breaking change :)

MischaPanch:
@sysradium to test openhands, I'm solving this with their app in #78 

Concerning maturity: I disagree, we will be doing lots of evaluations  and comparisons to other systems in the next weeks, stuff will change then. Then maybe we can go for a stable version and release a package to pypi :). But I appreciate the compliment

sysradium:
@MischaPanch ok, no problem. Will check the openhands you have mentioned.

My suggestion was to keep the rolling `master` but have reasonable version tags. All of them will be v0.0.x since you are not, as you say, stable enough, but that would at least allow people to pin the "version" and know how "far" they are from the tip.

I get that technically it is possible to use a commit hash or just create a local branch and keep rebasing it. So it is just a suggestion based on how rapidly growing projects operated that I contributed to.

Always happy to help if there is a need.

MischaPanch:
@sysradium maybe we will start having tags soon. We already have a sort of versioning based on the date and `__version__` in the init. I guess, git tags won't hurt

PR #88, opened 2025-04-25
Add support for clangd command-line arguments in configuration

HillLiu:
I make a quick PR to add clangd CLI custom arguments so that clangd can support remote indexing.

For example, see:
https://github.com/clangd/chrome-remote-index/blob/main/docs/index.md

Maybe it's not a good practice, but it's for reference.

### DIFF
--- a/scripts/mcp_server.py
+++ b/scripts/mcp_server.py
@@ -1,6 +1,4 @@
-from serena.mcp import create_mcp_server
-
-mcp = create_mcp_server()
+from serena.mcp import start_mcp_server
 
 if __name__ == "__main__":
-    mcp.run()
+    start_mcp_server()


MischaPanch:
Thanks! We can't do it this way, as it couples the multilspy code to Serena config specifics, and we want to merge back all improvements to multilspy.

I think what we need is a separate language server configuration support for all languages in multilspy. I'll think about how to best achieve this. 

Appreciate you PRs and the feedback!

HillLiu:
Hi @MischaPanch:

Could you also consider making _DEFAULT_MAX_ANSWER_LENGTH configurable?

https://github.com/oraios/serena/blob/main/src/serena/agent.py#L435

Issue #89, opened 2025-04-25
Support for different profiles


MischaPanch:
This issue will be used for comparing AI assistants.

## Problem Statement

In different scenarios the `SerenaAgent` will require different configurations (selected tools and prompts). For example, when started as an agent in something like Agno, the `InitialInstructionsTool` should be deactivated. In a planning mode, only the non-editing tools should be active, and some prompts should be adjusted. Some prompts and tool descriptions can be extended with examples (in context learning), and such examples should be both configurable by the user and dependent on the mode.

## Solution Structure

The serena-mcp-server commands will get two more options `--context <name_or_path_to_yaml>` and `--mode <name_or_path_to_yaml>`. Context and mode profiles are determined by yaml files, and Serena will include multiple of default ones that can be referenced by name. A user can use custom ones by defining new files and passing the paths to them insteaf of the names. Multiple modes can be passed, so one can do for example `--context desktop-chat-app --mode planning --mode interactive`.

The context and mode configurations will influence the system prompt as well as the tool selection. The context refers to the environment in which Serena is being used. For example, the context "within-ide-assistant" means that Serena is being used within something like Cline or Cursor, and should lead to disabling of tools related to non-symbolic file operations (read or write of whole files or sections, replacement of lines, the shell tool), since comparable tools are already available to the LLM inside the context anyway. The system prompt should be adjusted accordingly. The mode determines behavior of serena. For example, in planning mode only non-editing tools should be active, and the prompts should be adjusted to instruct Serena to not write code. In the `one-shot` mode, the system prompt should include instructions on finishing the task autonomously without focusing on explaining intermediate steps, while in the `interactive` mode, it should contain instructions on asking the user clarifying questions whenever needed, and generally provide informative results on intermediate steps.

Finally, no matter which tools end up being selected through context and mode, the configuration of excluded tools in `project.yml` will be prioritized.

## Technical Details

The system-prompt should become a template instead of a fixed string. The template will accept a string for the context and a list of strings for modes. The system prompt should continue being in the prompts.yaml, so you should make sure a template-string will be loaded from it.

It should be easy to programmatically discover which tools are active for a context, a mode, and an arbitrary context, list-of-modes configuration. When resolving conflicts between tool activations, the context should be prioritized. If a conflict arises in a list of modes (like one mode wanting to activate a tool while the other mode wanting to deactivate it), an error should be raised. When the server starts up, the logs should include a highlighted message about the selected context and mode and the final set of active tools.

[...]

PR #96, opened 2025-04-26
Claude desktop serena #89

MischaPanch:
```yaml

agent_name: ClaudeDesktopSerena
commit_at_start: f9cea02674a
serena_at_commit: f9cea02674a
acceptable: false
requirements_fulfilled: almost
one_shot: false
code_quality: poor
num_followup_prompts: 2
developer_frustration: high
cost: 0
comment: Essentially unusable due to poor editing tools, much worse than with Claude Code. 
  Instead of editing files, Claude went on to creating a lot of new modules and new classes.
  Created an overly complex initial solution. Editing failed in a dramatic fashion, tool use also often failed
  (find_symbol with something like `Foo.__init__`, which doesn't work), creating duplicated lines, inserted
  in the middle of a method, created "backup" files, started using `sed` for editing, the language server
  seems to have died a few times and was restarted, rarely understood to read lines before replacing them
  out of the box. `replace_symbol_body` failed very often.
conversations:
  - https://claude.ai/share/553ee4e6-4d93-4b4e-8498-fc05f14d8683
  - https://claude.ai/share/432d6bbc-1cc4-4755-9148-53024de09cef
  - https://claude.ai/share/67900e8f-0abb-44ee-b876-85e09bb2deb1
```

Lessons learned: apart from an urgently needed improvement in the editing tools, as well as better instructions for using the symbolic reading tools, following the conversations I now believe that the tendency of using mocks in tests comes from initially writing tests without mocks, then trying to make them pass at any cost.


MischaPanch:
You were solving the issue attached here. In the implementation you did multiple mistakes:

You should have modified the existing SerenaAgent instead of creating a new class
You also created various new modules which you should not have done. The issue is about extending existing modules and not writing a new mcp module, a new module for a tool, or a new module for an agent.
The test uses too much mocking. I don't want any mocking or monkeypatching. There is a valid serena_agent.config and you can create a temporary project.yaml for tests, otherwise you don't need to mock any tools or anything else. Just check that the existing things work as expected - so be more on the integration test side of things. I forbid you to do monkeypatching!
The prompt extensions in the profiles should be just a single string that is used in templating, no need for a list
Same for the mode, just a single string that is used in the system prompt
You adjusted the autogenerated file prompt_factory.py instead of adjusting the template from which it is created
Have a look at the last commit to see what you did and understand what I'm talking about. Remove all unnecessary new modules that you have written there and correct all the mistakes that I have outlined. Improve the tests. Continue working until the mistakes are fully resolved. Use the symbolic tools to perform minimal edits instead of rewriting whole files!

Issue #109, opened 2025-05-01
Support for all Unicode characters


shyraptor:
my project is often working with Czech characters like "ž", "š" etc., I'll try to create a PR soon

shyraptor:
@MischaPanch Can we close this, or do you want to make sure no such issue is yet to be resolved?

MischaPanch:
I'll check for all read/write operations again before closing

MischaPanch:
Done

PR #110, opened 2025-05-01
specified utf-8 encoding in symbol.py

shyraptor:
Hey, I added explicit utf-8 encoding to _edited_file() in symbol.py since it's been causing extensive problems. I believe it's been just overlooked, but if native encoding was intentional, be so kind to explain the reason :) #109


MischaPanch:
Thanks! We already fixed that in a bunch of places but not this one yet

opcode81:
This is not the ideal solution. The correct fix is to specify the encoding in the project configuration. We cannot just assume an encoding.

PR #117, opened 2025-05-16
Feature/add symbol end position

chknd1nner:
Enhancement to add the **ending line** of the symbol found to the output of the **find_symbol** tool.

This is an improvement because then the **read_file** tool can be called with the exact start and end lines.

Previously, I found Claude called **read_file** and **guessed** how many lines to read which potentially wastes tokens when it reads too many lines or reads too few and has to make a followup tool call to get the rest of the lines it missed.

This feature required the following changes:

1. Changes to src/serena/symbol.py

- Modified the Symbol.to_dict method to include end position information when the location parameter is True
- Added a conditional check to make sure body_end_position exists before trying to access it
- The end position is included in a new end_location field with the same structure as the existing location field

2. Changes to src/serena/agent.py

- Updated the docstring of the FindSymbolTool.apply method to document that the output now includes both start and end location information.
- Added a clear explanation of how this helps with accurate code range identification

3. Added Test Coverage

- Created a new test file test/serena/test_symbol.py with a test case that verifies the end position information is correctly included in the output
- The test mocks a symbol with known start and end positions and ensures both are present in the dictionary representation

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -1,6 +1,10 @@
 name: Tests on CI
 
-on: [pull_request]
+on:
+  pull_request:
+  push:
+    branches:
+      - main
 
 concurrency:
   group: ci-${{ github.workflow }}-${{ github.ref }}


chknd1nner:
image

opcode81:
Trying to understand the problem you are solving: Why does Claude want to read the lines corresponding to a class in the first place? If Claude wants to read the implementation of a class, it can simply pass `include_body=True` to the `find_symbol` tool. This is by far the easiest way to achieve this (in a single tool call, no less). When I use Serena, Claude does this consistently.

Also, the diff for symbol.py is broken, unfortunately.

chknd1nner:
> If Claude wants to read the implementation of a class, it can simply pass `include_body=True`

That's a very good point! But the reality is in my experience Claude will (often) neglect this obvious method and insist on using find_symbol without the include_body property and then fall back on using the read_file tool in an inefficient manner since it doesn't know how many lines to read!

In any case, it's a definite QoL improvement. I see no reason the find_symbol tool **shouldn't** return the entire set of location details.

See the screenshot of a typical interaction below where it should have used include_body, but didn't.

image
Should have used include_body

image
Admits it just guessed the end line number. Fails to even acknowledge it could've used include_body!

chknd1nner:
Sorry about the diff on symbol.py. I am but a humble vibe coder. When I was making changes, serena introduced some bugs to the symbol.py file including a duplicate @property decorator and getting rid of the logging import near the top.

I had serena investigate and it found those issues and also some "indentation" issues. I had serena fix the problems and it seems like it re-wrote the entire symbol.py file with the fixes. This must have introduced some linting issues and hence git is unable to recognise where the original is the same as what I put into the commit.

As a beginner git and github user, I have no idea how to fix this. I'd love to learn though...

MischaPanch:
@chknd1nner if you fix the diff I'm happy to proceed with the review :). It's probably just line endings, could you make sure set the git config to not commit clrf line endings to the repo? Should be `git config --global core.autocrlf true` if you're on windows. We mentioned that in the readme.

Might be easier to start a new branch and just copy-paste your changes there, then the line endings will be handled properly after setting the config. You can then open a new PR and close this one

chknd1nner:
Sure! No problem. I'll fix it up when I get a chance and submit a new PR. Thanks for the tip.

PR #123, opened 2025-05-20
Add Symbol.body_end_position with safe handling and end_location to t…

chknd1nner:
Fixing the diff for symbol.py


chknd1nner:
Dang, it's still not working.

PR #128, opened 2025-05-25
Improve execute_shell_command - Timeout and without shell=True

ShadovvBeast:
Improved execute_shell_command, because I noticed that it was hanging a lot for me (running on Windows, maybe because of that?)
shell=True is usually not necessary, so commented it out, was it there for a reason?
I also added a timeout, although without shell=True it doesn't hang anymore

### DIFF
--- a/src/serena/util/shell.py
+++ b/src/serena/util/shell.py
@@ -25,7 +25,7 @@ def execute_shell_command(command: str, cwd: str | None = None, capture_stderr: 
 
     process = subprocess.Popen(
         command,
-        shell=True,
+#        shell=True,
         stdout=subprocess.PIPE,
         stderr=subprocess.PIPE if capture_stderr else None,
         text=True,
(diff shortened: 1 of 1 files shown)


MischaPanch:
@opcode81 what was the reason for shell=True? For me it also hangs a lot on windows

@ShadovvBeast Thanks for the contribution! 
Could you pls pass the timeout as kwarg instead of hardcoding? Some commands may take long (like tests)

ShadovvBeast:
Sure, with that in mind, moved it to kwargs and raised the default to 60 👍

opcode81:
Disabling `shell` on non-Windows platforms is likely to cause problems, because many commands rely on the shell's configuration. E.g. on Linux, `shell=False` will mean that, in particular, the configuration in `.bashrc` and similar scripts does not apply.
Therefore, `shell = platform.system() != "Windows"` tends to cause the least trouble (and is what I usually use).

I am against introducing any sort of timeout. There is simply no time limit we can reasonably impose; a shell command can take arbitrarily long and raising an exception after some random duration is not going to be helpful.

opcode81:
I pushed the suggested change which sets `shell` to False only on Windows to `main`.
I don't want to add a timeout, so I am closing this PR.
Thanks for pointing out the issue, @ShadovvBeast.

PR #129, opened 2025-05-26
Feature/find symbol end position

chknd1nner:
Finally, some of the recent updates have fixed the line endings to symbol.py!

Now my change to add end_position shows the correct diffs for easy code review.

I also added some enhancements to gitignore for Mac users like myself. I also got Claude to put in every conceivable ignore for every other platform. It added quite a few and even some for Windows that weren't in the original gitignore, but should by like thumbs.db etc.

### DIFF
--- a/src/serena/symbol.py
+++ b/src/serena/symbol.py
@@ -268,22 +268,29 @@ class Symbol(ToStringMixin):
         Convert the symbol to a dictionary.
 
         :param kind: whether to include the kind of the symbol
-        :param location: whether to include the location of the symbol
+        :param location: whether to include the location of the symbol (will also include end_location if available)
         :param depth: the depth of the symbol
         :param include_body: whether to include the body of the top-level symbol.
         :param include_children_body: whether to also include the body of the children.
             Note that the body of the children is part of the body of the parent symbol,
             so there is usually no need to set this to True unless you want process the output
             and pass the children without passing the parent body to the LM.
-        :return: a dictionary representation of the symbol
+        :return: a dictionary representation of the symbol with fields like 'name', 'kind', 'location', 'end_location'
         """
         result: dict[str, Any] = {"name": self.name, "name_path": self.get_name_path()}
-
         if kind:
             result["kind"] = self.kind
 
         if location:
             result["location"] = self.location.to_dict()
+            # Add end position information if available
+            end_pos = self.body_end_position
+            if end_pos is not None:
+                result["end_location"] = {
+                    "relative_path": self.relative_path,
+                    "line": end_pos["line"],
+                    "column": end_pos["character"]
+                }
 
         if include_body:
             if self.body is None:
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,64 @@
-#
+# macOS specific files
+.DS_Store
+.AppleDouble
+.LSOverride
+._*
+.Spotlight-V100
+.Trashes
+Icon
+.fseventsd
+.DocumentRevisions-V100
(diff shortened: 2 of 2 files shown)


src/serena/symbol.py
```diff
@@ -268,22 +268,29 @@ def to_dict(
...
         """
         result: dict[str, Any] = {"name": self.name, "name_path": self.get_name_path()}
-
         if kind:
             result["kind"] = self.kind
 
         if location:
             result["location"] = self.location.to_dict()
+            # Add end position information if available
+            end_pos = self.body_end_position
```
MischaPanch, line 287: why is that needed or helpful? This info is retrieved from the location itself, so result["location"] will contain it. Adding end_location here is just duplicating the information

MischaPanch, line 287: Ah, I see now that self.location only contains the start location. I think it generally should contain more, as sometimes location also refers to info about both start and end location, which will be confusing to the model (and to humans alike).

I think just adding it to the dict is not the best way to solve this, will check out how to best add end location info tomorrow. Thanks for the PR!

PS: We are working on symbol-name based editing which should hopefully make the often fragile location based edits obsolete

chknd1nner: Exactly. The 'find_symbol' tool only returns the start line without this enhancement.

I know the system is supposed to use the include_body = true parameter to return the symbol body, BUT, I found that often the system eroneously falls back on the read_file tool which greatly benefits from having the end_line available.

Without the end_line information, Claude will try to guess the end line and either under read (requiring a second tool call to get more lines), or over read. Both of which are very wasteful of tokens. And this is my overiding goal, to have maximum token efficiency because I am poor and only on Claude pro plan ($20 one).

I asked Claude if it was possible to add the end_position info to the output of the find_symbol tool and this enhancement to ToDict was what it came up with after analysing the entire code base. It also said something about checking if the info was available as a safety measure and for backwards compatibility with existing serena tools/workflows, hence the if statement.

MischaPanch: As so often with LLM code, it's a working solution but not a very clean one :) (a clean one would require a small refactoring, I think).

We'll fix this in the next days, but maybe I'll close this PR and fix it directly on main.

In any case, I appreciate the contribution, thanks for it! It's the identification of the issue that's the main work in this case, not its solution

chknd1nner: No worries. I'd like to understand what a clean solution would look like so I can learn :)

Would also appreciate if you could add all the Mac specific patterns to gitignore so I don't keep getting warnings about untracked files on my repo. Thx.

MischaPanch:
Solved [here](https://github.com/oraios/serena/commit/6c121836bc10ad4fbbb50a47241d65155b34be6f), closing this PR.

Explanation: since we seem to need the end_line for LLM performance reasons (at least for now), it should form part of the SymbolLocation's representation instead of just adding it to the serialization ad hoc. Moreover, in the solution in this PR the relative path was repeated (unnecessary and token-consuming).

The changes in gitignore are unrelated to the issue, did you actually need them, or are they a remnant of generated code?

chknd1nner:
You're right, the gitignore patch is completely unrelated.

I'm such a git newb (but I'm learning fast with AI assistance!). When I created the feature branch for end_position, I should have rebased it to upstream main instead of my own main which has the gitignore change.

Could you please make the gitignore change directly on your main? It would be a QoL improvement for anyone on Mac platform.

MischaPanch:
@chknd1nner I pushed the changes to gitignore to main

PR #130, opened 2025-05-27
Fix args parsing in agno agent

aniliou-85:
Hello there, just a small fix that makes agno agent work with the changes introduced by https://github.com/oraios/serena/commit/9eaf02ee5536a4ebfc8b8faa4b753774596d700b

- Used mutually exclusive group when parsing to keep backwards compatibility
- Tested locally


MischaPanch:
Thx! We should probably also use click here to not have two different parsers. Merging this for now

PR #141, opened 2025-06-01
feat: support custom config file path

jinzhongjia:
This pull request introduces enhancements to the Serena agent and MCP server to support custom configuration files and override options for GUI log window and web dashboard settings. The changes improve flexibility for configuration management and command-line interface (CLI) usability.

### Enhancements to Configuration Management:

* Added an optional `config_file_path` parameter to the `SerenaConfig` class, allowing users to specify a custom configuration file path. This includes updates to the `get_config_file_path`, `from_config_file`, and `save` methods to handle the custom path. (`src/serena/agent.py`, [[1]](diffhunk://#diff-0e6fa239d7c03f1204c98723f6fc94c71b8daff44c7bb416435d7f64fa2169b5R293-R310) [[2]](diffhunk://#diff-0e6fa239d7c03f1204c98723f6fc94c71b8daff44c7bb416435d7f64fa2169b5L316-R321) [[3]](diffhunk://#diff-0e6fa239d7c03f1204c98723f6fc94c71b8daff44c7bb416435d7f64fa2169b5L380-R385)

### CLI Improvements:

* Updated the `create_mcp_server_and_agent` function to accept `config_file_path`, `gui_log_window_override`, and `web_dashboard_override` parameters, enabling runtime overrides of configuration settings. (`src/serena/mcp.py`, [src/serena/mcp.pyR105-R145](diffhunk://#diff-fe762e9741259d3a69e1e1bb71f4dce45980687ea2e3f9e22e2ff4062f310facR105-R145))
* Added new CLI options `--config`, `--gui-log-window/--no-gui-log-window`, and `--web-dashboard/--no-web-dashboard` to the MCP server command for specifying custom configuration files and toggling GUI log window and web dashboard features. (`src/serena/mcp.py`, [[1]](diffhunk://#diff-fe762e9741259d3a69e1e1bb71f4dce45980687ea2e3f9e22e2ff4062f310facR219-R224) [[2]](diffhunk://#diff-fe762e9741259d3a69e1e1bb71f4dce45980687ea2e3f9e22e2ff4062f310facR243-R252)
* Modified the `start_mcp_server` function to pass the new CLI options to `create_mcp_server_and_agent`, ensuring the overrides are applied. (`src/serena/mcp.py`, [[1]](diffhunk://#diff-fe762e9741259d3a69e1e1bb71f4dce45980687ea2e3f9e22e2ff4062f310facR277-R279) [[2]](diffhunk://#diff-fe762e9741259d3a69e1e1bb71f4dce45980687ea2e3f9e22e2ff4062f310facL259-R305)

### DIFF
--- a/src/serena/agent.py
+++ b/src/serena/agent.py
@@ -290,19 +290,24 @@ class SerenaConfig(SerenaConfigBase):
     """
 
     loaded_commented_yaml: CommentedMap
+    config_file_path: str | None = None
 
     CONFIG_FILE = "serena_config.yml"
 
     @classmethod
-    def get_config_file_path(cls) -> str:
+    def get_config_file_path(cls, custom_path: str | None = None) -> str:
+        if custom_path:
+            return str(Path(custom_path).resolve())
         return os.path.join(serena_root_path(), cls.CONFIG_FILE)
 
     @classmethod
-    def from_config_file(cls) -> "SerenaConfig":
+    def from_config_file(cls, config_file_path: str | None = None) -> "SerenaConfig":
         """
         Static constructor to create SerenaConfig from the configuration file
+        
+        :param config_file_path: Optional custom path to configuration file. If None, uses default path.
         """
-        config_file = cls.get_config_file_path()
+        config_file = cls.get_config_file_path(config_file_path)
         if not os.path.exists(config_file):
             raise FileNotFoundError(f"Serena configuration file not found: {config_file}")
 
@@ -313,7 +318,7 @@ class SerenaConfig(SerenaConfigBase):
             raise ValueError(f"Error loading Serena configuration from {config_file}: {e}") from e
 
         # Create instance
-        instance = cls(loaded_commented_yaml=loaded_commented_yaml)
+        instance = cls(loaded_commented_yaml=loaded_commented_yaml, config_file_path=config_file)
 
         # read projects
         if "projects" not in loaded_commented_yaml:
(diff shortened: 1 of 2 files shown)


jinzhongjia:
After adding this, we are allowed to use uvx to run serena directly. Now, if the main branch directly uses uvx to run, it will report that the configuration file cannot be found.

jinzhongjia:
This is my first PR. To be honest, I don't know much about the specific specifications of python projects. If there is any problem with my submission, please explain it directly

MischaPanch:
Thanks for the PR. I understand that one may want to override the config in the call to the command, and also be able to start without having a serena_config.yml. I was about to do this but you beat me to it :) 

But why do you need to have custom config files? What's the use case for that?

jinzhongjia:
When running directly with uvx, uv will run in a temporary directory

At this point, the configuration file cannot be found

MischaPanch:
> When running directly with uvx, uv will run in a temporary directory
> 
> At this point, the configuration file cannot be found

I see. Then the solution is to just create a default config file from the template on the fly. Or, if that doesn't work for some reason, you can use the in-memory version of the config, like we do in the tests. Check the subclasses of the base config.

However, we should never need custom paths to the config file. There should be only one place where it can be located, later on we want this place to be in ~/.serena

jinzhongjia:
So do I think it is better to close the pr, or to modify it directly by you

MischaPanch:
Yes, let's close this. I'll work on making serena installable with uvx and will ping you here when it's done

PR #152, opened 2025-06-05
fix: Agno not using supplied model (closes #149)

### LINKED ISSUE #149: Serena + OpenRouter + Agno (ERROR    OPENAI_API_KEY not set)
# Intention
To use OpenRouter and Serena with Agno.

# Observation

Image

This is confusing, because it seems to me that the Serena + Agno documentation in the README must not work for anyone, since only Claude and Gemini are configured out of the box.

# Expected
README instructions for Agno + Serena should work with models other than OpenAI. 

# Investigation
1. This seems to be caused by AgentMemory not supporting any models other than OpenAI-like models.
https://github.com/agno-agi/agno/issues/2607#issuecomment-2766530032

2. Serena is currently using agno `1.2.15` but the latest agno version is ` 1.5.6` which uses memory in a different way that *seems* to allow specification of the model. https://docs.agno.com/agents/memory#show-me-the-code%3A-memory-%26-storage-in-action

# Attempted Remediation
1. Attempted to specify OpenRouter with "openai/o3", but `AgentMemory` reaches for OpenAI by default, using the env `OPENAI_API_KEY` directly.
2. Attempted to modify `SerenaAgnoAgentProvider` to specify the OpenRouter model into the `classifier` argument of Agent, but no cigar.
3. Added OpenRouter model simultaneously for `classifier`, `summarizer`, and `manager`.
4. Updated agno to `1.5.6` and made the appropriate changes to `src/serena/agno.py` to use the supplied model.
5. Removed `memory` from Agent and set `enable_agentic_memory` to False.

sunwukonga:
feat(config): use OpenRouter/openai model with Agno Memory docs: add OpenAI and OpenRouter keys to `.env.example` chore: update `agno` dependency to version 1.5.6
 - pass model once to Memory, instead of three times

fix(deps): environment variable not available (closes #149)
 - replace dotenv (deprecated) with python-dotenv

feat: add memory model to agent initialization
 - differentiate general model from model used for Memory
 - ultimately this causes problems with reading memory, but may work in the future.

deps: add `openai` package dependency for OpenRouter functionality. add `pyright` for virtual environment support in `pyproject.toml`.
 - allows Mason:pyright in neovim to use correct .venv

chore: lock dependencies `uv.lock`
add persistent SQLite-based memory storage in `SerenaAgnoAgentProvider`
 - seems to be a conflict between Serena memory and Agno memory. Serena wins.

docs: add note to README, agno + serena + openai
 - memory through Agno seems not to be working right now


sunwukonga:
> [!Note]
> I did not attempt to change any `<CR>` `^M` line endings as this would pollute and confuse the PR.
> The README clearly describes how windows users should be configuring their local git to avoid committing carriage returns.

> [!Warning]
> [Seeking feedback]: ~I was unable to use two different models for Serena and Agno Memory, the result was the ability to write memories with Serena, but not read them~.
> EDIT: I have been unable to reproduce this:
> ```python
> model = OpenRouter(id="google/gemini-2.0-flash-001")
> memory_model = OpenRouter(id="openai/gpt-4.1")
> // AND
> enable_user_memories=False,
> ```

> [!Note]
> `gopls` was a test (`uv run poe test`) dependency not mentioned in the README.
> Resolved with: `go install golang.org/x/tools/gopls@latest`
> As a vim user, my language servers are not installed globally; instead, usually through Mason.

MischaPanch:
Thanks, looks good! I'll try it out and merge if everything works

sunwukonga:
Rebased on `main` and fixed an issue with the `.env` environment variables not loaded before OpenRouter module looks in `os.getenv`.

MischaPanch:
I never got around to trying this, but might be obsolete after a modernization of the integration in #702

PR #157, opened 2025-06-06
Add Docker support with GitHub Actions workflow and updated Readme.md

diazoxide:
- Introduced `.github/workflows/docker.yml` for building and pushing Docker images.
- Updated `Dockerfile` to split stages for `base`, `development`, and `production`.
- Added `compose.yaml` for Docker Compose setups.
- Enhanced `README.md` with Docker usage instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,6 +243,8 @@ Configure the MCP server in your client.
 For [Claude Desktop](https://claude.ai/download) (available for Windows and macOS), go to File / Settings / Developer / MCP Servers / Edit Config,
 which will let you open the JSON file `claude_desktop_config.json`. Add the following (with adjusted paths) to enable Serena:
 
+#### Local Installation
+
 ```json
 {
     "mcpServers": {
@@ -253,6 +255,23 @@ which will let you open the JSON file `claude_desktop_config.json`. Add the foll
     }
 }
 ```
+
+#### Docker Installation
+
+Alternatively, you can run Serena using Docker:
+
+```json
+{
+    "mcpServers": {
+        "serena": {
+            "command": "docker",
+            "args": ["run", "--rm", "-i", "--network", "host", "-v", "/path/to/your/projects:/workspaces/projects", "ghcr.io/oraios/serena:latest", "serena-mcp-server", "--transport", "stdio"]
+        }
+    }
+}
+```
+
+Replace `/path/to/your/projects` with the absolute path to your projects directory. The Docker approach has the advantage of not requiring local installation of dependencies.
 
 If you are using paths containing backslashes for paths on Windows
 (note that you can also just use forward slashes), be sure to escape them correctly (`\\`).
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -254,6 +256,23 @@ which will let you open the JSON file `claude_desktop_config.json`. Add the foll
...
+#### Docker Installation
+
+Alternatively, you can run Serena using Docker:
+
+```json
+{
+    "mcpServers": {
+        "serena": {
+            "command": "docker",
+            "args": ["run", "--rm", "-i", "--network", "host", "-v", "/path/to/your/projects:/workspaces/projects", "ghcr.io/oraios/serena:latest", "serena-mcp-server", "--transport", "stdio"]
```
MischaPanch, line 270: Thanks, overall the PR looks good and is a valuable addition!

However, the docker based setup requires more explanation. First, the project activation will not work directly with this for anything that is not mounted. Second, the serena_config.yml will be populated with paths from the container, not local paths, so after using the docker setup a non-docker setup cannot really be used. A solution would be to have a dedicated serena_config for docker, which would require code changes. 

The dashboard will run on a port that will need to be forwarded. You already did this in compose.yaml, but it should be mentioned here. The local port to which it forwards could be occupied, it should probably be read off the env. Or maybe there is a mechanism to take the next non-occupied port, like we are doing it now in the dashboard startup?

The gui window option should always be disabled for the docker setup.

Another concern I have is about the line-endings when using docker on windows. One will probably end up with inconsistent line-endings in edited files, right?

We were generally thinking of extending the current configuration setup which would also make it more easily compatible with using serena through docker. But for now, I think we should clearly document it as experimental and list all the caveats that I mentioned.

You can also add that for the docker setup, the shell tool will be safer, which is definitely a benefit

diazoxide: Updated, please check.

DOCKER.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Future Improvements
+
+We're working on:
+- Automatic config migration between environments
+- Better project path handling
+- Dynamic port allocation
+- Windows line-ending handling
+
+For updates and issues, please check the [GitHub repository](https://github.com/AbanteAI/serena).
```
MischaPanch: Pls rather create issues in the serena repo and link to them :). In the issues you can then link to the fork, if you want to

src/serena/agent.py
```diff
@@ -102,6 +102,19 @@ def get_serena_managed_dir(project_root: str | Path) -> str:
...
+def is_running_in_docker() -> bool:
+    """Check if we're running inside a Docker container."""
+    # Check for Docker-specific files
+    if os.path.exists('/.dockerenv'):
+        return True
+    # Check cgroup for docker references
+    try:
+        with open('/proc/self/cgroup', 'r') as f:
+            return 'docker' in f.read()
+    except:
```
MischaPanch: Can be narrowed to FileNotFound, right?

src/serena/agent.py
```diff
@@ -315,6 +328,7 @@ class SerenaConfig(SerenaConfigBase):
     loaded_commented_yaml: CommentedMap
 
     CONFIG_FILE = "serena_config.yml"
+    CONFIG_FILE_DOCKER = "serena_config.docker.yml"
```
MischaPanch: For passing custom configs, the mounted directory would need to contain such a file. This is achieved through the mount in compose if the user has created this file in serena root (I guess this is what you envisioned), but it's not clear.

Pls add an explanation that the file will be autogenerated if missing, that it is added to the container in the compose mount, and that users can adjust it.

MischaPanch (commented):
Apart from minor comments, looks good, thanks! Will merge when they are addressed

MischaPanch:
@diazoxide ping. Wanna finish it yourself? Otherwise I'd implement the small comments I made and merge this

Issue #160, opened 2025-06-08
Feature: Custom prompts without editing anything in repository


80avin:
Use case:
I am finding the prompts for memory are too generic, and some tools' prompts can be improvised.
But since I use serena using `uvx` directly from the github repository which copies the serena repository in a temp directory, I cannot edit the files in config/**/*.yml to customize the prompts.

Requirement
* copy the config files to a user config directory like `~/.local/share/serena/` for linux ( may use [platformdirs](https://pypi.org/project/platformdirs/) ) and read the configs from there
* (Optional if above is implemented) Provide a CLI argument to read config from a custom directory

MischaPanch:
Yes, I agree. @opcode81 and I were discussing extending the configuration options and will likely be implementing something permitting prompts and description overrides soon, at least before publishing Serena to pypi.

But for now, why is it inconvenient for you to clone and adjust the repo instead of using uvx? In the end, the prompts are essentially a part of the core functionality, and changing them is akin to changing code

80avin:
I am also working on a PR for this.

> why is it inconvenient for you to clone and adjust the repo instead of using uvx?

With uvx, I can use the `-U` argument to keep the `serena` up to date. Otherwise, it is a manual effort, and I will often lag behind.
Moreover, if I change them in the repo clone, then I will have unstaged files requiring conflict handling on every git pull. I cannot push them because they might be my personal preferences.

> changing them is akin to changing code

I don't think I can agree here.
configs reflect personal preferences, like I use [mise](https://mise.jdx.dev/) for my tool & venv management, so I want serena's prompts to convey that. I also want it to save different observations in memory than the default ones.
If I change these in code, then I have merge conflicts to manage on every pull.

MischaPanch:
Ok, that's a valid point. I tried to make things configurable with the possibility for custom modes and contexts, the memories and the project prompt, but it might not be enough. 

I think a solution should allow the user to override the prompts without having to copy-paste an entirely new config. So maybe:

1. tool description overrides in the context yaml
2. Prompt template overrides by allowing multiple yamls and merging them. This should happen in src/interprompt, since we still want to use the PromptFactory for loading the prompt templates. Then one can configure the overrides through CLI or serena_config

80avin:
@MischaPanch 
I thought of having more config files which can override the existing prompts or extend them.
But, realized that just providing the base config files is equivalent to it and saves the headache of managing & merging multiple configs.

[...]

PR #161, opened 2025-06-08
feat: copy config files to user directory

80avin:
fixes https://github.com/oraios/serena/issues/160

Copies the config files to user directory ( ~/.local/share/serena in linux ) on first initialization.

If the config is modified in the repository, we can upgrade the `.version` to force a re-write of config.

NOTE: Tested only on Linux.


80avin:
Moving to draft.
While the changes work, logging isn't working.

I spent some time, but apparently the log implementation is a bit complex.

On logging, I am curious to know why aren't we logging to file directly ? We are logging to stdout/stderr on start of program and later we switch to memorylogger.
Instead, I guess at the start of everything we should set log target as a file and serve the dashboard logs from there itself.

src/serena/config.py
```diff
@@ -144,3 +144,185 @@ def print_overview(self) -> None:
...
+    1. Creates user config directories if they don't exist
+    2. Copies config files from package to user data directory if they don't exist
+    3. Handles version-based migration with backup and logging
+    4. Copies serena_config.template.yml to serena_config.yml in user data directory
+
+    Raises:
+        Exception: If copying fails due to permissions or other IO issues
+
+    """
+    from pathlib import Path
```
MischaPanch, line 163: Pls don't use local imports without need

src/serena/config.py
```diff
@@ -144,3 +144,185 @@ def print_overview(self) -> None:
...
+        log.info("User configuration initialization completed successfully")
+
+    except Exception as e:
+        log.error(f"Failed to initialize user configuration: {e}")
+        raise Exception(f"Critical error during config initialization: {e}") from e
+
+
+def _copy_or_migrate_config_dir(config_name: str, src_dir: Path, dest_dir: Path, current_version: str):
+    """Copy or migrate a configuration directory."""
+    import shutil
```
MischaPanch, line 213: Also here, move up

src/serena/config.py
```diff
@@ -144,3 +144,185 @@ def print_overview(self) -> None:
...
+
+        # Copy the directory
+        shutil.copytree(src_dir, dest_dir, dirs_exist_ok=True)
+
+        # Write version file
+        version_file.write_text(current_version)
+        log.info(f"Successfully copied {config_name} config files")
+
+
+def _copy_or_migrate_main_config(template_file: Path, user_config_file: Path, current_version: str):
```
MischaPanch, line 254: Overall, the code in serena is object-oriented. This means config-related functionality should be added to the serena-config-related classes. They are currently in agent.py (we can move them to config.py later).

I suspect that quite some code here was written by AI, which is not a bad thing :). You can ask to rewrite it in an object-oriented way and use existing classes, maybe introduce new ones if needed. I will review then later.

To make the reviewing easier for me, pls also add info to the readme and changelog. It has to be done anyway, and will make it easier for me to understand the workflow you envisioned

MischaPanch (changes requested):
Overall, lot's of good stuff here - config versioning is a good idea, so is allowing the users to change the prompts. However, the structure of the added code doesn't fit the pre-existing project structure. We could push/modify some memories to make it easier to work on serena with LLMs.

For an easier review/merge I'd also propose to split the proposed functionality into two PRs. One for versioning and custome user config, and a separate one with backup/migration code. Anything that calls shutil.rmtree needs to be reviewed very carefully. You can certainly reach the desired functionality (passing a path to custom config files) without any migration, right? One could even split it in three PRs - versioned configs, desired functionality, migration/copying code.

There's also a change to uv.lock that shouldn't be there

MischaPanch:
> On logging, I am curious to know why aren't we logging to file directly ? We are logging to stdout/stderr on start of program and later we switch to memorylogger. Instead, I guess at the start of everything we should set log target as a file and serve the dashboard logs from there itself.

I need to check why that's the case, the logging was mainly written by @opcode81 who's on vacation now. But the functionality you need should not require any changes to logging

src/serena/config.py
80avin: It was me actually. I couldn't find where shall I fit this function, and didn't realize that the repo follows Object Oriented paradigm.

I'll refactor it.

MischaPanch: Cool, thanks! I suspected an LLM because of the local imports, which often appear as an artefact :)

MischaPanch: Pls make sure to read the overall PR comment [here ](https://github.com/oraios/serena/pull/161#pullrequestreview-2909837162) before you start refactoring. Just making sure you don't do unnecessary work and the next review goes smoothly

80avin: Sure.

Though I'm not sure when will I be able to complete them, mostly because of the day job.

> ...because of the local imports...

Probably I'm influenced by FreeCAD recently, but I have a habit of doing so, mostly in plugins. Reasoning being that a plugin whose startup may block the app's startup, it is better to do lazy imports.
In any case, it's probably not worth it here.

80avin:
> ...the functionality you need should not require any changes to logging

I found it weird that the logs I wanted to push weren't coming to the terminal nor in the http server. That's why I moved the PR to draft, to inspect the cause behind this.

src/serena/config.py
MischaPanch: Np, if you feel like you won't have time, let me know, then I'll continue where you stopped

80avin: That will be much better !!

80avin: It may take, at worst, one week for me.
But if you feel you have enough time to do it earlier, I won't mind at all.

MischaPanch: NP, it can wait a week. Since this touches end-user configuration, it's important to get this right.

I have more than enough other topics to work on for serena, so I'll happily accept your help on this, even if it may take a while

MischaPanch:
@80avin do you still want to work on this?

opcode81:
@80avin closing this in favour of #274, which also adds support for the configuration file being located in the user directory (and making this the new default).
Feel free to build on the new structure (and open a new PR) to support the goals of #160, though we should perhaps clarify what exactly it is that we want to support.

PR #168, opened 2025-06-13
Support multi-language projects

opcode81:
NOTE: This was created by Codex, not me.

## Summary
- allow project.yml to specify multiple languages and autogenerate them
- create `MultiLanguageServer` to manage multiple language servers
- start a language server for each detected language
- update symbol search tools with optional language parameter
- add mixed-language test repository and tests

## Testing
- `uv run poe format`
- `uv run poe type-check`
- `uv run poe test` *(fails: network access needed to download language server dependencies)*

------
https://chatgpt.com/codex/tasks/task_e_684bf765bf748320806ea9881a82801c

### DIFF
--- a/src/serena/agent.py
+++ b/src/serena/agent.py
@@ -39,6 +39,7 @@ from serena import serena_version
 from serena.config import SerenaAgentContext, SerenaAgentMode
 from serena.constants import PROJECT_TEMPLATE_FILE, REPO_ROOT, SELENA_CONFIG_TEMPLATE_FILE, SERENA_MANAGED_DIR_NAME
 from serena.dashboard import MemoryLogHandler, SerenaDashboardAPI
+from serena.multi_language_server import MultiLanguageServer
 from serena.prompt_factory import PromptFactory, SerenaPromptFactory
 from serena.symbol import SymbolManager
 from serena.text_utils import search_files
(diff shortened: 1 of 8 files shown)


MischaPanch:
Hi all

@opcode81 and I thought fairly long about this topic and decided to go into a slightly different direction. Managing multiple LS is resource intensive and finalizing this feature in a clean way will require rather large changes to Serena, given how flaky the language servers are and how annoying it is to deal with various race conditions and async task execution.

Moreover, isolated LS for multi-language projects will never work really well. One would not be able to find references from one language for another.

So, for now we decided to go a different way. We will write an IDE plugin, first for jetbrains, then possibly also for vscode, where we use the IDE's powerful symbolic analysis to provide the functionality that we currently use language servers for. With this, Serena will work out of the box on all frameworks supported by the IDE, and it will also provide a better integration.

Since we are a small independent team, and we receive no funding for Serena, we will make the plugin a paid service. If this goes well, it will allow us to continue investing time in Serena to make the open-source version amazing for all non-professional programmers or those who don't want to invest any money in such a tool, while others could consider purchasing the plugin in order to get enhanced functionality that is not easily reachable with language servers, and in order to support the project and ensure its long-term sustainability. We see this as a natural way to have the best of both worlds.

Closing this PR, the issue #192 will be resolved when the plugin is published. Feel free to follow the issue to get updates on that

PR #170, opened 2025-06-13
Fix deadlocks

### LINKED ISSUE #153: with claude-code the find_symbol and related mcp tools don't return any results
I added serena mcp server to claude using

```bash

 claude mcp add serena -- uvx \
   --from git+https://github.com/oraios/serena serena-mcp-server 
   --context ide-assistant
```

and then I run claude, and it correctly lists all  the serena mcp tools, but when I activate my project and search for various symbols, the tools (e.g. `find_symbol`) return empty results.

MischaPanch:
Closes #153

The issue was due to async event loop contamination, which is essentially a bug in asyncio and happens at random times, causing unrecoverable deadlocks.

It was quasi impossible to reproduce outside of the MCP server (thus impossible to debug), since it's mainly within the MCP server where the event loops cross-contaminated

We had a whole of three asyncio applicaions running in a single process: the mcp server, the language server and the dashboard app. They all shared one event loop and eventually, after enough requests, things would deadlock.

The solution was to let everything run in its own request. Since Tools and tool metadata needed for FastMCP are not serializable, a small refactoring was needed in the core code. Major new modules were introduced for creating process isolation, in particular a tool wrapper.

A major challenge was to terminate everything correctly. We also have to collect logs from one process and pass it to another.

The current status works. Names are suboptimal, and the solution is likely overly complex. I generated fairly large parts with claude-code (and serena within it, in the periods when it worked). I did have to go through the new code extensively manually, fix a lot of issues and remove a lot of complexity. Still, probably too much complexity remains. I already marked one TODO for later, but it's not very urgent.

Tests for the process isolation and a new marker for them have been added. I tested the dashboard shutdown and logging forwarding extensively manually,

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,68 @@
+name: Build and Push Docker Images
+
+on:
+  push:
+    branches: [ main ]
+    tags: [ 'v*' ]
+  pull_request:
+    branches: [ main ]
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  build-and-push:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
+
+    - name: Log in to Container Registry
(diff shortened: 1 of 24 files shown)


MischaPanch:
@opcode81 Finally finished, merging this now since it solves core functionality issues. We can do a post-merge review in pair programming

The work on this was probably the most frustrating and time-consuming task that I have ever done.

PR #180, opened 2025-06-19
Add codespell support (config, workflow to detect/not fix) and make it fix some typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,25 @@
+# Codespell configuration is within pyproject.toml
+---
+name: Codespell
+
+on:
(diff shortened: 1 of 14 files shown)


MischaPanch:
Thanks for the contribution! Merging, but if it codespell gets in the way, I'll disable it later ;). Let's see

yarikoptic:
Sure, just remember that it is easy to adjust configuration or provide inline comments for ignores.

PR #181, opened 2025-06-19
Option to specify a Serena config directory

zhaodong2013062:
I'm having issues using Serena in my organization-managed codebase, where I am not allowed to create a `.serena` configuration directory in the project itself. I've added an option to specify a config directory so I can create the config outside of my repo.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,25 @@
+# Codespell configuration is within pyproject.toml
+---
+name: Codespell
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: read
+
+jobs:
+  codespell:
+    name: Check for spelling errors
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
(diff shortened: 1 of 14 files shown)


MischaPanch:
@zhaodong2013062 I'm not a fan of extending the interfaces, though it may be necessary for your case. It's quite strange though, coding agents will write code files in your project, why is having some (gitignored) config files a problem? 

Your current PR is insufficient for not creating a .serena dir, since the language server cache will also land there. IMO it doesn't make much sense to keep the LS cache outside of the project


EDIT: below is something I wrote in the beginning because I misunderstood the issue and thought it was about serena_config.yml instead of the project.yml.


Would it work for you to read the config from ~/.serena instead? We were generally thinking of moving the config location there before publishing Serena to pypi.

The logic should be: take the config in ~/.serena if it's available, otherwise check if there is a config at its current position and move it it to home (migration for backwards compatibility), otherwise generate a new config in home.

@opcode81 wdyt?

MischaPanch (commented):
@zhaodong2013062 thanks for the PR.

Have you seen my comment above? Let's continue the conversation in that thread

MischaPanch:
Closing due to inactivity

PR #203, opened 2025-06-20
Update Omnisharp to 1.39.13

thomhurst:


MischaPanch:
Did it solve the indexing for you?

thomhurst:
Not entirely but it got past the file that was problematic before. So it might have helped

> thomh@Tom:~/dev/TUnit$ uvx --from git+https://github.com/thomhurst/serena index-project
Indexing symbols in project /home/thomh/dev/TUnit
Indexing (AfterTestHookMethod.cs):  10%|████▍                                        | 138/1394 [00:23<05:20,  3.92it/s]

MischaPanch:
Ah, man. Maybe it needs to be updated even more? The LS might also be using the wrong .NET version , the responsible code is in `get_dotnet_version()`

thomhurst:
I think I may have found the root cause?

For whatever reason, it's not always getting all the bytes from a file.

> Indexing (ClassConstructorTest.cs):  75%|███████████████████████████████▎          | 1049/1407 [00:03<00:01, 322.33it/s]    self.run()
  File "/home/thomh/.local/share/uv/python/cpython-3.11.13-linux-x86_64-gnu/lib/python3.11/threading.py", line 982, in run
    self._target(*self._args, **self._kwargs)
  File "/home/thomh/.cache/uv/archive-v0/iRFxO_YB6ZXYWXoOI7OBb/lib/python3.11/site-packages/solidlsp/ls_handler.py", line 325, in run_forever
    body = self._read_bytes_from_process(self.process, self.process.stdout, num_bytes)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/thomh/.cache/uv/archive-v0/iRFxO_YB6ZXYWXoOI7OBb/lib/python3.11/site-packages/solidlsp/ls_handler.py", line 298, in _read_bytes_from_process
    raise EOFError(f"Process terminated. Expected {num_bytes} bytes, got {len(data)}")
EOFError: Process terminated. Expected 41699 bytes, got 7024

MischaPanch (changes requested):
Thanks for the analysis and the PR! I have resolved the conflicts and added minor adjustments on the branch feature/omnisharp-1.39.13.

Please set your branch to it.

We are in the process of completely removing the code in multilspy, just FYI - no need to add anything there.

As the main change is in ls_handler, and it was not entirely clear to me whether it actually solves or just reports the issue, I asked Claude to analyze the diff. Its conclusion is below.

@opcode81 wrote the new ls_handler, so he's better suited to reviewing this part

Looking at the error message and the changes, this partially solves the problem but likely doesn't address the root cause completely. Here's the analysis:
What the Changes Fix:
1. Better Detection and Recovery
The new code would have provided much better diagnostics for this exact scenario:

Instead of just "Process terminated. Expected 41699 bytes, got 7024"
You'd get: exit code, stderr output, and more context about what went wrong

2. Timeout Protection
The 30-second timeout prevents the code from hanging forever when partial reads occur, but doesn't solve why partial reads happen.
3. Chunked Reading
The new code reads in chunks (up to 8KB) rather than trying to read all remaining bytes at once, which might help with large responses (41699 bytes is fairly large).
What It Doesn't Fix:
The Root Cause
The fundamental issue appears to be:

OmniSharp process is terminating unexpectedly during large responses
Or there's a communication protocol mismatch where the Content-Length header says 41699 bytes but only 7024 are actually available

Possible Underlying Issues:

OmniSharp crashes when processing large files/responses
Buffer overflow in OmniSharp's stdout handling
Encoding issues where byte count is calculated incorrectly
Process resource limits causing OmniSharp to terminate mid-response

So the changes are:

✅ Better error reporting - you'll know why OmniSharp died
✅ Graceful degradation - system won't hang on partial reads
✅ Recovery capability - can continue processing other requests
❌ Root cause fix - OmniSharp is still likely to crash on large responses

Next Steps Would Be:

Check OmniSharp logs for why it's terminating
Investigate memory limits or resource constraints
Consider chunking large requests before sending to OmniSharp
Add retry logic for failed large requests

The changes make the system more resilient and provide better debugging info, but you'd probably still see OmniSharp processes terminating during large file indexing operations.

MischaPanch:
@thomhurst did the changes you made actually solve this problem? And how was the behavior when this appeared - did the language server crash or did indexing continue once you had caught such errors?

It's not clear to us yet what could have caused it unless omnisharp crashed in the middle of reading the file

thomhurst:
So it didn't solve it completely. But I got to about 90% until it started failing again. Which is must more than the 6 or 7% I was getting originally.

I was using Claude to help with this PR, and the summary is pretty accurate. It helps narrow down the problem and give better errors, but I don't think the root cause is fixed.

What I did notice is that once a timeout happened, it'd return an error saying "The language server did not respond in 30 seconds" and then it would move on to the next file. But then every single file after that would hang too. So it's as if the first file is still holding everything up.

MischaPanch:
>What I did notice is that once a timeout happened, it'd return an error saying "The language server did not respond in 30 seconds" and then it would move on to the next file. But then every single file after that would hang too. So it's as if the first file is still holding everything up.

Ok, that's consistent with the hypothesis that omnisharp just dies while reading a file. It can be addressed by catching such cases and restarting it

MischaPanch:
We can probably switch to directly using `Microsoft.CodeAnalysis.LanguageServer` with dotnet and nuget instead of relying on omnisharp. I imagine that the former is more stable

thomhurst:
That sounds sensible

MischaPanch:
Closing, we're getting rid of omnisharp

Issue #246, opened 2025-06-27
ARM support for Docker images (M-series Macs)


SalvatoreT:
When I tried to pull the Docker image for Serena on my MacBook Pro, I got the following error:

```
❯ docker pull ghcr.io/oraios/serena:latest
latest: Pulling from oraios/serena
no matching manifest for linux/arm64/v8 in the manifest list entries
```

So, I checked the Docker images, and it looks like only the x86 version is being built (or at least tagged).
https://github.com/oraios/serena/pkgs/container/serena

![Image](https://github.com/user-attachments/assets/8e03ffec-78dd-485f-9f5a-c92a773dde91)

I'll open a PR to add support.

SalvatoreT:
🎉 

```
❯ docker pull ghcr.io/oraios/serena:latest
latest: Pulling from oraios/serena
34ef2a75627f: Already exists
6939e8b629d3: Already exists
8d8c9691b884: Already exists
4185d6dac69a: Already exists
ac8bfd75f443: Pull complete
79a94c4b7c2b: Pull complete
7c0a04b8774d: Pull complete
2ac968290ada: Pull complete
a659b80aa030: Pull complete
e957da5e18d6: Pull complete
a95bb55216e1: Pull complete
abe1ef915ec9: Pull complete
4f4fb700ef54: Pull complete
ff9c23f10a94: Pull complete
Digest: sha256:776ce6e7b4d40ef362d00bf6f0a4ef032678ec8445a4dfe361491c0f2d84de84
Status: Downloaded newer image for ghcr.io/oraios/serena:latest
ghcr.io/oraios/serena:latest
```

https://github.com/oraios/serena/pkgs/container/serena

![Image](https://github.com/user-attachments/assets/9054ad39-bd23-464b-b9e1-207ce290f751)

PR #247, opened 2025-06-27
ci: publish ARM-based Docker image too

SalvatoreT:
Without ARM support, this is harder to use on MacBook Pros.

I copied the `platforms` tag example from the Docker docs. https://docs.docker.com/build/ci/github-actions/multi-platform/

Closes https://github.com/oraios/serena/issues/246


MischaPanch:
Thanks! Merging

PR #251, opened 2025-06-28
Support for clojure via clojure-lsp

### LINKED ISSUE #162: Support for clojure via clojure-lsp
I just opened https://github.com/microsoft/multilspy/pull/113 adding support for clojure. Once the kinks are ironed out of that one, I can prepare a separate PR for serena if you'd like.

mdbenito:
Given the lack of attention received by https://github.com/microsoft/multilspy/pull/113 , here's the PR for serena. It's basically that one, plus a few extra tests.

This PR:

- Closes #162 by adding support for clojure-lsp
- Adds some basic tests and clojure test repo
- Adds java and clojure to the pytest github workflow (this _might_ benefit from some caching, but I haven't looked into it)
- Adds a mark to the clojure tests to skip if the cli is unavailable or not working properly
- Adds the custom marker config to `pytest.ini`: in contradiction with the claim in pytest docs, tests were throwing warnings about unknown custom marks 🤷🏽 

## Open questions

- [x] Windows tests fail (see my comment below)
- [ ] I don't know what the policy is for failing tooling in language servers: even though I'm downloading `clojure-lsp`, we still need the correct version of clojure cli and a java SDK installed. For now, I'm just raising exceptions. How would you like to fail in this situation? Log and fail? 
- [ ] Ditto for the download of `clojure-lsp` itself.

### DIFF
--- a/src/solidlsp/language_servers/clojure_lsp/clojure_lsp.py
+++ b/src/solidlsp/language_servers/clojure_lsp/clojure_lsp.py
@@ -94,7 +94,8 @@ class ClojureLSP(SolidLanguageServer):
         platform_id = PlatformUtils.get_platform_id()
 
         runtime_dependencies = [dependency for dependency in self.runtime_dependencies if dependency["platformId"] == platform_id.value]
-        assert len(runtime_dependencies) == 1
+        if len(runtime_dependencies) != 1:
+            raise RuntimeError(f"Could not find a suitable clojure-lsp runtime dependency for platform {platform_id.value}.")
         dependency = runtime_dependencies[0]
 
         clojurelsp_ls_dir = os.path.join(os.path.dirname(__file__), "static", "clojure-lsp")
@@ -103,9 +104,9 @@ class ClojureLSP(SolidLanguageServer):
             os.makedirs(clojurelsp_ls_dir, exist_ok=True)
             logger.log(f"Downloading and extracting clojure-lsp from {dependency['url']} to {clojurelsp_ls_dir}", logging.INFO)
             FileUtils.download_and_extract_archive(logger, dependency["url"], clojurelsp_ls_dir, dependency["archiveType"])
-        assert os.path.exists(clojurelsp_executable_path)
+        if not os.path.exists(clojurelsp_executable_path):
+            raise FileNotFoundError(f"Download failed? Could not find clojure-lsp executable at {clojurelsp_executable_path}")
         os.chmod(clojurelsp_executable_path, stat.S_IEXEC)
-
         return clojurelsp_executable_path
 
     def _get_initialize_params(self, repository_absolute_path: str) -> InitializeParams:


MischaPanch:
Thanks @mdbenito! Let me know when I should review

mdbenito:
> Thanks @mdbenito! Let me know when I should review

Well, this is ready-ish. Windows tests fail for some reason I cannot debug, lacking access to a windows machine / license. I suspected some issue with the hardcoded paths in the tests, but it seems to be something else. If anyone can help or suggest some workaround, I'd be grateful.

src/solidlsp/language_servers/clojure_lsp/initialize_params.json
```diff
@@ -0,0 +1,133 @@
+{
```
MischaPanch: I think there's no point for keeping these jsons separately, for the pyright server I put everything directly into the module. Wdyt? If you agree, would you mind moving them to dicts and then deleting the jsons?

test/solidlsp/clojure/test_clojure_basic.py
```diff
@@ -0,0 +1,230 @@
+from pathlib import Path
```
MischaPanch: Could you pls also add at least one test case for closure to test/test_serena_agent.py? I found it valuable to also test language server implementations using the high-level interfaces, even though the test logic is largely overlapping

pytest.ini
```diff
@@ -1,2 +1,14 @@
 [pytest]
 addopts = --snapshot-patch-pycharm-diff
+# this shouldn't be necessary since it's already in pyproject.toml
+# but tests show lots of warnings without it
+markers =
```
MischaPanch, line 5: the markers are in pyproject.toml, this file shouldn't exist (deleting it on main now)

src/solidlsp/language_servers/clojure_lsp/runtime_dependencies.json
```diff
@@ -0,0 +1,37 @@
...
+            "archiveType": "zip",
+            "binaryName": "clojure-lsp"
+        },
+        {
+            "id": "clojure-lsp",
+            "description": "clojure-lsp for Windows (x64)",
+            "url": "https://github.com/clojure-lsp/clojure-lsp/releases/latest/download/clojure-lsp-native-windows-amd64.zip",
+            "platformId": "win-x64",
+            "archiveType": "zip",
+            "binaryName": "clojure-lsp"
```
MischaPanch: maybe this should be clojure-lsp.exe? Could be the reason CI fails on windows.

I can check myself on windows soon

pytest.ini
MischaPanch: This file was introduced recently by mistake, this is why you encountered 

> Adds the custom marker config to pytest.ini: in contradiction with the claim in pytest docs, tests were throwing warnings about unknown custom marks 🤷🏽

src/solidlsp/language_servers/clojure_lsp/initialize_params.json
mdbenito: JSONs make no sense, agreed.

src/solidlsp/language_servers/clojure_lsp/runtime_dependencies.json
mdbenito: Seems likely! I'll push that just in case.

mdbenito: Good catch, that solved it.

mdbenito:
@MischaPanch I addressed the comments you left and fixed the windows issue with your suggstion. What would you like to do wrt. the couple of open points in the PR description?

MischaPanch:
Thx! @mdbenito 

For the other LS we try to download as much as possible, without requiring the user to already have things on their system. Only if it's too complicated or unreasonable to not do that (like e.g. installing go ourselves) we log and fail.

You know better where to put that threshold here. If closure cli is hard to install automatically, log and fail is reasonable.

The last point about closure-lsp I don't understand - you already download it in this PR, right?

mdbenito:
> For the other LS we try to download as much as possible, without requiring the user to already have things on their system. Only if it's too complicated or unreasonable to not do that (like e.g. installing go ourselves) we log and fail.

My question was rather about _how_ to fail. I'm raising an exception and letting serena die. Is that expected? Are exceptions caught elsewhere?

> You know better where to put that threshold here. If closure cli is hard to install automatically, log and fail is reasonable.

Anyone working with clojure needs a working cli. Installing a potentially different version for code analysis seems like a source for obscure problems for the user.

> The last point about closure-lsp I don't understand - you already download it in this PR, right?

Yes, I download `clojure-lsp`, but if the download fails, an assert just kills serena, which is kind of brutal IMO.

MischaPanch:
Failing and killing Serena is the right approach, the agent is useless without a language server. I generally prefer non-assert exceptions for this, but it's not too important. Approving this now, feel free to merge or to first improve the error handling, if you want to

MischaPanch (approved):

PR #255, opened 2025-06-29
Add pytest-xdist for parallel tests

mdbenito:
Just that. Uses `-n auto` in the test suite for max speed. Should halve test time on CI given that GH uses 2 cores in their runners. But it's mostly nicer for local testing

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,6 +53,7 @@ dev = [
   "mypy>=1.4.1",
   "poethepoet>=0.20.0",
   "pytest>=8.0.2",
+  "pytest-xdist>=3.5.0",
   "ruff>=0.0.285",
   "toml-sort>=0.24.2",
   "types-pyyaml>=6.0.12.20241230",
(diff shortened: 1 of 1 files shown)


MischaPanch:
We have session-scoped fixtures, and I generally made bad experiences with parallelizing tests (racing conditions leading to mysterious failures), so I'm somewhat reluctant.

We can add xdist to the dependencies, and you can do `poe test -n auto` locally, but let's not do it for CI for now, ok?

MischaPanch:
Already you see, on windows tests are slower and on MacOS one of them fails (might be just random and unrelated though)

mdbenito:
> We have session-scoped fixtures, and I generally made bad experiences with parallelizing tests (racing conditions leading to mysterious failures), so I'm somewhat reluctant.

I can relate to that :D 

> We can add xdist to the dependencies, and you can do poe test -n auto locally, but let's not do it for CI for now, ok?

👍🏽 

> Already you see, on windows tests are slower and on MacOS one of them fails (might be just random and unrelated though)

Weird, I was expecting a failure on windows, not on OSX. I'm rerunning the failed job, let's see.

PR #261, opened 2025-06-30
Elixir Support

dbernazal:


opcode81:
Thanks for the contribution! 
Please resolve the conflicts (note that `pytest.ini` was removed) and also add Elixir to the list of languages in `README.md`.

opcode81:
@dbernazal for the tests in CI to work, `pytest.yml` will need to be extended.
Quick search indicates that this could perhaps be used to install Elixir:
https://github.com/erlef/setup-beam

To fix the (stupid) spelling issue, you can add `# codespell: ignore` to the respective line.

dbernazal:
Got it, will address.

dbernazal:
Looking into the next-ls install next

opcode81:
I would suggest to download the OS-specific language server in the LS Python code and store it in the `static` subdirectory of the language server directory (see e.g. `EclipseJDTLS.setupRuntimeDependencies` for an example where this is done) - unless there is a specific reason to do this differently for Elixir.

Installing a well-defined version in this way for every Serena user will ensure a known/fixed version and thus reproducible results. Because of this, it is preferred to relying on a system-level installation.

dbernazal:
Still working through the CI errors. I also realized next-ls doesn't support Windows. I could explicitly note this and remove the Elixir tests from the Windows builds if we're ok with that or potentially try to switch it to a different Language server.

opcode81:
> I also realized next-ls doesn't support Windows. 

next-ls does seem to support Windows. The [latest release](https://github.com/elixir-tools/next-ls/releases/tag/v0.23.3) contains a .exe file, at least.

MischaPanch:
@dbernazal I added clojure to the list of languages in the readme, there's now a tiny conflict. Pls fix it so we can run the GH actions

dbernazal:
>I added clojure to the list of languages in the readme, there's now a tiny conflict. Pls fix it so we can run the GH actions

Y'all are so fast on the reviews! 😆 

I attempted to re-implement with ElixirLS but it requires Dialyzer to be run in order to have symbol references. This didn't seem great because of the time it takes for Dialyzer to run it's initial pass(can be minutes). 

Instead, I fixed up tests for the NextLS implementation and remove the test run for Windows. I'll monitor the CI and fix anything else that pops up.

dbernazal:
Just put into draft until I get CI sorted.

opcode81:
> remove the test run for Windows.

@dbernazal but doesn't next-ls support Windows?

MischaPanch:
The removing of the tests on windows has to happen with skipif, not just in CI (otherwise tests don't run through locally). Have a look how it was done for closure.

That is, if it can't be made to work for windows. You might want to make the skipif conditional on next-ls being available

dbernazal:
@opcode81 Sorry, meant to clarify. I confirmed with the maintainer, Mitch Hanberg(we work together), that NextLS does not support Windows despite the exe.

dbernazal:
I think this one is ready now. Thank you for your help and patience!

.gitignore
```diff
@@ -221,3 +221,8 @@ tmp/
 
 # Claude settings
 .claude/settings.local.json
+
+# Elixir
+/test/resources/repos/elixir/test_repo/deps
```
MischaPanch, line 226: Pls add to the gitignore in /test/resources/repos/elixir/test_repo instead

test/solidlsp/elixir/test_elixir_basic.py
```diff
@@ -0,0 +1,123 @@
...
+        user_new_symbol = None
+        for symbol in symbols[0]:  # Top level symbols
+            if symbol.get("name") == "User" and symbol.get("kind") == 2:  # Module
+                for child in symbol.get("children", []):
+                    if child.get("name", "").startswith("def new(") and child.get("kind") == 12:  # Function
+                        user_new_symbol = child
+                        break
+                break
+        
+        if not user_new_symbol or "selectionRange" not in user_new_symbol:
```
MischaPanch, line 103: This happens in many places in the tests. The tests should either fail or succeed but not be skipped if in case something unexpected happens

test/solidlsp/elixir/test_elixir_basic.py
```diff
@@ -0,0 +1,123 @@
...
+        if not create_user_symbol or "selectionRange" not in create_user_symbol:
+            pytest.skip("UserService.create_user function or its selectionRange not found")
+
+        sel_start = create_user_symbol["selectionRange"]["start"]
+        references = language_server.request_references(
+            file_path, sel_start["line"], sel_start["character"]
+        )
+
+        assert references is not None
+        assert len(references) > 0
```
MischaPanch, line 85: This is not enough, one should at least test that the references have the right names.

test/solidlsp/elixir/test_elixir_basic.py
```diff
@@ -0,0 +1,123 @@
...
+        user_new_symbol = None
+        for symbol in symbols[0]:  # Top level symbols
+            if symbol.get("name") == "User" and symbol.get("kind") == 2:  # Module
+                for child in symbol.get("children", []):
+                    if child.get("name", "").startswith("def new(") and child.get("kind") == 12:  # Function
+                        user_new_symbol = child
+                        break
+                break
+        
+        if not user_new_symbol or "selectionRange" not in user_new_symbol:
```
MischaPanch, line 43: same as other comment on usage of skip

test/solidlsp/elixir/test_elixir_basic.py
```diff
@@ -0,0 +1,123 @@
...
+        )
+
+        assert referencing_symbols is not None
+
+    @pytest.mark.parametrize("language_server", [Language.ELIXIR], indirect=True)
+    def test_timeout_enumeration_bug(self, language_server: SolidLanguageServer):
+        """Test that enumeration doesn't timeout (regression test)."""
+        # This should complete without timing out
+        symbols = language_server.request_document_symbols("lib/models.ex")
+        assert symbols is not None
```
MischaPanch, line 118: these kind of asserts are too weak, we should at least check for some expected names

test/solidlsp/elixir/test_elixir_integration.py
```diff
@@ -0,0 +1,162 @@
...
+    """Integration tests for Elixir language server with test repository."""
+
+    @pytest.fixture
+    def elixir_test_repo_path(self):
+        """Get the path to the Elixir test repository."""
+        test_dir = Path(__file__).parent.parent.parent
+        return str(test_dir / "resources" / "repos" / "elixir" / "test_repo")
+
+    def test_elixir_repo_structure(self, elixir_test_repo_path):
+        """Test that the Elixir test repository has the expected structure."""
```
MischaPanch, line 34: I don't think this test is needed

test/solidlsp/elixir/test_elixir_integration.py
```diff
@@ -0,0 +1,162 @@
...
+        content = language_server.retrieve_full_file_content(services_file)
+        lines = content.split("\n")
+        user_reference_line = None
+        for i, line in enumerate(lines):
+            if "alias TestRepo.Models.{User" in line:
+                user_reference_line = i
+                break
+                
+        if user_reference_line is None:
+            pytest.skip("Could not find User reference in services.ex")
```
MischaPanch, line 64: same as in other comments. You should probably never have this pattern

test/solidlsp/elixir/test_elixir_integration.py
```diff
@@ -0,0 +1,162 @@
...
+        if function_matches:
+            files_with_functions = set()
+            for match in function_matches:
+                if match.source_file_path:
+                    files_with_functions.add(os.path.basename(match.source_file_path))
+            
+            # Should find functions in multiple files
+            expected_files = {"models.ex", "services.ex", "examples.ex", "utils.ex", "test_repo.ex"}
+            found_files = expected_files.intersection(files_with_functions)
+            assert len(found_files) > 0, f"Expected functions in {expected_files}, found in {files_with_functions}"
```
MischaPanch, line 90: as in other tests, just checking for existence is not enough

test/solidlsp/elixir/test_elixir_integration.py
```diff
@@ -0,0 +1,162 @@
...
+                    all_symbols.extend(symbol_group)
+                else:
+                    all_symbols.append(symbol_group)
+            
+            symbol_names = [s.get("name", "") for s in all_symbols]
+            
+            # Should understand nested module structure
+            expected_modules = ["TestRepo.Models", "User", "Item", "Order"]
+            found_modules = [name for name in expected_modules if any(name in symbol_name for symbol_name in symbol_names)]
+            assert len(found_modules) > 0, f"Expected modules {expected_modules}, found symbols {symbol_names}"
```
MischaPanch, line 121: similar to other comments, check that we find exactly what we expect pls

test/solidlsp/elixir/test_elixir_symbol_retrieval.py
```diff
@@ -0,0 +1,346 @@
...
+        lib_files = [key for key in overview_keys if key.startswith('lib/')]
+        assert len(lib_files) > 0, f"Expected to find lib/ files in overview keys: {overview_keys}"
+        
+        # Should contain information about our modules
+        overview_text = str(lib_overview).lower()
+        expected_terms = ["models", "services", "user", "item"]
+        found_terms = [term for term in expected_terms if term in overview_text]
+        assert len(found_terms) > 0, f"Expected to find some terms from {expected_terms} in overview"
+
+    # @pytest.mark.parametrize("language_server", [Language.ELIXIR], indirect=True)
```
MischaPanch, line 302: pls remove or make functional

src/solidlsp/language_servers/elixir_tools/initialize_params.json
```diff
@@ -0,0 +1,97 @@
+{
```
MischaPanch, line 1: we're moving away from the jsons, see #265 . Pls move this and the runtime dependencies directly into python dicts, like in the pyright server. You can also have a look at the PR I mentioned

test/solidlsp/elixir/conftest.py
```diff
@@ -0,0 +1,169 @@
...
+    # Check if already compiled (optimization for repeated runs)
+    build_path = os.path.join(repo_path, "_build")
+    deps_path = os.path.join(repo_path, "deps")
+    
+    if os.path.exists(build_path) and os.path.exists(deps_path):
+        print(f"Elixir test repository already compiled in {repo_path}")
+        return
+        
+    try:
+        print(f"Installing dependencies and compiling Elixir test repository for optimal Next LS performance...")
```
MischaPanch, line 38: not sure we should have prints in tests or rather use logging, @opcode81 wdyt?

test/solidlsp/elixir/conftest.py
```diff
@@ -0,0 +1,169 @@
...
+"""
+import os
+import subprocess
+import pytest
+import time
+from pathlib import Path
+
+
+def ensure_elixir_test_repo_compiled(repo_path: str) -> None:
+    """Ensure the Elixir test repository dependencies are installed and project is compiled.
```
MischaPanch, line 12: Is that something that should be done on the general language server level?

test/solidlsp/elixir/conftest.py
```diff
@@ -0,0 +1,169 @@
...
+    All output is logged for transparency and debugging.
+    """
+    # Get the test repo path relative to this conftest.py file
+    test_repo_path = Path(__file__).parent.parent.parent / "resources" / "repos" / "elixir" / "test_repo"
+    ensure_elixir_test_repo_compiled(str(test_repo_path))
+    return str(test_repo_path)
+
+
+@pytest.fixture(scope="session")
+def elixir_test_repo_path(setup_elixir_test_environment):
```
MischaPanch, line 163: if setup_elixir_test_environment is autouse, why does it need to be passed? And what's the point of this fixture, it just returns what it gets

MischaPanch:
@dbernazal I'm gonna merge this now since I want to move on with #265 , and I don't want to force you to resolve conflicts because of that. If you have time, I'd appreciate a follow-up PR that addresses the comments on tests. Otherwise, I will try addressing them myself soon.

Thank you for bringing Elixir to Serena!

dbernazal:
Apologies for the delay, but yes will address the comments in a new PR.

PR #284, opened 2025-07-05
JetBrains plugin support

opcode81:


tylersatre:
If you'd like any outside testing of this I'd love to give it a shot.  I had been working on making serena use multiple LSPs but this seems like the much cleaner option in the end!

src/serena/config/serena_config.py
```diff
@@ -22,19 +23,90 @@
...
+    @classmethod
+    def default(cls) -> "ToolSet":
+        """
+        :return: the default tool set, which contains all tools that are enabled by default
+        """
+        from serena.tools import ToolRegistry
+
+        return cls(set(ToolRegistry().get_tool_names_default_enabled()))
+
+    def apply(self, *tool_inclusion_definitions: "ToolInclusionDefinition") -> "ToolSet":
```
MischaPanch, line 51: ToolInclusionDefinition should be imported at TYPE_CHECKING

src/serena/config/context_mode.py
```diff
@@ -75,6 +72,14 @@ def from_name(cls, name: str) -> Self:
             )
         return cls.from_yaml(yaml_path)
 
+    @classmethod
+    def from_name_internal(cls, name: str) -> Self:
```
MischaPanch, line 76: Reminder to future self - also load from `~/.serena/...`. Method should be renamed to `from_name`.

We can also do it in this PR, though it's unrelated to the plugin support

src/serena/project.py
```diff
@@ -0,0 +1,278 @@
...
+        :return: the content of the file
+        """
+        abs_path = Path(self.project_root) / relative_path
+        if not abs_path.exists():
+            raise FileNotFoundError(f"File not found: {abs_path}")
+        return abs_path.read_text(encoding=self.project_config.encoding)
+
+    def get_ignore_spec(self) -> pathspec.PathSpec:
+        """Returns the pathspec matcher for the paths that were configured to be ignored through
+        the multilspy config.
```
MischaPanch:
```suggestion
        the solidlsp config.
```

src/serena/project.py
```diff
@@ -0,0 +1,278 @@
...
+
+        This is is a subset of the full language-specific ignore spec that determines
+        which files are relevant for the language server.
+
+        This matcher is useful for operations outside of the language server,
+        such as when searching for relevant non-language files in the project.
+        """
+        return self._ignore_spec
+
+    def is_ignored_dirname(self, dirname: str) -> bool:
```
MischaPanch: make private?

src/serena/project.py
```diff
@@ -0,0 +1,278 @@
...
+        return dirname.startswith(".")
+
+    def is_ignored_relative_path(self, relative_path: str, ignore_unsupported_files: bool = True) -> bool:
```
MischaPanch: make private?

src/serena/symbol.py
```diff
@@ -490,24 +411,36 @@ def add_children(s: Self) -> list[dict[str, Any]]:
...
+    """
+    the column number in which the reference is located (0-based)
+    """
 
     @classmethod
     def from_lsp_reference(cls, reference: LSPReferenceInSymbol) -> Self:
-        return cls(symbol=Symbol(reference.symbol), line=reference.line, character=reference.character)
+        return cls(symbol=LanguageServerSymbol(reference.symbol), line=reference.line, character=reference.character)
 
+    def get_relative_path(self) -> str | None:
```
MischaPanch, line 439: when can this be None?

src/serena/config/serena_config.py
opcode81: ToolInclusionDefinition is defined in the same file, but below this usage.

src/serena/config/context_mode.py
opcode81: Why should the method be renamed? Internal modes are separate from regular modes which can be loaded through the other method `from_name`; this is in addition to `from_name` for loading internal modes that are not user-facing.

src/serena/symbol.py
opcode81: See defining class `LanguageServerSymbolLocation`:

> the relative path of the file containing the symbol; if None, the symbol is defined outside of the project's scope

src/serena/project.py
opcode81: The whole docstring was bogus, I updated it.

src/serena/project.py
opcode81: Done.
Oddly, the function `is_ignored_path` does not call this method.
Previously, this was not such an apparent oddity, as the methods were not as close to each other (and differently named iirc).

src/serena/project.py
opcode81: Done

src/serena/config/context_mode.py
MischaPanch: Didn't notice (or forgot) that we already have from_name

PR #313, opened 2025-07-17
Feature/enhanced cli

MischaPanch:
Support for custom modes and contexts, as well as restructuring an enhancement of CLI.

Run `uv run help` to see all options

I hacked a bit around click to make the code structured, mostly idiomatic with click, and still compatible with the pyproject entrypoint scripts


opcode81:
From an OOP perspective, it's not a very elegant design: Many static methods are introduced to replace what could have been handled more elegantly in abstractions like `RegisteredSerenaAgentMode` (reifying a registered mode with its meta-data) and `RegisteredSerenaAgentModeCollection` (singleton for the full set).

PR #318, opened 2025-07-19
Apply project-specific tool exclusions early, cc_vanilla

MischaPanch:
## Verdict

overly complex, creates an additional class without any need, and forgot to exclude the activate_project tool

## Reproduction & Stats
Tokens: 658,747
Prompt:
When the serena mcp server is started in ide-assistant mode and a project is passed, we should:                                                                                                                         │
│                                                                                                                                                                                                                           │
│   1. check the excluded tools from the project.yml. If no project.yml exists, there are no excluded tools there. You can use the Project or ProjectConfig classes for that.                                               │
│   2. exclude the tools there a-priori for the entire session, like the tools that are excluded from the context                                                                                                           │
│   3. exclude the activate_project tool                                                                                                                                                                                    │
│                                                                                                                                                                                                                           │
│   The code modification should happen in SerenaAgent, the problematic block is here:                                                                                                                                      │
│                                                                                                                                                                                                                           │
│           # determine the base toolset defining the set of exposed tools (which e.g. the MCP shall see),                                                                                                                  │
│           # limited by the Serena config, the context (which is fixed for the session) and JetBrains mode                                                                                                                 │
│           tool_inclusion_definitions = [self.serena_config, self._context]                                                                                                                                                │
│           if self.serena_config.jetbrains:                                                                                                                                                                                │
│               tool_inclusion_definitions.append(SerenaAgentMode.from_name_internal("jetbrains"))                                                                                                                          │
│           self._base_tool_set = ToolSet.default().apply(*tool_inclusion_definitions)                                                                                                                                      │
│           self._exposed_tools = {tc: t for tc, t in self._all_tools.items() if self._base_tool_set.includes_name(t.get_name())}                                                                                           │
│           log.info(f"Number of exposed tools: {len(self._exposed_tools)}")                                                                                                                                                │
│                                                                                                                                                                                                                           │
│   The exposed tools should be adjusted to reflect the logic I mentioned above

### DIFF
--- a/src/serena/agent.py
+++ b/src/serena/agent.py
@@ -179,7 +179,45 @@ class SerenaAgent:
         tool_inclusion_definitions = [self.serena_config, self._context]
         if self.serena_config.jetbrains:
             tool_inclusion_definitions.append(SerenaAgentMode.from_name_internal("jetbrains"))
+
+        # If we're in ide-assistant mode and a project is passed, exclude project-configured tools early
+        project_excluded_tools: set[str] = set()
+        if self._context.name == "ide-assistant" and project is not None:
+            # Get excluded tools from project config if it exists
+            try:
+                project_instance = self.serena_config.get_project(project)
+                if project_instance is None and os.path.isdir(project):
+                    # Try to load project config directly from the project path
+                    from serena.config.serena_config import ProjectConfig
+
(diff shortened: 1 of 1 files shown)


MischaPanch:
Discarding

PR #321, opened 2025-07-20
Capture startup logs and other logging improvements

opcode81:


MischaPanch:
Cool, the code is more orderly! And good that we have an abstraction for logging execution times

PR #339, opened 2025-07-23
Add Svelte language server support

screenfluent:
## Summary
  Adds full Svelte language server support to Serena with complete LSP integration for Svelte 5 development.

  ## Features Added
  - **SvelteLanguageServer**: Complete LSP integration with auto-installation of `svelte-language-server@0.17.16`
  - **Svelte 5 Compatibility**: Full support for modern syntax (`$props()`, `onclick` vs `on:click`)
  - **TypeScript Integration**: Proper handling of TypeScript in Svelte components
  - **Symbol Operations**: Complete symbol discovery, finding, and editing capabilities
  - **Test Coverage**: Comprehensive test suite validating all functionality

  ## Technical Implementation
  - Added `Language.SVELTE` enum and `.svelte` file pattern recognition
  - Implemented runtime dependency management with npm integration
  - Enhanced LSP initialization parameters for Svelte 5 and TypeScript support
  - Created realistic test repository with modern Svelte 5 components
  - Updated .gitignore to properly exclude Node.js dependencies

  ## Validation
  ✅ All tests pass (`uv run poe test -m "svelte"`)
  ✅ Type checking passes (`uv run poe type-check`)
  ✅ Code formatting passes (`uv run poe format`)
  ✅ Cross-language compatibility maintained
  ✅ Real-world SvelteKit project validation completed

  ## Files Changed
  - `src/solidlsp/language_servers/svelte_language_server.py` - New Svelte LSP implementation
  - `src/solidlsp/ls_config.py` - Added SVELTE language enum
  - `src/solidlsp/ls.py` - Added factory integration
  - `test/solidlsp/svelte/` - Complete test suite
  - `test/resources/repos/svelte/` - Test repository with Svelte 5 components
  - `.gitignore` - Added Node.js dependency exclusions
  - `README.md` - Updated language support documentation
  - `CHANGELOG.md` - Documented new features

  This enables Serena to work seamlessly with Svelte projects, providing the same powerful symbol-based editing capabilities available for other supported languages.

  **Tested and confirmed working on real SvelteKit projects with TypeScript integration.**


MischaPanch:
This is very cool, thank you for the PR!

But I don't understand how that would have typescript support. Does the svelte LS automatically analyze typescript files, finding cross references and so on, without starting a TS language server? I was under the impression that it doesn't do that.

Also, the filename matcher only contains .svelte files, so for a project containing ts and svelte files, the autodetected dominant language might end up being typescript.

Your tests don't include typescript files or reference finding. If typescript is really supported out of the box, please add such tests. I'd be happy to bring full svelte support to serena

screenfluent:
Yeah, you are right! I will work on this. Big thanks for awesome project!

MischaPanch:
I don't think it will be easy to reach full svelte support this way. In vscode, afaik, the svelte plugin passes the output of the typescript LS to the svelte LS for cross-referencing, that's not easy to replicate in Serena. Also, it would require starting the two LS for svelte projects, something not currently implemented anywhere else.

You might have seen that we're working on a jetbrains extension that we'll publish soon, with that Serena will get full support of svelte and most other frameworks

MischaPanch:
Don't want to discourage you on finding a solution to this, just saying it's probably not going to be trivial

screenfluent:
Seems to be working... 😅

MischaPanch:
Very cool, that's how I imagined it! But I think this doesn't fully work, right?

```

    @override
    def request_references(self, relative_file_path: str, line: int, column: int) -> list[ls_types.Location]:
        """Reference finding with intelligent language server routing."""
        if self._should_use_typescript_ls(relative_file_path):
            try:
                return self._get_ts_server().request_references(relative_file_path, line, column)
            except Exception as e:
                self.logger.log(f"TypeScript LS references failed: {e}", logging.WARNING)

        # Use Svelte LS
        return super().request_references(relative_file_path, line, column)
 ```

It won't find the references to typescript symbols within .svelte files.

Regardless, it will provide a lot of value and we can still merge it, but it was this part that I meant when I said "it's not gonna be simple". Do you want to try and solve this? Or do I misunderstand something?

**Comment by screenfluent** [comment]
Ahhh yep I missed it. Let me fight with that tomorrow! :)

**Comment by MischaPanch** [comment]
Wish you luck in that! :smile:

**Comment by screenfluent** [comment]
Still working on that with finishing touches and it was brutal fight haha. I had to create something like this:

```text
+--------------------------------+
|        Serena MCP Core         |
+--------------------------------+
             ^      |
             | LSP  |
             v      |
+--------------------------------+
| SvelteProxyLanguageServer (Py) |
|      (Adapter - The Brain)     |
+--------------------------------+
    |          |            |
    | LSP      | LSP        | stdio (svelte -> tsx)
    v          v            v
+----------+ +----------+ +-------------------+
| Svelte   | | TypeScript | svelte2tsx Bridge |
| LS       | | LS       | (Minimal Node.js) |
| (stdio)  | | (stdio)  | (stdio)           |
+----------+ +----------+ +-------------------+

**Comment by MischaPanch** [comment]
Cool! I had warned you that it wouldn't be easy ^^

Thanks for the work, that's gonna be awesome feature and many users will be thankful to you

**Comment by screenfluent** [comment]
Hey Mischa! I’m still playing with Svelte(Kit) support and the thing with .svelte-kit/types being in .gitignore keeps biting me... 🙃

Right now my local patch just touched the core like this:

src/serena/project.py
```py
def validate_relative_path(self, rel_path, language_server=None):
    ...
    if self.is_ignored_path(rel_path):
        if language_server and hasattr(language_server, "is_ignored_path"):
            if not language_server.is_ignored_path(rel_path, ignore_unsupported_files=False):
                return   # allowed for this LS
        raise ValueError("ignored")
```

…and every tool now passes its LS instance:

src/serena/tools/file_tools.py
```py
ls = getattr(self.agent, "language_server", None)
self.project.validate_relative_path(path, ls)
```

It works, but core suddenly knows too much about language server 😅

I’m thinking about cleaner roads:

1. allow-list in config

```yaml
allowed_paths:
  - ".svelte-kit/**/*.d.ts"
  - ".svelte-kit/ambient.d.ts"
  - ".svelte-kit/tsconfig.json"
```

Project.is_ignored_path() would first check this list.

2. pluggable ignore policy

```py
class IgnorePolicy(Protocol):
    def is_ignored(self, rel: str) -> bool: ...

Project(..., ignore_policy=GitIgnorePolicy()+SvelteKitPolicy())
```

3. capability-based approach (I think this is great balance)

Each LS could declare what framework files it needs:

```py
# In SolidLanguageServer base class
def get_framework_path_allowlist(self) -> list[str]:
    return [] # Default is empty

# In SvelteProxyLanguageServer
@override
def get_framework_path_allowlist(self) -> list[str]:
    return [".svelte-kit/**/*.d.ts", ".svelte-kit/tsconfig.json"]

# Core would call this during validation:
  if language_server:
      allowlist = language_server.get_framework_path_allowlist()
      if any(fnmatch(rel_path, pattern) for pattern in allowlist):
          return  # Framework needs this path
```

4. just use negation rules in `.gitignore (!/.svelte-kit/types/**/*.d.ts)` but I don't like to force users to do it...


The thing is, this is not just SvelteKit problem. Next.js has `.next/types/` Nuxt has `.nuxt/types/` or Astro generates types in ignored dirs. So having a clean solution would help many adapters.

What do you think? I like capability-based approach most. It keeps core clean and lets each framework adapter handle its own special paths. Happy to implement any approach you prefer!

MischaPanch:
Hey Szymon:)

Yes, option 3 seems seems like the right approach to me as well. We definitely shouldn't force the users to modify their gitignores or to couple parts of the code that don't belong together. So it makes sense to introduce a new concept and let language server implementations select the files they need. 

I'm a bit worried about how find symbol and so on will work with this though. For example, references and symbols (and generally anything) should never be found in these files. Moreover, the LS has a method request_analyzed_files (or something like that), which would also be affected.

So, while I agree with the option 3, I think just a straightforward whitelist won't do it. We need to have a dedicated concept for files that the language server absolutely needs for some reason but that don't form part of the code that the user wants analyzed. Wdyt?

@screenfluent

MischaPanch:
Hi @screenfluent . Are you still working on this?

screenfluent:
Hey @MischaPanch I pushed too hard recently and burned out a bit. I came up with framework files, renamed later to support files, as maybe it would be more future proof for some cases.

```py
    def get_support_files(self) -> list[str]:
        """
        Returns a list of glob patterns (using fnmatch syntax) for files that are
        essential for the language server's analysis but are not user-editable
        source code. Paths must be project-relative strings.

        These files will be accessible for reading but will be strictly read-only
        and will be hidden from user-facing tools like symbol searches.
        """
        return []

    def is_support_file(self, relative_path: str) -> bool:
        """
        Checks if a given path matches the glob patterns for support files.
        The path passed must be a project-relative string.
        """
        # Coerce to string and guard against directory traversal
        relative_path = str(relative_path)
        if relative_path.startswith(("../", r"..\\")):
            return False

        patterns = self.get_support_files()
        if not patterns:
            return False

        from fnmatch import fnmatch

        return any(fnmatch(relative_path, pattern) for pattern in patterns)
```

And introduced write flag to path validation to check write permissions 

```py
self.project.validate_relative_path(relative_path, write=True)
```

But I need to take some rest to finish it.

MischaPanch:
Take your time, no stress! Justed wanted to know whether you're still on it. 

Otherwise, we could merge the first version where everything is supported except finding cross-language references, this would already make most users happy.

But if you want to continue with this soon, we can wait till it's fully finished instead

MischaPanch:
If you don't want to continue on this, would you mind if we picked it up from here at some point later to bring in svelte support? Should we then keep your commits, or would you prefer your name to not be associated?

screenfluent:
Oh sorry, nothing has changed. I was cleaning up local mess on my computer to work with git worktrees to be able to restart my work and continue contributing long term. I completely forgot that those changes will mess up my PR.

MischaPanch:
NP, I got a bit worried :). Don't want to rush you at all, just that there's already very valuable work done here and I wouldn't want it to go to waste. The finding references is really a cherry on top, albeit an important one

screenfluent:
No worries, Serena is too important for me haha.

MischaPanch:
Hey @screenfluent. Wanted to ping and ask if you have an approximate timeline on this. We want to release 1.0.0 soon, it would be awesome to have svelte support included. If you can't work on that for now, I'd consider merging a slight variation of the current status, which has everything except cross-language reference finding

screenfluent:
Hi there, I was hoping to get back to you with good news, but honestly I don't know if I make it. I made some attempts but ended up with some hell. 😅 I want to try one last time.

MischaPanch:
Let me then take what you've already done here, maybe push some minor commits on top, and merge preliminary svelte support without cross-language reference finding. If you find the capacity to try again later, you can build off that. Would that work for you?

screenfluent:
I'm hoping to finish today

screenfluent:
I overestimated how fast I can deal with refactoring the monolith I ended up with... 😅 the pressure of your release isn't helping, so I just do the work and it will be ready when I finish. 🤷‍♂️

MischaPanch:
Np, we'll won't release before mid next week in the end ;)

screenfluent:
Unfortunately I have to give up on this. 😵‍💫 Feel free to use the initial version without TS cross-linking. I was close to having it working, but I don't have time for a tool that I needed for working on my projects, where it turned out that I work on the tool exclusively, not the projects... 😅

ketema:
Hey @MischaPanch - I saw your comment on #707 about Svelte integration. After analyzing #339 and #504, I think I can   help unblock this.                                                                                                     

**The Problem**: Finding TypeScript symbols referenced in Svelte `<script>` blocks.

**Why #339/#504 Failed**: Both attempted proxy LSP architecture (intercepting and merging LSP responses). This creates a monolithic wrapper that can route to ONE LSP at a time but can't merge results from BOTH.

**Proposed Solution (Tree-Sitter Approach)**:
1. **Svelte ↔ TypeScript ONLY** (not a generic FFI framework yet)
2. **Sidecar architecture**: Build parallel index that enhances (not replaces) existing LSP queries
3. **Tree-sitter parsing**: Semantic parsing handles aliases/re-exports/destructuring
4. **File-based cache**: Store index in `.serena/cache/svelte_ts_index.json` (no database dependency)
5. **Progressive indexing**: Parse open files immediately, background-scan workspace (don't block startup)
6. **Graceful degradation**: If tracker unavailable/fails, fall back to LSP-only mode (no breaking changes)

**Why This is Different from #339/#504**:
- Don't build a proxy LSP (too complex, as @screenfluent discovered)
- Don't intercept all LSP calls (coordinate queries instead)
- Use tree-sitter for semantic parsing (industry standard: JetBrains uses PSI trees, VSCode uses TypeScript service +  svelte2tsx)
- Modular integration as optional enhancement to existing LSP queries

**Addressing #168's Core Insight**:
You closed #168 noting that "isolated LS for multi-language projects will never work really well" because "one would   not be able to find references from one language for another."

**This is exactly the problem I'm proposing to solve.** The tree-sitter approach builds a cross-language symbol index  that connects TypeScript exports to Svelte imports WITHOUT the complexity of a MultiLanguageServer wrapper.

Your IDE plugin strategy makes sense for professional developers, but leaves the open-source MCP community (Claude     Desktop, Claude Code, Cursor, Copilot, Augment.ai) without a solution. The tree-sitter tracker can serve both: provide cross-language intelligence for MCP users NOW, and potentially be reused by your IDE plugin later.

**Approach is not new**:
- Feasibility confirmed by industry precedent (JetBrains PSI, VSCode TypeScript service)
- Known limitations documented upfront: grammar lag, index staleness, platform builds

**Broader Impact**:
If this works for Svelte, same pattern applies to:
- Vue ↔ TypeScript, Next.js, Astro
- Rust ↔ Python (PyO3), Rust ↔ JavaScript (wasm-bindgen)
- Protocol Buffers → multiple languages

**Why Robust > Simple for AI-Enhanced Workflows**:
The comparison of LSPManager (400 LOC) vs LanguageServerManager (87 LOC) in #703/#704 reflects a deeper philosophical  shift: In budding AI-enhanced development, robustness with graceful degradation is more valuable than simplicity in my opinion.

I can use AI to systematically verify 400 LOC of error handling more reliably than I can debug a crashed 87 LOC        implementation that disrupts my interactive workflow. When (not if, because as you yorself noted they are 'flaky') an  LSP fails mid-session, graceful degradation (continuing with working LSPs) maintains smooth AI-assisted development    flow. This isn't premature complexity—it's optimizing for the bottleneck that actually matters in human-AI             collaboration.

The tree-sitter approach follows this philosophy: sidecar architecture with fallback ensures users never experience    breaking changes, even if the tracker encounters edge cases.

**Proposed Next Steps**:
1. I will build a standalone TreeSitterReferenceTracker module with unit tests (no Serena integration yet)
2. Validate the approach with real Svelte/TS project (e.g., SvelteKit starter)
3. Benchmark and measure actual performance characteristics
4. If viable, integrate with LSPManager and open PR with documented limitations and fallback behavior

**Would you be open to this approach?** I want to validate it aligns with the project philosophy before investing      significant time.

opcode81:
@ketema the core assumption underlying your proposal is that Tree-sitter can indeed resolve references. As far as I am aware, Tree-sitter performs no name resolution whatsoever. Therefore, it will not know what is indeed a reference to a symbol definition. It would be highly problematic to assume that an occurence of an unqualified name in some local scope is definitely a reference to a symbol of the same name. There is a reason why language servers are complex beasts.

I could be wrong though, so if I am mistaken, it would be great if you could provide a tech demo (repo) where 
 * there is a mix of languages,
 * (unqualified) symbol names are not unique,
 * yet (cross-language) references can still be resolved correctly using Tree-Sitter.

> Why Robust > Simple for AI-Enhanced Workflows:
> The comparison of LSPManager (400 LOC) vs LanguageServerManager (87 LOC) in https://github.com/oraios/serena/pull/703/https://github.com/oraios/serena/pull/704 reflects a deeper philosophical shift:
> In budding AI-enhanced development, robustness with graceful degradation is more valuable than simplicity in my opinion.

This is a completely **false dichotomy**. Importantly, our solution could easily implement graceful degradation by changing just a few lines. We deliberately chose not to, because we prefer "fail fast" over "graceful degradation" in this particular case.
For instance, if a user wants to use a language server because he deems the language support essential, why would it be a good thing for Serena to completely ignore this (and pretend a language server failure is not a problem)? The LLM would see the errors (or perhaps not) and choose alternative solution paths (which could be hugely detrimental, less efficient and more error-prone) while the user remains unaware of the problem. I just fundamentally don't think that this would be a good thing. The user can instead make a deliberate decision to either (a) make the language server work or (b) disable the language server. The assumption that a user definitely always wants (b), which is what graceful degradation dictates, is unfounded. Whether the behaviour indeed qualifies as "graceful" in this case is certainly debatable.

The term **robustness** is multifaceted. It’s not only about resilience to runtime failures (by glossing over them in the spirit of more or less "graceful degradation"), but also about resilience to undesirable effects, which is precisely what "fail fast" optimizes for.

Also, the relevant property is **elegance** rather than simplicity.

MischaPanch:
Let's summarize it like this: 

@ketema if you want to build a PR or a technical demo for how cross-language referencing will work, we will be happy to have a look at it and can see how we can incorporate it in Serena.

We won't tell you how to develop the code, you can do it as you like.

However, for this particular project that is controlled by @opcode81 and me, we follow the paradigm of what is classically known as "good code". We firmly believe that this is one of the key components behind Serena's popularity. So, any code merged here has to follow that paradigm.

If your demo/PR doesn't follow it, it can still be very useful as a starting point, finished in further iterations either by you, by us, in collaboration, or by other contributors. But we will never merge a huge change if similar goals can be achieved by what would classically be known as "better code".

Now for svelte, there are other PRs, in particular one that uses a typescript codebase that forwards references from typescript to svelte. We didn't merge it since it's an external dependency, and hence unacceptable for this project which already has many users. But it could be a good starting point. 

We haven't experimented with tree sitter for code analysis but I'm aware that it's a rather powerful tool. If it can be used to serve the goal, great, we can add it to the technologies we rely on :)

ketema:
I dug into this and realized I was completely wrong. This is not even a problem.  Got too excited. Apologies.

Issue #186, opened 2025-06-19
Make use of diagnostics methods in LS


MischaPanch:
Should be easy and really help, get diagnostics after every edit. Don't accept edits that create syntax errors

Maybe make tool "get problems in code" like in IDEs.

Since this is central functionality for LS, should not require much/any hacking

dsyme:
@MischaPanch I'm kind of interested in adding this if #660 lands OK to your liking.

opcode81:
Diagnostics support was added in v1.3.0

PR #347, opened 2025-07-24
Use text_document_diagnostic

bingbing-sagittal:
i want to use text_document_diagnostic so added this.


MischaPanch:
Are you using serena/solidlsp as library?

The current change doesn't affect the application layer. Do you plan adding diagnostics to the SerenaAgent later on?

src/solidlsp/ls.py
```diff
@@ -709,6 +709,54 @@ def request_references(self, relative_file_path: str, line: int, column: int) ->
...
+                }
+            )
+
+        if response is None:
+            return []
+
+        assert isinstance(response, dict), f"Unexpected response from Language Server (expected list, got {type(response)}): {response}"
+        ret: list[ls_types.Diagnostic] = []
+        for item in response["items"]:
+            assert isinstance(item, dict), f"Unexpected response from Language Server (expected dict, got {type(item)}): {item}"
```
MischaPanch: nitpick - for performance reasons you can just construct it like this:

```python
diagnostic = ls_types.Diagnostic(
    severity=item["severity"]
    ....
)
```

It's also cleaner code like this. If something is missing, the raised KeyError is informative enough. Moreover, some non-optional fields are missing, like `uri`

src/solidlsp/ls_types.py
```diff
@@ -317,3 +317,19 @@ class Hover(TypedDict):
...
     visualize the hover, e.g. by changing the background color. """
+
+class Diagnostic(TypedDict):
+    """Diagnostic information for a text document."""
+
+    uri: DocumentUri
+    """ The URI of the text document to which the diagnostics apply. """
+    range: Range
+    """ The range of the text document to which the diagnostics apply. """
+    severity: NotRequired[int]
```
MischaPanch: Shouldn't it be an IntEnum rather than an int? We should have some interpretation of the severity level backed in (through the enum name)

bingbing-sagittal:
> Are you using serena/solidlsp as library?
> 
> The current change doesn't affect the application layer. Do you plan adding diagnostics to the SerenaAgent later on?

I'm currently using `serena/solidlsp` as a library.
And yes, I do think diagnostics is a valuable information for the agent.

MischaPanch:
I agree, we wanted to add them for a while and had codex spin up a draft in #186 .

We can merge this now, without extension of the Agent, but would you like to try adding the diagnostics feature to our editing tools?

Just out of curiosity - in which context are you using solidlsp?

src/solidlsp/ls_types.py
```diff
@@ -317,3 +317,19 @@ class Hover(TypedDict):
...
     visualize the hover, e.g. by changing the background color. """
+
+class Diagnostic(TypedDict):
+    """Diagnostic information for a text document."""
+
+    uri: DocumentUri
+    """ The URI of the text document to which the diagnostics apply. """
+    range: Range
+    """ The range of the text document to which the diagnostics apply. """
+    severity: NotRequired[IntEnum]
```
MischaPanch: You misunderstood me. I mean we should have a class like

```python

class DiagnosticsSeverity(IntEnum):
   severe = 1
    ....
```

I don't know the actual semantics of the numbers, pls add the right ones.

Then you can use

```
severity: NotRequired[DiagnosticsSeverity]
```

bingbing-sagittal:
> I agree, we wanted to add them for a while and had codex spin up a draft in #186 .
> 
> We can merge this now, without extension of the Agent, but would you like to try adding the diagnostics feature to our editing tools?
> 
> Just out of curiosity - in which context are you using solidlsp?

I'm doing some experiments that ask LLM to fix errors in LLM generated code, then came across this repo.
And yes, would love to explore more on your editing tools.

MischaPanch:
Cool, I'll merge this now, feel free to do a follow-up PR that includes diagnostics into serena's application layer. The logic should be:

check diagnostics prior to an edit operation, perform the edit, let the editing tool return any new problems caused by the edit.

Several (all?) language servers publish diagnostics by themselves, not sure whether we can use that by subscribing to some messages, or whether it makes sense to do that even if it's technically possible.

MischaPanch (approved):

PR #349, opened 2025-07-24
Improved performance of find_all_non_ignored_files 

MischaPanch:
@opcode81 pls have a look, if I'm not missing anything, this should have always been like that (which is why I'm suspicious about missing something)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 Status of the `main` branch. Changes prior to the next official version change will appear here.
 
+Fixes:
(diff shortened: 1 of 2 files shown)


MischaPanch:
I verified that it gives the same result as on main. The improvement in performance is drastic, moreover, we were scanning `.git` in the version on main. Merging this now, as it will improve life for all users. We can do a post-merge review if I did miss something

opcode81:
> if I'm not missing anything, this should have always been like that

@MischaPanch Agreed!

PR #350, opened 2025-07-24
LS: use home directory for language server installations

MischaPanch:
1. Added a migration logic that will omit unnecessary installation
2. Used the correct LS installation path for clojure
3. Refactored the elixir LS to match the structure of existing LS


MischaPanch:
@mdbenito this will affect you a tiny bit since the closure LS will be redownloaded (clojure was not using the default resources dir before that, so the migration won't work for it). Next startup of serena may be a minute longer

src/serena/project.py
```diff
@@ -271,4 +273,5 @@ def create_language_server(
             ls_logger,
             self.project_root,
             timeout=ls_timeout,
+            solidlsp_settings=SolidLSPSettings(solidlsp_dir=SERENA_MANAGED_DIR_IN_HOME),
```
opcode81, line 276: No subdirectory?

src/solidlsp/settings.py
```diff
@@ -4,21 +4,17 @@
...
-        global_cache_dir = os.path.join(str(pathlib.Path.home()), ".multilspy", "global_cache")
-        os.makedirs(global_cache_dir, exist_ok=True)
-        return global_cache_dir
+    def __post_init__(self):
+        os.makedirs(str(self.solidlsp_dir), exist_ok=True)
+        os.makedirs(str(self.ls_resources_dir), exist_ok=True)
+
+    @property
+    def ls_resources_dir(self):
+        return os.path.join(str(self.solidlsp_dir), "language_servers", "static")
```
opcode81, line 20: Is there still a point to the "static" subdirectory? Seems redundant now.

MischaPanch: Maybe eventually we'll have "dynamic" files (like caches) saved to .serena/language_servers ^^

src/serena/project.py
MischaPanch: It will be saved to <solidlsp_dir>/language_servers, so a further subdir is not needed IMO

src/solidlsp/settings.py
opcode81: @MischaPanch I don't think so. It serves no purpose whatsoever. Language servers already save dynamic information to their "static" folder, e.g. logs, etc.

src/serena/project.py
opcode81: It is not designed to contain a single folder only (otherwise the `language_servers` folder is actually redundant), so we will now have solidlsp data and serena data mixed, which I don't think is necessarily the cleanest solution.
If it were true that the `language_servers` folder will forever be the only thing in that folder, then we should remove that folder (but it won't be the case, I think).

opcode81:
@MischaPanch why merge before we could clarify the best folder structure?
It is not good to change this once more after it is merged ...

MischaPanch:
I saw you approved the PR, so I didn't think it was important to you. Let's quickly chat tomorrow and adjust it, if you want to

opcode81:
> I saw you approved the PR, so I didn't think it was important to you. Let's quickly chat tomorrow and adjust it, if you want to

Your interpretation was perfectly correct: It isn't important to me, and by approving, I did signal that it was basically fine to merge "as is". But I also thought that if the answer was really that "static" serves no purpose, that you would opt to remove it. Oh well... :-)

Issue #285, opened 2025-07-06
Can I have a manual LSP mode?


LeeroyDing:
I'd like to run a custom LSP provided as a binary - is there a way to configure that?

MischaPanch:
There's no such feature yet, but it could likely be done without too much effort. One could write a CustomLanguageServer that  takes as input a command starting the language server on stdio.

Do you want to try implementing that? 

Otherwise we will come back to this later, as there are currently higher priority issues that need to be worked on

LeeroyDing:
I can't find much documentation but I'll give it a go.

MischaPanch:
There isn't much documentation since it's a new case, not previously seen.

Have a look at the pyright ls, which is one of the simplest LS implementations, for inspiration. For configuration, check out SerenaConfig. The project language discovery happens in the `ProjectConfig.autogenerate`, you will have to bypass it or ignore the languages identified there (making sure it doesn't crash b/c of an unknown language in case the ls startup command is configured by the user)

The correct LS is identified from the `Language` Enum.

You will likely have to adjust them all, or you bypass them by handling a special case where the ls startup command is configured in `SerenaConfig` (may be the easier route).

If you use serena within some client for understanding the serena repo and crafting a plan, you should be able to go pretty far :). We had good experiences of contributions created in this way

@opcode81 FYI, maybe you also have some words of advice

LeeroyDing:
# **Software Design Document: Custom Language Server Support**

**1. Introduction**

*   **1.1. Purpose:** This document outlines the design for adding support for custom language servers to Serena. This will allow users to integrate language servers for languages not natively supported by Serena by providing a simple command to start the server.
*   **1.2. Scope:** The scope of this project is to implement the necessary changes to allow users to configure and use a custom language server within Serena. This includes changes to the configuration, the language server management, and the project setup.
*   **1.3. Definitions, Acronyms, and Abbreviations:**
    *   **LS:** Language Server
    *   **LSP:** Language Server Protocol
    *   **SDD:** Software Design Document
*   **1.4. References:**
    *   GitHub Issue #285: Support for custom language servers

**2. System Overview**

*   **2.1. System Context:** Serena currently supports a fixed set of language servers. This feature will extend Serena to allow users to specify a command to run any LSP-compliant language server. This will enable support for a wider range of languages and tools.
*   **2.2. Use Cases:**
    *   A user wants to use a language server for a language not currently supported by Serena (e.g., Swift, OCaml).
[...]

PR #361, opened 2025-07-26
feat: Add support for custom language servers

LeeroyDing:
Implementation of #285

### DIFF
--- a/src/serena/config/serena_config.py
+++ b/src/serena/config/serena_config.py
@@ -192,7 +192,11 @@ class ProjectConfig(ToolInclusionDefinition, ToStringMixin):
             raise FileNotFoundError(f"Project root not found: {project_root}")
         with LogTime("Project configuration auto-generation", logger=log):
             project_name = project_name or project_root.name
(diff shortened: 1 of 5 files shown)


LeeroyDing:
NOT manually tested yet!

MischaPanch:
It won't exactly work like that since initialization_options are not static - they need the process_id and the repo_uri. Also the checking for custom_ls.yml is not appropriate (there may be several custom LS)

If you don't need it for now, we can leave the PR open and later build on it. I think a better way to allow users to start custom LS is by allowing them to make an actual python implementation of an LS using whatever startup command they want (i.e. a python module, not a json), and let serena pick it up automatically. So something like a command `register_ls` which modifies the list of supported languages and the LS creation, and the user can provide a path to that command which is executed when the server is spun up. Something like that - mainly that users write python and not json for getting custom ls support.

LeeroyDing:
I did find a use for it, will take your suggestion and try it at some point in the next couple of weeks

opcode81:
Stale

PR #364, opened 2025-07-27
docs(readme): fix wrong sse url

tiennm99:


MischaPanch:
Wanted to fix that forever and kept forgetting it, thanks!

PR #369, opened 2025-07-27
fix: TypeScript, Intelephense server initialization on Windows under Claude Code (closes #368)

### LINKED ISSUE #368: TypeScript Language Server doesn't work on Windows with Claude Code
1. Install Serena MCP with the command: `claude mcp add serena -- uvx --from git+https://github.com/oraios/serena serena-mcp-server --context ide-assistant --project  Some/Path/To/The/Project`
2. Index the project with `uvx --from git+https://github.com/oraios/serena index-project` command
3. Start Claude Code
4. In the logs there are no messages after: `INFO  2025-07-27 13:50:44,822 [SerenaAgentExecutor_0] solidlsp.language_servers.common:_run_command:82 - Running command 'npm install --prefix ./ typescript@5.5.4' in 'C:\Users\username\AppData\Local\uv\cache\archive-v0\HdnSyCVzKUWPypyn2inj2\Lib\site-packages\solidlsp\language_servers\static\TypeScriptLanguageServer\ts-lsp'` and the mentioned folder is empty. It is clear that installation never finishes. So it is unable to install even a server under the Claude Code on Windows. I tried to Serena in MCP Inspector and all works perfectly fine.
5. Shutdown Serena Server, shutdown Claude Code. 
6. Install manually `typescript@5.5.4`, `typescript-language-server@4.3.3` in the folder mentioned before.
7. After installation succeeds run again Claude Code.
8. See the error log with the error:
```
INFO  2025-07-27 13:57:40,972 [SerenaAgentExecutor_0] serena.agent:start:324 - Language server initialization starting ...
INFO  2025-07-27 13:57:40,972 [SerenaAgentExecutor_0] serena.project:create_language_server:268 - Creating language server instance for D:\dev\copilot-review.
INFO  2025-07-27 13:57:41,022 [MainThread] serena.mcp:server_lifespan:213 - MCP server lifetime setup complete
INFO  2025-07-27 13:57:41,036 [MainThread] mcp.server.lowlevel.server:_handle_request:625 - Processing request of type ListToolsRequest
INFO  2025-07-27 13:57:41,039 [MainThread] mcp.server.lowlevel.server:_handle_request:625 - Processing request of type ListPromptsRequest
INFO  2025-07-27 13:57:41,041 [MainThread] mcp.server.lowlevel.server:_handle_request:625 - Processing request of type ListResourcesRequest
INFO  2025-07-27 13:57:41,655 [SerenaAgentExecutor_0] solidlsp:load_cache:1610 - Loading document symbols cache from D:\dev\project-name\.serena\cache\typescript\document_symbols_cache_v23-06-25.pkl
INFO  2025-07-27 13:57:41,773 [SerenaAgentExecutor_0] solidlsp:load_cache:1614 - Loaded 169 document symbols from cache.
INFO  2025-07-27 13:57:41,780 [SerenaAgentExecutor_0] serena.agent:reset_language_server:542 - Starting the language server for Project
INFO  2025-07-27 13:57:41,785 [SerenaAgentExecutor_0] solidlsp:start:1656 - Starting language server with language typescript for D:\dev\project-name
INFO  2025-07-27 13:57:41,789 [SerenaAgentExecutor_0] solidlsp:_start_server:219 - Starting TypeScript server process
INFO  2025-07-27 13:57:41,789 [SerenaAgentExecutor_0] solidlsp.ls_handler:start:157 - Starting language server process via command: C:\Users\username\AppData\Local\uv\cache\archive-v0\HdnSyCVzKUWPypyn2inj2\Lib\site-packages\solidlsp\language_servers\static\TypeScriptLanguageServer\ts-lsp\node_modules\.bin\typescript-language-server --stdio
INFO  2025-07-27 13:57:41,797 [SerenaAgentExecutor_0] solidlsp:_start_server:223 - Sending initialize request from LSP client to LSP server and awaiting response
INFO  2025-07-27 13:57:41,843 [SerenaAgentExecutor_0] solidlsp:logging_fn:263 - LSP: client -> logger: {"jsonrpc": "2.0", "method": "initialize", "id": 1, "params": {"locale": "en", "capabiliti...
INFO  2025-07-27 13:57:41,851 [SerenaAgentExecutor_0] solidlsp:logging_fn:263 - LSP: client -> logger: Waiting for response to request initialize with params: {"locale": "en", "capabilities": {...
INFO  2025-07-27 13:57:41,859 [LSP-stderr-reader] solidlsp:logging_fn:263 - LSP: client -> logger: LSP stderr: /usr/bin/bash: Files\Git\bin\bash.exe: No such file or directory ...
```
After investigation it looks like Claude Code starts the MCP under the bash on Windows, which is found in the Git directory, and `typescript-language-server` is just basic bash script, which runs a node with js script as an argument, anyway here are my observations:
1. The issue happens because of the parameter `shell=True`, but if we disable this parameter, it doesn't fix the issue, because we still trying to run the regular script, not the executable. And for some reason it is unable to find the path. Here is the log, when I added exception handling in the `_run_command` in `common.py`
```
ERROR 2025-07-27 14:11:49,370 [SerenaAgentExecutor_0] solidlsp:_run_command:98 - Command "npm install --prefix ./ typescript@5.5.4" failed with error: [WinError 2] The system cannot find the file specified
ERROR 2025-07-27 14:11:49,383 [SerenaAgentExecutor_0] solidlsp:_run_command:98 - Command "npm install --prefix ./ typescript-language-server@4.3.3" failed with error: [WinError 2] The system cannot find the file specified
```

I propose the fix where we will run the node executable and will pass there exact script it have to run

auric:
Implement direct node execution for TypeScript, Intelephense language servers

- Add `CommandUtils` class in `common.py` for cross-platform Node.js npm script resolution
- Replace `shell=True` `npm` commands with direct node + `npm-cli.js` execution
- Update `TypeScript`, `VTS`, and `Intelephense` language servers to use direct `node` execution
- Convert all `RuntimeDependency` commands to use `list[str]` instead of `str | list[str]`
- Add `command_shell` flag to `RuntimeDependency` for granular shell control

Fixes #368


src/solidlsp/ls_handler.py
```diff
@@ -150,10 +150,14 @@ def start(self) -> None:
...
         is_windows = platform.system() == "Windows"
-        if not isinstance(cmd, str) and not is_windows:
+        if not isinstance(cmd, str):
             # Since we are using the shell, we need to convert the command list to a single string
-            # on Linux/macOS
-            cmd = " ".join(cmd)
+            # on all platforms.
+            if is_windows:
+                # On Windows, it's important to quote arguments correctly for the shell.
+                cmd = subprocess.list2cmdline(cmd)
```
MischaPanch: can't we always do this, or does this cause problems on non-windows?

src/solidlsp/language_servers/common.py
```diff
@@ -63,39 +63,57 @@ def install(self, logger: LanguageServerLogger, target_dir: str) -> dict[str, st
...
-            stdout=subprocess.PIPE,
-            stderr=subprocess.STDOUT,
-            **kwargs,
-        )
-        if completed_process.returncode != 0:
-            log.warning("Command '%s' failed with return code %d", command, completed_process.returncode)
-            log.warning("Command output:\n%s", completed_process.stdout)
+
+        if isinstance(command, list):
+            command_parts = command
```
MischaPanch: let's keep simple interfaces. If you need a list for solving this issue, the interface should be changed to `command: list[str]` and the callers should be responsible for creating the list.

src/solidlsp/language_servers/common.py
```diff
@@ -104,3 +122,91 @@ def _install_from_url(dep: RuntimeDependency, logger: LanguageServerLogger, targ
...
+    @staticmethod
+    def find_npm_executable() -> str | None:
+        """
+        Find the path to the npm executable.
+        Returns the full path to npm executable or None if not found.
+        """
+        return shutil.which("npm")
+
+    @staticmethod
+    def get_npm_cli_script_path() -> str | None:
```
MischaPanch: I don't see a need for this class, all functionality can be in a top level function

`def get_npm_path`

If we want to have a class serving as namespace for this, we should rather have 

```python

class CommandUitls:
   # the only method for now
    def get_npm_path

```

Public methods form fart of the interface, and should be kept to a minimum. 

Everything else that this class is doing should be done by the caller, or within that method (no point in wrapping shutil.which in a separate method which is not required as parto of an interface)

MischaPanch (changes requested):

MischaPanch:
@auric I just did a very superficial review to give some early feedback, gonna review in more detail once the tests pass. Thanks for your contribution!

src/solidlsp/language_servers/common.py
auric: replaced with `CommandUtils`

src/solidlsp/language_servers/common.py
auric: done

src/solidlsp/ls_handler.py
auric: actually this function was needed only for Windows, but it is undocumented, so I deleted this call

MischaPanch:
@auric given that you seem to have difficulties finding the command on Unix - can't you use the code for finding npm that's currently on main? It seems to work without problems

opcode81:
> @auric given that you seem to have difficulties finding the command on Unix - can't you use the code for finding npm that's currently on main? It seems to work without problems

@MischaPanch the code on main does not look for the location of `npm` anywhere; it simply uses an execution mechanism which resolves the location automatically based on the PATH (`shell=True`).
Note that on Windows, using plain `npm` will also work, even though it is definitely not the name of an executable (`.exe`) when run in a shell that can resolve this based on PATH and PATHEXT.

@auric so why do we need to find the location of `npm`? Why isn't resolution by the shell sufficient?
Also, please note that as far as quoting arguments (that may contain spaces) is concerned: On Windows, it is possible to pass a list of arguments to Popen, even when using `shell=True` (but not on other platforms). This is made use of, in particular, in `SolidLanguageServerHandler.start`, where `ProcessLaunchInfo.cmd` can be a list or string. Perhaps a similar mechanism should be used for runtime dependency installation. Note that this is helpful because Windows is the only platform where spaces in paths are common.

auric:
@opcode81 unfortunately `shell=True` approach doesn't work on Windows with Claude Code, please read the #368 for more details, when run under Claude Code installation doesn't work, I tried to install it without Serena, but then starting of the server doesn't work, there is an error: `LSP stderr: /usr/bin/bash: Files\Git\bin\bash.exe: No such file or directory ...`

opcode81:
@auric see my comments in the issue.

auric:
@MischaPanch @opcode81 I believe that on the `main` previously tests didn't fail because there was no Node installation in the `github/workflows` and the error during packages installation was silent, there was no checks of the return value during installation.
I reverted the installation process to the regular `npm install` on Linux and MacOS and tests still fail with the error: `ERROR test/solidlsp/typescript/test_typescript_basic.py::TestTypescriptLanguageServer::test_find_referencing_symbols[typescript] - subprocess.CalledProcessError: Command '['npm', 'install', '--prefix', './', 'typescript@5.5.4', 'typescript-language-server@4.3.3']' returned non-zero exit status 1.`
I don't know what to do

MischaPanch:
We'll look into it when we find time (hopefully soon) and will build on this PR to close the issue

opcode81:
I currently don't think there actually is an issue - at least not with the changes in branch `fix/typescript-ls-commands` in place. (see #368)

> I believe that on the main previously tests didn't fail because there was no Node installation in the github/workflows and the error during packages installation was silent, there was no checks of the return value during installation.

There was no issue in the tests previously. The installation worked. Otherwise, the subsequent tests using the TypeScript language server would never have worked ...

src/solidlsp/language_servers/common.py
```diff
@@ -2,14 +2,16 @@
...
 import subprocess
 from collections.abc import Sequence
 from dataclasses import dataclass
+from pathlib import Path
 
 from solidlsp.ls_logger import LanguageServerLogger
 from solidlsp.ls_utils import FileUtils, PlatformUtils
 
-log = logging.getLogger(__name__)
+logger = logging.getLogger(__name__)
```
auric, line 14: I should delete this line of code, renamed it accidentally

auric:
@opcode81 
> There was no issue in the tests previously. The installation worked. Otherwise, the subsequent tests using the TypeScript language server would never have worked ...

but there was no Node installed previously, I added its installation into the workflow with this PR, previously tests didn't fail because `subprocess.run` failed silently, because there was `check=True` but was no exception handling, exception was eaten elsewhere
image

opcode81:
@auric No, npm was always present. Otherwise, the TypeScript tests would never have worked. But they did work. Note that the tests use the TypeScript language server in order to obtain symbolic information, and there is no way for the server to be present if it doesn't get installed appropriately.

MischaPanch:
This can be closed, right?

PR #370, opened 2025-07-27
Make mypy run through on solidlsp

MischaPanch:


MischaPanch:
@opcode81 just have a quick glance pls. No need to review, just informing you that I did this chore.
Resolved something like 250 mypy failures on solidlsp and included it into poe type-check.

The reason is that mypy helps Claude in performing refactorings in solidlsp without running tests. The last refactoring where SolidLSPSettings were added would have been much smoother with this. Claude did manage to make it right in the end, but it took forever and consumed tons of tokens.

It might add minimal value for us, or some minor annoyances (as usual with mypy), but IMO worth the price for having automated and fast feedback for coding agents working on serena.

opcode81:
After merging main and fixing a few issues, some still remain (pertaining to diagnostics).

MischaPanch:
@jetbrains-junie

Resolve conflicts with main and finish fixing typing errors. You can generously apply type: ignore where needed.
Make sure `uv run poe type-check` and `uv run poe test` run through.

opcode81:
We should finally get this merged. It was a lot of work to update this and make it pass. I ignored some new language server modules completely, because there were just too many "errors".

MischaPanch:
Yes, let's merge this. We can merge now, right? I'll look into what's going on with tests on CI, three different failures on the different systems. For Elixir at least there were recent changes that might be causing the failures...

MischaPanch:
I can't reproduce the failing/hanging tests on Ubuntu (tried on WSL)

.github/workflows/pytest.yml
```diff
@@ -38,6 +38,30 @@ jobs:
...
+          fi
+      - name: Install dependencies
+        shell: bash
+        run: uv pip install -e ".[dev]"
+      - name: Check formatting
+        shell: bash
+        run: uv run poe lint
+      - name: Type-checking with mypy
+        shell: bash
+        run: uv run poe type-check
```
opcode81: I deliberately did this last, because my goal was to get the greatest amount of (useful) information, not to fail early.
I care less about a linting/mypy failure than I do about the test results, so I want the test results first (and I want to get them even if there's a type annotation missing that nobody really would have needed, notably).

MischaPanch: Moved

PR #371, opened 2025-07-27
Feature/initial instructions without tool

### LINKED ISSUE #366: Initial instructions _can_ be automatically provided on server initialization
No need for a special `InitialInstructionsTool` for that, theoretically.

## Claude Code changelog

> v1.0.52
>
>    Added support for MCP server instructions

## MCP spec

[Lifecycle - Model Context Protocol](https://modelcontextprotocol.io/specification/2025-06-18/basic/lifecycle#initialization)

> The server **MUST** respond with its own capabilities and information:
> 
> 
>     {
>       "jsonrpc": "2.0",
>       "id": 1,
>       "result": {
>         "protocolVersion": "2024-11-05",
>         "capabilities": {
>            ...
>         },
>         "serverInfo": {
>           ...
>         },
>         "instructions": "Optional instructions for the client"
>       }
>     }

MischaPanch:
Closes #366 

This makes the life of many users easier - serena works automatically with subagents, new sessions, after compactifying and so on.

We could remove the initial instructions tool or disable it in our contexts. I tested with claude code, it works there, but I'm not sure how it is in other clients.

We could:

1. disable the tool in ide-assistant context, rollout automatically to the users. If it works with claude desktop as well, also disable it in that context.
2. merge this without disabling the tool, asking the users to modify their serena configs in the readme (few will notice it though)
3. completely remove the tool, hoping that all relevant clients will read the instructions from the mcp server
4. A specialized mode that disables this tool (not a fan of this, but technically works) 

If we leave the tool enabled, claude code will use it, reading the instructions twice and wasting tokens.
I have also not tested extensively whether it behaves well when receiving the instructions through the MCP server itself, but from testing a bit, it seemed well.

@jaens @opcode81 WDYT? I like options 1 or 3, but it may negatively affect users who use clients that don't actually read the instructions (not sure whether such clients currently exist). We can also say that even if such clients exist, we don't care about these and that users should wait for their client to update the handling of the MCP servers. If not happened already, it will happen soon, I guess


jaens:
I think in the long term it definitely makes sense to rely on the instructions field, as it's part of the spec and _technically_ not reading could be filed as a bug of the agent (and it's probably trivial to implement).

For me personally, I think the tool is currently _better_ in the sense that sometimes I don't want to apply the full instructions at all, which are quite aggressive and can theoretically interfere with some workflows (eg. _I_ don't care about the standard token wastage that much, I just want the best _engineering_ result).  
(I'd rather build custom instructions for specific workflows.)

opcode81:
It works with Claude Desktop also.
I think we should go with

5. Mark the tool with `ToolMarkerOptional`, such that it will be disabled by default and users must explicitly enable it to get it.

Given that the most important MCP clients can do without it, I think this would make sense. As @jaens has already said, other clients should also adhere to the standard and not doing so should be considered a bug. If such a bug cannot be resolved, users still have the option to enable the tool explicitly.

MischaPanch:
@opcode81
I implemented option 5 and I added support for customizing our prompts.

@jaens this means you will likely no longer need a custom command. You can override prompts by creating your own yaml files in your home/.serena directory.

I have extended the CLI to make this easier, you can for example do

```
<uv invocation> serena prompts create-override system_prompt
```

With this and the pre-existing ability to customize modes and contexts, essentially everything about serena becomes configurable.

MischaPanch:
We can consider also having per-project overrides by allowing prompt yamls in `<project>/.serena`, which would be easy to implement as a follow-up. But I wouldn't do that now

README.md
```diff
@@ -354,11 +354,14 @@ For example, when using `uvx`, you would run
...
 ```
 
-ℹ️ Serena comes with an instruction text, and Claude needs to read it to properly use Serena's tools. 
-  Once in Claude Code, you can ask to "read Serena's initial instructions" or run `/mcp__serena__initial_instructions` to load the instruction text. 
-  Do this whenever you start a new conversation and after any compacting operation to ensure Claude remains properly configured to use Serena's tools.
-
-ℹ️ **NEW**: an alternative to the above is adding the instructions as part of the system prompt, then you will not need to run the command above or to remember re-running it after compacting.
+ℹ️ Serena comes with an instruction text, and Claude needs to read it to properly use Serena's tools.
+  As of version `v1.0.52`, claude code reads the instructions of the MCP server, so this is handled automatically.
+  If you are using an older version, or if Claude fails to read the instructions, you can ask it explicitly
```
opcode81, line 357: These instructions no longer apply.

MischaPanch: We should comment that they have to enable the tool to do that

MischaPanch: Done

MischaPanch:
Merging this now since it's a serious quality of life improvement, we can correct things in a post-merge review if needed. Nothing here is really controversial, we might at most discuss some implementation details

opcode81:
> It works with Claude Desktop also.

I don't know why I came to this conclusion back then. It does not work in Claude Desktop, and we should probably re-enable the tool (disabling it only in `ide-assistant` mode, which mostly means "Claude Code mode" tbh).

Issue #385, opened 2025-08-01
Solargraph: can't activate json-2.7.1, already activated json-2.7.2 (Gem::LoadError)


dragonfax:
Can't launch solar graph.

Note that solargraph works fine locally, but serena is trying to do something weird. Launch a binary directly or something. I might work but the dependencies its injecting are broken.

Solargraph works fine locally.

```
% solargraph stdio
Solargraph is listening on stdio PID=39652
<CTRL-D>
% which solargraph
/Users/jstillwell/.rbenv/shims/solargraph
% ruby -v
ruby 3.3.6 (2024-11-05 revision 75015d4c1f) [arm64-darwin23]
% which ruby
/Users/jstillwell/.rbenv/shims/ruby
% cat .ruby-version
3.3.6
```

 But this is what serena is doing

```
% /Users/jstillwell/.rbenv/versions/3.3.6/lib/ruby/gems/3.3.0/gems/solargraph-0.51.1/bin/solargraph stdio
/Users/jstillwell/.rbenv/versions/3.3.6/lib/ruby/3.3.0/rubygems/specification.rb:2235:in `check_version_conflict': can't activate json-2.7.1, already activated json-2.7.2 (Gem::LoadError)
	from /Users/jstillwell/.rbenv/versions/3.3.6/lib/ruby/3.3.0/rubygems/specification.rb:1379:in `activate'
	from <internal:/Users/jstillwell/.rbenv/versions/3.3.6/lib/ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:107:in `each'
	from <internal:/Users/jstillwell/.rbenv/versions/3.3.6/lib/ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:107:in `block in require'
	from <internal:/Users/jstillwell/.rbenv/versions/3.3.6/lib/ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:39:in `synchronize'
	from <internal:/Users/jstillwell/.rbenv/versions/3.3.6/lib/ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:39:in `require'
	from /Users/jstillwell/.rbenv/versions/3.3.6/lib/ruby/3.3.0/json.rb:590:in `rescue in <module:JSON>'
	from /Users/jstillwell/.rbenv/versions/3.3.6/lib/ruby/3.3.0/json.rb:587:in `<module:JSON>'
[...]

PR #389, opened 2025-08-02
Solargraph (ruby): Use the framework provided by ruby gems to execute

dragonfax:
Don't subvert the gem system and try to run the binary directly.
Instead use the framework provided by ruby gems to execute the provide binary.

`gem exec` and similarly `bundle exec` gives the underlying package framework a chance to properly setup dependencies and the package env for the gem your trying to use.

Instigated by https://github.com/oraios/serena/issues/385


MischaPanch:
Looks good, thanks. Did it solve your problem?

dragonfax:
Definitely. Tested it a few different ways.

MischaPanch:
Great, happy to hear that!

Btw, all that's missing for Ruby to be "officially supported" in Serena is to add some tests on it, like for the other languages

PR #392, opened 2025-08-02
Connecting Serena to chatgpt

MischaPanch:
This basically works, but there are two restrictions

1. Chatgpt will accept a maximum of 30 tools, we have more.
2. Chatgpt only accepts short descriptions of tools, up to 300 chars. We have much longer ones

We can solve 1 easily by having a custom chatgpt context, but 2 is a problem. We could solve it by allowing tool description overrides in the context, something that we have discussed a while ago.

This feature is very sexy IMO and comes to us for free using mcpo. Especially witch GPT5 around the corner. It also allows people to use premium models for coding without more subscriptions.

@opcode81 pls have a look at the doc file (it's very short)

Rejected approaches:
- using mcpo as a library and adding a custom cli command to serena (not possible with mcpo)
- documenting sse mode. MCPO does support sse mode, but in this context it doesn't add a benefit, I believe, so I didn't mention it
- conflate with openwebui. Connecting to openwebui also uses MCPO, but runs locally (without cloudflared), so I'd write separate docs on this

### DIFF
--- a/docs/serena_on_chatgpt.md
+++ b/docs/serena_on_chatgpt.md
@@ -18,7 +18,7 @@ Run the following command to launch Serena as http server (assuming port 8000):
 ```bash
 uvx mcpo --port 8000 --api-key <YOUR_SECRET_KEY> -- \
   uvx --from git+https://github.com/oraios/serena \
-  serena start-mcp-server --project $(pwd)
+  serena start-mcp-server --context chatgpt --project $(pwd)
 ```
 
 - `--api-key` is required to secure the server.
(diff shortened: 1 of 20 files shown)


MischaPanch:
I now added tool description overrides, so the only thing missing is a custom chatgpt context with excluded tools and shortened descriptions

MischaPanch:
@OpenHands please fix the failing actions on PR #392

MischaPanch:
Ready for review, now works out of the box with the chatgpt context

PR #393, opened 2025-08-02
Allow system prompt to depend on available tools and tool categories …

opcode81:
…(as indicated by ToolMarkers)

* Introduce additional tool markers that represent relevant categories
* Apply some conditions to current system prompt

This PR applies it only to the system prompt, but similar mechanisms can straightforwardly be applied to modes and contexts.


src/serena/generated/generated_prompt_factory.py
```diff
@@ -1,36 +1,38 @@
...
-        return self._render_prompt("think_about_whether_you_are_done", locals())
-
-    def create_summarize_changes(self) -> str:
-        return self._render_prompt("summarize_changes", locals())
-
-    def create_prepare_for_new_conversation(self) -> str:
-        return self._render_prompt("prepare_for_new_conversation", locals())
-
-    def create_system_prompt(self, *, context_system_prompt: Any, mode_system_prompts: Any) -> str:
-        return self._render_prompt("system_prompt", locals())
```
MischaPanch, line 36: why is there a diff - line endings?

src/serena/agent.py
```diff
@@ -336,10 +337,23 @@ def get_active_modes(self) -> list[SerenaAgentMode]:
...
+                    available_markers.add(marker_class.__name__)
+
+        log.info("Generating system prompt with available_tools=(see exposed tools), available_markers=%s", available_markers)
+        system_prompt = self.prompt_factory.create_system_prompt(
             context_system_prompt=self._context.prompt,
             mode_system_prompts=[mode.prompt for mode in self._modes],
+            available_tools=available_tools,
+            available_markers=available_markers,
         )
+        log.info("System prompt:\n%s", system_prompt)
```
MischaPanch, line 370: This is fairly large, should we really always log it?

opcode81: Yes, we definitely should.
This replaces the previous initial instructions, where it was visible as the tool output.

src/serena/generated/generated_prompt_factory.py
opcode81: Diff is gone now. Yes, was line endings.

MischaPanch (approved):

PR #394, opened 2025-08-02
Docker: add compose.override.yaml for quick start setup

kght6123:
Using compose.override.yaml made setup a bit easier.

Added documentation and excluded it in .gitignore.

We're very impressed with the work you've done on the project and hope this PR helps.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -227,6 +227,3 @@ tmp/
 /test/resources/repos/elixir/test_repo/deps
 # Exception: Don't ignore Elixir test repository lib directory (contains source code)
 !/test/resources/repos/elixir/test_repo/lib
-
-# Docker Override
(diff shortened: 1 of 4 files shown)


MischaPanch:
Thanks for the compliment and the contribution! Since the compose.override is not a valid docker compose file, I propose to just include that snippet in the docker documentation and not add a separate file.

You could also add some commented lines in the actual docker compose yaml. Or you could document how the [compose merging mechanism](https://docs.docker.com/compose/how-tos/multiple-compose-files/merge/) could be used for this purpose and provide example files for that

kght6123:
@MischaPanch Thank you for your review. I've made some changes, what do you think?

MischaPanch:
That works, thanks! Merging

PR #398, opened 2025-08-03
Fix: Add HTML escaping to dashboard log messages

kazuki-hanai:
Escape special HTML characters in log messages to prevent XSS vulnerabilities and ensure proper display of code snippets containing HTML tags.

### DIFF
--- a/src/serena/resources/dashboard/dashboard.js
+++ b/src/serena/resources/dashboard/dashboard.js
@@ -1,5 +1,6 @@
 class LogMessage {
     constructor(message, toolNames) {
+        message = this.escapeHtml(message);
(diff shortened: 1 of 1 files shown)


MischaPanch:
Generally looks good, but is there no dedicated JS function for that? Do we have to write our own escape logic?

kazuki-hanai:
@MischaPanch Yes, I googled it for a while and found that there don't seem to be any standard functions for escaping HTML strings. We need to implement it manually. Alternatively, the most recommended approach is not to use dangerous rendering functions like `.html()`.

MischaPanch:
Ok, merging this now. If there are further issues, we can rethink using .html 

Thank you for the contribution

Issue #400, opened 2025-08-03
Adjust docs on agno


MischaPanch:
Agno now supports [MCP out of the box](https://docs.agno.com/tools/mcp/mcp), no more need for an adapter

We should adjust the docs to mention the adapter-based agno integration as a sidenote, or even in a separate document for "Serena as a Library".

Agno does not stand out, instead we should just mention that Serena also works with local agents/UIs providing agno, [jan](https://github.com/menloresearch/jan) and openwebui as example (the latter using mcpo).

PR #401, opened 2025-08-03
Update readme

MischaPanch:
Closes: #400

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,11 +6,13 @@
 * :rocket: Serena is a powerful **coding agent toolkit** capable of turning an LLM into a fully-featured agent that works **directly on your codebase**.
   Unlike most other tools, it is not tied to an LLM, framework or an interface, making it easy to use it in a variety of ways.
 * :wrench: Serena provides essential **semantic code retrieval and editing tools** that are akin to an IDE's capabilities, extracting code entities at the symbol level and exploiting relational structure. When combined with an existing coding agent, these tools greatly enhance (token) efficiency.
-* :free: Serena is **free & open-source**, enhancing the capabilities of LLMs you already have access to free of charge. 
+* :free: Serena is **free & open-source**, enhancing the capabilities of LLMs you already have access to free of charge.
 
 ### Demonstration 1 - Efficient Operation in Claude Code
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -169,6 +192,12 @@ IDE-based subscriptions (such as Windsurf or Cursor) that forced us to keep purc
 The substantial API costs incurred by tools like Claude Code, Cline, Aider and other API-based tools are similarly unattractive.
 We thus built Serena with the prospect of being able to cancel most other subscriptions.
 
+### Note
+
+We had built Serena before Claude Code was included in Antrhopic's pro tier. Now the cost saving aspect is less important,
+but Serena is still valuable as its tools are more powerful than those of Claude Code or other agents that we are aware of (at least for now), so many users
+find value in using Serena together with Claude Code or similar coding agents.
+
```
opcode81: I think we should remove the entire section "Free Coding Agents with Serena", as the key aspect is that it can turn any LLM into a competent coding agent - even free ones, so we've already said it above. This entire section is not really adding useful information and neither is the history of Claude Code not having been as "freely" available as it is now.

README.md
```diff
@@ -142,24 +147,42 @@ implementation.
...
+See the sections on [configuration](#configuration) and [modes and contexts](#modes-and-contexts) 
+for more information on how to adapt Serena to your needs,
+and explore the cli with `uvx --from git+https://github.com/oraios/serena serena --help`.
+
+### As Library
+While the lower level interfaces should not be considered as stable yet and have no extensive documentation,
+we do develop them with generality and ease of use in mind. 
+As a result, you can use Serena as a library to build your own agents using the provided tools
+and language server capabilities.
```
opcode81: This section should be very short and to the point; without any sub-headings. Adding so much information here is counter-productive and conflates usage modes (how?) with the purpose (what?). This section is supposed to be about the latter only (as the title suggests).

opcode81: I would also refrain from trying to mention "everything" "everywhere". We definitely don't need links to configuration or modes and contexts here, for example. This is overloaded.

It would be a bad idea for users to jump around the page using the links provided. The preferred way of reading the document should be in a mostly linear fashion..

opcode81: Actually, I think we can also delete this entire section ("What can I use Serena for?").
We've already said what Serena is above; The section does not really add anything.

README.md
MischaPanch: Agree, removed

README.md
MischaPanch: Agree, removed. I'll just briefly mention that serena can be used as library above.

PR #403, opened 2025-08-03
Fix MCP communication issues preventing client connections

blanham:
## Summary
  Fixes two critical issues preventing MCP clients (Claude Code, Cursor) from connecting to the Serena MCP server:

  - **TypeError on startup**: `'NoneType' object is not iterable` when parsing tool configurations
  - **Web dashboard breaking MCP**: Browser auto-opening interfered with stdin/stdout communication

  ## Issues Fixed

  ### 1. TypeError during server initialization
  **Problem**: When `excluded_tools` or `included_optional_tools` were `None` in configuration, the server crashed with
  `TypeError: 'NoneType' object is not iterable`.

  **Solution**: Added null checks with `or []` fallback to handle `None` values gracefully in `ToolSet.apply()`.

  ### 2. Web dashboard breaking MCP stdio communication
  **Problem**: When web dashboard was enabled, the browser auto-opening process used `multiprocessing.Process` with `os.dup2()`
  calls that interfered with the main process's stdin/stdout file descriptors, breaking MCP JSON-RPC communication.

  **Solution**: Replaced `multiprocessing.Process` with `subprocess.Popen` using completely isolated stdio streams (`DEVNULL`)
  to avoid file descriptor inheritance issues.

Okay, now human time. Neither Cursor nor Claude Code would see serena properly when the dashboard + opening a web browser occurred. I suspect this is actually the issue mentioned in serene_config.yml.

### DIFF
--- a/src/serena/agent.py
+++ b/src/serena/agent.py
@@ -188,11 +188,24 @@ class SerenaAgent:
             dashboard_url = f"http://127.0.0.1:{port}/dashboard/index.html"
             log.info("Serena web dashboard started at %s", dashboard_url)
             if self.serena_config.web_dashboard_open_on_launch:
-                # open the dashboard URL in the default web browser (using a separate process to control
-                # output redirection)
-                process = multiprocessing.Process(target=self._open_dashboard, args=(dashboard_url,))
-                process.start()
-                process.join(timeout=1)
+                # Fixed browser opening that doesn't interfere with MCP stdio
+                import subprocess
+                import webbrowser
+                
+                # Use subprocess instead of multiprocessing to avoid file descriptor inheritance issues
+                try:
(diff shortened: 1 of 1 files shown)


opcode81:
Neither of the issues you are solving is a known issue. Please describe the circumstances under which these issues are observed.

MischaPanch:
For the first issue, I suppose he configured the yaml with an empty entry instead of a list, which then was interpreted as None when loading to Python. That would be a user error IMO, the interface of the yaml expects a list.

The second one is reminiscent of a problem that we had already fixed (webbrowser command printing on stdout), is multiprocessing not already preventing any stdout contamination?

MischaPanch:
Closing, OP doesn't react and the PR doesn't seem to actually solve any problems

PR #405, opened 2025-08-04
Fix: Improve Solargraph compatibility for various Ruby environments

Keisuke05410:
## Summary
This PR improves Solargraph (Ruby Language Server) compatibility across different Ruby environments by modifying the binary discovery mechanism.

## Problem
The current implementation relies on `gem exec solargraph` command, which:
- Is not available in older RubyGems versions
- May not work correctly with Ruby version managers (asdf, rbenv, rvm) that provide their own shims

## Solution
- First attempt to find `solargraph` directly in PATH (which includes version manager shims)
- Fall back to `gem exec solargraph` only if direct PATH lookup fails
- This maintains backward compatibility while supporting more environments

## Affected Users
- Users with asdf/rbenv/rvm installed Ruby environments
- Users with older RubyGems versions that don't support `gem exec`
- Users who installed Solargraph via system package managers

## Testing
- Module import verified successfully
- Type checking passes
- No breaking changes to existing functionality

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/solidlsp/language_servers/solargraph.py
+++ b/src/solidlsp/language_servers/solargraph.py
@@ -57,17 +57,6 @@ class Solargraph(SolidLanguageServer):
         """
         Setup runtime dependencies for Solargraph and return the command to start the server.
         """
-        runtime_dependencies = [
-            {
-                "url": "https://rubygems.org/downloads/solargraph-0.51.1.gem",
-                "installCommand": "gem install solargraph -v 0.51.1",
-                "binaryName": "solargraph",
-                "archiveType": "gem",
-            }
-        ]
-
-        dependency = runtime_dependencies[0]
-
         # Check if Ruby is installed
         try:
             result = subprocess.run(["ruby", "--version"], check=True, capture_output=True, cwd=repository_root_path)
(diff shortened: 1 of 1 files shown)


src/solidlsp/language_servers/solargraph.py
```diff
@@ -78,6 +79,13 @@ def _setup_runtime_dependencies(logger: LanguageServerLogger, config: LanguageSe
             raise RuntimeError("Ruby is not installed. Please install Ruby before continuing.") from e
 
         # Check if solargraph is installed
+        # First, try to find solargraph in PATH (includes asdf shims)
```
MischaPanch, line 71: to make it clear that no dependencies are downloaded in that case, could you pls move this code below your addition? Thx!

```
runtime_dependencies = [
            {
                "url": "https://rubygems.org/downloads/solargraph-0.51.1.gem",
                "installCommand": "gem install solargraph -v 0.51.1",
                "binaryName": "solargraph",
                "archiveType": "gem",
            }
        ]

        dependency = runtime_dependencies[0]
```

Keisuke05410:
Thank you for the review @MischaPanch\! 

I've moved the `runtime_dependencies` definition to after the PATH check as requested. This makes it clearer that the dependencies are only needed when solargraph is not found in PATH.

The changes have been pushed and are ready for your review.

dragonfax:
Noice! 👍

Issue #230, opened 2025-06-25
The default rust_analyzer LSP configuration is out of date


waynr:
It looks to be about 1.5 years out of date:

- https://github.com/oraios/serena/blob/b1eba4563e7f892c96e2b8e97ca871984b4d2e08/src/solidlsp/language_servers/rust_analyzer/runtime_dependencies.json#L15

# Coding agent recommendation:

- Think about the version of dependencies you create.
- Consider making it possible to just use the system-installed rust analyzer

opcode81:
We want to be independent of system installations - and we need to use a fixed version such that we obtain well-defined behaviour.
If you want to try a newer version, please test it and issue a PR if it works.

waynr:
Thanks for the quick response! But for now I think it's best for me to avoid using this project at all. I'll close the issue since you don't seem to agree that this is a problem.

MischaPanch:
Reopening since the statement is "we don't have time to work on all issues and have to prioritize the things we do for Serena, pls try to help for issues that you care about" (Serena is an open source project from which we have no income or sponsors at the moment), and not "we don't care about rust"

PR #412, opened 2025-08-05
Added support for rust 2024 edition

kaibyao:
This updates the versions downloaded for rust-analyzer to the latest, which resolves https://github.com/oraios/serena/issues/230.

I was running into this error previously:

```
[LSP-stderr-reader] solidlsp.ls_handler:_read_ls_process_stderr:373 - [ERROR project_model::cargo_workspace] Unsupported edition `_E2024`
```

This has gone away with this update.

I've also added tests. Forgive me if the setup isn't correct; Python isn't my strong suit.


MischaPanch:
Looks good, thanks! I think minor tweaks in the tests would be beneficial, but I'll add them directly on main later.

Is there a reason that we need separate tests for the 2024 version? Are there incompatibilities in the syntax or something like that?

Regardless, merging this now, but pls let me know what you think about that question,

kaibyao:
Thanks for merging!

> Is there a reason that we need separate tests for the 2024 version? Are there incompatibilities in the syntax or something like that?

Rust 1) does add breaking changes between editions, but 2) rust-analyzer attempts to maintain backward-compatability, as far as I can tell. I added the separate folder just to make sure serena/lsp continues to work with both editions, but it's probably safe to just merge + use the 2024 edition (because of (2) above).

PR #425, opened 2025-08-06
Apply platform-specific default kwargs to subprocess calls

opcode81:
* On Windows, consistently add creationflags=CREATE_NO_WINDOW
  * On other platforms, consistently add start_new_session=True


opcode81:
@eLyiN it is not clear to me which problem is being solved by adding `start_new_session=True` for non-Windows platforms.

opcode81:
@MischaPanch so it turns out #422 didn't need fixing. But shall we still keep this to have a central way of setting flags for subprocesses? We could keep only the Windows thing for now. What do you think?

MischaPanch:
Sure, let's centralize this, feel free to merge

MischaPanch:
@opcode81 might affect #439, since more subprocess calls are introduced there. Should we merge now?

PR #427, opened 2025-08-07
Fix an error that occurred when trying to scan an inaccessible directory

tsukumijima:
Currently, if the project directory contains subdirectories that are not accessible with the current user's permissions, a `PermissionError` is raised during the directory scan. This causes the entire scan to fail, ultimately preventing Serena from starting.

For example, this can occur on Linux when developing a project within Docker, where persistent volumes (such as PostgreSQL's data directory) are mounted inside the project directory. These directories often have restrictive permissions like `600` or `700`, and when Serena (or Claude Code) is launched under a regular user account on Ubuntu, access to those directories fails.

Since Docker containers usually run as `root`, the files inside such directories require elevated privileges to be read. However, these system-specific or data directories are not relevant to Serena's operation and should be safely skipped during scanning.

This change ensures that Serena can start properly even when inaccessible directories exist within the project path.


src/serena/util/file_system.py
```diff
@@ -40,31 +40,41 @@ def scan_directory(
...
+                                sub_result = scan_directory(
+                                    entry_path,
+                                    recursive=True,
+                                    relative_to=relative_to,
+                                    is_ignored_dir=is_ignored_dir,
+                                    is_ignored_file=is_ignored_file,
+                                )
+                                files.extend(sub_result.files)
+                                directories.extend(sub_result.directories)
+                except (PermissionError, OSError) as ex:
```
opcode81: Catching both is not reasonable: `OSError` is the parent class of `PermissionError`. So this has the same effect as catching only `OSError`. Shouldn't we be catching only `PermissionError`?

tsukumijima: @opcode81 
I'm sorry. I'm embarrassed to say that I didn't know for the first time that PermissionError inherits OSError. I've corrected it, so please check it out.

MischaPanch:
I think this is ok now, merging. Thank you for the contribution!

PR #436, opened 2025-08-07
feat: add Erlang language server support via Erlang LS

dmbr0:
### Summary
Implements Erlang language support for Serena following the CONTRIBUTING.md guidelines.

### Changes
- **Language Server**: Created `ErlangLanguageServer` subclass using Erlang LS
- **Core Integration**: Added `ERLANG` to Language enum and SolidLanguageServer.create method
- **Test Coverage**: Added test repository with sample Erlang modules and test suite
- **File Extensions**: Supports `.erl`, `.hrl`, `.escript`, `.config`, `.app`, `.app.src`

### Dependencies
- Erlang/OTP runtime
- Erlang LS language server

### Testing
- ✅ Language server initialization test passes
- ✅ Integration test with actual Erlang project succeeds
- ✅ Serena successfully starts and analyzes Erlang code

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -263,6 +263,7 @@ markers = [
   "bash: language server running for Bash",
   "snapshot: snapshot tests for symbolic editing operations",
   "ruby: language server running for Ruby",
+  "erlang: language server running for Erlang",
 ]
 
 [tool.codespell]
--- a/src/solidlsp/language_servers/erlang_language_server.py
+++ b/src/solidlsp/language_servers/erlang_language_server.py
@@ -1,0 +1,53 @@
+"""Erlang Language Server implementation using Erlang LS."""
+
+import shutil
+import subprocess
+
+from solidlsp.ls import SolidLanguageServer
+from solidlsp.ls_config import LanguageServerConfig
+from solidlsp.ls_logger import LanguageServerLogger
+from solidlsp.lsp_protocol_handler.server import ProcessLaunchInfo
+from solidlsp.settings import SolidLSPSettings
+
+
+class ErlangLanguageServer(SolidLanguageServer):
+    """Language server for Erlang using Erlang LS."""
+
+    def __init__(
+        self,
+        config: LanguageServerConfig,
+        logger: LanguageServerLogger,
+        repository_root_path: str,
+        solidlsp_settings: SolidLSPSettings,
+    ):
+        """
+        Creates an ErlangLanguageServer instance. This class is not meant to be instantiated directly.
+        Use LanguageServer.create() instead.
+        """
+        self.erlang_ls_path = shutil.which("erlang_ls")
+        if not self.erlang_ls_path:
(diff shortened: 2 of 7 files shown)


MischaPanch:
You'll have to extend `pytest.yml` such that the erlang tests can run in CI.
Something like this might help
https://github.com/erlef/setup-beam/

MischaPanch:
Thank you for the contribution, happy to include Erlang support in Serena! I will review after tests run through in CI

dmbr0:
Okay sounds good I will update pytest.yml and see if I can get it running. Sorry this is my first PR! Most welcome. Look forward to contributing.

dmbr0:
Not really sure what the problem is here. the tests keeping hanging at various points for different machines whether I run them here or on my runners in my fork and its not always in the erlang tests section.

MischaPanch:
@dmbr0 tests pass, just that formatting was off. I pushed on your branch, make sure that you push before continuing.

Is this ready for review, or did you want to add anything else? If it's ready, I might push some minor cleanups and then merge

MischaPanch:
@dmbr0 ah, now I see, it hangs on ubuntu. The hanging seems to be caused by this test, which runs through on macos but hangs on linux

```
test/solidlsp/erlang/test_erlang_symbol_retrieval.py::TestErlangLanguageServerSymbols::test_symbol_methods_integration[erlang] PASSED [ 71%]
```

Could you look into it? Maybe you'll see what can cause the hang

dmbr0:
Yes ill look today but when I run it under my account Ubuntu passes and osx will hang

MischaPanch:
Yeah, this stuff is not fun. Maybe some kind of random race condition, or some dependency issue on Ubuntu. You can also create an experimental branch, remove all other tests and see if you can get the erlang suite running in CI - this way at least you experiment in a clean and reproducible environment.

Please do the latter outside of a PR though, so you don't consume our action budget too much.

dmbr0:
Yeah I'm doing it under my fork now. Originally thought it was just going to be easy to fix and then I didn't know how to stop it here. 

Tests run fine and pass on my machine and self hosted runners but as soon as I use githubs the hangs start and it goes back and forth between osx and Ubuntu. 

I'll keep working on it though

MischaPanch:
Sounds good, thanks!

MischaPanch:
Maybe some of the other actions we use screw up the env. Was happening with the setup_swift and ruby related actions, also either working on osx or on Ubuntu, but never on both at the same time. I ended up deleting the setup_swift action because of that. You can try removing all other actions to see if that's is the cause.

MischaPanch:
It runs locally, can't make it work on CI. I am merging this as "experimental support", which is better than no erlang support at all :)

erlang_ls was recently archived, if erlang users want to help out, updating the LS implementation to use https://github.com/whatsapp/erlang-language-platform would be nice

PR #439, opened 2025-08-08
Enhance Solargraph integration: Rails detection, exclude patterns, and error handling

kiakiraki:
## Summary

  This PR significantly improves Ruby language support in SolidLSP to bring it up to par with the Python implementation. The changes provide
  better performance, error handling, and project detection for Ruby/Rails projects, making Ruby a first-class citizen alongside Python in the
   Serena ecosystem.

  ## Motivation

  Currently, Ruby language support lacks several features that are available for Python:
  - No project-specific exclude patterns for better indexing performance
  - Limited error diagnostics during setup
  - Missing Rails-specific optimizations
  - Basic initialization handling compared to Python's robust approach

  This creates an inconsistent developer experience between Python and Ruby projects.

  ## Changes

  🚀 Performance & Project Detection

  - Ruby/Rails-specific exclude patterns: Automatically excludes vendor/, .bundle/, tmp/, log/, coverage/, etc.
  - Automatic Rails project detection: Detects Rails projects via config/application.rb, Rakefile, and Gemfile analysis
  - Dynamic configuration: Applies Rails-specific exclude patterns when Rails is detected

  🛠️ Enhanced Error Handling

  - Detailed Ruby installation diagnostics with version compatibility checks (recommends Ruby 2.6+)
  - Multiple Ruby manager support: Provides specific installation instructions for rbenv, RVM, asdf, and system packages
  - Improved Bundler error messages with clear resolution steps
  - Better Solargraph installation error handling with fallback options

  🔧 Robust Initialization

  - Rails-aware initialization options: Mirrors Python's initializationOptions structure with Ruby-specific patterns
  - Proper analysis completion detection: Uses analysis_complete event pattern consistent with Pyright
  - Enhanced LSP capability negotiation: Matches Python implementation's capability declarations

  ## Implementation Details

###  New Helper Methods

  - _detect_rails_project(): Automatically detects Rails projects by checking for Rails indicators
  - _get_ruby_exclude_patterns(): Generates context-aware exclude patterns based on project type
  - Enhanced _setup_runtime_dependencies(): Provides detailed error diagnostics and resolution steps

###  Improved Directory Handling

  - Expanded is_ignored_dirname() with Ruby/Rails-specific directories
  - Dynamic exclude patterns in initializationOptions based on automatic project detection

 ### Better Error Messages

 #### Before: Generic error
  "Ruby is not installed. Please install Ruby before continuing."

 #### After: Helpful with multiple options
  "Ruby is not installed or not found in PATH. Please install Ruby using one of these methods:
  - Using rbenv: rbenv install 3.0.0 && rbenv global 3.0.0
  - Using RVM: rvm install 3.0.0 && rvm use 3.0.0 --default
  - Using asdf: asdf install ruby 3.0.0 && asdf global ruby 3.0.0
  - System package manager (brew install ruby, apt install ruby, etc.)"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,12 @@ Status of the `main` branch. Changes prior to the next official version change w
     * Hybrid function detection (LSP + regex fallback) for reliable symbolic editing
     * Automatic Node.js/npm dependency management 
     * Cross-platform compatibility including Windows
+  * Enhance Solargraph (Ruby) integration
+    * Automatic Rails project detection via config/application.rb, Rakefile, and Gemfile analysis
+    * Ruby/Rails-specific exclude patterns for improved indexing performance (vendor/, .bundle/, tmp/, log/, coverage/)
+    * Enhanced error handling with detailed diagnostics and Ruby manager-specific installation instructions (rbenv, RVM, asdf)
+    * Improved LSP capability negotiation and analysis completion detection
+    * Better Bundler and Solargraph installation error messages with clear resolution steps
 
 Fixes:
 * Ignore `.git` in check for ignored paths and improve performance of `find_all_non_ignored_files`


src/solidlsp/language_servers/solargraph.py
```diff
@@ -142,25 +176,130 @@ def _setup_runtime_dependencies(logger: LanguageServerLogger, config: LanguageSe
...
-        initialize_params = {
-            "capabilities": {},
-            "trace": "verbose",
+        exclude_patterns = Solargraph._get_ruby_exclude_patterns(repository_absolute_path)
+
+        initialize_params: InitializeParams = {  # type: ignore
             "processId": os.getpid(),
             "rootPath": repository_absolute_path,
-            "rootUri": pathlib.Path(repository_absolute_path).as_uri(),
+            "rootUri": root_uri,
```
MischaPanch, line 263: Is the enhancement of initialize_params actually needed? Seems like it was working well with the default settings for most params

src/solidlsp/language_servers/solargraph.py
```diff
@@ -62,13 +74,30 @@ def _setup_runtime_dependencies(logger: LanguageServerLogger, config: LanguageSe
         """
         # Check if Ruby is installed
         try:
-            result = subprocess.run(["ruby", "--version"], check=True, capture_output=True, cwd=repository_root_path)
+            result = subprocess.run(["ruby", "--version"], check=True, capture_output=True, cwd=repository_root_path, text=True)
```
MischaPanch, line 77: why was text=True needed? Just out of curiosity

MischaPanch (commented):
Just two small questions, otherwise seems good.

Please also add an entry to the changelog about the better support for ruby (rails, bundler, tests, startup waiting etc)

src/solidlsp/language_servers/solargraph.py
kiakiraki: Thanks for the feedback!
I believe the exclude option is needed for performance; the other options were redundant with the defaults, so I’ve removed them.

src/solidlsp/language_servers/solargraph.py
kiakiraki: I think text=True ensures subprocess.run() returns string output instead of bytes (which I think is the default).

  Without it:
  `result.stdout  # would be: b"ruby 3.0.0p0 ..."  (bytes)`

  With text=True:
  `result.stdout  # becomes: "ruby 3.0.0p0 ..."   (string)`

  I think we need strings because the code immediately does:
  - result.stdout.strip()
  - re.search(r"ruby (\d+)\.(\d+)\.(\d+)", ruby_version)

kiakiraki:
@MischaPanch 
Thanks for the review — I’ve addressed the feedback.

MischaPanch:
Thanks, merging! Ruby support is first-class now I believe, and almost all of it came from the community, really cool

PR #442, opened 2025-08-08
Fix #432: Normalize path separators in GetSymbolsOverviewTool

### LINKED ISSUE #432: GetSymbolsOverviewTool may return name_paths with backslashes
> Also, name_paths should always have forward slashes 

 _Originally posted by @MischaPanch in [#424](https://github.com/oraios/serena/issues/424#issuecomment-3160915558)_

Example: https://claude.ai/share/0b79b853-a7ab-4239-81af-cef211f15960

Design issue: GetSymbolsOverviewTool is not using the abstractions which generate name_path correctly and instead directly returns low-level data from the LS.

liebertar:
Hi\! I came across issue #432 and thought I'd take a stab at fixing it.

## The Issue
GetSymbolsOverviewTool was returning paths with backslashes on Windows, which should always use forward slashes as mentioned in the issue.

## The Fix
Modified the `get_symbol_overview` method in `symbol.py` to normalize path separators by replacing backslashes with forward slashes when creating the name_path.

It's a one-line change that ensures consistent forward slash usage across all platforms. 

Let me know if this looks good to merge\!

Fixes #432


MischaPanch (changes requested):
Thanks for the PR! This should be handled on a deeper level, so `symbol[0]` should already come with the right slashes.

opcode81:
I will take this over.

opcode81:
#446 is the new PR.
@liebertar while your solution did fix the output, it didn't fix the underlying design issue.

PR #443, opened 2025-08-08
Remove duplicate Swift

ykws:
It seems that Swift was added in each of the following two PRs at different positions, so there was no conflict and both were merged, resulting in duplication.

- #401
- #437


MischaPanch:
It's not because of duplicate PRs, I was just blind yesterday ^^. thx!

PR #446, opened 2025-08-08
Make LS overview calls return the full symbol information

### LINKED ISSUE #432: GetSymbolsOverviewTool may return name_paths with backslashes

opcode81:
... instead of creating new low-level data structures.

This enables uniform handling of required conversions in our higher-level Serena data structures

Fixes #432

### DIFF
--- a/src/solidlsp/ls.py
+++ b/src/solidlsp/ls.py
@@ -25,6 +25,7 @@ from solidlsp.ls_config import Language, LanguageServerConfig
 from solidlsp.ls_exceptions import SolidLSPException
 from solidlsp.ls_handler import SolidLanguageServerHandler
 from solidlsp.ls_logger import LanguageServerLogger
+from solidlsp.ls_types import UnifiedSymbolInformation
 from solidlsp.ls_utils import FileUtils, PathUtils, TextUtils
 from solidlsp.lsp_protocol_handler import lsp_types
 from solidlsp.lsp_protocol_handler import lsp_types as LSPTypes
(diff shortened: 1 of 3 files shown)


src/solidlsp/ls.py
```diff
@@ -1201,28 +1190,19 @@ def process_symbol(symbol: ls_types.UnifiedSymbolInformation):
             process_symbol(root)
         return result
 
-    def request_document_overview(self, relative_file_path: str) -> list[tuple[str, ls_types.SymbolKind, int, int]]:
+    def request_document_overview(self, relative_file_path: str) -> list[UnifiedSymbolInformation]:
```
MischaPanch, line 1193: rename to `request_top_level_symbols`?

MischaPanch (approved):
Minor comment on renaming, feel free to either merge as is or perform the renaming if you agree

src/solidlsp/ls.py
opcode81: The name stems from the role the call plays in Serena, and we should probably consider moving it (and the other method) to Serena's code where the name fits.
But even if it stays here, we could, in the future, soften the semantics and make it return not only top-level symbols (as suggested in the TODO in Serena), so I am keeping the name for now.

PR #448, opened 2025-08-08
fix: discover all `.gitignore` files in project

blaugold:
Before this change, only `.gitignore` files at the project root and its direct subdirectories were found.

This issue prevented me from indexing my project because large JS files in build directories were getting the indexing process stuck.


opcode81:
Thanks for the fix!

PR #452, opened 2025-08-09
fix: remove incorrect description of supporting directories from `get_symbols_overview` docs

blaugold:
I noticed that models often use `get_symbols_overview` on a directory only to get an error because the tool does not support directories. The issue was that the description and usage instructions for the tool incorrectly mentioned support for directories.


MischaPanch:
Thanks! I recently removed support for directories because it can cause too large answers, but forgot to adjust all descriptions accordingly. Great that you noticed!

PR #459, opened 2025-08-09
Continue indexing on timeout errors, make timeout configurable in CLI

MischaPanch:
I didn't use the FileLogger since it would have been a bit cumbersome (or I didn't figure out a good way), so I just write the log file manually


MischaPanch:
Merging this since it's a quality of life improvement, we can adjust later if needed

PR #471, opened 2025-08-12
project: fix gather_source_files dir filtering; add test

Xopoko:
**Bug fix**: gather_source_files erroneously filtered out source files when the absolute path to the repo contained hidden directories (e.g., ".codex"). This happened because directory filtering used absolute paths, triggering the ".*" directory rule for ancestors outside the project.

**Change**
- Use relative paths to the project root when filtering directories in Project.gather_source_files
- Add focused test: test_project_gather_source_files.py verifying that Clojure sources under src/ are included alongside deps.edn

**Why**
- Prevents empty search results for valid source trees (observed zero matches for clj/py tests when running under a hidden parent dir)

**Before**
- project.gather_source_files() on a clojure test repo returned only ["deps.edn"], excluding src/**/*.clj

**After**
- Returns src files (e.g., src/ns/core.clj) and deps.edn

**Tests**
- Added unit test covering the regression
- Ran: `pytest -q -k test_project_gather_source_files` and the regular suite subset
  206 passed, 1 skipped, 188 deselected locally with markers to avoid external language servers

**Compatibility**
- No behavior change to ignore rules; only fixes a false positive path handling edge case

### DIFF
--- a/src/serena/project.py
+++ b/src/serena/project.py
@@ -193,9 +193,12 @@ class Project:
             return [relative_path]
         else:
             for root, dirs, files in os.walk(start_path, followlinks=True):
-                dirs[:] = [d for d in dirs if not self._is_ignored_relative_path(os.path.join(root, d))]
+                # Filter directories using relative paths to avoid false positives from absolute paths (e.g., parent dirs like .codex)
+                dirs[:] = [
+                    d for d in dirs if not self._is_ignored_relative_path(os.path.relpath(os.path.join(root, d), start=self.project_root))
+                ]
                 for file in files:
-                    rel_file_path = os.path.relpath(os.path.join(root, file), start=self.project_root)
+                    rel_file_path = os.path.relpath(os.path.join(root, file), start=self.project_root).replace(os.path.sep, "/")
                     try:
                         if not self._is_ignored_relative_path(rel_file_path):
                             rel_file_paths.append(rel_file_path)
--- a/test/serena/util/test_project_gather_source_files.py
+++ b/test/serena/util/test_project_gather_source_files.py
@@ -1,0 +1,25 @@
+from serena.project import Project
+from solidlsp.ls_config import Language
+
+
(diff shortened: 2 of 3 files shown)


src/serena/project.py
```diff
@@ -193,9 +193,12 @@ def gather_source_files(self, relative_path: str = "") -> list[str]:
             return [relative_path]
         else:
             for root, dirs, files in os.walk(start_path, followlinks=True):
-                dirs[:] = [d for d in dirs if not self._is_ignored_relative_path(os.path.join(root, d))]
+                # Filter directories using relative paths to avoid false positives from absolute paths (e.g., parent dirs like .codex)
+                dirs[:] = [
+                    d for d in dirs if not self._is_ignored_relative_path(os.path.relpath(os.path.join(root, d), start=self.project_root))
```
MischaPanch, line 198: I think the change should rather happen in _is_ignored_relative_path, right?

src/serena/project.py
```diff
@@ -193,9 +193,12 @@ def gather_source_files(self, relative_path: str = "") -> list[str]:
...
+                ]
                 for file in files:
-                    rel_file_path = os.path.relpath(os.path.join(root, file), start=self.project_root)
+                    rel_file_path = os.path.relpath(os.path.join(root, file), start=self.project_root).replace(os.path.sep, "/")
```
MischaPanch, line 201: why was the replace needed?

test/serena/util/test_project_gather_source_files.py
```diff
@@ -0,0 +1,25 @@
+from serena.project import Project
+from solidlsp.ls_config import Language
+
+
+def test_gather_source_files_includes_language_sources(tmp_path):
```
MischaPanch, line 5: I don't think this tests adds anything. We have integration tests that read language files, I propose to remove it.

MischaPanch (changes requested):

PR #480, opened 2025-08-13
Fix ListDirTool FileNotFoundError for non-existent directories

seuros:
Now when LLMs try to access non-existent paths, they get clear feedback about what went wrong and where they are in the filesystem.

Similar to fix in commit 8464f25 for CreateTextFileTool

### DIFF
--- a/src/serena/tools/file_tools.py
+++ b/src/serena/tools/file_tools.py
@@ -100,6 +100,15 @@ class ListDirTool(Tool):
             required for the task.
         :return: a JSON object with the names of directories and files within the given directory
         """
+        # Check if the directory exists before validation
+        if not self.project.relative_path_exists(relative_path):
+            error_info = {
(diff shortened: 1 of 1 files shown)


src/serena/tools/tools_base.py
```diff
@@ -271,7 +271,12 @@ def task() -> str:
             except Exception as e:
                 if not catch_exceptions:
                     raise
-                msg = f"Error executing tool: {e}"
+                # Include project root context in error messages for better LLM understanding
```
MischaPanch: Thanks for the PR! Let's remove this part though, we shouldn't rely on getattr in this way anywhere. The exception should already contain all relevant info, no need to further decorate it

MischaPanch (changes requested):
See comment

MischaPanch:

PR #489, opened 2025-08-15
ok

thethien8a:


MischaPanch:
I don't know what this is about, but it's not mergeable. I remember explaining somewhere else why we can't switch to union types but can't find it now.

In any case: we can't switch to union types, multiple clients have problems with this. Apart from that, if the clien't can't handle an integer, then it's really a bug in the client. If this is about openai compatibility, see how we handle codex and chatgpt in the readme

PR #495, opened 2025-08-18
refactor: use rustup-managed rust-analyzer instead of bundling

0xdeafbeef:
BREAKING CHANGE: rust-analyzer now requires rustup to be installed

## Motivation

The previous implementation bundled rust-analyzer binaries that quickly became outdated and required manual updates to Serena's codebase. This caused version mismatches between the bundled rust-analyzer and users' actual Rust toolchain versions.

## Changes

- Remove RuntimeDependency-based binary bundling for rust-analyzer
- Use rustup-managed rust-analyzer from the user's toolchain
- Add automatic installation of rust-analyzer component if missing
- Install rustup with minimal profile in Docker images

## Rationale

Rust development universally relies on rustup for toolchain management. Unlike some other languages where system packages or bundled binaries make sense, Rust developers already have rustup installed and expect tools to use their toolchain's versions.

This approach ensures:
- rust-analyzer always matches the user's Rust version
- No outdated binaries shipped with Serena
- Automatic updates via `rustup update`
- Consistency with Rust ecosystem expectations

The LSP API of rust-analyzer is stable enough that version differences are not a concern. If users don't have rustup, they'll get a clear error message - this is acceptable since you cannot meaningfully develop Rust without rustup.

## Docker Support

Added rustup installation to Docker images to ensure Rust support works in containerized environments. Uses minimal profile to reduce image size while still providing full rust-analyzer functionality.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # latest
 Status of the `main` branch. Changes prior to the next official version change will appear here.
 
+## Breaking Changes
+
+* **Rust support now requires rustup**: The rust-analyzer is no longer bundled with Serena. Instead, it uses the rust-analyzer from your Rust toolchain managed by rustup. This ensures compatibility with your Rust version and eliminates outdated bundled binaries. If you don't have rustup installed, you'll need to install it from https://rustup.rs/
+
 # 0.1.4
 
 ## Summary
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -13,7 +13,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 **Test Markers:**
 Available pytest markers for selective testing:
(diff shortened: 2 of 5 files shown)


0xdeafbeef:
btw the one shipped with serena had version `0.3.1689-standalone` which is 2yo :)
https://github.com/rust-lang/rust-analyzer/releases/tag/2023-10-09

```
 ~/.serena/language_servers/static/RustAnalyzer/RustAnalyzer/rust_analyzer --version
rust-analyzer 0.3.1689-standalone
```

MischaPanch:
Looks good, merging! In follow-up commits I will try to run the rust tests in CI, unfortunately, they were always hanging there.

We took over the rust support implementation from multilspy, there was too little interest from the rust community to seriously improve on it (until now)

MischaPanch:
Thank you for the contribution

PR #498, opened 2025-08-18
Feat: Add Support for Zig, Nix, Lua

rubas:
## Summary

This PR adds comprehensive language server support for **Zig**, **Nix** and **Lua** to Serena.

### 1. **Zig Language Support**
- Integrated ZLS (Zig Language Server)
- Manual installation required (binary download from ZLS releases)
- Full LSP capabilities for Zig development
- Test suite with Zig project examples
- Platform support: Linux, macOS, Windows

### 2. **Nix Language Support** 
- Integrated `nil` (NIx Language server) for Nix/NixOS development
- Automatic installation via cargo if not present
- Full LSP capabilities: symbols, references, completions, hover info
- Comprehensive test suite with Nix flake examples
- Platform support: Linux, macOS (Windows not supported due to Nix limitations)

### 3. **Lua Language Support**
- Integrated `lua-language-server` (LuaLS) 
- Automatic download and installation if not present
- Full LSP capabilities for Lua development
- Test suite covering calculator and utility modules
- Platform support: Linux, macOS, Windows
- Fixed Windows CI issues with proper PATH handling

## Dependencies

- **Zig**: Requires ZLS binary installation
- **Nix**: Requires Nix environment and cargo (for nil installation)
- **Lua**: No external dependencies (downloads binary directly)


MischaPanch:
Hi @rubas . Thank you for the PR, we'd be happy to add Zig support!

I wanted to review this today but noticed that you also added svelte-related commits here. Pls remove them from this branch, we should keep different LS contributions in separate PRs for easier review and a cleaner history.

Also, regarding svelte there was some work in the past (parts of it may be merged soon) in #339 and ongoing work in #504.

rubas:
@MischaPanch Sorry, yes. It was at the end much easier with the long running test to have all LSP integrated at once (and we are already using in internally). I hope this okay, as this makes merging easier.

Now, everything builds (except sometimes there is a flaky c# test).

👉 Let me know, if should remove the Svelte Implementation.

MischaPanch:
Ok, that's really a major contribution, so I don't want to bum you out. Great work here! Let's do the following to make this mergeable into Serena.

The main focus of the review will be the tests. Currently, they are incomplete and deviate in structure from the pre-existing tests. Please do the following:

1. Test that the LS actually finds the right symbols. There should be no tests just asserting that an output was produced (like an empty list or None).
2. Make sure that for every LS you test cross-file reference symbol finding. Currently, I think there's not a single test for that
3. Don't test functionality that may not be available and don't skip tests in that case. Either something works and is tested or it doesn't work and there is no test for it. 
4. The only exceptional case where tests may be skipped is when the LS really doesn't support a platform (like nix on windows).
5. Please remove svelte support from this PR, see my comment above as to the explanation. If I haven't misread it, this implementation doesn't give cross-language reference resolutions (between svelte and TS), right?

MischaPanch:
@rubas once you addressed the comment above, pls ping me for a review :)

src/solidlsp/language_servers/zls.py
```diff
@@ -211,6 +223,90 @@ def do_nothing(params):
...
 
+        # Auto-open workspace files if enabled
+        if self.auto_open_workspace:
+            self._open_workspace_files()
+
         # ZLS server is typically ready immediately after initialization
         self.server_ready.set()
         self.server_ready.wait()
+
+    def _open_workspace_files(self):
```
MischaPanch: This can take forever, right? Should be mentioned in the readme that startup for Zig can be slow.

I also find it weird that this is needed, no other LS behaves this way. Are you sure it's true and not a Claude hallucination? I would rather suspect that some initialization option like dynamicRegistration=True is missing

src/solidlsp/language_servers/zls.py
```diff
@@ -211,6 +223,90 @@ def do_nothing(params):
...
+                    continue
+
+                # Store the relative path as string (will use platform-specific separators)
+                zig_files.append(str(relative_path))
+
+            self.logger.log(f"Found {len(zig_files)} Zig files to open", logging.INFO)
+
+            # Open each file by sending didOpen notification to ZLS
+            for file_path in zig_files:
+                try:
```
MischaPanch: If it's really true that this is needed, should use multithreading here

src/solidlsp/language_servers/zls.py
rubas: I was working a long time on this and tried to make it pass the tests on windows, without any luck at the end. Was never reliable. Similar with some other tests that fail from time to time.

rubas:
### Changes made:
  - **Removed file watcher complexity** - Simplified ZLS implementation by removing the automatic file opening system
  - **Added comprehensive tests** - 11 tests covering symbols, references, go-to-definition, and cross-file navigation
  - **Windows compatibility** - ZLS disabled on Windows due to unreliable cross-file references (reason unknown). Follows same pattern as Elixir.

  ### Platform support:
  - ✅ **Lua**: All platforms (Windows, Linux, macOS)
  - ✅ **Nix**: Linux and macOS only (Nix doesn't support Windows natively)
  - ✅ **Zig**: Linux and macOS only (cross-file references unreliable on Windows)

rubas:
@MischaPanch Ping.

MischaPanch:
Great, thanks! Will do final review tomorrow, but summary sounds good

MischaPanch:
Alright, now that I call a PR, great work! Merging 3100 lines of external contribution containing support for three new languages in one go is almost surreal, but the tests are very expressive, so it seems like everything indeed does work.

Thanks and congrats on the big chunk of work being finalized here! With a bit of luck, the efforts of supporting Svelte will also be concluded in the near future

MischaPanch:
@rubas out of curiosity, what was up with the whole file-watching stuff for zig? Was that claude hallucinating problems? Or was it an attempt to force cross-file references working on windows?

rubas:
@MischaPanch Cool 🎸 

Kinda both. But calling it hallucinating would not be fair. The cross file reference did not work and he ran multiple experiments to get it working. Opening the files manually di the trick... but it was clear, there must be a better way. 

After digging more into the source code of ZLS (and LSP in general) and using it as context, he figured out that opening `build.zon` to give ZLS the project context works too and is obviously the better solution. After that it run numerous tests for windows, but never succeeded, till I called it a day 😆 

Imaging Serena would not be limited to one language per project, that would simplify a svelte integration. But that is for another day ...

xuhuanzy:
I think you shouldn't automatically activate luals. In fact, we are developing a new `lualsp`, which has completely surpassed luals in both functionality and ease of use, and has added some new syntax features (such as generics and attributes). Automatically activating luals is unacceptable to us.  
Link: https://github.com/EmmyLuaLs/emmylua-analyzer-rust

PR #507, opened 2025-08-21
Change kotlin lsp to official https://github.com/Kotlin/kotlin-lsp

cylee0909:
This MR updates the Kotlin Language Server (LSP) to use the official implementation from https://github.com/Kotlin/kotlin-lsp.

Key changes:

Replaced the previous LSP implementation with the official Kotlin LSP
Updated relevant configuration to point to the new official repository
Ensured compatibility with existing Kotlin project structures and tooling

Benefits of this change:

Access to official updates and maintenance directly from the Kotlin team
Improved stability and better alignment with Kotlin language features
Enhanced support for Kotlin-specific language capabilities

### DIFF
--- a/src/solidlsp/language_servers/kotlin_language_server.py
+++ b/src/solidlsp/language_servers/kotlin_language_server.py
@@ -43,7 +43,7 @@ class KotlinLanguageServer(SolidLanguageServer):
         self.runtime_dependency_paths = runtime_dependency_paths
 
         # Create command to execute the Kotlin Language Server script
-        cmd = [f'"{self.runtime_dependency_paths.kotlin_executable_path}"', "--stdio"]
+        cmd = [self.runtime_dependency_paths.kotlin_executable_path, "--stdio"]
 
         # Set environment variables including JAVA_HOME
         proc_env = {"JAVA_HOME": self.runtime_dependency_paths.java_home_path}


MischaPanch:
Have you tested it on Kotlin code, does it actually work well? The official lsp is still in pre-alpha status

cylee0909:
> Have you tested it on Kotlin code, does it actually work well? The official lsp is still in pre-alpha status

Yes, I tested it on my Mac and it works fine. Although kotlin-lsp is not released yet, it is much faster than the existing implementation.

MischaPanch:
Cool, great to hear that! Would you have time to add tests like for the other LS implementations? In particular, cross-file reference finding needs to be tested. Then we could promote Kotlin to "officially supported".

cylee0909:
> Cool, great to hear that! Would you have time to add tests like for the other LS implementations? In particular, cross-file reference finding needs to be tested. Then we could promote Kotlin to "officially supported".

Thanks for your tip, I didn't notice the relevant test cases at first. I have added the relevant tests now, and the basic test cases are copied from Java. They all pass on my device..

MischaPanch:
Thanks, I will run them locally and on CI later and will then merge this if everything works (after formatting).

If you want to help further, you could try making it run in CI yourself by setting up kotlin the the GH workflow, maybe there already is a GH action for that somewhere. Make sure to run poe format for the tests to start

MischaPanch:
The pre-alpha nature of the LS might be showing here - initialization doesn't work on windows

MischaPanch:
@cylee0909 If you can't make it work on windows due to immaturity, what we can do is instead of replacing the current kotlin LS implementation just adding a new one, called kotlin_official. User's can then specify it in their project yaml, just like for the other "experimental" LS implementations like typescript_vsc and omnisharp. We can then promote it to the default kotlin LS once it works on windows

MischaPanch:
It was just the quotes around the command that caused the problem on windows, so this actually works. Merging it now, thank you for the contribution!

cylee0909:
> It was just the quotes around the command that caused the problem on windows, so this actually works. Merging it now, thank you for the contribution!

I'm very happy to hear this news.

PR #509, opened 2025-08-21
Fix OpenAI tool descriptions

pwilkin:
This PR:
* modifies the `_sanitize_for_openai_tools` method in `mcp.py` to fix the oneOf/anyOf (null|X) type conversion
* adds an agent mode for OpenAI compatible servers like Llama.cpp (`openai-agent`) that requires OpenAI-compatible tool descriptions and updates the README
* adds a testcase to verify completeness of the conversion by checking for "type" in all tools' parameters


test/serena/test_tool_parameter_types.py
```diff
@@ -0,0 +1,49 @@
+import pytest
+from serena.mcp import SerenaMCPFactorySingleProcess
+from serena.config.serena_config import SerenaConfig
+
+def load_agent():
```
MischaPanch: should rather be load_config, but I think we don't need this function at all. You only have one test, pls create the config there. You should also not load it from disc but instantiate it directly with gui and dashboard disabled, like

```
SerenaConfig(gui_log_window_enabled=False, web_dashboard=False, log_level=logging.ERROR)
```

README.md
```diff
@@ -552,6 +552,8 @@ Choose the context that best matches the type of integration you are using.
 When launching Serena, specify the context using `--context <context-name>`.  
 Note that for cases where parameter lists are specified (e.g. Claude Desktop), you must add two parameters to the list.
 
+If you are using a local server (such as Llama.cpp) which requires you to use OpenAI-compatible tool descriptions, use context `openai-agent` instead of `agent`.
```
MischaPanch: it's not about tool descriptions but tool parameter types.

The current solution on main seems to work for openai models, so maybe the context should rather be called llama-cpp?

A context for this is also not the optimal solution, since this is about the LLM selected and not the actual execution context. But for now it's the best we can do, we can improve on it later if needed.

MischaPanch (changes requested):
Thanks, overall looks good, just small changes needed. Please also run `poe format` when you're done so that CI can run through

README.md
pwilkin: Yeah, you might be right, although I'm not sure about other servers, so maybe `oaicompat-agent`?

test/serena/test_tool_parameter_types.py
pwilkin: Aight, will make inline config.

README.md
MischaPanch: Sure, oaicompat-agent works

MischaPanch:
Tests are failing, you need to change the context file name

pwilkin:
> Tests are failing, you need to change the context file name

Yeah, sorry, fixed now, tests seem to be passing (the Mac one doesn't seem related to my changes).

MischaPanch:
Thank you for the contribution, have fun with Serena and local models :)

pwilkin:
Thanks :)

Issue #473, opened 2025-08-12
Ignore spec derived from nested .gitignore files is incorrect


opcode81:
> So I did a quick test, and the matching is definitely incorrect.
> 
> If in a subdirectory `foo/`, `.gitignore` contains
> * `foo.txt`, then foo.txt will (correctly) be ignored only in the subtree of foo/ (but not in /)
> * `/foo.txt`, then foo.txt will (incorrectly) be ignore anywhere in foo/ (e.g. also in foo/bar/), while it should only match foo/foo.txt
> *  `**/foo.txt`, then foo.txt will (incorrectly) be ignored everywhere (including /foo.txt)
> * `/**/foo.txt`, then foo.txt will (correctly) be ignored only in the subtree of foo/  

 _Originally posted by @opcode81 in [#453](https://github.com/oraios/serena/issues/453#issuecomment-3180944099)_

MischaPanch:
@jetbrains-junie Solve this in a test driven manner

ghost:
Hey, it’s Junie by JetBrains! I started [processing](https://github.com/oraios/serena/actions/runs/16932598523) your request 🚀

ghost:
### ✅ Generation Status

* Code generation was **successful**

Message:

> The implementation was updated to correctly handle nested `.gitignore` patterns, ensuring anchored patterns match only direct children and double-star patterns do not match outside subdirectories. A test script was created but could not be executed due to import errors. The changes were submitted without confirming test results.




More information:
* 🔗 [GitHub Action](https://github.com/oraios/serena/actions/runs/16932598523)
* 💬 [Report the issue on our Discord and help us improve](https://jb.gg/junie/github)

MischaPanch:
Junie was useless..

Scoping was wrong, fixed in de173ce078. Claude code did very on this IMO

opcode81:
Why was this closed? It is clearly still broken.

opcode81:
``

PR #510, opened 2025-08-21
Improve handling of ignored files and .gitignore discovery

### LINKED ISSUE #453: Don't discover `.gitignore` files that are ignored themselves
The current mechanism for finding `.gitignore` files also finds `.gitignore` files that are ignored by a `.gitignore` file that is actually part of the repo. For example, `node_modules` is ignored in the `.gitignore` file at the project root, but `.gitignore` files in `node_modules` packages are still being discovered and their patterns are added to the `GitignoreSpec`

This is not causing incorrect behavior, but it's probably not good for performance.

### LINKED ISSUE #470: Serena MCP Server Initialization Hangs During Gitignore Parsing in Directories with Virtual Environments
## Issue Summary

There is possibly a performance bottleneck or infinite loop in Serena's gitignore parsing logic that causes the MCP server to hang during initialization when scanning directories containing Python virtual environments.

## Environment

- **Operating System**: WSL2 (Windows Subsystem for Linux)
- **File System**: NTFS mount via `/mnt/c/`
- **Serena Version**: 0.1.3
- **Python Version**: 3.10
- **Project Structure**: Large Python project with virtual environments

### Initial Symptoms

The Serena MCP server appears to connect successfully in most directories but consistently fails to initialize in one specific directory structure. The failure pattern seemed directory-specific. It seems to fail containing large virtual environments

### Investigation Process

Through systematic testing, I discovered that Serena works in **every subdirectory** of the failing project **except** for virtual environment directories. This was puzzling because these directories are properly listed in `.gitignore` files.

#### 1. Hanging Location Identified

Using timeout debugging, I found that Serena consistently hangs at this specific log line:

```
INFO  2025-08-11 23:16:35,023 [MainThread] serena.project:__init__:31 - Parsing all gitignore files in /path/to/project
```

The process never progresses beyond this point, suggesting the gitignore parsing logic encounters an issue.

#### 2. Virtual Environment Impact

I suspect the issue is related to the complexity or structure of Python virtual environments because:

- Serena worked perfectly in the same directory **before** (with the same venv present)
- **Deleting the virtual environment directories immediately resolved the issue**
- **Recreating fresh virtual environments allows Serena to work again**

#### 3. Source Code Analysis

Looking at the Serena source code, I think there might be two potential issues:

**a) Recursive Gitignore Discovery (`file_system.py:154`)**
```python
relative_paths = glob.glob("**/.gitignore", root_dir=self.repo_root, recursive=True)
```

This appears to scan the **entire directory tree** to find all `.gitignore` files before applying any ignore rules. I suspect this could be problematic with large virtual environments because:
- It has to traverse all directories (including venv) **before** knowing what to ignore
- Virtual environments can contain thousands of files and complex symlink structures

**b) Symlink Following in File Scanning (`project.py`)**
```python
for root, dirs, files in os.walk(start_path, followlinks=True):
```

I noticed that `followlinks=True` is used in the file scanning code. Virtual environments often contain symlinks, and I suspect this could potentially cause:
- Infinite loops with circular symlinks
- Massive directory traversal following symlinks to system directories

### Reproduction Steps

1. Create a Python project with a large virtual environment (many packages installed)
2. Ensure the venv is properly listed in `.gitignore`
3. Try to initialize Serena MCP server: `uvx --from git+https://github.com/oraios/serena serena start-mcp-server --context ide-assistant --project $(pwd)`
4. Observe hanging at "Parsing all gitignore files" log line

### Temporary Workaround

The issue can be resolved by:
1. Removing virtual environment directories
2. Recreating fresh, minimal virtual environments
3. This suggests the issue may be related to accumulated complexity/state in the venv

### Potential Root Causes (Speculation)

I think the issue could be caused by:

1. **Performance bottleneck**: The recursive gitignore discovery (`glob.glob("**/.gitignore", recursive=True)`) might be too slow with complex directory structures
2. **Symlink loops**: Virtual environments might contain symlink structures that cause infinite traversal with `followlinks=True`
3. **Memory exhaustion**: Large directory structures might exhaust memory during scanning
4. **Race conditions**: Possible timing issues in directory scanning logic

### Suggested Investigation Areas

I suspect the following areas might need investigation:

1. **Add timeout mechanisms** to gitignore parsing operations
2. **Consider disabling `followlinks=True`** or adding circular symlink detection
3. **Implement early termination** for gitignore discovery in large directories
4. **Add progress logging** to identify exactly where the hang occurs
5. **Consider lazy loading** of gitignore files instead of scanning everything upfront

### Environment Details

- WSL2 with NTFS mounts might compound filesystem performance issues
- Large Python virtual environments with ML packages (PyTorch, OpenCV, etc.)
- Project contained thousands of Python packages in venv

## Additional Context

This appears to be a regression or edge case that emerges when virtual environments reach a certain complexity threshold. The same setup worked previously, suggesting the issue might be triggered by recent changes in the virtual environment structure or accumulated filesystem state.

I believe this could affect other users with large Python projects containing complex virtual environments, especially in WSL/Docker environments where symlink handling might behave differently.

opcode81:
This is a WIP

* Handles .gitignore discovery incrementally: .gitignore files are parsed as they are discovered via top-down traversal, such that .gitignore files in ignored directories are not discovered (and the directories not searched in the first place)

fixes #470, fixes #453 (or at least would do if #473 was actually fixed)


src/serena/util/file_system.py
```diff
@@ -138,21 +137,39 @@ def __init__(self, repo_root: str) -> None:
...
+        :return: an iterator yielding paths to .gitignore files (top-down)
         """
-        relative_paths = glob.glob("**/.gitignore", root_dir=self.repo_root, recursive=True)
-        return [os.path.join(self.repo_root, rel_path) for rel_path in relative_paths]
+        queue: list[GitignoreParser._QueueItem] = [self._QueueItem(abs_path=self.repo_root)]
+
+        def scan(abs_path: str, rel_path: str | None):
+            for entry in os.scandir(abs_path):
+                if entry.is_dir(follow_symlinks=follow_symlinks):
+                    dir_rel_path = rel_path + "/" + entry.name if rel_path else entry.name
```
MischaPanch: Is the truthy check on rel_path on purpose? If yes, a small comment on why empty string and None should be treated the same way would be nice. Though I don't see when empty string could ever be passed.

Also, shouldn't os.path.join be used here?

src/serena/util/file_system.py
```diff
@@ -138,21 +137,39 @@ def __init__(self, repo_root: str) -> None:
...
+            log.info("Processing gitignore file: %s", gitignore_path)
+            spec = self._create_ignore_spec(gitignore_path)
             if spec.patterns:  # Only add non-empty specs
                 self.ignore_specs.append(spec)
 
-    def _find_gitignore_files(self) -> list[str]:
+    @dataclass
+    class _QueueItem:
+        abs_path: str
+        rel_path: str | None = None
```
MischaPanch: None is only ever passed for the project root, right? Instead of a default value, we could explicitly pass empty string and have the type being string.

I don't fully understand why we need the absolute path at all, seems like duplicated information if we have the repo root available

MischaPanch: What's the reason for not having a queue of just relative paths?

src/serena/util/file_system.py
opcode81: This line was auto-completed by Copilot; I would not have written it that way myself, but it is ultimately fine, so I kept it.

Using "/" is fine for all platforms and a path can never be an empty string.
The algorithm clearly satisfies the invariant that the relative path is either None or an actual relative path.

src/serena/util/file_system.py
```diff
@@ -138,21 +137,39 @@ def __init__(self, repo_root: str) -> None:
...
+        Directory paths are skipped if they match any already loaded ignore patterns.
 
-        :return: List of absolute paths to .gitignore files
+        :return: an iterator yielding paths to .gitignore files (top-down)
         """
-        relative_paths = glob.glob("**/.gitignore", root_dir=self.repo_root, recursive=True)
-        return [os.path.join(self.repo_root, rel_path) for rel_path in relative_paths]
+        queue: list[GitignoreParser._QueueItem] = [self._QueueItem(abs_path=self.repo_root)]
+
+        def scan(abs_path: str, rel_path: str | None):
```
MischaPanch: On one hand it's better since we're not searching ignored dirs, on the other hand we're back to Python for loops and are undoing the performance benefits of glob. Maybe there's some way to get the best of both worlds? Or maybe we can introduce a flag for controlling behavior here, though of course no user will ever set it..

src/serena/util/file_system.py
opcode81: The reason is: It would be semantically confusing to use "" for the root, because "" is not a path, and using "." would lead to incorrect paths. 
What's the reason you object to the current solution? It is correct and explicit.

src/serena/util/file_system.py
opcode81: Using scandir is very efficient. I don't see a better option.

src/serena/util/file_system.py
MischaPanch: It seems overengineered, I don't see why not just a list of relative paths. Introducing a data structure just for the root to not be listed as "" seems like an overkill. It also allows incorrect state in principle

opcode81: The algorithm is correct and does not produce incorrect states.

MischaPanch: Not arguing with the correctness of the algorithm :)

PR #520, opened 2025-08-22
feat: Add allowed_external_paths configuration for accessing files outside project root

RussEnnisTS:
This feature allows projects to explicitly configure additional directories that Serena can access beyond the project root, addressing use cases like:
- Shared libraries across multiple projects
- External configuration files
- Monorepo scenarios with inter-project dependencies

Implementation details:
- Added allowed_external_paths field to ProjectConfig dataclass
- Enhanced validate_relative_path() to check allowed external paths
- Added path_is_in_allowed_external() method for validation
- Implemented startup warnings when external paths are configured
- Maintains existing security checks (gitignore, symlink resolution)

Security considerations:
- External paths must be explicitly configured in project.yml
- Clear warnings displayed at startup when enabled
- Path traversal attacks still prevented via resolve()
- Gitignore rules still apply to external paths

Testing:
- Comprehensive test suite in test_external_paths.py
- Tests cover absolute/relative paths, symlinks, traversal prevention
- Added serialization test in test_serena_agent.py

Documentation:
- Updated README.md with configuration examples and security warnings
- Added detailed comments in project.template.yml
- Documented in both security and configuration sections

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,9 +336,16 @@ in case the project was activated in the past, by its name. The default project 
 * "Activate the project my_project"
 
 All projects that have been activated will be automatically added to your `serena_config.yml`, and for each
-project, the file `.serena/project.yml` will be generated. You can adjust the latter, e.g., by changing the name
-(which you refer to during the activation) or other options. Make sure to not have two different projects with the
-same name.
+project, the file `.serena/project.yml` will be generated. You can adjust the latter to configure project-specific settings:
+- `project_name`: The name used to activate the project (must be unique)
+- `language`: The primary programming language (python, typescript, java, etc.)
+- `ignored_paths`: Additional paths to ignore (uses gitignore syntax)
+- `read_only`: Set to true to disable all editing tools
+- `allowed_external_paths`: List of paths outside the project that Serena can access (see security warning below)
+- `excluded_tools`: List of tool names to disable for this project
+
+Make sure to not have two different projects with the same name.
+For a complete list of options, see the project template in `src/serena/resources/project.template.yml`.
 
 ℹ️ For larger projects, we recommend that you index your project to accelerate Serena's tools; otherwise the first
 tool application may be very slow.
@@ -527,6 +534,15 @@ If you only want to use Serena purely for analyzing code and suggesting implemen
 without modifying the codebase, you can enable read-only mode by setting `read_only: true` in your project configuration file.
 This will automatically disable all editing tools and prevent any modifications to your codebase while still
 allowing all analysis and exploration capabilities.
+
+In some cases, you may need to allow Serena to access files outside the project directory, such as shared libraries
+or configuration files. You can configure this by adding paths to the `allowed_external_paths` list in your project configuration file:
+```yaml
+allowed_external_paths:
+  - /home/user/shared/libraries
+  - ../shared-config
+```
+**⚠️ Security Warning**: Allowing external paths relaxes Serena's security restrictions. Only add paths that you trust and that are necessary for your project. Serena will display warnings at startup when external paths are configured.
 
 In general, be sure to back up your work and use a version control system in order to avoid
 losing any work.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,6 +62,7 @@ dev = [
   "ruff>=0.0.285",
   "toml-sort>=0.24.2",
   "types-pyyaml>=6.0.12.20241230",
+  "types-requests>=2.32.0",
   "syrupy>=4.9.1",
 ]
 agno = ["agno>=1.2.6", "sqlalchemy>=2.0.40"]
--- a/src/serena/agent.py
+++ b/src/serena/agent.py
@@ -27,6 +27,7 @@ from serena.dashboard import SerenaDashboardAPI
 from serena.project import Project
 from serena.prompt_factory import SerenaPromptFactory
 from serena.tools import ActivateProjectTool, Tool, ToolMarker, ToolRegistry
+from serena.util.file_system import match_path
 from serena.util.inspection import iter_subclasses
 from serena.util.logging import MemoryLogHandler
 from solidlsp import SolidLanguageServer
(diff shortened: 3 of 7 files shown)


MischaPanch:
Hi, thanks for the PR. @opcode81 should be the one to review this, but before reviewing: what problem does this solve? The language server that Serena wraps always runs on one dir. Does this solve a particular need you had, did you need to find symbols in files outside the project?

RussEnnisTS:
> Hi, thanks for the PR. @opcode81 should be the one to review this, but before reviewing: what problem does this solve? The language server that Serena wraps always runs on one dir. Does this solve a particular need you had, did you need to find symbols in files outside the project?

I do a lot of work in an environment where I can't put .dotdirs in the project root, so I have to put claude and serena one level down from where I might naturally want them, and then all the stuff I would want to be in scope as just children of the project root are instead sibling dirs. Claude has the 'additionalDirectories' config that solves this problem for me (I just list the parent or the siblings as additional directories) but I wanted a similar capability for serena so I could have claude and serena have the same project scope & visibility. I imagine others might have similar uses whenever they have cause to the use 'additionalDirectories' feature of claude settings.

MischaPanch:
This would enable some tools of Serena to work on files outside the dir (like search for pattern), but the main tools are using the language server, which is instantiated for a single directory. So I don't think allowing files outside really makes sense. Not allowing dotdirs in the root is a very strange requirement, I don't understand in what situation it makes sense.

If you want to bypass Serena's project level config mechanism, I suggest that you write a custom script where you instantiate the MCP server in sse mode by yourself and don't load the project config from disk, or consider some alternative solution.

I think your case is too specific to include it into our code, we already have many config options and supporting this feature means more maintenance overhead in the future.

@opcode81 wdyt?

RussEnnisTS:
I think I would get most of the benefit I'm looking for by being able to control the project root as being something other than the parent dir of the .serena directory (in all my use cases I actually want it to be one level higher, but the ability to index other directories is potentially useful).

Do you not see the ability to reference (and ideally index) external libraries as useful?

MischaPanch:
> Do you not see the ability to reference (and ideally index) external libraries as useful?

I do, but your changes might not deliver on that. I don't know exactly how the LS and our tools will behave when the repo_dir is not the actual top-level directory. 

The JetBrains extension of Serena that we will publish soon should alleviate all such things as it will use the IDE for providing symbol information and won't even need the .serena directory.

Earlier we had the possibility to configure where the `project.yml` lived, but then decided to fix the location since it is tied to the repo. Instead of permitting other directories, we could add "advanced" settings that allow loading the project.yml from some other dir, as was possible before. Would that solve your case? Alternatively, you could wait for the extension - we have already finished it and are in the process of putting it on the jetbrains marketplace.

RussEnnisTS:
I can see how that would help JetBrains IDE users, but serena is actually pretty good at being useful in other contexts that would not be helped by that. I think if I could have serena "project config dir" (i.e. .serena location) be separate from the "project root" used for indexing (perhaps as separate CLI arguments I could put in my `claude mcp add` arguments) That would solve 90% of my use case. It doesn't help cases where I want better indexing of heavily used external libraries, but I'm fine with losing that.

MischaPanch:
I don't currently know how external libraries are handled, that's kinda up to the LS. For example, intelephense does index files even when they are ignored. No idea how pyright would handle something like a conda env for example, or whether it checks in .venv automatically. So that's a whole separate topic IMO.

For your case, I propose that we close this PR and you open another one that permits starting the MCP server without having a .serena dir in the repo. Would that work for you? It involves customizing things around Project, where the memories live and so on. You'd need to be a bit careful and do things in a backwards compatible way, but from the comments I get the impression that you know how to program, right? :)

RussEnnisTS:
Do you mind if we discuss the proposed API here first? If we can get broad agreement there, then the next PR will have fewer roundtrips.

Two approaches suggest themselves:
1. Have all of the project.yml contents overridable on the CLI, and additionally the cache and memory roots (and still use cwd as index root, though for completeness maybe that should also have a cli arg). Default is to use $(cwd)/,serena/project etc in back-compatible ways, but you could selectively override *anything*.
2. Something more minimal, like cli args to be explicit about where index root and .serena dir are (both defaulting to cwd for back-compatibility)

I lean more towards 2, something like `serena --project /path/to/index/root --project-config /path/to/.serena` where project-config defaults to project/.serena, but I would use --project-config /path/to/index/root/stupid_container/.serena

Does the latter seem reasonable?

MischaPanch:
Yes, agree, let's do that. Time for a video call this weekend? You can reach me at michael.panchenko@oraios-ai.de

MischaPanch:
I think it would be more effective to talk about it in a call if you don't mind, and then @opcode81 could also join if he has time. If you don't want that, we can talk more in this issue and finalize an API proposal, but I feel like it will be a lot of sending messages back and forth

RussEnnisTS:
Synchronous might be tricky in the next couple days. The key seems to be the interaction of these:

```
Usage: serena-mcp-server [OPTIONS]

  Starts the Serena MCP server.

Options:
  --project [PROJECT_NAME|PROJECT_PATH]
                                  Path or name of project to activate at
                                  startup.
  --project-file [PROJECT_NAME|PROJECT_PATH]
                                  [DEPRECATED] Use --project instead.
```

with a proposed new `--project-config`. I would propose we leave those options as current, and add

```
  --project-config [PROJECT_CONFIG_DIR]
                                  Path to serena config directory (defaults to PROJECT_PATH/.serena).
```

The syntax of ~/.serena/serena_config.yml would be extended to support :-delimited registrations listing both the project root and  project config dir (necessary to preserve 'lookup by project name' functionality), back-compatible with just providing the project root as currently:

```
projects:
- /home/rennis/git/project1 # Back-compatible, 
- /home/rennis/git/project2:/home/rennis/git/project2/agentic/.serena # New flexible .serena location deeper in project
```

It becomes a very tightly scoped change to optionally decouple .serena location from the project root. Thoughts?

opcode81:
This is a lot of content to digest - and given the lack of structure in the discussion, I didn't yet.
As a general remark, let me start by saying that such changes should never be made by directly issuing a PR.

First, you need to create an issue that clearly explains which problem needs to be solved. It should also lay out exemplary circumstances under which it occurs (your typical setup).
Then we can devise a solution (software design).
Then someone (perhaps you) can implement it.

Doing it any other way is not efficient, because for any bigger change, the probability of it being implemented in an acceptable way is very low.

opcode81:
From what I understand, your problem is:
* You cannot put `.` directories like `.serena` in the actual project root (say `project/`)
* You need to index several directories but they are actually all in one folder (e.g. `project/dir1`, `project/dir2`)

Your proposed solution is to
* put Serena's data in `project/surrogate/.serena`
* somehow make Serena index `project/` instead of `project/surrogate/`.

This is rather contrived, as literally all IDEs follow the notion of a project directory that contains all relevant content - and Serena follows the same approach.

Potential solutions:

1. Add a new base directory `root/` such that
    * the actual project is in `root/project/`,
    * Serena (and other such things) can be in `root/.serena/`; everything will be indexed correctly.

2. Create a new directory somewhere (could be `project/surrogate/`) and add links there to all directories that shall be indexed. We then need only to make sure that Serena can handle links correctly (which it currently doesn't though).

3. We could reintroduce `project_root` as an optional setting in `project.yml` (we had this initially). It can simply default to `..` relative to the location of `.serena`.

MischaPanch:
After sleeping on it, I woke up feeling that it's too specific for us to address, so I agree with @opcode81 . Every IDE I know, Claude code itself and tons of other related things create a .dir in the project root on startup and this behavior is not configurable. Serena is like an IDE for an agent, so it does the same. In the end, I don't think it should be configurable for us either, it just complicates the config API and maintenance for us. I think for this case you'll have to workaround since it's very specific to you.

MischaPanch:
@opcode81 do you think it's worth reintroducing project root to support this?

The cache and memories are loaded from .serena, changes might be needed there

opcode81:
Reintroducing `project_root` is a simple solution, I'd say. Should be a small change.

But it is ultimately a very weird constraint that leads us to do this. How is it possible that `.` directories can be disallowed?

opcode81:
But it's still not very nice, because we have a discrepancy:
When specifying command line args, users are expected specifiy the directory containing `.serena` (and we call that the *project root*). The notion of a project root that differs from that will lead to confusion.
I suppose it can be done with corresponding explanations, but if it isn't "necessary", I'd rather avoid it.

opcode81:
For the specific case here, I can also think of another solution:

4. Make the directory name ".serena" configurable in `serena_config.yml`.

RussEnnisTS:
The specific constraint I'm working with comes from using https://github.com/twosigma/git-meta to manage a very large codebase with 100s or 1000s of git submodules, and there are restrictions on the location of particular kinds of files and directories. Most everything has to live in a git submodule, so if I want my claude session and serena project to span several related submodules, I have to pick/create one submodule for the config etc to live (and submodules cannot be nested). I am constrained from just putting random things at a convenient 'root node' unless I have it outside the project hierarchy entirely, and that is also less good because I want things like the serena memories to be in the same version control system.  Claude has the 'additionalDirectories' config that allows me to stuff it in a module and still see the sibling modules, but for serena because of the language server needing a singular root, I want to be able to configure its root for indexing (which does not need to be in a module) to be a distinct directory to the root where I have to host the serena config and actual index files (which must be in a module so I can add the serena config and memories to git; the index caches I .gitignore and rebuild for each new clone).

Basically anything that allows me to decouple the project/index root from where the config files & memories I need to commit to version control would solve my problem.

opcode81:
Can you comment specifically on the four solutions I suggested?

RussEnnisTS:
> Potential solutions:
> 
> 1. Add a new base directory root/ such that
> 
> the actual project is in root/project/,
> Serena (and other such things) can be in root/.serena/; everything will be indexed correctly.
> 
> 2. Create a new directory somewhere (could be project/surrogate/) and add links there to all directories that shall be indexed. We then need only to make sure that Serena can handle links correctly (which it currently doesn't though).
> 
> 3. We could reintroduce project_root as an optional setting in project.yml (we had this initially). It can simply default to .. relative to the location of .serena.
> 
> 4. Make the directory name ".serena" configurable in serena_config.yml
 
1. Doesn't work with my monorepo constraints.
2. Symlinks typically have a host of issues with them that mean we are strongly discouraged (perhaps forbidden, I can't remember) from using them inside our monorepo structure.
3. This I think might solve my problem
4. This might solve my problem, I'm not sure if I've understood the suggestion; are you proposing something like the config change I proposed at the bottom of this comment above? https://github.com/oraios/serena/pull/520#issuecomment-3215934444 :

> The syntax of ~/.serena/serena_config.yml would be extended to support :-delimited registrations listing both the project root and project config dir (necessary to preserve 'lookup by project name' functionality), back-compatible with just providing the project root as currently:
> 
> projects:
> - /home/rennis/git/project1 # Back-compatible, 
> - /home/rennis/git/project2:/home/rennis/git/project2/agentic/.serena # New flexible .serena location deeper in project
> It becomes a very tightly scoped change to optionally decouple .serena location from the project root. Thoughts?

I thought you might need 3 & 4 together (which is basically what I suggested in the linked comment above) in order to maintain referential integrity between the user config in ~/.serena/serena_config.yml and the now decoupled project root and project config dir.

RussEnnisTS:
You could simplify the serena_project.yml I guess if you could have it point to either the project root (as currently) or the project.yml exactly (not the directory, the file itself) if the project.yml specifies an index root other than ".."; then you are maintaining single-point of administration of the index root (in the project.yml) which is better than what I originally suggested.

RussEnnisTS:
To clarify in a single comment, what if serena_config.yml could have:

```
projects:
- /home/rennis/git/project1 # Back-compatible, 
- /home/rennis/git/project2/agentic/.serena/project.yml # New flexible exact project.yml anywhere you want
```

and then in project.yml for the second:
```
# /home/rennis/git/project2/agentic/.serena/project.yml
project_root: "../.."
```

and the first /home/rennis/git/project1/.serena/project.yml in need not specify project_root because it is the default "..". Relative paths are used because the git clone root might vary by user/fix.

MischaPanch:
Yes, that's exactly how the config was before we had fixed the project config location to the repo root. You'd just need to check that the language server cache and the memories are also handled appropriately, as they are all written to `project_root/.serena`.

I don't mind introducing this option, @opcode81 wdyt?

RussEnnisTS:
Would the cache and memory roots also need config relative to the project.yml, e.g. 

```
cache_root: "./cache"
memory_root: "./memories"
```

? I don't need that configurability, but I would want to make sure the memories were stored next to the project config, not as a child of the project root.

MischaPanch:
I am afraid we must add that too, otherwise there may be path conflicts. E.g., a user may have multiple project.yml within the same directory, so we can't just assume where other saved files should go

MischaPanch:
When doing this, please make sure to add explanations in the `project.template.yml` that for the absolute majority of users, the default values should be left untouched, and maybe a small explanation on why one would ever want to touch them (like the git-meta case, or claude code with additional directories)

opcode81:
> Would the cache and memory roots also need config relative to the project.yml, e.g. 
>
>```
>cache_root: "./cache"
>memory_root: "./memories"
>```

> I am afraid we must add that too

I think we should instead configure just one folder, the serena data folder (which is in `project_root/.serena`) by default,
i.e.

```
serena_data_path: $project_root/.serena
```

This is also more scalable, should we add further subfolders later. There is really no reason to store Serena's data for a project in anything other than a single folder.

opcode81:
> 4. This might solve my problem, I'm not sure if I've understood the suggestion

I was suggesting to make the default name of the Serena data folder (".serena") configurable.
Since your problem was that you cannot have `.serena` in your project root, maybe you can have `serena` or something else...

opcode81:
> 1. Doesn't work with my monorepo constraints.

I don't understand why it would violate your contraints. `root/project/` can still be exactly the monorepo you had before. You just have an added parent folder.

opcode81:
Closing this, as it is clear that the code being added by this PR is not something we want to merge. We can discuss other options to address the problem in a new issue.

PR #524, opened 2025-08-23
Allow passing LS implementation specific options.

MischaPanch:
Made use of in csharp and intelephense LS

@opcode81 : it would probably have been better to configure ls_specifics in serena_config instead of the project config, but that would have required a refactoring and I didn't want to do that (not important enough IMO)

### DIFF
--- a/src/serena/project.py
+++ b/src/serena/project.py
@@ -306,5 +306,9 @@ class Project:
             ls_logger,
             self.project_root,
             timeout=ls_timeout,
-            solidlsp_settings=SolidLSPSettings(solidlsp_dir=SERENA_MANAGED_DIR_IN_HOME, project_data_relative_path=SERENA_MANAGED_DIR_NAME, ls_specific_settings=ls_specific_settings),
+            solidlsp_settings=SolidLSPSettings(
+                solidlsp_dir=SERENA_MANAGED_DIR_IN_HOME,
+                project_data_relative_path=SERENA_MANAGED_DIR_NAME,
+                ls_specific_settings=ls_specific_settings,
+            ),
         )
--- a/src/solidlsp/language_servers/csharp_language_server.py
+++ b/src/solidlsp/language_servers/csharp_language_server.py
@@ -437,7 +437,9 @@ class CSharpLanguageServer(SolidLanguageServer):
         logger.log("Downloading .NET 9 runtime...", logging.INFO)
         dotnet_dir.mkdir(parents=True, exist_ok=True)
 
-        custom_dotnet_runtime_url = solidlsp_settings.ls_specific_settings.get(cls.get_language_enum_instance(), {}).get("dotnet_runtime_url")
+        custom_dotnet_runtime_url = solidlsp_settings.ls_specific_settings.get(cls.get_language_enum_instance(), {}).get(
+            "dotnet_runtime_url"
+        )
         if custom_dotnet_runtime_url is not None:
             logger.log(f"Using custom .NET runtime url: {custom_dotnet_runtime_url}", logging.INFO)
             url = custom_dotnet_runtime_url
--- a/src/solidlsp/ls_config.py
+++ b/src/solidlsp/ls_config.py
@@ -143,78 +143,103 @@ class Language(str, Enum):
         match self:
             case self.PYTHON:
                 from solidlsp.language_servers.pyright_server import PyrightServer
+
                 return PyrightServer
             case self.PYTHON_JEDI:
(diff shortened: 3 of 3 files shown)


MischaPanch:
@johngambleubind took a bit longer, but here it is: the possibility to configure the download path through adjusting project settings.

If some setting is missing, you can build on top of this and add it in a PR

opcode81:
> @opcode81 : it would probably have been better to configure ls_specifics in serena_config instead of the project config, but that would have required a refactoring and I didn't want to do that (not important enough IMO)

We discussed that it should be in `serena_config`.
Why would a refactoring be required? `Project.create_language_server` can be passed the dictionary from `SerenaConfig`, just like it is passed `trace_lsp_communication` from there, too. We could even just pass the `SerenaConfig` instance.

src/solidlsp/language_servers/intelephense.py
```diff
@@ -123,12 +126,21 @@ def _get_initialize_params(repository_absolute_path: str) -> InitializeParams:
...
         }
-
+        initialization_options = {}
         # Add license key if provided via environment variable
         license_key = os.environ.get("INTELEPHENSE_LICENSE_KEY")
         if license_key:
-            initialize_params["initializationOptions"] = {"licenceKey": license_key}
+            initialization_options["licenceKey"] = license_key
+
+        custom_intelephense_settings = self._solidlsp_settings.ls_specifics.get(self.__class__.__name__, {})
```
opcode81: We originally discussed that we would use the enum item's value that is associated with the language server.
Do you think the class name is a better option?
Right now, the `Language` enum is actually a `LanguageServer` enum - and perhaps we should rename it accordingly - and use it accordingly.

opcode81: I think the enum value would be preferable, because the user is already familiar with the respective values.
And we should probably document the possible options in `serena_config.yml` to make this accessible.

MischaPanch: Done

MischaPanch:
@opcode81 Since I did not want to repeat the code connecting the enum to the class, I have "simplified" the `SolidLanguagServer.create` method and clearly commented on the unusual pattern used there. Not super nice, but IMO better than copy-pasting the whole logic and with the comment it's rather clear. Later on we can generally improve the create/init separation

src/solidlsp/language_servers/intelephense.py
opcode81: The documentation in `serena_config.template..yml` is not updated to describe the new usage and does not list the possible options. Still says this:
> Maps the language server class name to the options. Have a look at the docstring of the constructors of the LS implementations ...

src/serena/project.py
```diff
@@ -266,6 +267,7 @@ def create_language_server(
         log_level: int = logging.INFO,
         ls_timeout: float | None = DEFAULT_TOOL_TIMEOUT - 5,
         trace_lsp_communication: bool = False,
+        ls_specifics: dict[Language, Any] | None = None,
```
opcode81: Shall we just add SerenaConfig as an argument? What do you think?
(I am not sure what I prefer)

src/serena/resources/serena_config.template.yml
```diff
@@ -33,6 +33,12 @@ trace_lsp_communication: False
 # whether to trace the communication between Serena and the language servers.
 # This is useful for debugging language server issues.
 
+ls_specifics: {}
```
opcode81: Do you like the name "ls_specifics"? I think it's a bit vague.
How about "ls_specific_settings" or "ls_advanced_settings"?

opcode81:
> @opcode81 Since I did not want to repeat the code connecting the enum to the class, I have "simplified" the `SolidLanguagServer.create` method and clearly commented on the unusual pattern used there. Not super nice, but IMO better than copy-pasting the whole logic and with the comment it's rather clear. Later on we can generally improve the create/init separation

Agreed. Not super nice, but good enough.

src/serena/resources/serena_config.template.yml
MischaPanch: renamed to ls_specific_settings

src/serena/project.py
MischaPanch: we can change that later if needed, leaving as is for now

ctf0:
@MischaPanch  how to use the new config `ls_specific_settings` ?

MischaPanch:
@ctf0 it's documented in the serena_config.template.yml

I'll add some info in the changelog and readme in the next commits to main

PR #526, opened 2025-08-23
Add nix flake

jcdickinson:
Adds a nix flake to the project so that nix users can easily execute it.

Prettier seems to have cleaned up some whitespace in the README.


MischaPanch:
Sure, why not, it doesn't hurt. Thanks for the contribution, nix users will be happy

PR #533, opened 2025-08-25
Fix: Extend Nix symbol ranges to include trailing semicolons

splack:
This fixes the double semicolon issue in replace_symbol_body when replacing Nix attributes with content that includes semicolons.

Root cause: nixd provides expression-level ranges (excluding semicolons) but serena expects statement-level ranges (including semicolons).

Solution: Override request_document_symbols() in NixLanguageServer to detect and include trailing semicolons in symbol ranges.

### DIFF
--- a/src/solidlsp/language_servers/nixd_ls.py
+++ b/src/solidlsp/language_servers/nixd_ls.py
@@ -15,6 +15,7 @@ from pathlib import Path
 
 from overrides import override
 
+from solidlsp import ls_types
 from solidlsp.ls import SolidLanguageServer
 from solidlsp.ls_config import LanguageServerConfig
 from solidlsp.ls_logger import LanguageServerLogger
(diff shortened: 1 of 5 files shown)


src/solidlsp/language_servers/nixd_ls.py
```diff
@@ -28,6 +29,82 @@ class NixLanguageServer(SolidLanguageServer):
...
+        if not symbol.get("range"):
+            return symbol
+
+        range_info = symbol["range"]
+        end_line = range_info["end"]["line"]
+        end_char = range_info["end"]["character"]
+
+        # Split file content into lines
+        lines = file_content.split("\n")
+        if end_line >= len(lines):
```
MischaPanch, line 47: when can that happen?

src/solidlsp/language_servers/nixd_ls.py
```diff
@@ -28,6 +29,82 @@ class NixLanguageServer(SolidLanguageServer):
...
+    def _extend_nix_symbol_range_to_include_semicolon(
+        self, symbol: ls_types.UnifiedSymbolInformation, file_content: str
+    ) -> ls_types.UnifiedSymbolInformation:
+        """
+        Extend symbol range to include trailing semicolon for Nix attribute symbols.
+
+        nixd provides ranges that exclude semicolons (expression-level), but serena needs
+        statement-level ranges that include semicolons for proper replacement.
+        """
+        if not symbol.get("range"):
```
MischaPanch: can that ever happen?

src/solidlsp/language_servers/nixd_ls.py
```diff
@@ -28,6 +29,82 @@ class NixLanguageServer(SolidLanguageServer):
...
+
+        # Split file content into lines
+        lines = file_content.split("\n")
+        if end_line >= len(lines):
+            return symbol
+
+        line = lines[end_line]
+
+        # Check if there's a semicolon immediately after the current range end
+        if end_char < len(line) and line[end_char] == ";":
```
MischaPanch, line 53: is the first condition not always true?

src/solidlsp/language_servers/nixd_ls.py
splack: Should not as long as nixd is doing it's thing correctly. I'll nuke it.

src/solidlsp/language_servers/nixd_ls.py
splack: Maybe not. If the code being replaced it malformed, i.e. no trailing semicolon then the end would be the eol. I'm not sure if the selection would always require a trailing semicolon. If so, then yes, then the first check is redundant.

MischaPanch:
Almost there, thanks for the snapshot test!

There are some differences in whitespace in the snapshot and the result, that's why it's failing. It should also be skipped on windows

splack:
nixd runs the formatter so I let it change the source so it should match the snapshot.

splack:
The formatting changed the line for one of the other nix tests.

MischaPanch:
The C# test failure is not on you, it happens randomly, I will increase the sleep time on main

splack:
Odd, I'm not sure what's going on with that c# test. Maybe something in the snapshot changed for it or a timeout? I only run nix and there were issues with getting dotnet in the flake devshell so I can only test this via ci. Sorry for the fail noise.

MischaPanch:
Everything looks good now, thanks for the contribution!

splack:
Thanks for the great tool! It's much better then my collection of tree-sitter stuff I was using.

Issue #535, opened 2025-08-25
Migrate Ruby LSP from Solargraph to ruby-lsp


kiakiraki:
## Background

Currently, Serena MCP server uses Solargraph as the language server for Ruby.
However, in our very large codebase, the following issues have become apparent:

- Symbol search does not work properly and causes the server to hang
- Performance degradation and slow responses due to the project’s massive scale

As a result, it has become difficult to maintain a stable and practical development experience.

On the other hand, ruby-lsp is actively maintained by Shopify and provides better compatibility with modern LSP features and IDEs.

## PoC Results

To address the problems with Solargraph, we created a PoC by integrating ruby-lsp into the kiakiraki/serena fork.
The results were:

- All LSP features, including Symbol search, worked smoothly
- The hanging issue was completely resolved
- Code completion and go-to-definition worked reliably even in a very large codebase

These improvements strongly suggest ruby-lsp is a better fit for our environment.

## Proposal

We propose migrating Ruby support from Solargraph to ruby-lsp.
Expected benefits include:

- Improved stability and responsiveness in very large codebases
- Long-term maintainability through active development
- Stronger support for Ruby 3.x and beyond
- Smooth adoption of advanced LSP features (code actions, diagnostics, etc.)

## Next Steps

If a decision is made to proceed, we are ready to open a PR migrating to ruby-lsp.

MischaPanch:
Yes, sounds good! You can do it like this: adjust the enum value in Language from ruby to ruby_solargraph and mark it as experimental.

Then add ruby-lsp as the main ruby LS implementation. This way users can still manually configure using solargraph, if they so desire

kiakiraki:
resolved by https://github.com/oraios/serena/pull/540

PR #540, opened 2025-08-25
Add Ruby dual language server support with ruby-lsp as primary and Solargraph as experimental

kiakiraki:
## Summary

This PR implements dual language server support for Ruby, addressing the reviewer feedback in issue #535. It adds [Shopify's
ruby-lsp](https://github.com/Shopify/ruby-lsp) as the modern primary Ruby language server while maintaining backward compatibility
with Solargraph as an experimental legacy option.

### Language Server Configuration and Ruby-LSP Implementation (`src/solidlsp/language_servers/ruby_lsp.py`)
- **Added `ruby-lsp`** as the primary Ruby language server (`Language.RUBY`)
- **Marked Solargraph as experimental** (`Language.RUBY_SOLARGRAPH`) for legacy users
- **File extension support**: ruby-lsp supports both `.rb` and `.erb` files, Solargraph supports `.rb` only

### Backward Compatibility
- Existing Solargraph users can continue using `ruby_solargraph` language setting


MischaPanch:
Looks good, thanks for the clean PR! I'll push a minor follow-up change on main, merging this now. Ruby folks should be happy about this

Issue #121, opened 2025-05-19
Add text file encoding to project configuration (with utf-8 as default)


opcode81:
This is wrong. The correct fix is to specify the encoding in the project configuration. We cannot just assume an encoding.

_Originally posted by @opcode81 in https://github.com/oraios/serena/issues/110#issuecomment-2892003322_

MischaPanch:
I thought we had agreed on only using utf-8 everywhere and not asking the user to configure it? The loop over encodings was there before my last commit, it slipped my sight. We can just remove it

opcode81:
We agreed that utf-8 is a sensible default until we handle this properly. This ticket is about handling it properly.
There is no alternative to allowing the user to configure this (unless you want to enter a heuristics-based guessing game). We will not ask (= require) the user to configure it but we will allow it. 
For the vast majority of users, utf-8 is fine.

MischaPanch:
@opcode81 this is finished, right?

opcode81:
Yes, this was done quite some time ago.

PR #550, opened 2025-08-27
fix: improve file encoding support and error handling during indexing

grumpygrover:
- Add multi-encoding fallback support (utf-8, utf-8-sig, latin-1, cp1252) to FileUtils.read_file()
- Change indexing error handling from TimeoutError to Exception to catch encoding failures
- Files with unsupported encodings now skip gracefully instead of crashing indexing

Fixes issues with indexing projects containing non-UTF-8 encoded files (e.g., legacy Windows files with special characters).

### DIFF
--- a/src/serena/cli.py
+++ b/src/serena/cli.py
@@ -464,7 +464,7 @@ class ProjectCommands(AutoRegisteringGroup):
                 try:
                     ls.request_document_symbols(f, include_body=False)
                     ls.request_document_symbols(f, include_body=True)
-                except TimeoutError as e:
+                except Exception as e:
                     log.error(f"Failed to index {f}, continuing.")
                     collected_exceptions.append(e)
                     files_failed.append(f)
--- a/src/solidlsp/ls_utils.py
+++ b/src/solidlsp/ls_utils.py
@@ -167,16 +167,28 @@ class FileUtils:
     def read_file(logger: LanguageServerLogger, file_path: str) -> str:
         """
         Reads the file at the given path and returns the contents as a string.
+        Tries multiple encodings in order: utf-8, utf-8-sig, latin-1, cp1252
         """
         if not os.path.exists(file_path):
             logger.log(f"File read '{file_path}' failed: File does not exist.", logging.ERROR)
             raise SolidLSPException(f"File read '{file_path}' failed: File does not exist.")
-        try:
-            with open(file_path, encoding="utf-8") as inp_file:
-                return inp_file.read()
-        except Exception as exc:
-            logger.log(f"File read '{file_path}' failed to read with encoding 'utf-8': {exc}", logging.ERROR)
-            raise SolidLSPException("File read failed.") from None
+
+        encodings = ["utf-8", "utf-8-sig", "latin-1", "cp1252"]
+        last_exception = None
+
+        for encoding in encodings:
+            try:
+                with open(file_path, encoding=encoding) as inp_file:
+                    content = inp_file.read()
+                    if encoding != "utf-8":
+                        logger.log(f"File read '{file_path}' succeeded with encoding '{encoding}'", logging.DEBUG)
+                    return content
+            except Exception as exc:
+                last_exception = exc
+                continue
+
+        logger.log(f"File read '{file_path}' failed with all encodings {encodings}: {last_exception}", logging.ERROR)
+        raise SolidLSPException("File read failed.") from None
 
     @staticmethod
     def download_file(logger: LanguageServerLogger, url: str, target_path: str) -> None:


src/solidlsp/ls_utils.py
```diff
@@ -167,16 +167,28 @@ class FileUtils:
...
             logger.log(f"File read '{file_path}' failed: File does not exist.", logging.ERROR)
             raise SolidLSPException(f"File read '{file_path}' failed: File does not exist.")
-        try:
-            with open(file_path, encoding="utf-8") as inp_file:
-                return inp_file.read()
-        except Exception as exc:
-            logger.log(f"File read '{file_path}' failed to read with encoding 'utf-8': {exc}", logging.ERROR)
-            raise SolidLSPException("File read failed.") from None
+
+        encodings = ["utf-8", "utf-8-sig", "latin-1", "cp1252"]
```
MischaPanch: We were considering making the encoding configurable in #121 . Would that solve your use case? How did you run into this problem?

src/serena/cli.py
```diff
@@ -464,7 +464,7 @@ def _index_project(project: str, log_level: str, timeout: float) -> None:
                 try:
                     ls.request_document_symbols(f, include_body=False)
                     ls.request_document_symbols(f, include_body=True)
-                except TimeoutError as e:
+                except Exception as e:
```
MischaPanch, line 467: I added this to main now, it's somewhat independent of the encoding issue

src/solidlsp/ls_utils.py
grumpygrover: I was trying to make Serena work on a legacy codebase where some files had encoding issues (cp-1252) with some strings, and it was crashing the indexing. After applying this fix, the process does not crash, but the list of encodings is probably very short. I don't know enough about encodings to make this a config option; my approach was to try with a few encodings and then skip the file if it was not able to work with it.

grumpygrover:
@MischaPanch summary of new changes

  1. Enhanced File Encoding Detection (src/solidlsp/ls_utils.py)

  Major Enhancement:
  - Replaced basic encoding fallback with intelligent encoding detection using charset-normalizer
  - Two-phase approach:
    a. Fast path: Try UTF-8 first (most common case)
    b. Smart detection: Use charset-normalizer for reliable encoding detection on UTF-8 failure

  Key Features:
  - Confidence-based fallbacks: When detection confidence is <50%, tries common encodings (latin-1, cp1252, utf-8-sig) before using detected encoding
  - Comprehensive error handling: Better error messages and exception handling throughout the process
  - Performance optimized: UTF-8 fast path for the majority of files
  - Detailed logging: Informative messages about encoding detection process (at DEBUG level)

  Technical Details:
  - Added charset_normalizer import for robust encoding detection
  - Replaced simple encoding list iteration with sophisticated detection algorithm
  - Added confidence scoring and fallback logic
  - Enhanced exception handling with proper error propagation
 
  Impact

  - More reliable file reading for projects with mixed encodings
  - Better error messages when encoding issues occur
  - Reduced log noise during normal operation
  - Improved performance with UTF-8 fast path
  - Enhanced maintainability with better type annotations
 
 
 Maybe this approach is a bit more predictable, and is not tied to specific encoding conversions, using chardet

MischaPanch:
Thanks! @opcode81 will review this when he has time, he has thought more about encoding issues with Serena

gsw945:
Many thanks for this PR, particularly useful in the context of Windows system. We look forward to its early merged.

opcode81:
This PR makes too many assumptions, which are inappropriate.

@gsw945 Serena supports explicitly defining the encoding in your project.yml (see #121). For example, you could set:

    encoding: "windows-1250"

Your project should consistenly use a single encoding.

Find the list of supported encodings [here](https://docs.python.org/3.11/library/codecs.html#standard-encodings).

I have just added a commit (563e42efbdf4d6424bc4af65f926a984b32f953a) on `main` which explicitly exposes the setting in the project.yml template.

I am closing this PR, because the fallback logic is arbitrary and therefore undesirable.

opcode81:
> Serena supports explicitly defining the encoding in your project.yml

The project's encoding setting was not consistently passed on; this is addressed in #674.

gsw945:
Yes! Your project should consistenly use a single encoding.

However, in certain projects, even when the majority of files have been standardised to UTF-8 encoding, individual files may still retain different encodings.
Take the project I am currently analysing as an example—this large-scale Unity project contains numerous Windows batch files. Enabling UTF-8 encoding for these batch files would cause issues when running on Windows system in CJK language environment (such as garbled characters appearing in comments and CJK text output displaying incorrectly).

gsw945:
I have create a new PR https://github.com/oraios/serena/pull/682 , It does not make many assumptions about encoding; it merely provides an additional opportunity when reading fails via the default encoding.

PR #566, opened 2025-08-28
test: skip Go tests when gopls is not installed

thilanga:
## Summary
- Add gopls availability check utility to detect when gopls is not installed
- Update Go tests to skip gracefully when gopls is unavailable instead of failing
- Follow the same pattern used by other language tests (Elixir, Erlang, Clojure)

## Problem
Go tests were failing when gopls (Go language server) was not installed, causing CI/test failures for developers who don't have Go development tools set up.

## Solution
Added a skipif condition that checks for gopls availability before running Go-specific tests. This matches the pattern already established in the codebase for other language servers.

## Test Plan
- [x] Tests skip correctly when gopls is not installed
- [x] Tests run normally when gopls is installed
- [x] Other language tests continue to work as expected

### DIFF
--- a/src/serena/agent.py
+++ b/src/serena/agent.py
@@ -603,21 +603,63 @@ class SerenaAgent:
         assert self.lines_read is not None
         self.lines_read.invalidate_lines_read(relative_path)
 
+    def __enter__(self) -> "SerenaAgent":
+        """Context manager entry - returns self for use in with statements"""
+        return self
+
+    def __exit__(self, exc_type: type[BaseException] | None, exc_val: BaseException | None, exc_tb: Any) -> None:
+        """Context manager exit - ensures proper cleanup of language server"""
+        self.cleanup()
+
+    def cleanup(self) -> None:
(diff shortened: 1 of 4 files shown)


src/serena/agent.py
```diff
@@ -603,21 +603,63 @@ def mark_file_modified(self, relative_path: str) -> None:
...
         self.lines_read.invalidate_lines_read(relative_path)
 
-    def __del__(self) -> None:
-        """
-        Destructor to clean up the language server instance and GUI logger
-        """
-        if not hasattr(self, "_is_initialized"):
-            return
-        log.info("SerenaAgent is shutting down ...")
+    def __enter__(self) -> "SerenaAgent":
```
MischaPanch, line 606: This change has nothing to do with go and you don't also mention it in the PR description. Is it accidental?

MischaPanch: I see, it builds on your other PR, so probably an accident. Pls remove the unrelated commit from this, then we can merge the skipping part now

PR #567, opened 2025-08-28
fix: improve TypeScript server cleanup to prevent process accumulation

thilanga:
## Summary
- Refactor TypeScript server cleanup to prevent zombie processes
- Improve exception handling and add proper type hints
- Add debug logging for better troubleshooting

## Problem
TypeScript server processes (tsserver, typescript-language-server, typingsInstaller) were not being properly cleaned up, leading to process accumulation that could cause system crashes over time.

## Solution
- Extracted TypeScript cleanup logic into a dedicated `_cleanup_typescript_processes()` method
- Improved exception handling with specific exception types (OSError, TimeoutError, etc.)
- Added proper type hints for the `__exit__` method parameter
- Added debug logging to track cleanup failures
- Ensures all TypeScript-related processes are killed on Unix-like systems

## Changes
- Added `_cleanup_typescript_processes()` method to handle TypeScript server cleanup
- Updated exception handling to use specific exception types instead of bare `except`
- Added `TracebackType` type hint for `__exit__` method
- Added debug logging for failed cleanup operations

## Test Plan
- [x] Code passes type checking (`mypy`)
- [x] Code is properly formatted
- [x] Existing tests continue to pass

### DIFF
--- a/src/serena/agent.py
+++ b/src/serena/agent.py
@@ -5,6 +5,7 @@ The Serena Model Context Protocol (MCP) Server
 import multiprocessing
 import os
 import platform
+import subprocess
 import sys
 import threading
 import webbrowser
@@ -13,6 +14,7 @@ from collections.abc import Callable
 from concurrent.futures import Future, ThreadPoolExecutor
 from logging import Logger
 from pathlib import Path
+from types import TracebackType
 from typing import TYPE_CHECKING, Any, Optional, TypeVar
 
 from sensai.util import logging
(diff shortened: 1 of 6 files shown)


src/serena/agent.py
```diff
@@ -603,21 +605,74 @@ def mark_file_modified(self, relative_path: str) -> None:
...
         self.lines_read.invalidate_lines_read(relative_path)
 
-    def __del__(self) -> None:
-        """
-        Destructor to clean up the language server instance and GUI logger
-        """
-        if not hasattr(self, "_is_initialized"):
-            return
-        log.info("SerenaAgent is shutting down ...")
+    def __enter__(self) -> "SerenaAgent":
```
MischaPanch: SerenaAgent is never used with `with`, and I don't see why it should. Do you actually need this for a more graceful shutdown?

thilanga: I was implementing a better GC logic. I have been testing the code a bit. Will update the PR soon. Looks like I got the matter under control. Now Serena doesn't use much memory

thilanga:
PR has been updated - rebased on main with the latest dead code cleanup changes

thilanga:
I need to revisit my implementation. This is not good enough

PR #589, opened 2025-09-02
fix(terraformls): add linux arm64 runtime dependency

snoord:
Fixes an issue where serena could not use the terraform-ls lsp when running on linux/arm64


MischaPanch:
Thanks! Merging :)

PR #593, opened 2025-09-02
Add support for AL programming language

SShadowS:
Windows and WSL tested.

Don't have a pure Linux dev env for AL, as it is not really used with this language, yet.


MischaPanch:
This is very interesting - not because of the functionality (I don't mind supporting AL), but because of the structure.

Could you give a summary on what exactly happens here? Why is a vscode extension needed and what role does it play? Tests run through in CI without VSCode being involved, right?

Why did you have to overwrite the request_document_symbols? No other LS does this.

So, overall a summary of the structure of the solution would help me in understanding what's going on :)

SShadowS:
Here's what's unique about the AL language server implementation:

1. VSCode Extension as Language Server

The AL language server isn't a standalone binary like gopls or rust-analyzer. Instead, it's embedded within the VS
 Code AL extension (ms-dynamics-smb.al). The extension contains:
- Platform-specific executables (Microsoft.Dynamics.Nav.EditorServices.Host.exe for Windows)
- These executables ARE the language server - they communicate via LSP over stdio
- It can also be downloaded via NuGet, but didn't want to have that as a prerequisite.
- I look for the Extension, so we don't have to download the total 280MB file if they already have it.

2. Automatic Download Mechanism

Since AL doesn't have a standalone package manager installation:
- The implementation downloads the entire VS Code extension (VSIX file) from the marketplace
- Extracts it to ~/.serena/language_servers/al-extension/
- Uses the embedded language server executable directly
- This allows tests to run in CI without VS Code or manual installation, but CAN use them.

3. Why request_document_symbols Override?

AL's language server has a unique quirk:
- Most language servers: Accept symbol requests for any file path, whether opened or not
- AL server: Returns empty/minimal symbols unless files are explicitly opened via textDocument/didOpen, think it is a VSCode thing.
- The override opens files before requesting symbols, then closes them
- This is why request_full_symbol_tree also needed overriding - to handle multi-file symbol collection

4. Custom AL-Specific Commands

AL server uses non-standard LSP extensions (Took me a while to sort out):
- al/gotodefinition instead of standard textDocument/definition
- al/setActiveWorkspace for workspace management
- al/hasProjectClosureLoadedRequest to check project loading status
- These require special handling during initialization

5. Complex'ish Initialization Sequence

AL requires:
1. Standard LSP initialization
2. Workspace configuration via workspace/didChangeConfiguration
3. Opening app.json to trigger project loading (Reads which keywords are supported in a given version, I think.)
4. Setting active workspace
5. Waiting for project to be fully indexed

Key Difference from Other Language Servers

Most language servers (gopls, rust-analyzer, pyright) are:
- Standalone binaries
- Installed via package managers
- Work immediately after initialize/initialized
- Don't require file opening for symbols

AL is unique because it's:
- Part of a VS Code extension ecosystem
- Requires explicit file management
- Has Business Central-specific initialization
- Designed primarily for VS Code integration

Hope that answered your questions. However, I just noticed that I made a merge error in the README.md file of all things, which I will sort out tomorrow. But any feedback is appreciated, as Python isn't my strong side.

MischaPanch:
Thanks for the detailed explanation, very helpful!

> AL server uses non-standard LSP extensions (Took me a while to sort out)

Amazing stuff, especially given that ALL related technologies (LSP, VSCode, AL and the extension) are created by the same company.

I'm inclined to merging this, will give it a proper review asap (likely won't have time before next week).

But out of curiosity - would the same approach work with any VSCode language extension (assuming they, unlike AL, actually implement the LSP)? Would it make sense to have a general abstraction for that such that any VSCode extension could be turned into a language server for Serena?

MischaPanch:
Btw, I think it makes a lot of sense to add many more comments to the code, since it's very different from support for other languages and will raise eyebrows

SShadowS:
I haven't checked other extensions layout, was pretty focus on just this one. Any specific languages you would like to add? Then I can check a few. Just doesn't make sense to pick any of the existing ones.

I will smother it with comments. 🙂

MischaPanch:
> I haven't checked other extensions layout, was pretty focus on just this one. Any specific languages you would like to add? Then I can check a few. Just doesn't make sense to pick any of the existing ones.

It's just a thought, I am not entirely sure that even makes sense. I guess in most or maybe all cases it would be better to download a binary instead of a vscode extension.

But the approach you took just struck really close to several ideas that I was investigating. For people using VSCode, one could write a simple VSC extension that exposes the language server API via websockets or http. Then one could make a `class VSCodeLS(SolidLanguageServer)` that connects to this. Voila - now you can use Serena on any language without having to download anything. A few more details are needed (which instance of vsc connects to which instance of serena), but overall this should work. I think, it might even be not all too much effort.

It's similar to what is done in [this project,](https://github.com/jerry426/token-saver-mcp) I believe. [This extension](https://github.com/estruyf/vscode-remote-control) could also be a good starting point.

Anyway, that's the background of the question, let me know what you think. If by any chance you want to try out implementing what I outlined above, that would be an awesome contribution to Serena. Otherwise, we will probably get to do that at some point in the future.

> I will smother it with comments. 🙂

Thanks, lol

SShadowS:
Now with a lot more comments in the code.

src/solidlsp/language_servers/al_language_server.py
```diff
@@ -0,0 +1,1045 @@
...
+        # Check if the extension is already installed
+        if not os.path.exists(executable_path):
+            logger.log(f"AL Language Server not found at {executable_path}. Downloading from VS Code marketplace...", logging.INFO)
+
+            # AL extension version - using latest stable version
+            AL_VERSION = "latest"
+            url = f"https://marketplace.visualstudio.com/_apis/public/gallery/publishers/ms-dynamics-smb/vsextensions/al/{AL_VERSION}/vspackage"
+
+            # Try to download the extension
+            if cls._download_al_extension(logger, url, al_extension_dir):
```
MischaPanch: Maybe reverse is a better logic - first try finding the installed one, then downloading? It would prevent unnecessary downloads.

I'd also prefer the error handling and logging to be handled in the methods, like:

```python
extension_path = find_extension()
if extension_path is None:
    logger.log("Failed to find extension on disc", ...)
    extension_path = download_extension()
```
All further logging and raising handled in `download_extension`.

Just a nitpick, but results in cleaner code IMO

src/solidlsp/language_servers/al_language_server.py
```diff
@@ -0,0 +1,1045 @@
...
+                logger.log("AL Extension downloaded successfully.", logging.INFO)
+            else:
+                logger.log("Failed to download AL extension, falling back to manual search.", logging.WARNING)
+                # Fall back to the old method of finding the extension
+                return cls._get_language_server_command_fallback(logger)
+
+        # Verify executable exists after download
+        if not os.path.exists(executable_path):
```
MischaPanch: verification should happen after download, raising then. There shouldn't be two calls to the fallback

src/solidlsp/language_servers/al_language_server.py
```diff
@@ -0,0 +1,1045 @@
...
+            st = os.stat(executable_path)
+            os.chmod(executable_path, st.st_mode | stat.S_IEXEC)
+
+        logger.log(f"Using AL Language Server executable: {executable_path}", level=5)
+
+        # The AL Language Server uses stdio for LSP communication by default
+        # Unlike many other language servers, it doesn't require a --stdio flag
+        # On Windows, we need to properly quote the path if it contains spaces to avoid shell issues
+        if system == "Windows" and " " in executable_path:
+            return f'"{executable_path}"'
```
MischaPanch: Is it harmful to quote for linux and mac?

The check on ` " " in executable_path` is unnecessary, let's always quote on windows. This is even worth a util method in `common.py`

src/solidlsp/language_servers/al_language_server.py
```diff
@@ -0,0 +1,1045 @@
...
+                        return str(item)
+
+        return None
+
+    @staticmethod
+    def _get_initialize_params(repository_absolute_path: str) -> dict:
+        """
+        Returns the initialize params for the AL Language Server.
+        """
+        import os
```
MischaPanch: let's avoid local imports like this, pls make global

src/solidlsp/language_servers/al_language_server.py
```diff
@@ -0,0 +1,1045 @@
...
+        The AL server requires additional setup after initialization:
+        1. Set the active workspace - tells AL which project to work with
+        2. Send workspace configuration - provides AL settings and paths
+        3. Open app.json to trigger project loading - AL uses app.json to identify project structure
+        4. Wait for project to be loaded - ensures symbols are indexed before operations
+
+        This special initialization sequence is unique to AL and necessary for proper
+        symbol resolution and navigation features.
+        """
+        import time
```
MischaPanch: move to global

src/solidlsp/language_servers/al_language_server.py
```diff
@@ -0,0 +1,1045 @@
...
+                        },
+                    }
+                    result.append(dir_symbol)
+
+            return result
+        else:
+            self.logger.log("AL: No symbols found in any files", level=3)
+            return []
+
+    def request_document_symbols(self, relative_file_path: str, include_body: bool = False) -> tuple[list[dict], list[dict]]:
```
MischaPanch, line 859: pls add `@override` (imported from `overrides`, not `typing`), a helpful util that we use throughout

src/solidlsp/language_servers/al_language_server.py
```diff
@@ -0,0 +1,1045 @@
...
+            "Reference",  # Reference assemblies - .NET dependencies
+            ".netpackages",  # .NET packages - NuGet packages for AL
+            "bin",  # Binary output - compiled binaries
+            "obj",  # Object files - intermediate build artifacts
+        }
+
+        # Check parent class ignore list first, then AL-specific
+        return super().is_ignored_dirname(dirname) or dirname in al_ignore_dirs
+
+    def request_full_symbol_tree(self, within_relative_path: str | None = None, include_body: bool = False) -> list[dict]:
```
MischaPanch, line 699: pls add `@override` (imported from `overrides`, not `typing`), a helpful util that we use throughout

src/solidlsp/language_servers/al_language_server.py
```diff
@@ -0,0 +1,1045 @@
...
+            # Some large projects may take longer to fully index
+            self.logger.log("Project load not confirmed, waiting additional time for AL workspace indexing", level=5)
+            time.sleep(5)
+        else:
+            # Even when loaded, give a bit more time for symbol indexing to complete
+            # This ensures all symbols are available for navigation
+            self.logger.log("Project loaded, waiting for symbol indexing", level=5)
+            time.sleep(2)
+
+    def is_ignored_dirname(self, dirname: str) -> bool:
```
MischaPanch, line 668: pls add `@override` (imported from `overrides`, not `typing`), a helpful util that we use throughout

src/solidlsp/language_servers/al_language_server.py
```diff
@@ -0,0 +1,1045 @@
...
+            # If direct request failed, try parent method as fallback
+            self.logger.log("AL: Falling back to parent method", level=5)
+            return super().request_document_symbols(relative_file_path, include_body)
+
+        except Exception as e:
+            self.logger.log(f"AL: Error in request_document_symbols: {e}", level=3)
+            return ([], [])
+
+    def _convert_lsp_symbol(self, lsp_symbol: dict) -> dict:
+        """Convert LSP symbol format to Serena format."""
```
MischaPanch: to lsp or to solidlsp format. This package is an independent library for language servers, not tied to Serena as MCP server. Over time, this may become the main aspect of this project, especially for power users.

Is the input somehow AL specific? Then it should be rather called `al_ls_symbol`

pyproject.toml
```diff
@@ -271,11 +271,17 @@ markers = [
...
   "erlang: language server running for Erlang",
+  "al: language server running for AL (Microsoft Dynamics 365 Business Central)",
 ]
 
 [tool.codespell]
 # Ref: https://github.com/codespell-project/codespell#using-a-config-file
 skip = '.git*,*.svg,*.lock,*.min.*'
 check-hidden = true
+
+[dependency-groups]
```
MischaPanch: we already have a dev dependencies group. You should also update and commit the uv.lock

README.md
```diff
@@ -128,44 +129,46 @@ Several videos and blog posts have talked about Serena:
 
 <!-- toc -->
 
+- [Table of Contents](#table-of-contents)
 - [Quick Start](#quick-start)
-  * [Running the Serena MCP Server](#running-the-serena-mcp-server)
-    + [Usage](#usage)
+  - [Running the Serena MCP Server](#running-the-serena-mcp-server)
```
MischaPanch: pls undo the style changes here

README.md
```diff
@@ -96,6 +96,7 @@ With Serena, we provide direct, out-of-the-box support for:
   * Nix (requires nixd installation)
   * Elixir (requires installation of NextLS and Elixir; **Windows not supported**)
   * Erlang (requires installation of beam and [erlang_ls](https://github.com/erlang-ls/erlang_ls), experimental, might be slow or hang)
+  * AL (requires VS Code AL extension for Business Central development; set `AL_EXTENSION_PATH` environment variable or install ms-dynamics-smb.al extension)
```
MischaPanch: It doesn't really require it, will be downloaded on the fly, right? Otherwise, it wouldn't work in CI

MischaPanch (changes requested):
After seeing so many vibe-coded PRs, this is blessing for my eyes. A large and fairly complex contribution, very well readable, well structured, well documented and tested. Thank you and really well done! Especially considering Python not being your dominant language.

I only have small comments, most of aesthetic nature, except the one on the when-to-download logic. When these are addressed, this can be merged.

If you want to add more languages in this way, you're more than welcome. Did I manage to spike your interest with the VSCode integration, or are you just here for adding AL and then be on your way? :)

src/solidlsp/language_servers/al_language_server.py
SShadowS: Ofc, added the download part later.
I scrapped the "One to rule them all" method and pulled them into their own seperate as you asked.

src/solidlsp/language_servers/al_language_server.py
SShadowS: Pulled it out into common.

src/solidlsp/language_servers/al_language_server.py
SShadowS: Removed all locals, just a habit from AL.

src/solidlsp/language_servers/al_language_server.py
SShadowS: Removed all locals, just a habit from AL.

src/solidlsp/language_servers/al_language_server.py
SShadowS: Had to read doc to find out what you meant, but done. :)

Do you also want them in front of start and _start_server then?

src/solidlsp/language_servers/al_language_server.py
SShadowS: Wrong naming from my part. Fixed

pyproject.toml
SShadowS: Moved. Now about the uv.lovk, my first time contributing to a shared Python project. So just uv lock and commit?

README.md
SShadowS: Removed

pyproject.toml
MischaPanch: yes, exactly

src/solidlsp/language_servers/al_language_server.py
MischaPanch: Yes, pls. It's not yet used everywhere in solidlsp since much of the cody is copy-paste from its multilspy past, so not all of it is using the same conventions and standards. But at least all new code should

README.md
SShadowS: Stupid VSCode extension which auto formats on save. Reverted.

PR #598, opened 2025-09-03
show diagnostic info when writng ts/js files

kolodny:
This PR enables serena to report any syntax or linting error when modifying a TS or JS file

### DIFF
--- a/src/serena/tools/file_tools.py
+++ b/src/serena/tools/file_tools.py
@@ -79,6 +79,18 @@ class CreateTextFileTool(Tool, ToolMarkerCanEdit):
         answer = f"File created: {relative_path}."
         if will_overwrite_existing:
             answer += " Overwrote existing file."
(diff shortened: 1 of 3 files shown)


MischaPanch:
Hi @kolodny, sorry for the late reply. Somehow this escaped our radar, there's a similar PR in #661 where we also had a discussion on this topic. We will see how to combine the contributions from here and there and add first support for diagnostics

opcode81:
Reporting diagnostics upon edits is mostly an anti-pattern which creates noise for the LLM, as intermediate states are often expected to have problems.
While support for this feature has, in fact, been added to Serena (see class `EditingToolWithDiagnostics`), the feature is disabled, and will, most likely, remain so.
We have opted for explicit diagnostics requests instead, such that the LLM can request diagnostics when it is ready.

PR #603, opened 2025-09-05
fix: enhance SearchForPatternTool param description

retanoj:
# Description

I setup serena as Claude Code mcp server. 
When I use Claude Code to search code, it will call serena search_for_pattern tool like: 
`search_for_pattern (MCP)(substring_pattern: "Yaml", paths_include_glob: "*.java", restrict_search_to_code_files: true)`

But sometimes Claude Code will set paths_include_glob params as "\*\*/\*.java,\*\*/\*.yaml", which is not satisfied param format rule. So this pr try to enhance tool description to limit Agent behavior.

# Changes

- Add format limit description to `paths_include_glob` param.
- Add format limit description to `paths_exclude_glob` param.

### DIFF
--- a/src/serena/tools/file_tools.py
+++ b/src/serena/tools/file_tools.py
@@ -339,10 +339,10 @@ class SearchForPatternTool(Tool):
         :param substring_pattern: Regular expression for a substring pattern to search for
         :param context_lines_before: Number of lines of context to include before each match
         :param context_lines_after: Number of lines of context to include after each match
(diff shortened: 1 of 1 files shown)


opcode81:
#652 added brace expansion, which allows the LLM to pass `**/{*.java,*.yaml}`. This new addition should make the model refrain from causing the issue that motivated this PR. Therefore closing.

PR #614, opened 2025-09-09
Remove 'read_file' from excluded tools list

ElNiak:
Inconsistent instruction and excluded tools: " The call of the read_file tool on an entire source code file should only happen in exceptional cases, usually you should first explore the file (by itself or as part of exploring the directory containing it) using the symbol_overview tool, and then make targeted reads using find_symbol and other symbolic tools."

### DIFF
--- a/src/serena/resources/config/contexts/ide-assistant.yml
+++ b/src/serena/resources/config/contexts/ide-assistant.yml
@@ -18,7 +18,6 @@ prompt: |
 
 excluded_tools:
   - create_text_file
-  - read_file
   - execute_shell_command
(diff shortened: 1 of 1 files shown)


opcode81:
Thanks for noticing the discrepancy (which we had already noticed but didn't yet have time to fix).
In the commit I just added to `main`, I have implemented another solution: To not mention the tool's name but still refer to the operation of reading a full file. 
Note that in IDE assistant contexts, a tool allowing to read files is virtually always provided through other means and therefore Serena's tool is redundant (which is why we disabled it).

PR #618, opened 2025-09-10
Improved performance and made code a bit easier to read

SShadowS:
I have removed a lot of sleeps, which were part of the initial reverse engineering of the language server.
Also gone through all my logging, as I was a bit too verbose. Again, left over from development.

### DIFF
--- a/src/solidlsp/language_servers/al_language_server.py
+++ b/src/solidlsp/language_servers/al_language_server.py
@@ -198,7 +198,7 @@ class ALLanguageServer(SolidLanguageServer):
         # Check environment variable
         env_path = os.environ.get("AL_EXTENSION_PATH")
         if env_path and os.path.exists(env_path):
-            logger.log(f"Found AL extension via AL_EXTENSION_PATH: {env_path}", level=5)
+            logger.log(f"Found AL extension via AL_EXTENSION_PATH: {env_path}", logging.DEBUG)
             return env_path
         elif env_path:
             logger.log(f"AL_EXTENSION_PATH set but directory not found: {env_path}", logging.WARNING)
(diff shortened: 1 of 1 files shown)


SShadowS:
@MischaPanch Can you give the failing test a kick?

src/solidlsp/language_servers/al_language_server.py
```diff
@@ -852,22 +828,16 @@ def request_full_symbol_tree(self, within_relative_path: str | None = None, incl
...
-
-        The AL server requires an exact sequence:
-        1. Open the file with textDocument/didOpen
-        2. Wait for the server to process the file
-        3. Request symbols with textDocument/documentSymbol
-        4. Close the file with textDocument/didClose
+        Override to ensure AL server properly handles symbol requests.
 
-        URI matching must be exact - any mismatch will result in empty symbols.
+        The AL server works best with the parent class's file buffer management system,
```
MischaPanch: The additional logging doesn't warrant an override IMO. If you feel like debug logging is missing, you can add it to the base class

src/solidlsp/language_servers/al_language_server.py
```diff
@@ -1039,42 +939,49 @@ def check_project_loaded(self) -> bool:
             self.logger.log("Cannot check project load - server not started", logging.DEBUG)
             return False
 
+        # Check if we've already determined this request isn't supported
+        if hasattr(self, "_project_load_check_unsupported") and self._project_load_check_unsupported:
```
MischaPanch: why the hasattr? All attributes should be set at init, is there a reason it can't be done for this one?

MischaPanch:
Thanks, we can merge once the minor comments are resolved. I'll try to find time to have a look at why swift installation fails on ubuntu, seems like a version pinning is missing somewhere at first glance

src/solidlsp/language_servers/al_language_server.py
SShadowS: Must confess, Google made me do it. Som random search. It is moved, I have nothing to defend it. Gotta up my Python game. :)

src/solidlsp/language_servers/al_language_server.py
SShadowS: Done, also removed all the helper functions I had referring it before. So now way more clean.

src/solidlsp/language_servers/al_language_server.py
MischaPanch: NP, lol, I have seen way worse, even in what was claimed to be "production code"

MischaPanch: One more nitpick - could we avoid the double negative? Easy to make a mistake there.

So it should be `self._project_load_check_supported = True`

Also, to up your python game, vars should be commented with a docstring, then the IDE will display info on hover. So

```python
self._project_load_check_supported = True
"""Info on what it means for project load check can be supported goes here and not in block comment"""
```

SShadowS: Thanks, will add that to my list of improvements. Nice to know, and yeah, adds a lot on readability.

SShadowS:
Done and thanks for the feedback.

PR #621, opened 2025-09-11
Add Android language server with mixed Java/Kotlin support

fantasyRqg:
I’ve only tested it on osx-arm64, and it works.


MischaPanch:
Thanks for the contribution. This is large and complex, and unfortunately, we are very time restricted this month, so we will not have time to review it for now.

The architecture seems to go in a similar direction like attempts for svelte support - using multiple LS, transferring the results from one to another for cross-referencing, and aggregating the results. Which can make sense for this use case.

We will definitely come back to this when we have more time again

Issue #587, opened 2025-09-02
ruby-lsp: how do you install this thing?


dragonfax:
I didn't know serena had switched from solargraph to ruby-lsp. Never used the latter.

Until this happened.

```
INFO  2025-09-02 01:23:59,178 [SerenaAgentExecutor_0] solidlsp:_start_server:375 - Starting ruby-lsp server process
INFO  2025-09-02 01:23:59,178 [SerenaAgentExecutor_0] solidlsp.ls_handler:start:189 - Starting language server process via command: ['ruby-lsp']
INFO  2025-09-02 01:23:59,181 [SerenaAgentExecutor_0] solidlsp:_start_server:379 - Sending initialize request from LSP client to LSP server and awaiting response
...
DEBUG 2025-09-02 01:23:59,181 [SerenaAgentExecutor_0] solidlsp.ls_handler:send_request:466 - Starting: Request[status='pending', method='initialize', request_id=1]
INFO  2025-09-02 01:23:59,192 [LSP-stderr-reader] solidlsp.ls_handler:_read_ls_process_stderr:387 - /bin/sh: ruby-lsp: command not found

ERROR 2025-09-02 01:23:59,192 [LSP-stderr-reader] solidlsp.ls_handler:_read_ls_process_stderr:391 - Language server stderr reader thread terminated unexpectedly
INFO  2025-09-02 01:23:59,192 [LSP-stdout-reader] solidlsp.ls_handler:_read_ls_process_stdout:362 - Language server stdout reader thread has terminated
ERROR 2025-09-02 01:23:59,193 [LSP-stdout-reader] solidlsp.ls_handler:_read_ls_process_stdout:366 - LanguageServerTerminatedException: Language server stdout read process terminated unexpectedly
INFO  2025-09-02 01:23:59,193 [LSP-stdout-reader] solidlsp.ls_handler:_cancel_pending_requests:451 - Cancelling 1 pending language server requests
INFO  2025-09-02 01:23:59,193 [LSP-stdout-reader] solidlsp.ls_handler:_cancel_pending_requests:453 - Cancelling Request[status='pending', method='initialize', request_id=1]
DEBUG 2025-09-02 01:23:59,193 [SerenaAgentExecutor_0] solidlsp.ls_handler:send_request:475 - Completed: Request[status='error', method='initialize', request_id=1]
[...]

PR #628, opened 2025-09-16
feat: add rbenv integration for Ruby LSP server

kiakiraki:
Add support for rbenv Ruby version management in the Ruby LSP language server. This allows Serena to respect project-specific .ruby-version files and use the appropriate Ruby and bundler versions as specified by rbenv.

Changes:

- Add _get_rbenv_ruby_command() helper to detect rbenv and use correct Ruby version
- Add _get_rbenv_bundle_command() helper for rbenv-aware bundle command execution
- Modify _setup_runtime_dependencies() to use rbenv commands when available
- Update bundle exec ruby-lsp command to respect rbenv configuration
- Fixes Ruby LSP initialization failures when projects use different Ruby versions than the system default, particularly resolving bundler version mismatch issues.

related: https://github.com/oraios/serena/issues/587#issuecomment-3264107156


src/solidlsp/language_servers/ruby_lsp.py
```diff
@@ -97,6 +97,51 @@ def _find_executable_with_extensions(executable_name: str) -> str | None:
...
+        """
+        Get the appropriate Ruby command considering rbenv configuration.
+        Returns the Ruby command that respects .ruby-version file if present.
+        """
+        # Check for .ruby-version file
+        ruby_version_file = os.path.join(repository_root_path, ".ruby-version")
+
+        if os.path.exists(ruby_version_file):
+            try:
+                with open(ruby_version_file) as f:
```
MischaPanch: Why opening here? Seems unnecessary, will rbenv not fail with a suitable error?

src/solidlsp/language_servers/ruby_lsp.py
```diff
@@ -97,6 +97,51 @@ def _find_executable_with_extensions(executable_name: str) -> str | None:
...
+                    f.read().strip()  # Verify file is readable
+
+                # Check if rbenv is available
+                try:
+                    rbenv_path = shutil.which("rbenv")
+                    if rbenv_path:
+                        # Use rbenv exec to run with specified version
+                        return ["rbenv", "exec", "ruby"]
```
MischaPanch: Unclear how this is related to `ruby_version_file`, which it doesn't use

src/solidlsp/language_servers/ruby_lsp.py
```diff
@@ -97,6 +97,51 @@ def _find_executable_with_extensions(executable_name: str) -> str | None:
...
+                except Exception:
```
MischaPanch: this is too broad. 

I also don't see what can cause exceptions in the above block at all, why the try-except?

src/solidlsp/language_servers/ruby_lsp.py
```diff
@@ -97,6 +97,51 @@ def _find_executable_with_extensions(executable_name: str) -> str | None:
...
+        return ["ruby"]
+
+    @staticmethod
+    def _get_rbenv_bundle_command(repository_root_path: str) -> list[str]:
+        """
+        Get the appropriate bundle command considering rbenv configuration.
+        """
+        ruby_version_file = os.path.join(repository_root_path, ".ruby-version")
+
+        if os.path.exists(ruby_version_file):
```
MischaPanch: same comments as above. 

Some code is repeated, and since both methods are called in the same scope, I think it might be better to inline them (remove the methods, just execute the code with some block comments if needed)

MischaPanch (changes requested):
I can't say I fully understand the logic here, `ruby_version_file` seems crucial for control flow but is actually unused within the code.

In any case, you know this topic better, so I trust it makes sense :). We can merge after the minor comments are resolved, consider writing some block comments on the role of `ruby_version_file` and why the attempted ruby/rbenv/bundler command finding is better than the fallback, and what the consequences (downsides) of using the fallback may be. Pls also add log messages so we can see in the logs which ruby, which bundler and so on were chosen and why - recommending to add the ruby_version_file to the user (if that's indeed the right recommendation)

MischaPanch:
Btw, @kiakiraki you put a lot of work into Serena by now. Would you like us to mention you somewhere in the acknowledgments? :)

MischaPanch:
@kiakiraki ping

src/solidlsp/language_servers/ruby_lsp.py
kiakiraki: You're right - the file content isn't actually used. What matters is whether the file exists, not its contents, since rbenv automatically reads .ruby-version to switch versions. Opening the file is unnecessary. I'll update the code.

src/solidlsp/language_servers/ruby_lsp.py
kiakiraki: The .ruby-version file indicates the project uses rbenv for version management. When present, we use `rbenv exec bundle` to ensure bundler runs with the correct Ruby version specified by rbenv.

src/solidlsp/language_servers/ruby_lsp.py
kiakiraki: The exception handling was redundant. I'll remove it.

kiakiraki:
@MischaPanch Thank you for the review. I'll address the feedback now.
(Apologies for the delayed response - I've been busy with my day job.)

> Would you like us to mention you somewhere in the acknowledgments? :)

I'd be honored, no problem at all.

rclark72:
This has introduced a bug when the `.ruby-version` is included as part of the project, but `rbenv` is not used as the version manager. 

Additionally, I'm not entirely sure if this is even necessary or will even solve the bug the user was experiencing. The original submitter of the bug [said](https://github.com/oraios/serena/issues/587#issuecomment-3358120090) it was resolved by installing ruby-lsp for the version of ruby they were running. I think this is the correct solution. It allows Serena to execute the compatible version of `ruby-lsp` for the ruby version their project is using. 

If there needs to be support for users that don't configure their shell to use the version manager, but still want to have Serena use the version manager, would configuration be a better approach?

MischaPanch:
@rclark72 could you open an issue for the bug? I suppose the idea is that using `rbenv` is preferred when it's available, and we should have a fallback behavior for when `.ruby_version` is included and `rbenv` is missing.

Concerning configuration - you're very welcome to open a PR that extends configurability. We recently merged #524 which allowed passing language server specific configurations through `serena_config`, this is likely what you would want to use.

@kiakiraki WDYT?

kiakiraki:
@rclark72 Thank you for reporting. I'd like to understand the issue better.

Looking at the current implementation (lines 132-137), there's already fallback logic: when .ruby-version exists but rbenv is not installed, it falls back to system Ruby commands (ruby and bundle) with a warning message.

Could you provide more details about the bug you encountered? For example:
 - What error messages did you see?
 - What Ruby version manager are you using (if any)?
 - What was the expected vs actual behavior?

PR #632, opened 2025-09-18
chore: add support for streamable-http transport for mcp

kosmoz:
Streamable HTTP is the sucessor of the SSE transport, which is considered legacy and no longer part of the [latest specification](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports).

This PR simply adds "streamable-http" as an option for `--transport` when calling `start-mcp-server`. This is already supported by FastMCP, so no further changes are necessary. 

The MCP endpoint is available under `http://localhost:$PORT/mcp`.

### DIFF
--- a/src/serena/cli.py
+++ b/src/serena/cli.py
@@ -119,7 +119,13 @@ class TopLevelCommands(AutoRegisteringGroup):
         show_default=True,
         help="Built-in mode names or paths to custom mode YAMLs.",
     )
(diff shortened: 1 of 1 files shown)


MischaPanch:
Thx, makes sense, but we should also update the readme accordingly. Would you like to do that as well?

kosmoz:
Hi, I updated the readme file. Feel free to make changes as necessary :slightly_smiling_face:

MischaPanch:
Thanks for the contribution!

PR #646, opened 2025-09-30
docs(readme): web dashboard disable option

jjangga0214:
Added command to disable web dashboard at startup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,6 +211,12 @@ This and other settings can be adjusted in the [configuration](#configuration) a
 uvx --from git+https://github.com/oraios/serena serena start-mcp-server
 ```
 
+By default, Serena automatically opens the web dashboard at startup. To disable this behavior, launch it with the following command.
+
+```shell
+uvx --from git+https://github.com/oraios/serena serena-mcp-server --enable-web-dashboard false start-mcp-server
+```
+
 Explore the CLI to see some of the customization options that serena provides (more info on them below).
 
 #### Local Installation
@@ -238,7 +244,7 @@ Explore the CLI to see some of the customization options that serena provides (m
    When running from outside the serena installation directory, be sure to pass it, i.e., use
 
    ```shell
-    uv run --directory /abs/path/to/serena serena start-mcp-server
+   uv run --directory /abs/path/to/serena serena start-mcp-server
    ```
 
 #### Using Docker (Experimental)


opcode81:
You are providing an arbitrary example of a command with an additional option.

jjangga0214:
@opcode81 

There are two perspectives:

1.	Strict separation by type: The basic usage is introduced in a Getting Started style as concisely as possible. All other configurations, regardless of usefulness, are introduced separately in a Reference section.

2.	Separation by usefulness: The basic usage is introduced in a Getting Started style, but configurations that have high demand are exceptionally included in Getting Started instead of being left only in Reference.

Option 1 separates content by category. It’s a systematic, formalist approach.
Option 2 separates content by usefulness. It’s a flexible, user-oriented approach.

Developers are often short on time. Option 1 is excellent in principle, but for tools where the real need is to get up and running quickly, Option 2 can be superior. That’s why styles like batteries-included, tips (annotations in technical docs), and recipes exist in the first place.

For example, as we see even in serena's issue template, 

스크린샷 2025-10-01 오후 2 16 21

how to disable the dashboard is already a top-level concern. The intention here is to allow people to skim only the beginning of the README and immediately apply what they need.

Just like the Pareto principle, where 80% of users only needs 20% of the settings features, highlighting that 20%—such as disabling the dashboard—directly in Getting Started is a good idea. (Though I personally think disabling the dashboard should be the sane default—but that’s outside the scope of this discussion.)

How do you think?

README.md
```diff
@@ -214,7 +214,7 @@ uvx --from git+https://github.com/oraios/serena serena start-mcp-server
 By default, Serena automatically opens the web dashboard at startup. To disable this behavior, launch it with the following command.
 
 ```shell
-uvx --from git+https://github.com/oraios/serena serena-mcp-server --enable-web-dashboard false start-mcp-server
+uvx --from git+https://github.com/oraios/serena serena --enable-web-dashboard false start-mcp-server
```
opcode81: I don't think this will work. The command `start-mcp-server` needs to come before additional arguments.

jjangga0214: @opcode81 Fixed :)

MischaPanch:
It's true that the example is not entirely arbitrary in the sense that many users ask for it. Maybe instead of one example we can add a small section about the dashboard high up in the readme. Including one sentence that we recommend the dashboard to be active to see logs, tool calls, tool stats, and any updates on Serena (we will start adding them to the dashboard). The dashboard can be more conveniently disabled through the global serena_config, this should also be explained there.

jjangga0214:
@MischaPanch 

> Maybe instead of one example we can add a small section about the dashboard high up in the readme.

Good. Placement suggestion? Then I'll edit. Or feel free to open another PR.

MischaPanch:
I'll adjust the PR, easier than review ping pong

MischaPanch:
We will be enhancing the dashboard in its functionality in the next weeks, so it is counterproductive to advertise turning it off at the top of the readme. Closing this

PR #652, opened 2025-10-04
feat(text_utils): add brace expansion for glob patterns

drewvolz:
`fnmatch` module in python does not appear to support brace expansion `{js,jsx,ts,tsx}` natively being a shell feature and not a standard glob pattern in https://github.com/oraios/serena/blob/main/src/serena/text_utils.py#L290-L319

```
[SerenaAgentExecutor_0l serena.text_utils:search_files: 319 - 
Skipping src/initialize.ts: does not match include pattern **/*.{js, jsx, ts, tsx}
```

It would be nice if we supported glob patterns that were wrapped with braces which commonly appear in projects that include javascript and typescript files.

The changes in this pull request aim to add that support by:

* Adding an `expand_braces` function to handle glob patterns
* Updating `search_files` to use it for include/exclude patterns
* Adding to the tool docstring such that the LM is aware

How this was tested

- Ran Serena from `main` as well as locally from this branch
- Set logging level to DEBUG
- New tests added in `test/Serena/test_text_utils.py`
- Healthcheck passed

--- 

SerenaAgent examples invoking `include_glob` patterns

```py
from serena.agent import SerenaAgent
from serena.constants import REPO_ROOT
from serena.tools import SearchForPatternTool

agent = SerenaAgent(project=REPO_ROOT, serena_config=SerenaConfig(gui_log_window_enabled=False, web_dashboard=False))
search_pattern_tool = agent.get_tool(SearchForPatternTool)

# Case 1: Find language server initialization in language server modules
print("Brace expansion with multiple file types")
result_language_servers = agent.execute_task(
    lambda: search_pattern_tool.apply(
        substring_pattern="def __init__",
        paths_include_glob="src/**/language_servers/**/*.{py,ts}",
        context_lines_before=1,
        context_lines_after=2
    )
)

# Case 2: Search test implementations
print("Search across test implementations")
result_test_patterns = agent.execute_task(
    lambda: search_pattern_tool.apply(
        substring_pattern="class.*Test",
        paths_include_glob="test/solidlsp/{python,typescript,rust}/**/*.{py,ts,rs,js}",
        context_lines_before=0,
        context_lines_after=1,
    )
)
```  

#### Output from `main`

First case matches: 0
Second case matches: 0

#### Output from `feat/dvolz/brace-expansion-glob`

First case matches: 1
Second case matches: 5

### DIFF
--- a/src/serena/text_utils.py
+++ b/src/serena/text_utils.py
@@ -274,6 +274,13 @@ def glob_match(pattern: str, path: str) -> bool:
     - ? matches a single character except /
     - [seq] matches any character in seq
 
+    Supports brace expansion:
+    - {a,b,c} expands to multiple patterns (including nesting)
+
+    Unsupported patterns:
+    - Bash extended glob features are unavailable in Python's fnmatch
+    - Extended globs like !(), ?(), +(), *(), @() are not supported
+
     :param pattern: Glob pattern (e.g., 'src/**/*.py', '**agent.py')
     :param path: File path to match against
     :return: True if path matches pattern
--- a/src/serena/tools/file_tools.py
+++ b/src/serena/tools/file_tools.py
@@ -335,26 +335,6 @@ class SearchForPatternTool(Tool):
             and provide glob patterns to include or exclude certain files on top of that.
             The globs are matched against relative file paths from the project root (not to the `relative_path` parameter that
             is used to further restrict the search).
-
-            Supported Glob Pattern Syntax:
-                Both `paths_include_glob` and `paths_exclude_glob` support the following pattern types:
-
-                Standard glob patterns (via Python's fnmatch):
-                  *           - Matches zero or more characters (e.g., "*.py" matches "test.py" in root)
-                  ?           - Matches exactly one character (e.g., "src/?.py" matches "src/a.py")
-                  [seq]       - Matches any character in seq (e.g., "**/[abc]*.py" matches paths with a, b, or c)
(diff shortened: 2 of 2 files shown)


MischaPanch:
Thanks! Looks good to me but I think we should also add it to the tool docstring such that the LM is aware of this possibility - or is it completely obvious to the LLM that this would be the case?

I only had a brief look, @opcode81 could you also review pls (you are more up to date on that part of the code than I am)

opcode81:
I think it will be confusing to LLMs that glob patterns are no longer actually glob patterns. The explanations you had to add to compensate for this are very long - perhaps even unacceptably long, increasing token usage disproportionately.

drewvolz:
Thanks for the feedback @opcode81. Cutting that down sounds extremely important, and I want to get this correct and to your liking.

opcode81:
@drewvolz I think this sentence you added should be enough for LLMs to understand what to do:

    Supports standard glob patterns (*, ?, [seq], **, etc.) and brace expansion {a,b,c}.
    
"brace expansion" is a well-known concept from shells, so it should be clear to LLMs without any additional explanations.
So let's drop all additional descriptions in the tool's docstring and instead move them to the docstring of the glob matching function (where it doesn't consume tokens).

MischaPanch:
@opcode81 LGTM, I think this can be merged

opcode81:
@drewvolz, thanks for your contribution. Merged!

drewvolz:
Nice! Thanks for working with me on that. Appreciate both of your reviews and what you are doing with Serena!

PR #656, opened 2025-10-07
add Perl::LanguageServer support

asunaro276:
## Summary

Adds support for Perl5 via Perl::LanguageServer integration.

## Changes

- **Language Server Integration**: Implemented `PerlLanguageServer` class in `src/solidlsp/language_servers/perl_language_server.py`
  - Support for document symbols, go-to-definition, and find-references
  - Cross-file reference tracking with proper configuration

- **Core Updates**:
  - Added `PERL` to `Language` enum in `src/solidlsp/ls_config.py`
  - Updated factory method in `src/solidlsp/ls.py` to instantiate PerlLanguageServer
  - Fixed request ID type handling in `src/solidlsp/ls_handler.py` to handle string/int ID mismatches

- **File Pattern Support**: Registered `.pl`, `.pm`, `.t` extensions for Perl files

- **Ignore Patterns**: Added Perl-specific ignore directories (blib, local, .carton, vendor, _build, cover_db)

## Dependencies

This implementation requires:
- **Perl 5** (preferably 5.10 or later)
- **Perl::LanguageServer**: Manual installation is recommended due to complex dependency resolution in CPAN. Automatic installation may fail depending on the environment configuration. Users should install it manually using: `cpanm Perl::LanguageServer` or `cpan Perl::LanguageServer`

## Why Perl::LanguageServer Over PLS

We chose Perl::LanguageServer instead of the alternative Perl Language Server (PLS) for the following reasons:

- **Full LSP Feature Support**: Perl::LanguageServer implements `textDocument/references`, which is critical for Serena's cross-file reference tracking capabilities. PLS lacks this method, significantly limiting its utility for code navigation and refactoring workflows.

- **Performance Trade-off**: While PLS is more lightweight, the performance difference is negligible for practical use cases. The additional capabilities of Perl::LanguageServer provide substantially more value than the marginal resource savings of PLS.

## Testing

Created comprehensive tests in `test/solidlsp/perl/test_perl_basic.py`:
- Document symbols test
- Cross-file definition lookup test
- Cross-file references test
- Test repository with multi-file Perl project in `test/resources/repos/perl/`

### DIFF
--- a/src/solidlsp/ls_handler.py
+++ b/src/solidlsp/ls_handler.py
@@ -523,7 +523,7 @@ class SolidLanguageServerHandler:
             if request is None and isinstance(response_id, str) and response_id.isdigit():
                 request = self._pending_requests.pop(int(response_id), None)
 
-            if request is None: # need to convert response_id to the right type
+            if request is None:  # need to convert response_id to the right type
                 log.debug("Request interrupted by user or not found for ID %s", response_id)
                 return
 


src/solidlsp/ls_handler.py
```diff
@@ -517,8 +517,20 @@ def _response_handler(self, response: StringDict) -> None:
...
         """
+        response_id = response["id"]
+        # Normalize ID type: some servers return string IDs even when sent as integers
+        # Try to match by converting string to int if possible
         with self._response_handlers_lock:
-            request = self._pending_requests.pop(response["id"])
+            if response_id in self._pending_requests:
+                request = self._pending_requests.pop(response_id)
+            elif isinstance(response_id, str) and response_id.isdigit():
+                # Try integer version of string ID
```
MischaPanch: Nitpick comment:

This is not really "try" since it will raise a KeyError. However, the KeyError is appropriate, we want the pending_requests to be adjusted.

So the only changes needed are:
1. remove the block comments saying "try", they are confusing
2. instead of if response_id in self._pending_requests pls do 

```
request = self._pending_requests.pop(response_id, None)
if request is None: # need to convert response_id to the right type
   log.debug(....)
```

.github/workflows/pytest.yml
```diff
@@ -213,6 +213,23 @@ jobs:
             echo "$LUA_LS_DIR/bin" >> $GITHUB_PATH
             rm lua-ls.zip
           fi
+      - name: Install Perl::LanguageServer
```
MischaPanch, line 216: Can we use the [setup perl action](https://github.com/marketplace/actions/setup-perl-environment) instead?

src/solidlsp/language_servers/perl_language_server.py
```diff
@@ -0,0 +1,225 @@
...
+        if platform_id not in valid_platforms:
+            raise RuntimeError(f"Platform {platform_id} is not supported for Perl at the moment")
+
+        perl_version = cls._get_perl_version()
+        if not perl_version:
+            raise RuntimeError(
+                "Perl is not installed. Please install Perl from https://www.perl.org/get.html and make sure it is added to your PATH."
+            )
+
+        perl_ls_version = cls._get_perl_language_server_version()
```
MischaPanch, line 90: Is it feasible to install Perl language server on the fly? Seems like they distribute the binaries. Pls have a look at how other LS implementations use `RuntimeDependencyCollection` to handle this

.github/workflows/pytest.yml
asunaro276: Thanks for the suggestion!
However, since Linux and macOS come with Perl pre-installed and Windows is not supported, the setup-perl action isn't necessary. We only need to install cpanminus and system dependencies for building Perl::LanguageServer.

src/solidlsp/language_servers/perl_language_server.py
asunaro276: The language server I targeted for support is Perl::LanguageServer, which is not distributed as a binary and can only be installed as a CPAN module. On the other hand, another language server implementation, Perl Navigator, distributes binaries.

I chose Perl::LanguageServer because it has a longer history, it's what most people around me are using, and it supports the textDocument/references method.

That said, if you think Perl Navigator would be a better fit for Serena MCP's policy given the binary distribution, I'm happy to rewrite the implementation to use it instead. What do you think?

src/solidlsp/ls_handler.py
asunaro276: Thanks for the feedback! I've addressed this in ee4636a.

src/solidlsp/language_servers/perl_language_server.py
MischaPanch: Ah, I see, I got confused by the big download button on the website, but the installation instructions are actually a bit more involved.

I don't have an opinion on which LS is better, if you think Perl::LanguageServer is better, I trust your evaluation. Note that we can support multiple LS for a language using the experimental LS option (not that it's needed, just mentioning). Merging this now, thanks for the contribution!

PR #658, opened 2025-10-07
Add Nim-lang support via nim-langserver

a5m0:
## Summary

- Implement a `NimLanguageServer` class that handles downloading/installing `nimlangserver`, sets up environment variables, writes auxiliary Nim/nimsuggest config files, and improves logging/readiness checks so the process starts reliably.
- Wire the new server into Serena’s language registry and document official Nim support in the README and changelog.
- Provide pytest coverage plus a sample Nim project that exercises symbols, definitions, references, and completions, along with a dedicated nim pytest marker.

## Testing

- `uv run pytest -m nim` (passes locally)


test/solidlsp/nim/test_nim_basic.py
```diff
@@ -0,0 +1,244 @@
...
+            if "uri" in first_def:
+                assert "utils.nim" in first_def["uri"], "Definition should point to utils.nim"
+
+    @pytest.mark.parametrize("language_server", [Language.NIM], indirect=True)
+    def test_nim_find_references(self, language_server: SolidLanguageServer) -> None:
+        """Test find references functionality for Nim."""
+        # Test finding references to the Person type
+        references = language_server.request_references("main.nim", 20, 3)  # Person type definition
+
+        if references:
```
MischaPanch: Since this is the most common point of failure, pls make sure cross-file references finding actually works by ensuring the expected references are found, not just their existence. Pls remove all `assert isinstance` and `assert len(...) > 0`, these are too generic and don't really add value. There's a dedicated section about this in the guide (memory) on how to add support for new languages

test/solidlsp/nim/test_nim_basic.py
```diff
@@ -0,0 +1,244 @@
...
+"""
+
+import pytest
+
+from solidlsp import SolidLanguageServer
+from solidlsp.ls_config import Language
+
+
+@pytest.mark.nim
+class TestNimLanguageServerBasics:
```
MischaPanch, line 22: tests on windows time out - is the language server not supported for windows?

src/solidlsp/language_servers/nim_language_server.py
```diff
@@ -0,0 +1,590 @@
...
+    """Custom handler for nimlangserver that only sends Content-Length header."""
+
+    def _send_payload(self, payload):
+        """Override to send only Content-Length header for nimlangserver compatibility."""
+        if not self.process or not self.process.stdin:
+            return
+        self._log(payload)
+
+        # Create simplified message for nimlangserver (only Content-Length header)
+        import json
```
MischaPanch: pls use global imports (here and below)

src/solidlsp/language_servers/nim_language_server.py
```diff
@@ -0,0 +1,590 @@
...
+            try:
+                # Write as single operation
+                self.process.stdin.write(header + body)
+                self.process.stdin.flush()
+            except (BrokenPipeError, ConnectionResetError, OSError) as e:
+                if self.logger:
+                    self.logger("client", "logger", f"Failed to write to stdin: {e}")
+                return
+
+    def _read_ls_process_stderr(self) -> None:
```
MischaPanch: pls use @overrides for all overridden methods (importes from overrides, not typing)

src/solidlsp/language_servers/nim_language_server.py
```diff
@@ -0,0 +1,590 @@
...
+
+        log = logging.getLogger(f"{self.__class__.__module__}.{self.__class__.__name__}")
+        ENCODING = "utf-8"
+
+        # Track nimsuggest errors to prevent endless restarts
+        error_count = 0
+        max_errors = 10
+
+        try:
+            while self.process and self.process.stderr:
```
MischaPanch: I suppose it is this what causes the hanging on windows - we shouldn't have a while loop forever, pls add some condition to terminate, even in case of failure

src/solidlsp/language_servers/nim_language_server.py
```diff
@@ -0,0 +1,590 @@
...
+        self, config: LanguageServerConfig, logger: LanguageServerLogger, repository_root_path: str, solidlsp_settings: SolidLSPSettings
+    ):
+        """
+        Creates a NimLanguageServer instance. This class is not meant to be instantiated directly.
+        Use LanguageServer.create() instead.
+        """
+        nim_lsp_executable_path = self._setup_runtime_dependencies(logger, solidlsp_settings)
+
+        # Ensure nimble bin is in PATH for nimsuggest
+        nimble_bin = os.path.expanduser("~/.nimble/bin")
```
MischaPanch: is this going to work on windows? Maybe using `shutile.where` is more reliable

src/solidlsp/language_servers/nim_language_server.py
```diff
@@ -0,0 +1,590 @@
...
+            logger,
+            repository_root_path,
+            ProcessLaunchInfo(cmd=nim_lsp_executable_path, cwd=repository_root_path, env=env),
+            "nim",
+            solidlsp_settings,
+        )
+
+        # Override with custom handler for nimlangserver after parent initialization
+        # First stop the default handler if it's already running
+        if hasattr(self, "server") and self.server and hasattr(self.server, "process"):
```
MischaPanch: that's a rather nasty workaround. Instead, we could add a method `_setup_server_process` in the base class which is called in init and can be overridden here. You can add a comment that the method can be overridden for implementations requiring custom handlers

src/solidlsp/language_servers/nim_language_server.py
```diff
@@ -0,0 +1,590 @@
...
+        # Replace with custom handler
+        self.server = NimLanguageServerHandler(
+            process_launch_info,
+            logger=logging_fn,
+        )
+
+    def _create_nim_config_if_needed(self):
+        """Create or supplement Nim configuration to help nimsuggest work better."""
+        try:
+            nim_cfg_path = os.path.join(self.repository_root_path, "nim.cfg")
```
MischaPanch, line 293: should we really create these files in the user's repository? I suppose this will even create files in nim/test_repo during test runs.

If there's really no way around it, pls factor out _NIMSUGGEST_CONTENT and other such things as multiline strings.

src/solidlsp/language_servers/nim_language_server.py
```diff
@@ -0,0 +1,590 @@
...
+
+                self.logger.log("Created nimsuggest.cfg to improve stability", logging.INFO)
+
+            # Check if nim.cfg exists and if it needs suggestions
+            if not os.path.exists(nim_cfg_path):
+                # Only create nim.cfg if this is clearly a Nim project (has .nimble file)
+                nimble_files = list(pathlib.Path(self.repository_root_path).glob("*.nimble"))
+                if nimble_files:
+                    # Check common project structures to determine paths
+                    has_src = os.path.exists(os.path.join(self.repository_root_path, "src"))
```
MischaPanch: this seems very fragile - why is that needed? I suppose project structure is not enforced in nim, right?

src/solidlsp/language_servers/nim_language_server.py
```diff
@@ -0,0 +1,590 @@
...
+    @classmethod
+    def _setup_runtime_dependencies(cls, logger: LanguageServerLogger, solidlsp_settings: SolidLSPSettings) -> str:
+        """
+        Setup runtime dependencies for Nim Language Server and return the command to start the server.
+        """
+        # Store settings for later use if needed
+        cls._solidlsp_settings = solidlsp_settings
+
+        # First check if nimlangserver is already installed via nimble
+        nimble_bin = os.path.expanduser("~/.nimble/bin")
```
MischaPanch: is this OS independent?

src/solidlsp/language_servers/nim_language_server.py
```diff
@@ -0,0 +1,590 @@
...
+            ]
+        )
+
+        try:
+            # Install to nimble's default location
+            deps.install(logger, nimble_bin)
+        except Exception as e:
+            raise RuntimeError(
+                f"Failed to install nimlangserver via nimble: {e}\nPlease try installing manually with: nimble install nimlangserver"
+            )
```
MischaPanch: pls use `raise ... from e`

src/solidlsp/language_servers/nim_language_server.py
```diff
@@ -0,0 +1,590 @@
+"""
+Provides Nim specific instantiation of the LanguageServer class using nimlangserver.
+Contains various configurations and settings specific to Nim language.
+"""
```
MischaPanch, line 4: There's a whole bunch of tricks used here that are typically not needed in an LS implementation, pls expand the docstring and mention the most unexpected ones (like the custom handler, the while loop, the writing of config files etc) and why they are needed

MischaPanch (changes requested):
Thanks for the contribution! A good step towards adding nim support, but a bit of "polishing" work is still needed. I haven't reviewed all details yet (in particular, the logic of the while loop or the config writing), and will do so once these first comments are addressed.

a5m0:
Thank you for the feedback, I will work on all your suggestions. This pull request may have been a little premature, I have been using this implementation successfully but there are a lot of shortcomings with the nim-langserver that I'm not happy with and trying to work around rather than waiting for Nimony.

linxiaotutututu1123 (approved):

linxiaotutututu1123 (approved):

linxiaotutututu1123 (approved):

linxiaotutututu1123 (approved):

MischaPanch:
@a5m0 thanks for the updates to the PR. Could you pls merge main into this and resolve the conflicts? It would also help with the review if you answered on my previous comments and briefly mention you addressed them.

src/solidlsp/language_servers/nim_language_server.py
```diff
@@ -0,0 +1,555 @@
...
+                    with open(nim_cfg_path, "w") as f:
+                        f.write(_NIM_CFG_TEMPLATE.format(extra_paths=extra_paths))
+                    log.warning("Created nim.cfg in project root with path hints for nimsuggest. Consider adding it to .gitignore.")
+            else:
+                log.debug("Found existing nim.cfg, respecting project configuration")
+        except Exception as e:
+            log.debug("Could not create config files: %s", e)
+
+    @classmethod
+    def _setup_runtime_dependencies(cls, config: LanguageServerConfig, solidlsp_settings: SolidLSPSettings) -> str:
```
MischaPanch: pls use the newer `LanguageServerDependencyProvider` interface for downloading dependencies, all new LS should build on that. You can find many examples for how to use it in the exiting LS (it's usually a very small refactoring to change existing code to use the new interface)

src/solidlsp/language_servers/nim_language_server.py
```diff
@@ -0,0 +1,555 @@
...
+            self.server_ready.set()
+        else:
+            log.info("Nim server initialization complete")
+
+    @override
+    def request_document_symbols(self, relative_file_path: str, file_buffer: "LSPFileBuffer | None" = None) -> "DocumentSymbols":
+        """Override to add bounded retry when nimsuggest has port issues.
+
+        When nimsuggest crashes and restarts, the first request may return empty results
+        because nimsuggest has port=0. This override retries with backoff, waiting for
```
MischaPanch: what does port=0 mean in this context?

src/solidlsp/language_servers/nim_language_server.py
```diff
@@ -0,0 +1,555 @@
...
+        nimsuggest to become functional.
+        """
+        from solidlsp.ls import DocumentSymbols
+
+        max_retries = 3
+        retry_delays = [2.0, 5.0, 10.0]
+
+        for attempt in range(max_retries + 1):
```
MischaPanch, line 796: that's really quite a hack, would be sad if there's no other way... Why does it crash in the first place? Can we somehow detect the crash and restart instead of retrying requests? And why does it only affect this method and not the remaining LS methods?

test/solidlsp/nim/test_nim_basic.py
```diff
@@ -0,0 +1,199 @@
...
+                break
+        assert format_symbol is not None, "Could not find 'formatNumber' symbol in utils.nim"
+
+        sel_start = format_symbol["selectionRange"]["start"]
+        refs = language_server.request_references("utils.nim", sel_start["line"], sel_start["character"])
+        # nimlangserver may return empty references or only the definition site for cross-file lookups
+        ref_paths = [ref.get("relativePath", "") for ref in refs]
+        assert (
+            any("utils.nim" in p for p in ref_paths) or refs == []
+        ), f"If references are returned, utils.nim (definition site) should be among them, got: {ref_paths}"
```
MischaPanch: this should assert that found a reference in a different file, but it only checks the same file as the source symbol

test/solidlsp/nim/test_nim_basic.py
```diff
@@ -0,0 +1,199 @@
...
+                target_line, target_col = i, col
+                break
+        assert target_line is not None, "Could not find formatNumber call in main.nim"
+
+        definition = language_server.request_definition("main.nim", target_line, target_col)
+        assert definition, "Should find definition for formatNumber call"
+        assert "utils.nim" in definition[0]["uri"], "Definition should point to utils.nim"
+
+    @pytest.mark.parametrize("language_server", [Language.NIM], indirect=True)
+    def test_nim_completions(self, language_server: SolidLanguageServer) -> None:
```
MischaPanch, line 200: pls add a test that hover works, we use it in our tools now.

test/solidlsp/nim/test_nim_unit.py
```diff
@@ -0,0 +1,155 @@
+"""
+Unit tests for Nim language server components that don't require a running language server.
+"""
+
+import json
+import os
```
MischaPanch: I think this module should be deleted. I don't see the value in these tests, and they use private methods (which no test should do)

MischaPanch (changes requested):

src/solidlsp/language_servers/nim_language_server.py
a5m0: The DependencyProvider class (line 130) now extends LanguageServerDependencyProviderSinglePath and implements the standard interface methods:
  - _get_or_install_core_dependency() — finds or installs nimlangserver
  - _create_launch_command() — returns [core_path]
  - create_launch_command_env() — ensures nimble bin is in PATH

It uses RuntimeDependencyCollection / RuntimeDependency for the nimble install step (lines 169-178), and raise ... from e for proper exception chaining (line 185)

src/solidlsp/language_servers/nim_language_server.py
a5m0: port=0 means nimsuggest has crashed or hasn't started listening yet.

The class docstring (line 120-121) and inline comments (lines 223-226) now explain:
```
  # Track nimsuggest port health to avoid marking server as ready when port=0.
  # port=0 means nimsuggest has crashed or hasn't started listening yet.
```
And in the docstring: "When nimsuggest crashes or is still initializing, it reports port=0 (meaning 'no port assigned / not listening')."

src/solidlsp/language_servers/nim_language_server.py
a5m0: The class docstring (lines 108-128) now explains this thoroughly:

  - Why it crashes: nimlangserver delegates to nimsuggest, which runs as a separate process on a TCP port. When nimsuggest crashes (malformed code, internal errors) or is still initializing, it reports port=0.
  - Why we can't just restart: nimlangserver already handles restart internally and we just need to wait for nimsuggest to come back with a valid port.
  - Why document symbols is most affected: It's typically the first heavy request after server init, hitting the initialization window. Later requests (definition, references, hover) are called after nimsuggest has stabilized.

That said, we do now also retry hover, references, and definition via a shared _retry_on_empty helper (lines 535-567) because nimsuggest can occasionally need warmup time for those too.

test/solidlsp/nim/test_nim_basic.py
a5m0: test_find_references_across_files (lines 150-181) now:
  - Finds formatNumber call in main.nim (defined in utils.nim)
  - Asserts request_references returns results including main.nim
  - Also calls request_definition and asserts it resolves to utils.nim (cross-file verification)

Additionally, a dedicated test_request_referencing_symbols_cross_file test (lines 249-275) queries formatNumber's definition in utils.nim and asserts that request_referencing_symbols finds a reference in main.nim.

test/solidlsp/nim/test_nim_unit.py
a5m0: test_nim_unit.py has been deleted

test/solidlsp/nim/test_nim_basic.py
a5m0: test_nim_hover now exists at lines 219-247 of test_nim_basic.py. It:
  - Finds the greet symbol via request_document_symbols
  - Uses selectionRange to position on the symbol name
  - Calls request_hover and asserts non-None result
  - Validates the hover content contains the proc name "greet"

a5m0:
Rebased on latest origin/main, resolved conflicts and addressed comments, thanks!

src/solidlsp/language_servers/nim_language_server.py
```diff
@@ -0,0 +1,1056 @@
...
+        DefinitionParams,
+        DocumentSymbol,
+        Location,
+        LocationLink,
+        SymbolInformation,
+    )
+
+log = logging.getLogger(__name__)
+
+ENCODING = "utf-8"
```
MischaPanch, line 44: encoding should be read off the LS's field `_encoding`

src/solidlsp/language_servers/nim_language_server.py
```diff
@@ -0,0 +1,1056 @@
...
+
+        first: dict = symbols[0]  # type: ignore[assignment]  # LSP returns TypedDicts
+
+        # Determine where the range lives depending on the LSP response format:
+        # - DocumentSymbol: top-level "range" key
+        # - SymbolInformation: nested "location" -> "range"
+        is_document_symbol = "range" in first
+        location = first.get("location")
+        is_symbol_information = not is_document_symbol and isinstance(location, dict) and "range" in location
+        if not is_document_symbol and not is_symbol_information:
```
MischaPanch, line 651: what is this case semantically? Why can we just return here?

src/solidlsp/language_servers/nim_language_server.py
```diff
@@ -0,0 +1,1056 @@
...
+            # nimlangserver's original range covers just the symbol name; after we
+            # extend it to the full body, consumers need selectionRange to locate
+            # the name (e.g. for hover requests).
+            if "selectionRange" not in symbol:
+                symbol["selectionRange"] = copy.deepcopy(range_info)
+
+            range_start_line = range_info["start"]["line"]
+            range_end_line = range_info["end"]["line"]
+
+            # If the range already spans multiple lines, assume it is correct
```
MischaPanch, line 702: The logic here doesn't seem very robust.  It would be better to use a proper parser, like https://github.com/alaviss/tree-sitter-nim, to get the actual symbol range. I understand that it may be too much to ask from you, but in this state it's an experimental implementation that can only be a temporary step before a proper solution.

The best would be if there was an actually functioning LS for nim, but I guess that's outside the realm of possibilities, right?

src/solidlsp/language_servers/nim_language_server.py
```diff
@@ -0,0 +1,1056 @@
...
+            if line_indent <= base_indent:
+                # Reached a line at the same or lower indent — body ends
+                break
+
+            last_content_line = line_idx
+
+        return last_content_line
+
+    @override
+    def request_document_symbols(self, relative_file_path: str, file_buffer: "LSPFileBuffer | None" = None) -> "DocumentSymbols":
```
MischaPanch, line 774: You override both the public and the private method, is that really needed?

src/solidlsp/language_servers/nim_language_server.py
```diff
@@ -0,0 +1,1056 @@
...
+        column: int,
+        include_imports: bool = True,
+        include_self: bool = False,
+        include_body: bool = False,
+        include_file_symbols: bool = False,
+    ) -> list[ReferenceInSymbol]:
+        """Override to add a goto-definition fallback for Nim.
+
+        nimsuggest's ``textDocument/references`` can return incomplete results,
+        especially for cross-file references.  When the standard path returns
```
MischaPanch, line 892: this is really a major bug in the LS, have you considered fixing it there?

src/solidlsp/language_servers/nim_language_server.py
```diff
@@ -0,0 +1,1056 @@
...
+
+        name_pattern = re.compile(r"\b" + re.escape(target_name) + r"\b")
+
+        result: list[ReferenceInSymbol] = []
+        seen: set[tuple[str, int, int]] = set()
+        lsp_request_count = 0
+
+        overview = self.request_overview("")
+
+        cap_reached = False
```
MischaPanch, line 962: I can't actually review this complicated logic. I guess having something for nim is better than having nothing, so if you say there's no better way than to hack around on almost every corner (repeating requests, adjusting selection ranges) and implementing essential LS functionality here in python, then I'll trust you on that. But we will have to flag nim as experimental, write a short document on the sorry status of the nim LS in our documentation (as warning to users) and keep an open issue to improve all of this or to wait for a better LS

MischaPanch (changes requested):

src/solidlsp/language_servers/nim_language_server.py
a5m0: Unfortunately the tooling available is not good (part of what encouraged me to add Nim to Serena), and I agree that the current nim-langserver support should be marked as experimental. I will try to address your new comments, but ultimately I expect Nim support to be sub-par until https://github.com/nim-lang/nimony is ready.

MischaPanch: Ok, I get it. We can merge this best-effort support for Nim. My only criterion (apart from the tests) is that the problems and the shortcuts (and their limitations) taken to address them are clearly documented in the code, and that expectation management towards nim support is done in the user-facing documentation (like we have done for scala).

zerone0x (approved):
Pulled the branch locally to make sure nothing obvious regressed. The new `NimLanguageServerProcess` and config helpers are wired through `Language.NIM`, the factory override looks clean, and the docs/changelog entries are in place. Ran `uv run pytest -m nim` — everything skips on this machine because neither Nim nor nimlangserver is installed, but there were no unexpected failures. From my side this looks good; thanks for pushing the full compatibility layer over the finish line.

MischaPanch:
There are still some small things left to be fixed before it can be merged.

@a5m0 said
> I will try to address your new comments

zerone0x:
I compared this PR with #1150 as requested.

Summary: #658 is the right base to continue with. #1150 is much smaller/simpler, but it does not cover the nimlangserver-specific problems that this PR already handles.

What #658 has that #1150 lacks:

- Dependency handling via `LanguageServerDependencyProviderSinglePath`, including discovery in `~/.nimble/bin`, PATH setup for `nimsuggest`, and optional `nimble install nimlangserver`.
- A custom `NimLanguageServerProcess` that sends only `Content-Length`. #1150 uses the default process, so it likely still hits the nimlangserver `Content-Type` parsing issue discussed here.
- Project/config handling: `nimsuggest.cfg`, `nim.cfg`, and `projectMapping` detection to keep nimsuggest pointed at a real `.nim` entry point instead of e.g. `config.nims`.
- Readiness/status handling for `extension/statusUpdate`, `port=0`, and nimsuggest restart/warmup states.
- Workarounds for the currently weak LS behavior: symbol range extension, retries for symbols/hover/definition/references, and the guarded goto-definition fallback for cross-file referencing symbols.
- Broader tests: symbols, types, goto-definition, references, completions, hover, and `request_referencing_symbols`. #1150 mainly tests basic symbols plus two direct cross-file reference checks.

What might be reusable from #1150:

- The simpler test repo shape (`src/calculator.nim`, `src/utils.nim`, `main.nim`) is a bit more conventional, but #658's current fixture is intentionally richer and better exercises the hacks/fallbacks.
- The user-facing install wording in #1150 is shorter, but #658's dependency provider makes manual installation less central.

I do not see any core implementation from #1150 that should replace parts of this PR. The remaining work looks more like polishing #658 than merging/cherry-picking #1150.

Remaining items I would prioritize before merge, based on the latest review thread:

1. Mark Nim support as experimental in user-facing docs (README language list and `docs/01-about/020_programming-languages.md`), similar to Scala/Kotlin expectation-setting.
2. Add a short Nim note documenting the current nimlangserver/nimsuggest limitations: generated `nim.cfg`/`nimsuggest.cfg`, `Content-Type` workaround, range-fix heuristic, retries/fallbacks, and possible incomplete cross-file references.
3. Optionally mention in `CHANGELOG.md` that initial Nim support is experimental.
4. If touching code again, address the small review nits that are straightforward; otherwise I think the important behavior differences vs #1150 are already covered here.

I also ran `uv run --extra dev pytest -m nim` locally on this branch; the Nim tests were all skipped because this machine does not have the Nim toolchain/nimlangserver installed, but collection succeeded and there were no unexpected failures.

MischaPanch:
Seems like the nim community is not particularly eager in this regard. Closing as stale

Issue #191, opened 2025-06-19
Add rename_symbol method


MischaPanch:
Maybe not urgent, but should do it only if it's easy.

Could be useful for one-shot setting a la codex where for some reason a refactoring of this type is needed

Will require regex replacement after find_referencing_lines. Sounds easy enough but there might be pitfalls

Good issue for community

Omar-Ballesteros:
Hi!
I'd like to take on this issue. I found that find_referencing_lines is located in SymbolManager, so I propose to implement rename_symbol as a method in the same class.

It will:
- Use find_referencing_lines to get the line numbers where the symbol appears.
- Read the file and perform a safe regex replacement using \b and re.escape() to avoid unintended matches.
- Write the updated content back to the file.

Let me know if you'd like me to proceed with this approach. Happy to adjust based on feedback!
Also, this would be my first open-source contribution, so I’m excited to get involved and open to any guidance 😁

MischaPanch:
Hey @Omar-Ballesteros . Sounds great, appreciate the help!

You'll have to add it a layer deeper, to the `LanguageServer`, and then through `SyncLanguageServer`, through `SymbolManager` and finally in a sublass of `Tool`.

Have a look how another tool was added in #193 (btw, the entire PR was created with claude code using Serena). The PR is not perfect, but the tool works and was added in the right abstractions.

Also on that branch, a memory was added with [a guide on how to add a tool ](https://github.com/oraios/serena/blob/diagnostics/.serena/memories/language_server_architecture_and_tool_propagation.md)that is fundamentally an extension of the `LanguageServer` class capabilities, should be helpful.

I suspect that claude code with serena will bring you quite far with this issue as well, especially if you also make use of that document as a memory.

Looking forward to your PR!

Omar-Ballesteros:
Hi @MischaPanch , thanks a lot for the detailed guidance!
I'll review the changes in #193 and go through the guide, that should give me a clearer picture.
Appreciate the support. 
I'll follow up soon with a PR or any questions that come up along the way!

dsyme:
Would love to see this addressed. @Omar-Ballesteros did you make progress on this?

PR #660, opened 2025-10-09
Add rename_symbol method

dsyme:
Implements #191 

Per advice I pointed VSCode Agent Mode at it and it seems to have done the right thing, judging from the tests. I iterated a bit to add integration tests for Rust and Typescript

---

AI summary: This pull request introduces a comprehensive "Rename Symbol" feature to the codebase, enabling symbols to be renamed throughout a project using language server capabilities. It adds a new tool for renaming, extends language server support for rename operations, and provides robust integration and unit tests to ensure correctness and error handling. The changes are grouped by feature implementation, language server support, and testing.


dsyme:
1. The windows failure looks unrelated

2. This failure looks unrelated:

```
FAILED test/solidlsp/csharp/test_csharp_basic.py::TestCSharpLanguageServer::test_find_referencing_symbols_across_files[csharp] - AssertionError: Should find reference in Models/Person.cs where Calculator.Subtract is called
assert False
 +  where False = any(<generator object TestCSharpLanguageServer.test_find_referencing_symbols_across_files.<locals>.<genexpr> at 0x7f913e422340>)
= 1 failed, 461 passed, 1 skipped, 45 deselected, 1 xpassed, 7 warnings in 846.14s (0:14:06) =
```

src/solidlsp/ls.py
```diff
@@ -1626,6 +1626,92 @@ def request_workspace_symbol(self, query: str) -> list[ls_types.UnifiedSymbolInf
...
+
+        params = ls_types.RenameParams(
+            textDocument=ls_types.TextDocumentIdentifier(
+                uri=pathlib.Path(os.path.join(self.repository_root_path, relative_file_path)).as_uri()
+            ),
+            position=ls_types.Position(line=line, character=column),
+            newName=new_name,
+        )
+
+        return self.server.send.rename(params)
```
MischaPanch, line 1654: Thing is, many language servers don't support this newer functionality AFAIK and I don't really like the idea of depending on it in Serena. So I think we should have an own fallback in case the LS does not support it - search all appearances of the name and replace it ourselves. What do you think?

src/serena/tools/symbol_tools.py
```diff
@@ -293,3 +293,84 @@ def apply(
...
+
+        if rename_result is None:
+            raise ValueError(f"Language server returned no rename edits for symbol '{name_path}'. The symbol might not support renaming.")
+
+        # Apply the workspace edit through the language server
+        language_server.apply_workspace_edit(rename_result)
+
+        # Mark all modified files for the agent
+        modified_files = set()
+        if hasattr(rename_result, "changes") and rename_result.changes:
```
MischaPanch: There is a whole lot of hasattr now, here and in the LS. If I see it correctly, the only purpose is to mark the files as modified. And the only purpose of that marker in turn is to update the `LinesRead` class, which was needed before the switch to using `name_path` everywhere and no longer relying on line numbers (see also my comment on this in your other PR about diagnostics).

So overall, maybe now the time has come to remove the marking mechanism and `LinesRead` since we don't use this anyway. It would significantly reduce the complexity of this feature.

If we do want to keep the marking mechanism, we should find some other way to accomplish this which does not rely on hasattr.

@opcode81 wdyt?

opcode81: Yes, definitely in favour of removing LinesRead if it makes things easier here. It's obsolete as far as I am concerned.
But we don't have to do this in this PR, we can do it later.

MischaPanch: I'll open a separate PR to remove LinesRead, then we can proceed with this.

src/solidlsp/ls.py
dsyme: @MischaPanch It's your call as designers. But from experience of using coding agents, they would probably do pretty well at adjusting gracefully and making use of other tools, or else returning the error up.  It's non-trivial to get a rename right in all situations - there are reasons why it's built into LSP - and if asked I think I would just recommend returning a tool failure in this case.

As one random example, if renaming from a symbol without spaces to one with spaces in F# the LSP should be responsible for enclosing the F# symbol with ``` ``...`` ``` quotes.  Doing that sort of thing from the outside is not fun

opcode81: > It's non-trivial to get a rename right in all situations - there are reasons why it's built into LSP - and if asked I think I would just recommend returning a tool failure in this case.

I agree with this assessment.

MischaPanch: Ok, then instead of the fallback, this tool should be disabled automatically if the project's language server does not provide the desired functionality. Something like a method (in Component)

```python
def is_available(self) -> bool:
   return True
```

which can be overridden for the new tool

```python
@override
def is_available(self) -> bool:
   if not self.agent.is_using_language_server():
       return True  # jetbrains mode, rename tool will be implemented 
   return self.agent.language_server.supports_rename()
```

which is then used to filter out active tools

opcode81: This is not really possible, because, during startup, we cannot wait for the language server to be available for the answer to be provided. And even if we could, this could only apply to cases where a project is provided at startup. For these cases, the information will have to be available statically (based on meta-information), e.g. in the `Language` enum (which is basically the LS enum).

Also note that we don't need the check for `agent.is_using_language_server()`, because it will be an entirely different tool in the JetBrains case.

opcode81: Perhaps the tool simply failing is acceptable; the number of tokens wasted on a failing tool call is not excessive. Especially if the major LS do support this operation, I would propose to go for this very simple solution. If they don't, we can reconsider our options...

dsyme: If the error message indicates not supported, yes, it seems acceptable.

MischaPanch: It's not only about a falling tool call, the presence of the tool uses tokens and also confuses weaker llms,.which sometimes can't deal with that (observed for qwen).

The information can be available statically, it's known at implementation time. The special case of starting with a known active project is, as I mentioned above, a very important special case (all Claude code users belong to that)

opcode81: I don't think it's worth it. Why make things more complicated only to save a few tokens in cases that will hardly ever occur? The rename function is widely implemented and supported by all of the most relevant language servers (Pyright, RustAnalyzer, typescript-language-server, Gopls, Roslyn, Intelephense, etc.).

src/serena/tools/symbol_tools.py
MischaPanch: LinesRead and associated functionality was removed in #666, all hasattr here can now be removed

src/solidlsp/ls.py
```diff
@@ -1626,6 +1626,92 @@ def request_workspace_symbol(self, query: str) -> list[ls_types.UnifiedSymbolInf
...
+
+    def apply_workspace_edit(self, workspace_edit: "ls_types.WorkspaceEdit") -> None:
+        """
+        Apply a WorkspaceEdit by making the changes to files.
+
+        :param workspace_edit: The WorkspaceEdit containing the changes to apply
+        """
+        if hasattr(workspace_edit, "changes") and workspace_edit.changes:
```
MischaPanch: This code is completely wrong, workspace_edit is a dict and hasattr is always false. The AI-generated tests used mocks and thus didn't catch anything, even the core functionality of the feature.

Unfortunately, the entire implementation was rather far from a working one

src/solidlsp/ls.py
MischaPanch: Ok, agree. We can later extend the tests and see how an LS which lacks this feature reports the problem

MischaPanch:
@opcode81 @dsyme 

As per my comment above, this was quite off. The AI generated code looked more convincing than it was, as unfortunately still happens so often. I have now implemented this feature using the `CodeEditor` and tested it with our previously established snapshot testing mechanism. I also added some sanity tests to the snapshot tests such that we can have failures even while snapshots are being generated

MischaPanch:
I have not tested how this behaves on other language servers - only on python. IMO, we can still roll this out and hope that LS that don't support this trigger this:

```
        try:
            rename_result = self._lang_server.request_rename_symbol_edit(
                relative_file_path=relative_file_path, line=symbol.location.line, column=symbol.location.column, new_name=new_name
            )
        except Exception as e:
            raise Exception(f"The current language server {self._lang_server.__class__} might not support renaming.") from e
        if not rename_result:
            raise Exception(
                f"Failed to retrieve rename edits for symbol '{name_path}'. "
                f"The current language server {self._lang_server.__class__} might not support renaming."
            )
```

dsyme:
@MischaPanch Apologies for the slop. Had thought the Go integration test in https://github.com/oraios/serena/pull/660/commits/9df9f0d4431b9a70b4e0fda930e9610ce196697b was sufficient to cover the end to end. 

I'll assess and see if I can make a suggestion for a change to the agent instructions on this repo to close the validation loop.

MischaPanch:
> I'll assess and see if I can make a suggestion for a change to the agent instructions on this repo to close the validation loop.

@dsyme sounds good, thanks! With the "reward hacking" properties of agents, they will create wrong code and then mock things in the tests to cover it up. That's (one of the reasons) why we have almost no mocks at all in the tests, instead relying on actual instances of agents and actual file system changes (using tempfiles). If you want to contribute some instructions (e.g., in form of a memory), this is worth considering

opcode81 (approved):
LGTM. Propose to squash-merge this, as we really don't need the history here.

dsyme:
🎉🎉🎉

PR #661, opened 2025-10-10
Add support for getting diagnostics

dsyme:
# Add GetDiagnosticsTool for Language Server Integration

## Overview
Adds a new `GetDiagnosticsTool` that retrieves diagnostic information (errors, warnings, hints) from language servers and returns structured JSON output with human-readable severity levels.

## Key Features
- Integrates with existing `solidlsp` infrastructure
- Returns diagnostics as structured JSON with line/character positions
- Handles missing fields gracefully
- Works with all supported language servers (Go, Python, TypeScript, etc.)

## Example Output
~~~json
[
  {
    "severity_name": "error",
    "message": "undefined: undefinedVariable", 
    "range": {"start": {"line": 10, "character": 5}},
    "source": "gopls"
  }
]
~~~

## Testing
- **14 tests total**: 10 unit tests + 4 Go integration tests
- Real language server integration with gopls
- Comprehensive error handling coverage

## Bug Fix
Fixed solidlsp diagnostic parsing to handle optional fields that some language servers (like Go) may omit.

Enhances Serena's code analysis capabilities with direct access to language server diagnostics across all supported languages.

### DIFF
--- a/test/serena/test_get_diagnostics_tool.py
+++ b/test/serena/test_get_diagnostics_tool.py
@@ -214,14 +214,16 @@ class TestGetDiagnosticsTool:
     def test_path_construction(self):
         """Test that file paths are constructed correctly."""
         relative_path = "src/module/file.py"
-        
-        with patch('os.path.exists', return_value=True), \
-             patch('os.path.isdir', return_value=False), \
-             patch.object(self.tool, '_limit_length', side_effect=lambda x, _: x):
-            
+
+        with (
+            patch("os.path.exists", return_value=True),
+            patch("os.path.isdir", return_value=False),
+            patch.object(self.tool, "_limit_length", side_effect=lambda x, _: x),
+        ):
+
             self.mock_language_server.request_text_document_diagnostics.return_value = []
             self.tool.apply(relative_path)
-            
+
             # Verify os.path.join was called with correct arguments
             # The actual path checking happens in the method, we verify the LS call
             self.mock_language_server.request_text_document_diagnostics.assert_called_once_with(relative_path)


dsyme:
The ubuntu test failure looks unrelated

```
FAILED test/serena/test_serena_agent.py::TestSerenaAgent::test_find_symbol_references[kotlin-Model-src/main/kotlin/test_repo/Model.kt-src/main/kotlin/test_repo/Main.kt] - json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

MischaPanch:
Thanks, this is a very good start!

But adding diagnostics is a somewhat major extension in functionality, and we should discuss the design in a bit more detail before proceeding. There are various possibilities here, a separate tool being only one of them.

We should think about these options:

1. Always call diagnostics after an edit and add the info to the editing tool output.
2. Allow diagnostics only for a part of the file. If the LSP doesn't natively support that, we could filter file-level diagnostics ourselves
3. In addition to 1, only report new errors (added by the LLM) by checking the diagnostics before the edit. Reporting pre-existing errors can lead the LLM to perform changes unrelated to the user provided task in order to "fix" all the errors.
4. The LS publishes diagnostics if I'm not mistaken, we could maybe listen to that instead of call? (implementation detail, but still)
5. We should add tests for more languages, just checking it for go is not sufficient
7. How to best report diagnostics? Reporting line numbers can be very problematic since they get outdated quickly and confuse the LLM. One reason behind Serena's good performance is that we rarely output line numbers, instead relying on symbol identifiers (name_path). Maybe for diagnostics line numbers have to be reported, but maybe there is another way or a hybrid solution (also report name_path of the symbol containing the error).

There may be other points worth discussing. @opcode81 could you pitch in here pls? And @dsyme would be happy to have your opinion on the points I mentioned. Once we decided on a design, implementing will probably be quite painless

dsyme:
@MischaPanch Cool, yes, a super-interesting puzzle. 

VSCode Agent Mode is clearly using diagnostics scraped from VSCode, and it's really powerful for code correctness I think.

Let's close this and transfer the questions and learnings to a new issue, since the implementation will need to be thrown away

MischaPanch:
@dsyme

@opcode81 and I have discussed this, and we'd like to address this in multiple steps. As a first step, we will merge the new tool but mark it as optional, so users will have to enable it explicitly. This means only powerusers will become aware of this.

Then we can address the proposed points in a separate issue. Always checking diagnostics is a double edged sword - in my experience it quite often it led the agent to perform unrelated changes, but it also sometimes saved it from editing errors or not adhering to the project's standards.

Do you want to continue this PR, or should I take over and merge after finishing?

eyehia7:
Have you integrated  diagnostics feature that lsp offers with Serena ?

MischaPanch:
It's on its way, we are finalizing the integration of it in the next days. Closing this PR, the diagnostic feature we will incorporate is significantly more extensive

PR #673, opened 2025-10-15
Add initial guidelines for agents

dsyme:
This tries to capture the learning from #660 for future use by agents.

This is AI written, under guidance to include learnings about how to implement and test `rename_symbol`, without mentioning that feature explicitly. 

@MischaPanch I've found that working in a codebase with a strong AGENTS.md makes a huge difference. I tried not to replicate details from CONTRIBUTING.md

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,79 @@
+# Agents Development Guide
+
+This guide outlines best practices for AI agents contributing to the Serena codebase. Serena is a powerful coding agent toolkit that provides IDE-like semantic code analysis and editing capabilities across many programming languages.
+
+## Core Development Principles
+
+### 1. Real-World Testing - Minimize Mocking Philosophy
+
+Serena **strongly prefers real integration testing** over mocking, especially for core language server functionality. However, mocks are used strategically for specific scenarios.
+
+**When to Use Real Integration:**
+- Language server interactions (symbol retrieval, code editing, LSP communication)
+- File system operations and repository structures
+- Cross-language feature testing
+- End-to-end tool workflows
+
+**When Mocks Are Acceptable:**
+- External dependencies (network calls, system utilities)
+- Environment detection and platform-specific behavior
+- Infrastructure setup that doesn't affect core functionality
+- Test utilities and file readers for isolated unit tests
+
+**Test Infrastructure:**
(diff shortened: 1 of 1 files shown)


MischaPanch:
I like the idea of making this repo more amenable to AI workflows. We already started along these lines, using the committed memories (in `.serena/memories`). The memories should be extended to cover more aspects of tests or architecture.

I think the way forward is for the AGENTS.md (and CLAUDE.md) to instruct the agent to read the memory file names and be aware of them, loading them when needed either using Serena's tool, or just its own read_file tool (for agents that don't use Serena). This is better than putting everything into AGENTS.md for the following reasons:

1. compatible with all agents without having to repeat information
2. for any task at hand, the agent will not need to know all aspects of the project. Memory files allow loading only the relevant aspects (deemed as such by the agent itself)
3. the current proposal for AGENTS.md is not quite suitable. Some aspects are overrepresented (like the EditingTest and snapshots, which form only a small part of the test suite), some are too general and not needed IMO (like using ABC - any smart enough agent will know that after reading our code), and many important aspects are not mentioned at all (using uv and poe, testing with existing markers, using `poe type-check` or `mypy`, the various test strategies, and most architectural aspects).

Putting everything needed for an agent to succeed into a single AGENTS.md will blow it up, and will also make it hard to maintain. Serena's memory management is very simple, but it's already better than a simple file. If it's not enough, we can consider adding summaries to the memory files, and instructing the agent (through AGENTS.md) to read the file names and summaries in order to have a better basis for knowing which memories are relevant. We can also instruct the agent to always read some general memories (no mocks, coarse architecture overview, the most important design principles of Serena etc).

A good test as to whether we have the right memories and whether the instructions in AGENTS.md regarding memories work would be to reproduce the `rename_symbol` feature automatically. It is a well defined and easily testable task, and we have the ground truth in #660 . Also, it is a task at which the copilot agent (that was the one you had used, right?) completely failed.

If you would like to invest time into this (I think Serena is a pretty good test case for how far one-shot development with AI can be pushed), we'd be happy to collaborate on making agents more efficient here

dsyme:
Totally the notes should just go in a memory file, and this should direct to read those

dsyme:
I'll close this, when I return to any dev work in the repo I'll bring it over

PR #674, opened 2025-10-16
Consistently use the project's encoding setting, also passing it on to the LS

opcode81:


src/serena/agent.py
```diff
@@ -47,6 +47,7 @@ class MemoriesManager:
     def __init__(self, project_root: str):
         self._memory_dir = Path(get_serena_managed_in_project_dir(project_root)) / "memories"
         self._memory_dir.mkdir(parents=True, exist_ok=True)
+        self._encoding = "utf-8"
```
MischaPanch: why always utf-8 and not the project's encoding? Maybe worth a comment

src/solidlsp/ls_utils.py
```diff
@@ -165,18 +165,18 @@ class FileUtils:
     """
 
     @staticmethod
-    def read_file(logger: LanguageServerLogger, file_path: str) -> str:
+    def read_file(logger: LanguageServerLogger, file_path: str, encoding: str) -> str:
```
MischaPanch, line 168: maybe utf-8 as default? solidlsp is a "library" after all, would make it a bit more user friendly

opcode81: No, this is a util function which is used internally only. Adding a default is detrimental. It would be wrong not to pass the specific encoding that is actually supposed to be used by the language server.
Actually, it is always wrong to try to read a text file without explicitly defining the encoding. utf-8 is not a default anyone should want to use.

src/serena/agent.py
opcode81: All of Serena's own files use utf-8 (i.e. all config files and memories), because it would be strange to allow a mix here, and our templates are written in utf-8. Serena does not need to follow the project's encoding.

I was thinking about adding a constant for this "Serena file encoding"...
So we could have `DEFAULT_SOURCE_FILE_ENCODING` (which is now `DEFAULT_ENCODING`) and `SERENA_FILE_ENCODING` (for everything pertaining to Serena itself).

opcode81: We now have these two constants, establishing more clarity.

PR #679, opened 2025-10-16
[README-only-change]: Add quotes around /home/nickolaykondratyev/git_…

nickolay-kondratyev:
…repos/nickolay-kondratyev_serena per shell best practices

If someone uses spaces in file names /home/nickolaykondratyev/git_repos/nickolay-kondratyev_serena is likely to not work as expected due to argument splitting (hopefully people arent using spaces, but its best practice to quote everything in shell)


opcode81:
Thanks!

PR #682, opened 2025-10-17
enhancing file read robustness

gsw945:
For the project where encoding cannot be consistent, when reading file content via the default encoding fails, attempt to determine the file encoding before re-reading.

### DIFF
--- a/src/solidlsp/ls_utils.py
+++ b/src/solidlsp/ls_utils.py
@@ -13,8 +13,8 @@ import zipfile
 from enum import Enum
 from pathlib import Path, PurePath
 
-import requests
 import charset_normalizer
+import requests
 
 from solidlsp.ls_exceptions import SolidLSPException
 from solidlsp.ls_logger import LanguageServerLogger
@@ -181,7 +181,10 @@ class FileUtils:
                 results = charset_normalizer.from_path(file_path)
                 match = results.best()
                 if match:
-                    logger.log(f"Could not decode {file_path} with encoding='{encoding}'; using best match '{match.encoding}' instead", logging.WARNING)
+                    logger.log(
+                        f"Could not decode {file_path} with encoding='{encoding}'; using best match '{match.encoding}' instead",
+                        logging.WARNING,
+                    )
                     return match.raw.decode(match.encoding)
                 raise ude
         except Exception as exc:


src/solidlsp/ls_utils.py
```diff
@@ -14,6 +14,7 @@
 from pathlib import Path, PurePath
 
 import requests
+from charset_normalizer import from_path
```
opcode81: Let's import `charset_normalizer` instead of directly `from_path`, such that we have the more semantically meaningful `charset_normalizer.from_path` below.

src/serena/project.py
```diff
@@ -85,7 +86,14 @@ def read_file(self, relative_path: str) -> str:
...
             raise FileNotFoundError(f"File not found: {abs_path}")
-        return abs_path.read_text(encoding=self.project_config.encoding)
+        try:
+            return abs_path.read_text(encoding=self.project_config.encoding)
+        except UnicodeDecodeError as ude:
+            results = from_path(abs_path)
+            match = results.best()
+            if match:
+                return match.raw.decode(match.encoding)
+            raise ude
```
opcode81: Let's make this call `FileUtils.read_file` from `ls_utils`, such that we have a single place for this.

src/solidlsp/ls_utils.py
```diff
@@ -173,8 +174,15 @@ def read_file(logger: LanguageServerLogger, file_path: str, encoding: str) -> st
...
-            with open(file_path, encoding=encoding) as inp_file:
-                return inp_file.read()
+            try:
+                with open(file_path, encoding=encoding) as inp_file:
+                    return inp_file.read()
+            except UnicodeDecodeError as ude:
+                results = from_path(file_path)
+                match = results.best()
+                if match:
+                    return match.raw.decode(match.encoding)
```
opcode81, line 188: Let's log a warning here, e.g.

    log.warning(f"Could not decode {file_path} with encoding='{encoding}'; using best match '{match.encoding}' instead")

opcode81 (changes requested):
Thanks! This is a much better way of handling this than the other PR.
I have made some suggestions that we should integrate before merging.

opcode81:
Just to be sure: You have verified in an example project that this solves your issue with mixed encodings?

src/serena/project.py
```diff
@@ -138,9 +139,8 @@ def read_file(self, relative_path: str) -> str:
         :return: the content of the file
         """
         abs_path = Path(self.project_root) / relative_path
-        if not abs_path.exists():
-            raise FileNotFoundError(f"File not found: {abs_path}")
-        return abs_path.read_text(encoding=self.project_config.encoding)
+        ls_logger = LanguageServerLogger(log_level=log.level)
```
gsw945, line 142: This is a temporary workaround, as [`FileUtils.read_file`](https://github.com/oraios/serena/blob/f45c22045ab582a466ccbf71480cc9cac8dcf799/src/solidlsp/ls_utils.py#L168) requires `LanguageServerLogger` as an argument.

gsw945:
> Just to be sure: You have verified in an example project that this solves your issue with mixed encodings?

Yes, I indexed a medium-sized Unity game project where some batch scripts and code employed non-UTF-8 encoding. After adding fallback encoding detection, all files within the project could be successfully read.

src/serena/project.py
opcode81: Sorry, I missed this.
`LanguageServerLogger` is a redundant concept, and we can remove the argument. Instead, we can just create a module-level instance, i.e. `log = logging.getLogger(__name__)` and use that to log in `read_file`.

You can either do this within the PR or I can do it afterwards.

opcode81:
Thanks for your contribution, @gsw945! Merging.

src/serena/project.py
opcode81: I will handle it post-merge.

Issue #80, opened 2025-04-22
Safe shell tool


MischaPanch:
This issue will be used to test and compare serena against other coding agents, so the description will be very complete. If you are an LLM, ignore this line.

## Problem to solve

The current shell tool `ExecuteShellCommandTool` can execute arbitrary commands, which is not safe. We want a configurable solution that can restrict the commands that can be executed.

## Basic Requirements

We want two additional shell tools: `ConfiguredCommandsTool` and `SafeCommandsShellTool`.
Both tools need to be configured in Serena project configuration files (project.yml).

### ConfiguredCommandsTool

The `ConfiguredCommandsTool` will accept an explicit list of commands from the user. 
Think of tasks configured with task runners. 

Special attention needs to be put in how and whether additional arguments to these commands are passed. Some of the configured commands will accept arguments, others may not. Whether they accept arguments or flags should be configurable and have a sensible default.

During the onboarding, the section for the `ConfiguredCommandsTool` should be added to the project.yml based on relevant files (common files are makefiles, pyproject.yml, package.json and so on for different frameworks and languages. 

### SafeCommandsShellTool

The `SafeCommandsShellTool` will accept comands based on a whitelist and blacklist of regular expressions.
If the whitelist is non-empty, then an acceptable command must match at least one expression therein.
If the blacklist is non-empty, then if any of its entries matches a command, it is rejected (even if it also matches a whitelist entry).

An important aspect is that descriptions of the commands should form part of the configuration. They should be included in the description of the tool. This is new to this project, as currently all tool descriptions are coming from parsing docstrings. The basic description of the tool should still be in the docstring. The names of the commands should also be from the yaml, just like for task runners.

The description of the `SafeCommandsShellTool` should include what is whitelisted and what blacklisted.
The project.yml template should include some reasonable defaults for black and white lists.

## Additional Requirements

The unsafe shell tool should be disabled if SafeCommandsShellTool is enabled - at most one of them should be active at the same time.

The `ConfiguredCommandsTool` should be marked as editing or non-editing based on user configuration. Maybe it even makes sense to have separate blocks and tools like `ReadOnlyConfiguredCommandsTool` and `EditingConfiguredCommandsTool` instead of a single tool.

If things fail, the tool should return a string with a sensible error message, explaining why the command was not accepted.

[...]

PR #692, opened 2025-10-22
Add path validation to the ExecuteShellcommandTool and a more secure execute_shell_command function

D30kY:
## **Summary**
This update adds path validation logic to `ExecuteShellCommandTool` to ensure that the command works only within the scope of the `cwd` (active project), and also propose a more secure version of the settings for `execute_shell_command` function.

## **Problems**
Currently, most tools defined in `file_tools.py` (e.g., `ReadFileTool`, `CreateTextFile`, `ListDirTool`, `FindFileTool`) include security logic to verify that any file system access occurs within the scope of the `cwd` (Active Project).

However, `ExecuteShellCommandTool` lacks this validation, allowing a user-provided `command` to perform operations on paths outside the `cwd` scope.

A more critical issue is that the `execute_shell_command()` function this tool relies on is implemented as `subprocess.Popen(command, shell=True, ...)`. The `shell=True` option is highly vulnerable to **Command Injection**.

The combination of these two issues creates a critical security vulnerability. An attacker can inject shell meta-characters (e.g., `;`, `|`) along with Path Traversal syntax (e.g., `..`) into the `command` argument, enabling arbitrary command execution on paths completely outside the `cwd` scope (e.g., `cat ../../etc/passwd`).

## **Changes**
The following changes address this issue:

1. **Add more secure function to defense command injection (`safe_execute_shell_command`):**
    - The `subprocess.Popen` call is changed from `shell=True` to `shell=False`.
    - The `command` string is safely parsed into a list using `shlex.split()` before being passed to `subprocess.Popen`. This fundamentally blocks shell meta-characters from being interpreted as commands.
    - To accommodate users who may still rely on the old behavior, the original `execute_shell_command` function is not deleted. Instead, its invocation in `cmd_tools.py` is commented out.
2. **Add path validation to `ExecuteShellCommandTool`:**
    - Before the tool executes, the `command` string is parsed using `shlex.split()` to inspect individual arguments.
    - If any argument is identified as a potential path (e.g., starting with or containing `/`, `.`, or ), it is checked using logic similar to `_validate_path_containment` (or `is_path_in_project`) to ensure it is within the `cwd` scope.
    - If any path is detected outside the permitted scope, the command execution is blocked, and an exception is raised.

## **Notes**
While this update blocks basic shell injection and explicit path traversal, it serves as a primary, code-level defense. Limitations still exist, and caution should be exercised when using this Tool.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -286,5 +286,10 @@ markers = [
 # Ref: https://github.com/codespell-project/codespell#using-a-config-file
 skip = '.git*,*.svg,*.lock,*.min.*'
 check-hidden = true
+
+[dependency-groups]
+dev = [
+    "poethepoet>=0.36.0",
+]
 # ignore-regex = ''
 # ignore-words-list = ''
--- a/src/serena/tools/cmd_tools.py
+++ b/src/serena/tools/cmd_tools.py
@@ -6,27 +6,26 @@ import os.path
 import shlex
 
 from serena.tools import Tool, ToolMarkerCanEdit
-from serena.util.shell import execute_shell_command, safe_execute_shell_command
+from serena.util.shell import safe_execute_shell_command
 
 
 class ExecuteShellCommandTool(Tool, ToolMarkerCanEdit):
     """
     Executes a shell command.
     """
+
     def _validate_path_containment(self, path: str, base_dir: str):
 
         real_base_dir = os.path.realpath(base_dir)
-        
+
         if not os.path.isabs(path):
             path = os.path.join(real_base_dir, path)
-            
+
         real_path = os.path.realpath(path)
         common_path = os.path.commonpath([real_base_dir, real_path])
         if common_path != real_base_dir:
-            raise PermissionError(
-                f"Path traversal attempt blocked. The path '{path}' is outside the allowed directory '{base_dir}'."
-            )
-        
+            raise PermissionError(f"Path traversal attempt blocked. The path '{path}' is outside the allowed directory '{base_dir}'.")
+
     def apply(
         self,
         command: str,
@@ -64,10 +63,10 @@ class ExecuteShellCommandTool(Tool, ToolMarkerCanEdit):
         try:
             parts = shlex.split(command)
             for part in parts:
-                if '/' in part or part.startswith('.') or part.startswith('-'):
+                if "/" in part or part.startswith(".", "-"):
                     potential_path = os.path.join(_cwd, part)
                     if os.path.exists(os.path.dirname(potential_path)):
-                         self._validate_path_containment(part, _cwd)
+                        self._validate_path_containment(part, _cwd)
 
         except PermissionError:
             raise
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -280,12 +280,16 @@ markers = [
...
   "markdown: language server running for Markdown",
-  "julia: Julia language server tests",
 ]
 
 [tool.codespell]
 # Ref: https://github.com/codespell-project/codespell#using-a-config-file
 skip = '.git*,*.svg,*.lock,*.min.*'
 check-hidden = true
+
+[dependency-groups]
```
MischaPanch, line 290: pls undo

uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 1
+revision = 3
```
MischaPanch, line 2: pls undo the change to this file

src/serena/util/shell.py
```diff
@@ -42,6 +43,38 @@ def execute_shell_command(command: str, cwd: str | None = None, capture_stderr:
     return ShellCommandResult(stdout=stdout, stderr=stderr, return_code=process.returncode, cwd=cwd)
 
 
+def safe_execute_shell_command(command: str, cwd: str | None = None, capture_stderr: bool = False) -> ShellCommandResult:
```
MischaPanch, line 46: @opcode81 could you have a look pls? I don't think we need two functions, maybe this should replace the already existing one?

src/serena/tools/cmd_tools.py
```diff
@@ -47,6 +60,22 @@ def apply(
                         f"Specified a relative working directory ({cwd}), but the resulting path is not a directory: {_cwd}"
                     )
 
-        result = execute_shell_command(command, cwd=_cwd, capture_stderr=capture_stderr)
+        try:
+            parts = shlex.split(command)
```
MischaPanch, line 64: I'm really not sure that this works, there can be many edge cases with commands. There might be too many false positives.

At the very least, the command checking logic should be factored out into a separate util function and extensively tested.

@opcode81 WDYT? Does it make sense to attempt this? I am rather skeptical

MischaPanch (changes requested):
Hi hi. As we discussed per email, when the agent can write code and execute it, we cannot speak of safety (outside of a sandbox), so this is emphatically **not a critical safety issue** (!).

I am not sure that the path checking you have in mind can be reliably performed, but maybe I'm wrong. If it is indeed possible, the logic should be extensively tested, as I mentioned in a comment above.

@opcode81 pls pitch in :)

MischaPanch:
After discussing internally (and also since there's no answer), I'm closing this. The checks on paths are brittle and add complexity. My recommendation to look at #80 as essentially the only way to have a truly safe shell tool is not reflected in the code.

Checking paths will not give safety, and pseudo safety is not worth the additional complexity. If the LLM is assumed to be a bad actor, no amount of superficial safeguarding will suffice anyway - only a sandbox or something along the lines of #80 can help

PR #696, opened 2025-10-25
Add support for Godot Gdscript Language Server

nguyenchiencong:
## Description

Adds support for the Godot GDScript language server, including automatic Godot setup, TCP communication, and safe shutdown handling.

### Core Implementation

* **New**: `src/solidlsp/language_servers/gdscript_language_server.py`
  
  * Implements `GDScriptLanguageServer`, complete with Godot discovery, download, and launch logic
  * Supports external TCP endpoints and optional version overrides via `ls_specific_settings`
  * Tracks spawned Godot processes, adds atexit cleanup, and ensures graceful termination on shutdown

* **Modified**: `src/solidlsp/ls_handler.py`
  
  * Enhances TCP transport (socket setup, recv loop, header parsing) and tightens concurrency handling
  * Adds request logging refinements and supports toggling Content-Type headers per transport

* **Modified**: `src/solidlsp/lsp_protocol_handler/server.py`
  
  * Allows callers to suppress the `Content-Type` header when needed (Godot TCP startup quirk)

### Language Registration & Runtime Wiring

* **Modified**: `src/solidlsp/ls_config.py`
  
  * Registers GDScript language metadata, file associations, and ties the enum to the new LSP class

* **Modified**: `src/serena/agent.py`
  
  * Adds web dashboard shutdown callback to stop the LSP and GUI log window, preventing orphaned Godot processes

* **Modified**: `README.md`
  
  * Documents GDScript as an officially supported language

* **Modified**: `pyproject.toml`
  
  * Adds `gdscript` pytest marker to the global test configuration

### Tests

* **New**: `test/solidlsp/godot`

## Testing

```
uv run pytest test/solidlsp/godot -v
```

_All tests passing ✅_

### Godot Installation & Versioning

* Automatically locates Godot executables via environment variable, PATH, or OS-specific directories
* Downloads the requested release (default `4.5.1-stable`; override via `ls_specific_settings.gdscript.godot_version`)
* Ensures executables are symlinked/copied with proper permissions on non-Windows platforms

### TCP Transport Considerations

* Serena now omits the `Content-Type` header for raw TCP payloads, matching Godot’s expectations
* The reader loop switches to buffered header parsing to handle partial TCP frames reliably
* Idle socket logs are throttled to avoid spamming during long waits

### Process Cleanup

* Every spawned Godot PID is tracked; shutdown and atexit hooks terminate the entire process tree
* Dashboard shutdown callback ensures the language server stops even when users close the UI abruptly

## Checklist

* [x] Tests added and passing
* [x] LSP handler changes reviewed for other languages (stdin/stdout paths unaffected)
* [x] Code formatted with existing tooling
* [x] Documentation updated (`README.md`)
* [ ] Changelog entry prepared (pending release planning)


MischaPanch:
Hi. Thanks for the PR, I appreciate the effort and we'd be happy to have Godot support in Serena.

However, the PR is currently in an unreviewable state. There's large changes to core logic, accompanied by tests which are partly nonsensical (assert hasattr and so), use mocks, don't test the core functionality of the language server (pls see our guide on adding support for a new language) and generally are not sufficiently expressive while being overly verbose at the same time.

It is clear that AI was used to write large chunks of the code, and while that's not a problem, we (the reviewers) cannot take on the job of reviewing AI generated code and bringing it up to the necessary quality. This has to be done by the PR author.

The changes to the core logic introduce a lot of complexity and are done in a procedural way, which is not the way to go. It seems like you want to permit an alternative to stdio communication on a process between Serena and the language server in solidlsp, here using TCP sockets. This is valuable in itself, and I have prototyped something along these lines recently (for another use case).

I am not sure why that is needed for Godot, could you explain that?

If the TCP based connection is really needed and you want to see Godot support merged, I propose to do the following:

1. Abandon this PR, open an issue about adding an abstraction to permit support of other form of communication between SolidLanguageServer and the language server process, and implement this abstraction for TCP sockets. Since this affects core components of Serena, this has to be designed and done well. If you feel up for it, you can make a proposal for a design. Alternatively, you could wait until I finish the one I was prototyping in the past, but it will take a while, since we have higher priority topics at the moment.
2. After the above extension to solidlsp is merged, a new PR is written to support Godot which uses this and avoids using mocks and actually tests the core LS functionality.

nguyenchiencong:
Sorry, I was just setting the stage for the pull request. Didn't expect that someone would review it on the weekend. Retrospectively, it seems premature. 

As I was integrating with Godot, I found out that Godot LSP does not support the standard stdio transport mode. [This PR](https://github.com/godotengine/godot/pull/35864) switches the LSP from WebSocket to TCP. 

I recently added some cleanup and some core functionality testing.
Feel free to accept the pull request or not, it's really up to you, I'm ok with using a fork and finish cleaning up the integration over there.

MischaPanch:
All good, take your time. We can't just merge this as it touches the core logic, and any new code doing that has to be of sufficient quality. That's why I proposed doing it in two parts. As a bonus, adding a proper abstraction for the communication with the LS process would also enable websocket support, which would be a very nice feature (and a large step towards using VSCode as a replacement for the language servers).

PR #698, opened 2025-10-26
Add Fortran language support via fortls

suvarchal:
This PR adds support for the Fortran programming language to Serena using the fortls (Fortran Language Server) implementation.

Implementation Details

  The Fortran support implementation includes:

  Core Implementation (Previous commits)

  - Language Server: src/solidlsp/language_servers/fortran_language_server.py
    - Uses fortls for LSP integration
    - Supports comprehensive Fortran file extensions (.f90, .F90, .f95, .F95, .f03, .F03, .f08, .F08, .f, .F, .for, .FOR, .fpp, .FPP)
    - Configures Fortran-specific ignored directories (build, mod, obj, CMakeFiles)
    - Implements wait time (3.0s) for workspace indexing
  - Language Registration: src/solidlsp/ls_config.py
    - Added FORTRAN to Language enum
    - Configured file matchers for all modern and legacy Fortran extensions
    - Registered FortranLanguageServer in factory method
  - Test Repository: test/resources/repos/fortran/test_repo/
    - Minimal but representative test project
    - Demonstrates modules, functions, subroutines, and cross-file usage
    - Files: main.f90 and lib/math_utils.f90
  - Test Suite: test/solidlsp/fortran/test_fortran_basic.py
    - 6  tests covering:
        - Language server initialization
      - Module symbol finding
      - Function/subroutine symbol finding
      - Program symbol finding
      - Cross-file module usage
    - All tests verify actual symbol names (not just non-null responses)
- update README.md: Added Fortran to supported languages list with installation instructions
- update CHANGELOG.md: Documented Fortran support addition

Tests can be run selectively with: `pytest -m fortran`

### DIFF
--- a/test/solidlsp/fortran/test_fortran_basic.py
+++ b/test/solidlsp/fortran/test_fortran_basic.py
@@ -71,14 +71,12 @@ class TestFortranLanguageServer:
 
         assert add_numbers_symbol is not None, "Could not find 'add_numbers' function symbol in math_utils.f90"
 
-        # WORKAROUND: fortls bug - selectionRange.start is at character 0 (line start),
-        # but references only work when queried from the function name (character 13).
-        # Line 5 (0-indexed: 4) is: "    function add_numbers(a, b) result(sum)"
-        #                                      ^13 characters from start
+        # Use selectionRange to query for references
+        # Note: FortranLanguageServer automatically fixes fortls's incorrect selectionRange
         sel_start = add_numbers_symbol["selectionRange"]["start"]
 
-        # Use character 13 to query from the function name, not the "function" keyword
-        refs = language_server.request_references(file_path, sel_start["line"], 13)
+        # Query from the function name position using corrected selectionRange
+        refs = language_server.request_references(file_path, sel_start["line"], sel_start["character"])
 
         # Should find references (usage in main.f90 + definition in math_utils.f90)
         assert len(refs) > 0, "Should find references to add_numbers function"
@@ -123,8 +121,7 @@ class TestFortranLanguageServer:
         the containing symbols that have the references. This is different from
         test_find_references_cross_file which only returns locations.
 
-        NOTE: This test works around a fortls bug where selectionRange points to the start
-        of the line instead of the function name.
+        Note: FortranLanguageServer automatically fixes fortls's incorrect selectionRange.
         """
         # Get the add_numbers function symbol from math_utils.f90
         file_path = "modules/math_utils.f90"
@@ -139,11 +136,10 @@ class TestFortranLanguageServer:
 
         assert add_numbers_symbol is not None, "Could not find 'add_numbers' function symbol"
 
-        # WORKAROUND: fortls bug - selectionRange.start is at character 0 (line start),
-        # but references only work when queried from the function name (character 13).
+        # Use selectionRange to query for referencing symbols
+        # FortranLanguageServer automatically corrects fortls's incorrect selectionRange
         sel_start = add_numbers_symbol["selectionRange"]["start"]
-        # Use character 13 to query from the function name, not the "function" keyword
-        referencing_symbols = language_server.request_referencing_symbols(file_path, sel_start["line"], 13)
+        referencing_symbols = language_server.request_referencing_symbols(file_path, sel_start["line"], sel_start["character"])
 
         # Should find referencing symbols (not just locations, but symbols containing the references)
         assert len(referencing_symbols) > 0, "Should find referencing symbols when querying from function name position"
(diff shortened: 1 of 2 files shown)


test/solidlsp/fortran/test_fortran_basic.py
```diff
@@ -0,0 +1,100 @@
...
+    # Find all symbol names
+    all_names = [s.get("name") for s in all_symbols]
+
+    assert "test_program" in all_names, f"Program 'test_program' not found. Found symbols: {all_names}"
+
+
+def test_cross_file_module_usage(fortran_ls):
+    """Test that the language server can detect cross-file module usage."""
+    # Open both files to ensure they're indexed
+    fortran_ls.open_file("main.f90")
```
MischaPanch: this should not be necessary, we can't expect a user to open files before calling the tools. Does fortls actually need this?

.github/workflows/pytest.yml
```diff
@@ -241,6 +241,9 @@ jobs:
           echo "PERL_MB_OPT=--install_base \"$HOME/perl5\"" >> $GITHUB_ENV
           echo "PERL_MM_OPT=INSTALL_BASE=$HOME/perl5" >> $GITHUB_ENV
           echo "$HOME/perl5/bin" >> $GITHUB_PATH
+      - name: Install Fortran Language Server
+        shell: bash
+        run: pip install fortls
```
MischaPanch: Since fortls is in python, we could install it as part of serena (like pyright). Maybe as an extra, although it would require users to modify their uvx command.

@opcode81 wdyt? It would increase package size, but maybe not unacceptably so

MischaPanch (changes requested):
Thanks for the PR! We need to make sure cross-file referencing works without the user requesting to open the files beforehand before merging it.

test/solidlsp/fortran/test_fortran_basic.py
suvarchal: Oh thanks good catch, fortls doesn't need it to be open could do this instead, will update the test in a bit.
```
    all_symbols, root_symbols = fortran_ls.request_document_symbols("main.f90")
```

.github/workflows/pytest.yml
suvarchal: made changes but for some reason tests on mac-os fail because of a unrelated C#  test, i will mark this resolved.

test/solidlsp/fortran/test_fortran_basic.py
```diff
@@ -0,0 +1,99 @@
...
+    # Import here to avoid import errors if fortls is not installed
+    from solidlsp.language_servers.fortran_language_server import FortranLanguageServer
+
+    ls = FortranLanguageServer(config, logger, fortran_repo_path, settings)
+    ls.start()  # Start the language server
+    yield ls
+    ls.stop()
+
+
+def test_fortran_ls_initialization(fortran_ls):
```
MischaPanch: All these tests just call `request_document_symbols`, so cross-file reference finding is not tested. Pls adjust the tests to be like for the other language servers, have a look at the memory on how to add a language server or on some other test module.

Group the functions into one test class, and make sure you test request_referencing_symbols to do its job - this is usually the brittle part of language servers.

Reduce the number of tests that just test `request_document_symbols` to a single one, no need to repeat for different symbols.

If you want to be really thorough, pls also check that `request_defining_symbol` works across files (the definition is in a different file than the usage)

MischaPanch:
@suvarchal I pushed the installation of fortls as part of the serena package (you closed the conversation thread without doing that), so make sure to pull my commit before you push additional ones :)

suvarchal:
Sorry about that I know I shouldn't resolve conversations on my mobile, tenseness creeps in. Will restructure tests and get back.

test/solidlsp/fortran/test_fortran_basic.py
```diff
@@ -1,99 +1,222 @@
...
+        # Find the add_numbers function
+        add_numbers_symbol = None
+        for sym in symbols[0]:
+            if sym.get("name") == "add_numbers":
+                add_numbers_symbol = sym
+                break
+
+        assert add_numbers_symbol is not None, "Could not find 'add_numbers' function symbol in math_utils.f90"
+
+        # WORKAROUND: fortls bug - selectionRange.start is at character 0 (line start),
```
MischaPanch: This is a problem - we use the exact character to find references in high-level interfaces, specifically in `LanguageServerSymbol.column`.

It means that finding references for symbols will fail for fortran in the MCP server.

I had a similar problem for the Go LS in the past which I solved by configuring

```
                    "documentSymbol": {
                        "dynamicRegistration": True,
                        "hierarchicalDocumentSymbolSupport": True,
                        "symbolKind": {"valueSet": list(range(1, 27))},
                    },
```

Maybe some config option is missing for fortls? It is paramount that the LS returns the correct selection range for the MCP server to work

MischaPanch: If there is no way to solve this through configuration, the fortran LS implementation in Serena has to add a workaround by overriding some methods and ensuring that the location is adjusted in all outputs

suvarchal: I will have to dig a bit more, I was contemplating on fixing at fortls first but may be address it in fortran LS implementation could be quicker. Thanks for the hints will get back.

MischaPanch: Great, thanks for the work on this and good luck! Hopefully fortls won't pose you too much opposition

suvarchal: Implemented a workaround in LS implementation in serena to deal with fortls selectionRange bug overriding request_document_symbols, checking for most commonly used fortran constructs. I used this version as serena mcp with claude code with relatively large fortran code base code: https://github.com/FESOM/fesom2 to find types, subroutines and functions it seems to work as  expected.

PR #701, opened 2025-10-27
feat: Support user's Maven and Gradle configuration files in JDTLS

Rheasilvia:
Add detection and usage of user's custom Maven and Gradle configuration files to allow JDTLS to respect user-specific build tool settings.

Changes:
- Detect and use ~/.m2/settings.xml for Maven user settings
- Detect and use ~/.gradle for Gradle user home directory
- Set Maven updateSnapshots to false for better performance

This enhancement allows JDTLS to:
- Use custom Maven repositories and mirrors
- Respect proxy and authentication configurations
- Apply user-specific build tool settings
- Improve initialization performance by disabling automatic snapshot updates

The implementation gracefully falls back to None if configuration files don't exist, maintaining backward compatibility with existing setups.

### DIFF
--- a/src/solidlsp/language_servers/eclipse_jdtls.py
+++ b/src/solidlsp/language_servers/eclipse_jdtls.py
@@ -39,8 +39,24 @@ class RuntimeDependencyPaths:
 
 
 class EclipseJDTLS(SolidLanguageServer):
-    """
+    r"""
     The EclipseJDTLS class provides a Java specific implementation of the LanguageServer class
+
+    You can configure the following options in ls_specific_settings (in serena_config.yml):
+        - maven_user_settings: Path to Maven settings.xml file (default: ~/.m2/settings.xml)
+        - gradle_user_home: Path to Gradle user home directory (default: ~/.gradle)
+
+    Note: Gradle wrapper is disabled by default. Projects will use the bundled Gradle distribution.
+
+    Example configuration in ~/.serena/serena_config.yml:
+    ```yaml
+    ls_specific_settings:
+      java:
+        maven_user_settings: "/home/user/.m2/settings.xml"  # Unix/Linux/Mac
+        # maven_user_settings: 'C:\\Users\\YourName\\.m2\\settings.xml'  # Windows (use single quotes!)
+        gradle_user_home: "/home/user/.gradle"  # Unix/Linux/Mac
+        # gradle_user_home: 'C:\\Users\\YourName\\.gradle'  # Windows (use single quotes!)
+    ```
     """
 
     def __init__(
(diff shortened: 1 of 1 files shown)


src/solidlsp/language_servers/eclipse_jdtls.py
```diff
@@ -326,6 +326,16 @@ def _get_initialize_params(self, repository_absolute_path: str) -> InitializePar
             repository_absolute_path = os.path.abspath(repository_absolute_path)
         repo_uri = pathlib.Path(repository_absolute_path).as_uri()
 
+        # Check for user's Maven and Gradle configuration files
+        # This allows JDTLS to respect user's custom build tool configurations
+        # such as custom repositories, proxies, and authentication settings.
+        # If these files don't exist, JDTLS will use its default settings.
+        maven_settings_path = os.path.expanduser("~/.m2/settings.xml")
```
MischaPanch: Will this work on all platforms? And is it common for these files to be somewhere else, or are they always in these locations? 

I think it would be better to use the ls_specific_settings for this, then the user will be able to configure alternative locations through serena_config.yml. The usual locations can be the default then

src/solidlsp/language_servers/eclipse_jdtls.py
```diff
@@ -549,7 +559,9 @@ def _get_initialize_params(self, repository_absolute_path: str) -> InitializePar
                             ],
                             "generatesMetadataFilesAtProjectRoot": False,
                         },
-                        "maven": {"downloadSources": True, "updateSnapshots": True},
+                        # Set updateSnapshots to False to improve performance and avoid unnecessary network calls
+                        # Snapshots will only be updated when explicitly requested by the user
+                        "maven": {"downloadSources": True, "updateSnapshots": False},
```
MischaPanch, line 621: I don't know why this was set to True before, does it have any advantage for Serena's purposes to download snapshots? If yes, also this could be configurable through ls_specific_settings

src/solidlsp/language_servers/eclipse_jdtls.py
Rheasilvia: Yes, it works on all platforms (Windows, macOS, and Linux). While these files are typically in the default locations, users sometimes need to specify custom paths (e.g., when using non-standard installation directories or custom project setups).

I've added configuration options to `ls_specific_settings` in `serena_config.yml`, so users can override the default locations when needed. The standard paths remain as defaults, which should work for most cases.

Rheasilvia: I tested with the Kafka repo(https://github.com/apache/kafka). Using a mirror repository significantly speeds up Gradle/Maven project imports. Below is a comparison of first vs. second startup times. If not, it will timeout for me.

WXWorkCapture_17616244494719

WXWorkCapture_17616244982189

src/solidlsp/language_servers/eclipse_jdtls.py
Rheasilvia: Setting it to True would auto-check for snapshot updates on every build, causing unnecessary network calls and slower imports. For Serena's use case (code analysis, not dependency management), this doesn't add value. Developers typically update snapshots explicitly via CLI or IDE when needed.

I've defaulted to False for better performance. Can add it to ls_specific_settings if you think it's useful, but I suspect most users won't need it.

src/solidlsp/language_servers/eclipse_jdtls.py
```diff
@@ -326,6 +342,23 @@ def _get_initialize_params(self, repository_absolute_path: str) -> InitializePar
             repository_absolute_path = os.path.abspath(repository_absolute_path)
         repo_uri = pathlib.Path(repository_absolute_path).as_uri()
 
+        # Load user's Maven and Gradle configuration paths from ls_specific_settings["java"]
+        custom_java_settings = self._solidlsp_settings.ls_specific_settings.get(self.get_language_enum_instance(), {})
+
+        # Maven settings: default to ~/.m2/settings.xml
+        default_maven_settings = os.path.join(os.path.expanduser("~"), ".m2", "settings.xml")
```
MischaPanch, line 349: Almost there, but we should fail if the user actually provided an explicit path but it doesn't exist. Currently, this would just silently be ignored. So pls adjust to handle that case, failing with `FileNotFoundError` if something is explicitly set and the path doesn't exist, and only using the default if nothing was set (here and for gradle below), then just using None if no the file was not found (but logging it on info level, I suspect most users would want to be informed on settings not being found)

After that, this can be merged

MischaPanch:

Issue #611, opened 2025-09-08
Support for Projects with Multiple Languages


NaikSoftware:
I set up Serena in a cross-platform project and by default, LSP is now set up for Dart. But the project also has Java and TypeScript. It would be good to have the ability to configure multiple LSPs in the project.

bayasdev:
+1

piotr-piatkowski:
Same here, I have single project (git repo) with backend written in python (django) and frontend in JS (vue+) - should I run two serena instances with seperate configuration, or what?

NaikSoftware:
I found solution: use OpenCode, it has built-in LSP multilingual support))

avral:
yeah. .vue js that contains js/ts, css/scss, html/pug is not working. it's annoying

taqtiqa-mark:
Another option is https://github.com/jonrad/lsp-mcp
However you'd have to disable the relevant serena tools and instruct the agent to use the relevant `lsp-mcp` tools.... not sure how that would work out?

MischaPanch:
We know that this is one of  the top requested features and we are working on supporting this. There are several conversations in past issues and PRs on this, so I'm not going to repeat things said there. We will give an update in this issue once progress has been made.

taqtiqa-mark:
> There are several conversations in past issues and PRs on this, so I'm not going to repeat things said there.

Ack. Thanks. In case anyone here wants a potential workaround, see https://github.com/oraios/serena/issues/208#issuecomment-3448865499.

opcode81:
The feature is now implemented in #704.

opcode81:
The feature is now on main, and you even have the option to dynamically add new languages via the dashboard:

<img width="577" height="340" alt="Image" src="https://github.com/user-attachments/assets/42817943-5ab1-4f24-a11a-0c37fda32200" />

Issue #492, opened 2025-08-16
Monorepo/multirepo procedure unclear


bitnom:
I run the MCP from my monorepo/multirepo (It's a multi-language multirepo) root. I see a .serena/project.yml with `language:typescript` at the top. This gives me anxiety, and I uninstall the MCP.

Am I going to have to maintain per-subrepo configs? That's fine but it will mean I can only use serena in a more limited way. Most of my AI-assisted development begins at the monorepo root since many of the projects contained within are interdependent.

TimAidley:
I am just trying serena out, and I have a similar issue. I am porting an old game written in C to C# in Unity. My root folder looks something like this:

```
Root
  -- UnityProject     // Root of the Unity Project
  -- original         // original source lives here
```

But it looks like I can only have one language in a project?

MischaPanch:
Monorepo support will be added through the jetbrains extension, we are essentially finished with it and are working on releasing it asap.

See discussion here for reference: #192

opcode81:
Monorepo support will be added by #704

opcode81:
Multiple languages are now fully supported on main; you even have the option to dynamically add new languages for the current project via the dashboard:

<img width="577" height="340" alt="Image" src="https://github.com/user-attachments/assets/70ad60ef-c505-4f35-8f2c-c58cd5f14e26" />

PR #704, opened 2025-10-28
Add LanguageServerManager, adding simultaneous multi-language support

opcode81:
Allows multiple language servers to be used by a single project, adding mono-repo support and support for projects that simply use more than one language extensively.

Projects can now define more than one language (server) in project.yml:

```yaml
languages:
  - python
  - typescript
```

Projects still defining `language` continue to work as expected.

Implements #611
Implements #492

### DIFF
--- a/src/serena/ls_manager.py
+++ b/src/serena/ls_manager.py
@@ -1,5 +1,8 @@
 import logging
+import threading
 from collections.abc import Iterator
+
+from sensai.util.logging import LogTime
 
 from serena.constants import SERENA_MANAGED_DIR_IN_HOME, SERENA_MANAGED_DIR_NAME
 from solidlsp import SolidLanguageServer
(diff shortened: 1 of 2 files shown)


MischaPanch:
@opcode81 ready for review, right? Could you format pls so CI can run through

src/serena/tools/symbol_tools.py
```diff
@@ -42,7 +42,7 @@ def apply(self) -> str:
         """Use this tool only on explicit user request or after confirmation.
         It may be necessary to restart the language server if it hangs.
         """
-        self.agent.reset_language_server()
+        self.agent.reset_language_server_manager()
```
MischaPanch, line 45: I think we should be able to reset only one LS - the one that failed. Can be done in a follow-up PR since it's not urgent, there were no reports of the LS process failing and all previous failures that led us to adding this tool at all were likely due to asyncio

opcode81: If a specific LS fails and this is known, it is restarted individually. I specifically added code to track the failing LS:
https://github.com/oraios/serena/blob/388e01a3f72a6ad042e0ea91b3666104f9c2842a/src/serena/tools/tools_base.py#L247-L265

Also, if a required LS is not actually running, it is restarted upon retrieval:

https://github.com/oraios/serena/blob/388e01a3f72a6ad042e0ea91b3666104f9c2842a/src/serena/ls_manager.py#L34-L49

This tool is, therefore, mostly obsolete, as the automatic mechanisms above will mostly do the right thing. But the equivalent operation to restarting the single language server is definitely to restart the entire manager.

MischaPanch: Then let's remove the tool, or at least make it optional, so it doesn't pollute the default toolset

opcode81:
> we could be smarter about autodetection of languages now, automatically inferring a list (instead of just the main one) if some threshold of files for the subdominant languages is found (excluding markdown). I would even suggest a fairly low threshold. This can also be implemented in a follow-up, but I think it would be better to do it immediately, such that new monorepo users have a smooth experience out of the box

I don't agree that what you are suggesting is indeed "smarter". This is a slippery slope. If a language is mostly irrelevant to day-to-day development, yet it is a language where the LS cannot seamlessly be installed on the fly (on all the relevant systems), then this will cause more problems than it solves. Not all LS can just be installed everywhere without any problems or without requiring system dependencies. If this affects mostly irrelevant languages, then it becomes a nuisance.

Three alternative routes:

**Less is more.** It might be better to be very conservative about the languages we just install. Using only the main language is the "safe" default. 

**A user-defined set of generally supported languages**. Auto-detection just isn't a very good mechanism. Ultimately, I would prefer for this to be a deliberate decision by the user. In an IDE, this problem does not exist, because an IDE never just installs arbitrary extensions without asking the user; instead, all languages which the user has *once* declared as relevant (by installing the respective extension) will always work for any new project. We could do something similar for Serena: The user could declare which set of languages, when detected, will just always be enabled.

**Dynamically adding a language**. Another way to approach this is for the set of language servers to be dynamic. It would be straightforward to support the dynamic addition of a language server while the project is already activated, e.g. via a tool `AddLanguageServer`. This could be a part of the onboarding process, for example. The list of detected languages can just be extended dynamically.

But there is really no reason to conflate this with this PR. It can be added at any time.

MischaPanch:
> But there is really no reason to conflate this with this PR. It can be added at any time.

I agree, let's postpone the discussion. Maybe a useful default behavior can be found that does detect the languages for a monorepo, maybe not, no need to decide now.

src/serena/tools/symbol_tools.py
opcode81: I don't recall ever seeing this tool being a problem solver, so disabling it seems like a reasonable thing to do.

opcode81: @MischaPanch you already disabled this tool on August 9.

PR #711, opened 2025-10-30
feat: Add CLI tool to download language server dependencies

### LINKED ISSUE #664: Run selena in a container without egres
I'm trying to run serena in a container which has only limited access to the net. Only a few hosts are allowed. When serena initializes the project it starts accessing hosts like: services.gradle.org, visualstudioexptteam.gallery.vsassets.io, github.com, plugins.gradle.org, etc.. 

Provide a 'build' command that pre-fetches all of serena's dependencies. I don't care for disk space.

Opening each and every one of the hosts it tries to access. I don't know which those are and it might also break in a later version if new dependencies are added. It seems that depending on the project language different tools are downloaded.

Another alternative would be to create a few language specific fake project just for the sake of downloading the dependecies. I can do this at container build time, when there is network access.

**Additional context**

giovanimds:
Closes #664

Rebased onto current `main` and reimplemented from scratch — the previous version of this branch predated the `LanguageServerDependencyProvider` refactoring and no longer applied.

## What this PR does

Adds the CLI command

```shell
serena download-ls-dependencies <language server>...   # or --all
```

which downloads the runtime dependencies of the given language servers ahead of time, so that they need not be fetched on demand later on. This is the use case from #664: a container that may only reach a small, unstable-across-versions set of hosts at runtime, but has full network access at build time.

Following @opcode81's suggestion, the download is triggered by *instantiating* the language server (in a temporary directory, since the repository contents are never looked at), which constructs its launch command and thereby installs the dependencies.

## The uvx case

Instantiation alone is not sufficient for language servers run through `LanguageServerDependencyProviderUvx` (e.g. `python`/pyright): constructing the launch command only assembles `uvx --from <pkg>==<version> <entrypoint>`; the package environment is materialised when the command is *executed*. Such a server would report success without anything having been downloaded — precisely the case that then fails inside the offline container.

The dependency provider abstraction therefore gains an explicit part of its interface:

* `LanguageServerDependencyProvider.install_dependencies()`, whose default implementation covers the regular case (dependencies are installed while the launch command is constructed),
* overridden in `LanguageServerDependencyProviderUvx` to install the pinned package as a uv tool, so the subsequent `uvx` invocation reuses the existing installation. A user-provided `ls_path`/`ls_base_cmd` bypasses it, as there is then nothing for Serena to install.
* `SolidLanguageServer.install_dependencies()` exposes this on the language server itself.

## Behaviour

* Failures are collected per language server rather than aborting the run — a missing external toolchain (e.g. Go, Java) must not prevent the remaining downloads — and are reported with a non-zero exit code, so a container build fails early instead of silently producing an image with missing dependencies.
* The global `ls_specific_settings` are taken into account (without generating a config file if none exists), so users who point Serena at an existing installation don't download anything they don't need.

## Verification

* `ruff format` / `ruff check` / `ty check` clean (the pre-existing `ty` diagnostics on `main` are untouched).
* New unit tests in `test/serena/test_cli_download_ls_dependencies.py` covering selection, `--all`, partial failure + exit code, and argument validation.
* Manually verified end-to-end: `python` (uvx path → `uv tool install pyright==1.1.403`), `lua` (binary download), and `go` on a machine without Go (fails with the language server's own error message, exit code 1, remaining downloads unaffected).

Documentation added under "Language Support" and an entry in the CHANGELOG.

### DIFF
--- a/src/serena/cli.py
+++ b/src/serena/cli.py
@@ -5,6 +5,7 @@ import os
 import shutil
 import subprocess
 import sys
+import tempfile
 import time
 from collections.abc import Iterator, Sequence
 from logging import Logger
@@ -123,6 +124,36 @@ def _open_in_editor(path: str) -> None:
         print(f"Failed to open {path}: {e}")
 
 
+def _download_ls_dependencies(ls_id: LanguageServerId, ls_specific_settings: dict, repository_root_path: str) -> None:
+    """
+    Downloads the runtime dependencies of the given language server.
+
+    :param ls_id: the language server whose dependencies to download
+    :param ls_specific_settings: the user-provided language server-specific settings, which can affect which
+        dependencies are required (or whether Serena manages them at all)
+    :param repository_root_path: path of the directory to use as the language server's repository root;
+        since the language server is instantiated but never started, the directory's contents are irrelevant.
+        It also receives the (empty) project-specific data, which is thus discarded along with the directory.
+    :raises Exception: if the dependencies could not be downloaded
+    """
+    from solidlsp import SolidLanguageServer
+    from solidlsp.ls_config import LanguageServerConfig
+    from solidlsp.settings import SolidLSPSettings
+
+    # instantiating the language server already downloads most dependencies (because the launch command
+    # must be constructed), and install_dependencies covers the remaining ones
+    ls = SolidLanguageServer.create(
+        LanguageServerConfig(ls_id=ls_id),
+        repository_root_path,
+        solidlsp_settings=SolidLSPSettings(
+            solidlsp_dir=SerenaPaths().serena_user_home_dir,
+            project_data_path=os.path.join(repository_root_path, ".solidlsp"),
+            ls_specific_settings=ls_specific_settings,
+        ),
+    )
+    ls.install_dependencies()
+
+
 class ProjectType(click.ParamType):
     """ParamType allowing either a project name or a path to a project directory."""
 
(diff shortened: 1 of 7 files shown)


src/solidlsp/ls.py
```diff
@@ -1688,20 +1687,107 @@ def start(self) -> "SolidLanguageServer":
...
 
     @property
     def language_server(self) -> Self:
         return self
 
     def is_running(self) -> bool:
         return self.server.is_running()
+
+    @classmethod
+    def prepare_runtime_dependencies(
```
MischaPanch: Hey. Thanks for tackling this. The implementation is not how we do things in this project - we rely on clear, typed interfaces. gettattr should be used only in absolute exceptional situations, inspect essentially never, fallbacks also essentially never.

Instead, pls make this part of the interface. Introduce an abstract classmethod `def download_dependencies(cls, solidlsp_settings: SolidLSPSettings | None)` and implement it in all language servers. You might have to refactor a bit along the way to reduce code duplications. It's a bit of manual labor, the logic for downloading is present in each implementation, but it doesn't always follow a unified interface.

src/serena/cli.py
```diff
@@ -413,26 +413,109 @@ def __init__(self) -> None:
...
+        total = success = failed = skipped = 0
+        details: list[tuple[str, str, str]] = []
+
+        try:
+            pkg = importlib.import_module("solidlsp.language_servers")
+        except Exception as e:
+            click.echo(f"Error importing package solidlsp.language_servers: {e}")
+            return
+
+        for finder, mod_name, ispkg in pkgutil.iter_modules(pkg.__path__):
```
MischaPanch: don't use inspect, pkgutil and so on. These are essentially always codesmells that quickly lead to disaster.

We have a `Language` enum, you can just use `Language[l].get_ls_class`. If we didn't already have this, the correct way to address it would be to introduce something like that

MischaPanch (changes requested):

MischaPanch:
@giovanimds do you plan on working towards this? Otherwise I'd close the PR, since the urrent implementation is far off from hitting the target and likely won't help other people who might want to address this issue

giovanimds:
Yes, I do! I have a few commits on my local branch already. Sorry for the delay, I'll update the PR soon.

opcode81:
@giovanimds @MischaPanch this is actually overly complicated.
For all language servers that auto-install themselves, dependencies will be downloaded by simply instantiating the language server; I think I have previously stated this elsewhere.

Note: You can instantiate a server with any directory; it shouldn't matter. Instantiation will not actually start the server, so the contents of the directory will not even be looked at.

opcode81:
@giovanimds this is still marked as a draft. Is it abandoned?

MischaPanch:
I'm going to go ahead and close this. The implementation is still fairly off, and in two years of Serena's existence nobody else asked for this feature

PR #719, opened 2025-10-31
Support updating tools from dashboard

MischaPanch:
Using a dedicated tool _update_tool_list which is added to the server at the very end. Unfortunately, there seems to be no better way to do that, as the session object is not exposed in the mcp sdk in other places than the tool's context

**Doesn't work in Claude Code or Claude Desktop**

Anthropic's products don't follow Anthropic's own specifications...
https://github.com/anthropics/claude-code/issues/4118

We probably shouldn't merge until Anthropic resolves it

### DIFF
--- a/.vscode/mcp.json
+++ b/.vscode/mcp.json
@@ -1,0 +1,19 @@
+{
+	"servers": {
+		"serena": {
+			"type": "stdio",
+			"command": "uv",
+			"args": [
+				"run",
+				"--directory",
+				"${workspaceFolder}",
+				"serena",
+				"start-mcp-server",
+				"--context",
+				"ide-assistant",
+				"--project",
+				"${workspaceFolder}"
+			]
+		}
+	}
+}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,6 +35,7 @@ dependencies = [
   "tqdm>=4.67.1",
   "tiktoken>=0.9.0",
   "anthropic>=0.54.0",
+  "janus>=2.0.0",
 ]
 
 [[tool.uv.index]]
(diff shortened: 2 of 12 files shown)


opcode81:
> We probably shouldn't merge until Anthropic resolves it

Was this successfully tested with any client? Because if it can be shown that the implementation works correctly, we could merge it but disable it in the frontend, so it cannot actually be used. (Otherwise, we might get too many conflicts, especially if we refactor the Dashboard.)

MischaPanch:
Now tested with VSCode.

**Limitations**: currently only works in stdio mode (see below). They can be overcome, but that requires even more hackery, so let's do that later

My previous approach (an `_update_tool_list` tool which would be called by `SerenaAgent`) didn't work because you just can't call tools that use context from outside the mcp server. The context is extracted dynamically on each call from the client. The context contains the `Session` object which one can use to send the update tool list notification. This `Session` object is also created dynamically and never persisted. So we can't call the tool, and we also can't access the session....

So after a lot of crying and deliberation about life choices, I now have solved it by extending `FastMCP` in the following way:

1. We override `run_stdio_async` which persists the read and write stream (meaning the current solution will only work for stdio servers). This is because the `Session` instance is created much deeper in the code and it would be harder to hook into that part.
2. Copied the necessary methods from `Session` to send notifications

Since fastmcp is async first, we can't call these methods from outside (not thread safe) without the following trickery:

3. Override the main entrypoint `run` to also set up a sync-async communication queue and trigger the processing of it in parallel to the normal server startup. Luckily, python async annoyed enough people for the [janus](https://github.com/aio-libs/janus) project to exist, which serves for exactly that purpose.
4. Finally add a sync method `update_tool_list` which places the notification call into the queue

As part of this PR, I also added an mcp configuration that uses serena on serena in VSCode

Btw, serena is on the VSC marketplace:
https://github.com/mcp/oraios/serena

Not sure what exactly the install button there will do...

opcode81:
Serena v2 will fully resolve the issue of updating tools dynamically, so closing this.

PR #723, opened 2025-11-04
Fix memory name parameter inconsistency 

tylersatre:
I ran into an issue where Claude Code called the read memory tool incorrectly, the snippet from it is:
```
⏺ serena - read_memory (MCP)(memory_name: "project_overview")
  ⎿  Error: Error executing tool read_memory: 1 validation error for applyArguments
     memory_file_name
       Field required [type=missing, input_value={'memory_name': 'project_overview'}, input_type=dict]
         For further information visit https://errors.pydantic.dev/2.11/v/missing
```

It seems likely that the cause was one instance of inconsistent parameter naming for memory tools.  The `WriteMemoryTool` used `memory_name` while `ReadMemoryTool` and `DeleteMemoryTool` both use `memory_file_name`.

I think it makes sense for `WriteMemoryTool` to also use `memory_file_name` so there's not different parameter names for what are essentially the same thing.

### DIFF
--- a/src/serena/tools/memory_tools.py
+++ b/src/serena/tools/memory_tools.py
@@ -8,7 +8,7 @@ class WriteMemoryTool(Tool):
     Writes a named memory (for future reference) to Serena's project-specific memory store.
     """
 
-    def apply(self, memory_name: str, content: str, max_answer_chars: int = -1) -> str:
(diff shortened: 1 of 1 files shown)


MischaPanch:
Thanks! Merging

Issue #688, opened 2025-10-20
Scala language server: failed to find references of a datamember using `find_referencing_symbols` operation


eranmizrahi1:
Just tried Serena with Scala (💪 ).

i am working with `intellij` and not `VS Code`, so i configured it manully based on the [docs](https://github.com/oraios/serena/blob/main/docs/scala_setup_guide_for_serena.md#manual-setup-no-vs-code):
- add the plugin manually (```addSbtPlugin("ch.epfl.scala" % "sbt-bloop" % "2.0.16")```)
- run the following commands:
```
sbt -Dbloop.export-jar-classifiers=sources bloopInstall
sbt clean compile
```

- activated the project

- then tried it with the following prompt:
`using serena mcp, show me all references of "field1" data member of class "Config" under "SomeApp" object`

for the following example code:
```
object SomeApp {
    case class Config(field1:String)
}
```

But the mcp failed to find references to the symbol, but only with `search_for_pattern` operation.
it did tried to use  `find_symbol` & `find_referencing_symbols` first, did find the case class, but coudlnt find the references for the field.

I have:

- [ x ] Read the readme and verified that the issue cannot be solved by adjusting configuration
- [ x ] Understood that Serena's dashboard can be disabled through the config
- [ x ] Understood that by default a client session will start a separate instance of a Serena server. 
- [ x ] Understood that for multi-agent setups, the SSE mode should be used.
- [ x ] Verified that non-project files are ignored using either gitignore or the corresponding setting in `.serena/project.yml`
- [ x ] Have looked for similar issues and discussions, including closed ones
- [ x ] Made sure it's an actual issue and not a question - those should be opened as discussion instead.

If you have encountered a real and new issue:

- [ x ] I performed `<uv invocation> serena project health-check`
- [ x ] I indexed the project as described in the readme
- [ x ] Added sufficient explanation of my setup: the MCP client, the OS, the programming language, any config adjustments or relevant project specifics
- [ x ] Explained how the issue arose, added instructions on how to reproduce it (if possible)
- [ x ] If the issue happens on an open source project, I have added the link
- [ x ] Wrote a meaningful title and description

MischaPanch:
@YuMuuu would you have time to look into this?

MischaPanch:
[...]

PR #733, opened 2025-11-08
Use selection range in scala

YuMuuu:
I resolved an issue where find_referencing_symbol did not work properly in Serena’s Scala support. The images below show it in action.

resolve: https://github.com/oraios/serena/issues/688#issuecomment-3481783775 (It started working once I changed it to return serectionrange.)

image

image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -276,7 +276,6 @@ markers = [
   "dart: language server running for Dart",
   "erlang: language server running for Erlang",
   "scala: language server running for Scala",
-  "scala: language server running for Scala",
   "al: language server running for AL (Microsoft Dynamics 365 Business Central)",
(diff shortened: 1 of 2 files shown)


MischaPanch:
Did it solve the problem? For go this had helped

YuMuuu:
> Did it solve the problem? For go this had helped

Some of the issues preventing `find_referencing_symbol` from working have been resolved. (The problem caused by the position mismatch between `range` and `selectionRange` has been fixed.)

However, the issue where the fields of a case class cannot be retrieved as symbols (issue 688) is a problem with Metals, and therefore will not be resolved in this pull request.
I have separately filed an issue with Metals to investigate this further.

https://github.com/scalameta/metals/discussions/7951

PR #736, opened 2025-11-09
Add support for overloads, extending name paths with overload indices

### LINKED ISSUE #515: Make name paths support overloaded methods
Languages like Java make heavy use of overloaded methods, but Serena does not support this, requiring unique paths.
We could fix this by allowing an additional index to be specified, i.e. `MyClass/myMethod[idx]`, where the index follows the order of the method declarations in the class.

### LINKED ISSUE #578: find_referencing_symbols Incorrectly Handles Method Symbols, Leading to False "No symbol found" Warnings
I have:

- [x] Read the readme and verified that the issue cannot be solved by adjusting configuration
- [x] Understood that Serena's dashboard can be disabled through the config
- [x] Understood that by default a client session will start a separate instance of a Serena server. 
- [x] Understood that for multi-agent setups, the SSE mode should be used.
- [x] Verified that non-project files are ignored using either gitignore or the corresponding setting in `.serena/project.yml`
- [x] Have looked for similar issues and discussions, including closed ones
- [x] Made sure it's an actual issue and not a question - those should be opened as discussion instead.

If you have encountered a real and new issue:

- [x] I performed `<uv invocation> serena project health-check`
- [x] I indexed the project as described in the readme
- [x] Added sufficient explanation of my setup: the MCP client, the OS, the programming language, any config adjustments or relevant project specifics
- [x] Explained how the issue arose, added instructions on how to reproduce it (if possible)
- [x] If the issue happens on an open source project, I have added the link
- [x] Wrote a meaningful title and description

### LINKED ISSUE #735: Java LS (EclipseJDTLS) does not work if paths contain spaces

opcode81:
This adds support for overloaded symbols by
* expanding the symbol representation to include an `overload_idx` (0-based), which is added to the end of the name path (e.g. `MyClass/myOverloadedMethod[1]`)
* adjusting the language server symbol retrieval (`SolidLanguageServer.request_document_symbols`) accordingly
* adjusting the name path matching logic to include the overload index (now reified as `NamePathMatcher` for efficiency)
* raising the LS symbol cache version (version now stored within cache to avoid creating obsolete files)
* modifying symbol names in EclipseJDTLS such that they are the pure name only (thus making the overload logic apply to it)

Fixes #735 
Resolves #515 
Fixes #578


src/solidlsp/ls.py
```diff
@@ -85,6 +85,7 @@ class SolidLanguageServer(ABC):
     """
 
     CACHE_FOLDER_NAME = "cache"
+    CACHE_VERSION = 2
```
MischaPanch: This will mark previous caches as outdated, right? It will be a nuisance for users where indexing took long. We could do it in a backwards compatible way by keeping the cache versioned with dates, inferring the previous date from the filename, and bumping the dates only for languages where it's needed (only Java), thus introducing the (necessary) concept of a per-language cache version.

src/solidlsp/ls.py
```diff
@@ -830,22 +831,39 @@ def request_completions(
...
+        Sends a [documentSymbol](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_documentSymbol)
+        request to the language server to find symbols in the given file, which is assumed to be opened while the call is active.
+        Waits for the response and returns it.
+
+        :param relative_file_path: the relative path of the file that has the symbols.
+            IMPORTANT: It must be ensured that the file is already opened in the Language Server before calling this method.
+        :param include_body: whether to include the body of the symbols in the result.
+        :return: the list of symbols in the file.
+        """
+        self.logger.log(f"Requesting document symbols for {relative_file_path} from the Language Server", logging.DEBUG)
```
MischaPanch: maybe checking that it's open instead of just documenting?

src/solidlsp/ls.py
opcode81: This is not feasible at all. We need a single data structure, not 10 different data structures for different languages.
Versioning the cache using dates makes no difference. When you changed the date several times, you also invalidated all previous caches (because you only ever looked for the newest date).
In general, the only way not to discard old caches is to do a migration for cases where that's possible.

And I don't think Java is the only affected language. There are probably many more languages where the overloads mechanism already applies (without any changes to the LS, which were only necessary for Java, because they tried to resolve the overload issue in their own way). Why do you think Java is the only one?

src/solidlsp/ls.py
opcode81: This is an internal method. Documentation is sufficient.
All usage sites do it correctly.

src/solidlsp/ls.py
MischaPanch: The date was just as a measure for backwards compatibility. I thought of java since it was the only affected one  since you have adjusted _request_document_symbols only there, but nvm that, since it is the processed result that is cached, and that is indeed affecting many languages.

Since we are invalidating now, let's do the caching right for future updates of the cache. The processing in python is probably always fast, it is the LS that can be slow. We could cache just the raw LS results and always process, this will make invalidation less frequent.

I still think a per-language versioning makes sense. We could have a default version, and a language needing special care could just add something to the version, which can then be bumped independently.

MischaPanch: To clarify - since the previous version was a date, the most straightforward way to maintain compatibility for languages where nothing was changed was to read off that date from the filename and continue using it as the version (which is what I meant).

opcode81: As I suspected, my changes already work for C++, for example. The C++ LS does not try to make symbol names unique, so the new mechanism I added immediately makes a difference, allowing the overloads to be differentiated.

In this example

```cpp
int add(int a, int b) {
    return a + b;
}

int add(int a, int b, int c) {
    return a + b + c;
}
```

you will now get `add[0]` and `add[1]` as name paths, instead of `add` twice.

MischaPanch: Good, I agreed with invalidating now :). Let's just make sure we don't invalidate much in the future.

We have ruby users who have done a >30min indexing. Similar for PHP, I suspect. Generally, merging this will not be harmless - they will not know to reindex their project manually and instead will notice `find_symbol` hanging.

I am not sure about the best way to do this. We could do some kind of announcement in the readme. We could also display the cache version in the dashboard. In principle, this PR is a breaking change of Serena (in terms of user experience).

MischaPanch: TBH, maybe it's better to merge this together with the release of the jetbrains plugin, so users have the opportunity to jump onto that one instead of rebuilding their cache. It also fits naturally with the readme announcement

opcode81: This was resolved by merging #749 into this PR.

opcode81:
@MischaPanch I have now added a commit that saves the new cache and subsequently deletes the old cache after migration:
https://github.com/oraios/serena/pull/736/commits/70e64de890bd1998970e51b86438b5f53b571f3c

MischaPanch:
Perfect, thanks! From my side, this PR can be merged

PR #743, opened 2025-11-10
Intelephense: Add configurable setting `ignore_vendor`

### LINKED ISSUE #731: PHP: "vendor" folder should not be ignored by default
In intelephense.py, vendor directories are hardcoded to be ignored:

  def is_ignored_dirname(self, dirname: str) -> bool:
      # For PHP projects, we should ignore:
      # - vendor: third-party dependencies managed by Composer
      # - node_modules: if the project has JavaScript components
      # - cache: commonly used for caching
      return super().is_ignored_dirname(dirname) or dirname in ["node_modules", "vendor", "cache"]

  The Problem

  This creates a contradiction with Serena's core value proposition:

  Serena's instructions say: "Don't read entire files - use symbolic tools like find_symbol to explore code efficiently"

  But in practice: When working with third-party libraries, I'm forced to read entire vendor files because find_symbol doesn't work on them.

  Concrete example:
  - I'm using brick/date-time and need to understand how ZonedDateTime::parse() works
  - find_symbol with depth: 1 returns empty for vendor classes
  - I have to fall back to reading the entire ZonedDateTime.php file (700+ lines)
  - This defeats Serena's purpose of efficient code exploration

  Why This Matters for PHP Projects

  1. Third-party library exploration is crucial - Understanding external APIs is often MORE important than understanding your own code (which you wrote)
  2. Current workarounds don't work:
    - Setting ignore_all_files_in_gitignore: false in .serena/project.yml has no effect
    - The hardcoded exclusion overrides configuration
  3. The LLM gets stuck: Claude Code receives instructions to "use symbolic tools, don't read files" but then can't use those tools on vendor code, creating confusion

  Proposed Solution

  Make this configurable via ls_specific_settings in .serena/project.yml:

  ls_specific_settings:
    php:
      indexVendorDirectory: true  # default: false

  Or alternatively as a project-level setting:
  index_vendor_directories: true  # default: false

  Performance Tradeoff

  Yes, indexing would be slower with vendor included. However:

  - PHP developers are accustomed to slow indexing (PHPStorm: ~10 minutes)
  - This should be opt-in for teams that need it
  - The performance cost is worth it for proper third-party API exploration
  - Indexing happens once; exploration happens continuously

  Testing Notes

  I've locally patched line 39 to remove "vendor" from the exclusion list and can confirm:
  - Vendor indexing works correctly
  - find_symbol successfully returns methods from vendor classes
  - No issues with Intelephense handling vendor files

  Would be happy to test any proposed implementation!

opcode81:
Made setting configurable (default behaviour remains True)

Resolves #731 

Peripheral changes:
 * Refactor handling of LS-specific settings, ensuring that setting usage is logged

### DIFF
--- a/src/solidlsp/language_servers/intelephense.py
+++ b/src/solidlsp/language_servers/intelephense.py
@@ -28,15 +28,12 @@ class Intelephense(SolidLanguageServer):
     You can pass the following entries in ls_specific_settings["php"]:
         - maxMemory: sets intelephense.maxMemory
         - maxFileSize: sets intelephense.files.maxSize
+        - ignore_vendor: whether or ignore directories named "vendor" (default: true)
     """
 
     @override
     def is_ignored_dirname(self, dirname: str) -> bool:
-        # For PHP projects, we should ignore:
-        # - vendor: third-party dependencies <managed by Composer
-        # - node_modules: if the project has JavaScript components
-        # - cache: commonly used for caching
-        return super().is_ignored_dirname(dirname) or dirname in ["node_modules", "vendor", "cache"]
+        return super().is_ignored_dirname(dirname) or dirname in self._ignored_dirnames
 
     @classmethod
     def _setup_runtime_dependencies(cls, logger: LanguageServerLogger, solidlsp_settings: SolidLSPSettings) -> list[str]:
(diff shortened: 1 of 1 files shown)


opcode81:
@MischaPanch did you review the refactoring? Since this is not tested in CI (I think), please take a good look.

MischaPanch:
I did take a look, seems correct to me. I tested this manually with adjusting the settings and debugging - I suppose you did the same? (although now it would also show up in logs)

PR #749, opened 2025-11-11
Two-tier document symbol cache

opcode81:
Introduces a new two-tier cache
* raw document symbol cache (stores the raw LS response)
* (high-level) document symbol cache (stores instancs of newly introduced abstraction `DocumentSymbols`)

It was possible to convert the old cache into the raw document symbol cache, avoiding the problem of having to fully invalidate the old cache data.

Implied changes:
* Adapt all call sites of `request_document_symbols` to use to the new interface. For the tests, where we don't care about performance, I just wrote an adapter method and applied it everywhere.
* Remove `include_body` from the interface; all cached results store the body; we want the cache to be general-purpose (contain all relevant information)
* To make the raw symbol be "raw" indeed (save exactly what the LS returns), I abandoned the concept of mutating the returned dictionaries; instead, I am creating new augmented objects for the high-level cache. (Otherwise, the raw cache would essentially contain exactly the same objects as the full cache, of course.) 

Additional changes:
* Fix severe performance issue: Requesting document symbols with include_body=True previously re-read the entire source file for every single symbol within it

### DIFF
--- a/src/solidlsp/ls.py
+++ b/src/solidlsp/ls.py
@@ -1081,9 +1081,7 @@ class SolidLanguageServer(ABC):
 
             return document_symbols
 
-    def request_full_symbol_tree(
-        self, within_relative_path: str | None = None, include_body: bool = False
-    ) -> list[ls_types.UnifiedSymbolInformation]:
+    def request_full_symbol_tree(self, within_relative_path: str | None = None) -> list[ls_types.UnifiedSymbolInformation]:
         """
         Will go through all files in the project or within a relative path and build a tree of symbols.
         Note: this may be slow the first time it is called, especially if `within_relative_path` is not used to restrict the search.
@@ -1098,8 +1096,6 @@ class SolidLanguageServer(ABC):
         :param within_relative_path: pass a relative path to only consider symbols within this path.
             If a file is passed, only the symbols within this file will be considered.
             If a directory is passed, all files within this directory will be considered.
-        :param include_body: whether to include the body of the symbols in the result.
-
         :return: A list of root symbols representing the top-level packages/modules in the project.
         """
         if within_relative_path is not None:
--- a/src/serena/symbol.py
+++ b/src/serena/symbol.py
@@ -491,7 +491,6 @@ class LanguageServerSymbolRetriever:
     def find_by_name(
         self,
         name_path: str,
-        include_body: bool = False,
         include_kinds: Sequence[SymbolKind] | None = None,
         exclude_kinds: Sequence[SymbolKind] | None = None,
         substring_matching: bool = False,
@@ -504,7 +503,7 @@ class LanguageServerSymbolRetriever:
         """
         symbols: list[LanguageServerSymbol] = []
         for lang_server in self._ls_manager.iter_language_servers():
-            symbol_roots = lang_server.request_full_symbol_tree(within_relative_path=within_relative_path, include_body=include_body)
+            symbol_roots = lang_server.request_full_symbol_tree(within_relative_path=within_relative_path)
             for root in symbol_roots:
                 symbols.extend(
                     LanguageServerSymbol(root).find(
--- a/src/serena/tools/symbol_tools.py
+++ b/src/serena/tools/symbol_tools.py
@@ -150,7 +150,6 @@ class FindSymbolTool(Tool, ToolMarkerSymbolicRead):
         symbol_retriever = self.create_language_server_symbol_retriever()
         symbols = symbol_retriever.find_by_name(
             name_path,
-            include_body=include_body,
             include_kinds=parsed_include_kinds,
             exclude_kinds=parsed_exclude_kinds,
             substring_matching=substring_matching,
--- a/src/solidlsp/language_servers/al_language_server.py
+++ b/src/solidlsp/language_servers/al_language_server.py
@@ -680,7 +680,7 @@ class ALLanguageServer(SolidLanguageServer):
         return super().is_ignored_dirname(dirname) or dirname in al_ignore_dirs
 
     @override
-    def request_full_symbol_tree(self, within_relative_path: str | None = None, include_body: bool = False) -> list[dict]:
+    def request_full_symbol_tree(self, within_relative_path: str | None = None) -> list[dict]:
         """
         Override to handle AL's requirement of opening files before requesting symbols.
 


src/solidlsp/ls.py
```diff
@@ -77,6 +79,50 @@ class LSPFileBuffer:
...
+    def iter_symbols(self) -> Iterator[ls_types.UnifiedSymbolInformation]:
+        """
+        Iterate over all symbols in the document symbol tree.
+        Yields symbols in a depth-first manner.
+        """
+        if self._all_symbols is not None:
+            yield from self._all_symbols
+            return
+
+        def traverse(s: ls_types.UnifiedSymbolInformation):
```
MischaPanch, line 106: Is the traverse of the tree significantly slower than iterating through the already flattened list? If not, then I suggest removing the `_all_symbols`  field entirely. If yes, then this method, being the main entrypoint for new code, should populate `_all_symbols` by itself instead of delegating the population to `get_all_symbols_and_roots`

src/solidlsp/ls.py
```diff
@@ -85,7 +131,14 @@ class SolidLanguageServer(ABC):
...
 
     CACHE_FOLDER_NAME = "cache"
-    CACHE_VERSION = 2
+    RAW_DOCUMENT_SYMBOLS_CACHE_VERSION = 1
+    """
+    global version identifier for raw symbol caches; an LS-specific version is defined separately and combined with this
+    """
+    RAW_DOCUMENT_SYMBOL_CACHE_FILENAME = "raw_document_symbols.pkl"
+    RAW_DOCUMENT_SYMBOL_CACHE_FILENAME_LEGACY_FALLBACK = "document_symbols_cache_v23-06-25.pkl"
+    DOCUMENT_SYMBOL_CACHE_VERSION = 3
```
MischaPanch, line 143: why do we start with 3?

src/solidlsp/ls.py
```diff
@@ -162,6 +215,7 @@ def __init__(
         process_launch_info: ProcessLaunchInfo,
         language_id: str,
         solidlsp_settings: SolidLSPSettings,
+        cache_version_raw_document_symbols: int = 1,
```
MischaPanch, line 221: Maybe rename to `ls_specific_cache_version_raw_document_symbols`? Sounds too similar to `RAW_DOCUMENT_SYMBOLS_CACHE_VERSION ` now. Alternatively, rename the classvar to `GLOBAL_RAW_DOCUMENT_SYMBOLS_CACHE_VERSION`.

src/solidlsp/ls.py
```diff
@@ -85,7 +131,14 @@ class SolidLanguageServer(ABC):
     """
 
     CACHE_FOLDER_NAME = "cache"
-    CACHE_VERSION = 2
+    RAW_DOCUMENT_SYMBOLS_CACHE_VERSION = 1
+    """
+    global version identifier for raw symbol caches; an LS-specific version is defined separately and combined with this
```
MischaPanch: Maybe extend to explain that this is only bumped if some fundamental changes to what all LS cache are made? Otherwise the role of this version is difficult to understand, I think

src/solidlsp/ls.py
MischaPanch: Ah, lol, you actually named the field almost exactly this :D

src/solidlsp/ls.py
```diff
@@ -1234,9 +1343,10 @@ def retrieve_symbol_body(self, symbol: ls_types.UnifiedSymbolInformation | LSPTy
         symbol_start_line = symbol["location"]["range"]["start"]["line"]
         symbol_end_line = symbol["location"]["range"]["end"]["line"]
         assert "relativePath" in symbol["location"]
-        symbol_file = self.retrieve_full_file_content(symbol["location"]["relativePath"])
-        symbol_lines = symbol_file.split("\n")
-        symbol_body = "\n".join(symbol_lines[symbol_start_line : symbol_end_line + 1])
+        if file_lines is None:
```
MischaPanch, line 1345: nice! Thanks for addressing the old todo about the double read related to symbol bodies

src/solidlsp/ls.py
```diff
@@ -1566,56 +1676,110 @@ def request_defining_symbol(
...
-            self.cache_path.parent.mkdir(parents=True, exist_ok=True)
+        if not cache_file.exists():
+            # check for legacy cache to load to migrate
+            legacy_cache_file = self.cache_dir / self.RAW_DOCUMENT_SYMBOL_CACHE_FILENAME_LEGACY_FALLBACK
+            if legacy_cache_file.exists():
+                try:
+                    legacy_cache: dict[
+                        str, tuple[str, tuple[list[ls_types.UnifiedSymbolInformation], list[ls_types.UnifiedSymbolInformation]]]
+                    ] = load_pickle(legacy_cache_file)
+                    log.info("Migrating legacy document symbols cache with %d entries", len(legacy_cache))
```
MischaPanch, line 1710: Maybe on warning level? I also think we should delete the legacy cache file and create a new one, so I propose adding a separate method `_migrate_legacy_cache` that takes care of that, so loading looks like

```python
def _load_raw_document_symbols_cache(self):
    if not cache_file.exists() and legacy_file_cache.exists():
        self._migrate_legacy_cache()
    if cache_file_exists():
         ...
```

src/solidlsp/ls.py
opcode81: 1 was the original version, 2 was the new version on the other branch, 3 is the new version on this branch

src/solidlsp/ls.py
opcode81: I think it's common sense:
* If the LS output changes, you change the LS-specific version
* If the fundamental structure of the cache changes, you change this.

src/solidlsp/ls.py
opcode81: I think there is no significant speed difference, but logically, this should be slightly faster. Also, in most cases, the traversal happens only once, and you would never be able to amortise the costs.

I am keeping the additional interface around to avoid having to change all code. Especially in the tests, it wouldn't be worthwhile to change anything. Since the field is transient, it causes no problems; if you do use it, at least we don't keep reallocating it (and you might get a minor speedup).

So, fundamentally, if we want to keep the additional interface, then it needs to be supported properly - without any detrimental effects, and this is what my implementation achieves.

src/solidlsp/ls.py
opcode81: But I have now added this explanation to the docstring.

src/solidlsp/ls.py
opcode81: Why warning level? Warnings are reserved for something the user should probably pay attention to, because there might be a problem. This is just a normal event, so warning level would be inappropriate.

> we should delete the legacy cache

We can think about doing the deletion once we are done completely with this topic. Right now, keeping it around is better when switching between branches and testing the migrations.

> and create a new one

Note that the new cache is written as soon as it changes (but only if it changes).

src/solidlsp/ls.py
opcode81: Yeah, it seemed necessary to address it at this point.

src/solidlsp/ls.py
opcode81: Note that it would be rather illogical/redundant to use the naming also for this parameter, because when you are instantiating a subclass, everything is, of course, specific to that subclass.

MischaPanch: still, a bit confusing given the classvar IMO

src/solidlsp/ls.py
MischaPanch: Ok, then deletion at 1.0.0 release

src/solidlsp/ls.py
opcode81: To me, it is not confusing but very clear. The context is important.

If in PyrightLanguageServer, you write `super().__init__(cache_version_raw_document_symbols=2)`, then what could the version refer to other than a version relating to PyrightLanguageServer? Everything you pass to `super().__init__` relates *only* to the specifics of PyrightLanguageServer.

However, from the perspective of the super-class, there are two versions: one which relates to the format, and one which relates to the specific language server (instance/subclass). So, from that perspective, the more specific naming is required; from the other, it is clearly not.

opcode81: In a language like Java, this could be even more clear, because the class-level version could be private. This would make it completely clear that the class-level version has no relevance outside of the superclass whatsoever.

opcode81: In the same spirit, most people would, for example, write `setExternalDisplayName(String name)` rather than `setExternalDisplayName(String externalDisplayName)`, even though the internal variable that is set is `this.externalDisplayName`. The context makes more specific naming redundant.

src/solidlsp/ls.py
opcode81: In general, traversal will be dominated by the things done during the traversal, so worrying about the speed of this is not reasonable in general. However, if an optimisation is basically free (2 lines, as in this case), it is reasonable to do it anyway.

MischaPanch: I just found it weird that the attribute is not computed here but instead in the other method

opcode81: Do you still find it weird?
It is computed in the other method because that's the only case where it is really needed.
The other usage is just a minor optimisation.

MischaPanch: nvm, to me if there is an optimization, it should be computed at the first call of this method and not at the other method. But it's not important, no need to change. I'm resolving the thread

src/solidlsp/ls.py
MischaPanch: Ok

src/solidlsp/ls.py
opcode81: @MischaPanch it is only an optimisation *if* it is already computed. As I explained above, if it is not already computed, computing it first would defeat the optimisation; the effort would not be amortised.

src/solidlsp/ls.py
MischaPanch: @opcode81 I want to reopen this discussion and propose to make the "remove old cache migration" already.

If a user who has already indexed a project calls `serena project index` after merging this, they will create a new-style cache and an old-style cache which will never be read/migrated/deleted again. I don't think this is good behavior, let us just delete the old cache as part of the migration.

I tested it now, stepping through with the debugger, everything works (apart from the unrelated #765 )

opcode81: My proposal was to handle this when we are done with this topic, i.e. when we merge this into main.


But this statement doesn't make sense:

> they will create a new-style cache and an old-style cache

The old-style cache is not created by the new code.

MischaPanch: I mean it will create a situation where you will end up both with new-style cache (newly generated) and old-style cache (pre-existing, not deleted)

MischaPanch: So you want to handle it on the other branch?

opcode81: Yes. I am merging this branch.

Issue #591, opened 2025-09-02
Use the official Elixir Expert language server instead of the deprecated NextLS


humberaquino:
The [NextLS repo](https://github.com/elixir-tools/next-ls) is now deprecated and marked as archived since Aug 29th 2025 because the focus has shifted to working on the official language server for Elixir: [Expert](https://github.com/elixir-lang/expert).

MischaPanch:
A good one for the community, can be done in a similar manner as the migration from solargraph to ruby-lsp in #535.

MischaPanch:
@dbernazal FYI

dbernazal:
I can take care of this once Expert moves past the Alpha stage. Speaking with the maintainers, AFAIK, they're not quite ready for broad consumption.

MischaPanch:
Sounds good, thanks! Then let's continue this issue when the new LS is more stable

Dom-MN:
@dbernazal Does the current nextls version work? It fails to initialise in all of my project:

```
16:21:39.519 [error] Task #PID<0.218.0> started from #PID<0.215.0> terminating
** (KeyError) key :work_done_progress not found in: nil

If you are using the dot syntax, such as map.field, make sure the left-hand side of the dot is a map
    (next_ls 0.23.4) lib/next_ls/progress.ex:11: anonymous fn/3 in NextLS.Progress.start/3
    (elixir 1.17.3) lib/task/supervised.ex:101: Task.Supervised.invoke_mfa/2
Function: #Function<0.11992973/0 in NextLS.Progress.start/3>
    Args: []

16:23:41.411 [error] Runtime for folder test_repo failed to initialize

16:23:41.414 [debug] sent notification server -> client window/logMessage

16:23:41.414 [info] The runtime for test_repo has successfully shut down.

16:23:41.415 [error] ** (KeyError) key :show_message not found in: nil

If you are using the dot syntax, such as map.field, make sure the left-hand side of the dot is a map
    (next_ls 0.23.4) lib/next_ls.ex:1287: NextLS.handle_info/2
    (gen_lsp 0.11.0) lib/gen_lsp.ex:573: anonymous fn/2 in GenLSP.loop/3
    (telemetry 1.3.0) /Users/runner/work/next-ls/next-ls/deps/telemetry/src/telemetry.erl:324: :telemetry.span/3
    (gen_lsp 0.11.0) lib/gen_lsp.ex:572: anonymous fn/4 in GenLSP.loop/3
[...]

PR #752, opened 2025-11-12
fix: Adding window support for ElixirTools on init

dbernazal:
Fixes [#591](https://github.com/oraios/serena/issues/591)


MischaPanch:
Thanks! Merging this, but it doesn't fix #591 - that issue was about switching to a whole new language server

dbernazal:
Ack, yup you're right!

PR #763, opened 2025-11-17
Elixir, Rust, and YAML language server enhancements

alichtl:
- Fixes Elixir language server hang (should be using on_request instead of on_notification)
- Fixes Rust language server install-needed check (sometimes `rust-analyzer` is a symlink to `rustup`, which means it still needs to be installed)
- Adds support for YAML, including documentation and tests

### DIFF
--- a/test/solidlsp/yaml_ls/test_yaml_basic.py
+++ b/test/solidlsp/yaml_ls/test_yaml_basic.py
@@ -30,7 +30,7 @@ class TestYAMLLanguageServerBasics:
     @pytest.mark.parametrize("repo_path", [Language.YAML], indirect=True)
     def test_yaml_config_file_symbols(self, language_server: SolidLanguageServer, repo_path: Path) -> None:
         """Test document symbols detection in config.yaml with specific symbol verification."""
-        all_symbols, root_symbols = language_server.request_document_symbols("config.yaml", include_body=False)
+        all_symbols, root_symbols = language_server.request_document_symbols("config.yaml").get_all_symbols_and_roots()
 
         assert all_symbols is not None, "Should return symbols for config.yaml"
         assert len(all_symbols) > 0, f"Should find symbols in config.yaml, found {len(all_symbols)}"
@@ -64,7 +64,7 @@ class TestYAMLLanguageServerBasics:
     @pytest.mark.parametrize("repo_path", [Language.YAML], indirect=True)
     def test_yaml_services_file_symbols(self, language_server: SolidLanguageServer, repo_path: Path) -> None:
         """Test symbol detection in services.yml Docker Compose file."""
-        all_symbols, root_symbols = language_server.request_document_symbols("services.yml", include_body=False)
+        all_symbols, root_symbols = language_server.request_document_symbols("services.yml").get_all_symbols_and_roots()
 
         assert all_symbols is not None, "Should return symbols for services.yml"
         assert len(all_symbols) > 0, f"Should find symbols in services.yml, found {len(all_symbols)}"
@@ -92,7 +92,7 @@ class TestYAMLLanguageServerBasics:
     @pytest.mark.parametrize("repo_path", [Language.YAML], indirect=True)
     def test_yaml_data_file_symbols(self, language_server: SolidLanguageServer, repo_path: Path) -> None:
         """Test symbol detection in data.yaml file with array structures."""
-        all_symbols, root_symbols = language_server.request_document_symbols("data.yaml", include_body=False)
+        all_symbols, root_symbols = language_server.request_document_symbols("data.yaml").get_all_symbols_and_roots()
 
         assert all_symbols is not None, "Should return symbols for data.yaml"
         assert len(all_symbols) > 0, f"Should find symbols in data.yaml, found {len(all_symbols)}"
@@ -113,7 +113,7 @@ class TestYAMLLanguageServerBasics:
     @pytest.mark.parametrize("repo_path", [Language.YAML], indirect=True)
     def test_yaml_symbols_with_body(self, language_server: SolidLanguageServer, repo_path: Path) -> None:
         """Test request_document_symbols with body extraction."""
-        all_symbols, root_symbols = language_server.request_document_symbols("config.yaml", include_body=True)
+        all_symbols, root_symbols = language_server.request_document_symbols("config.yaml").get_all_symbols_and_roots()
 
         assert all_symbols is not None, "Should return symbols for config.yaml"
         assert len(all_symbols) > 0, "Should have symbols"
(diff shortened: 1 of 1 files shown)


src/solidlsp/language_servers/elixir_tools/elixir_tools.py
```diff
@@ -346,12 +356,6 @@ def work_done_progress(params):
         if self.server_ready.wait(timeout=ready_timeout):
             self.logger.log("Next LS is ready and available for requests", logging.INFO)
 
-            # Add a small settling period to ensure background indexing is complete
```
alichtl, line 349: This is no longer needed with the on_notification -> on_request fix above.

src/solidlsp/language_servers/yaml_language_server.py
```diff
@@ -0,0 +1,196 @@
+"""
```
MischaPanch, line 1: pls write tests ensuring that basic stuff works, like for all other LS that we support

src/solidlsp/ls_config.py
```diff
@@ -142,6 +143,8 @@ def get_source_fn_matcher(self) -> FilenameMatcher:
                 return FilenameMatcher("*.swift")
             case self.BASH:
                 return FilenameMatcher("*.sh", "*.bash")
+            case self.YAML:
```
MischaPanch, line 149: pls mark the yaml LS as experimental, similar to markdown. An entry in changelog and readme is also needed

src/serena/cli.py
```diff
@@ -534,7 +534,11 @@ def _index_project(project: str, log_level: str, timeout: float) -> None:
                 click.echo(f"Indexing for language {ls.language.value} ({k}/{len(servers)}) …")
                 collected_exceptions: list[Exception] = []
                 files_failed = []
-                for i, f in enumerate(tqdm(files, desc="Indexing")):
+
+                matcher = ls.language.get_source_fn_matcher()
```
MischaPanch: Pls undo, I fixed this behavior in #766

src/serena/config/serena_config.py
```diff
@@ -209,9 +209,12 @@ def autogenerate(
                         f"  project_name: {project_name}\n"
                         f"  language: python  # or typescript, java, csharp, rust, go, ruby, cpp, php, swift, elixir, terraform, bash\n"
                     )
-                # find the language with the highest percentage
-                dominant_language = max(language_composition.keys(), key=lambda lang: language_composition[lang])
-                languages_to_use = [dominant_language]
+                # Include all languages with at least 0.5% representation
+                # This ensures multi-language projects get full language server support
+                MIN_PERCENTAGE_THRESHOLD = 0.5
```
MischaPanch: @opcode81 and I were discussing how to handle this and wanted to address this later. One option is to have a threshold on the fraction of files compared to the dominant language.

The behavior you propose is too generous, many markdown, shell and other kinds of files will be included. This all will lead to serena requiring more resources.

Pls undo, the best way to autogenerate a config for a multi-language setup will be handled separately in the future. It is also unrelated to other changes here

alichtl:
Hi @MischaPanch. Thank you very much for reviewing this. I'll be sure to break work out into separate PRs in the future. I've made the requested changes here. Could you please let me know if any other changes are needed?

src/solidlsp/language_servers/yaml_language_server.py
alichtl: done

src/serena/cli.py
alichtl: done

src/serena/config/serena_config.py
alichtl: done

src/solidlsp/ls_config.py
alichtl: done.

test/solidlsp/yaml_ls/test_yaml_basic.py
```diff
@@ -0,0 +1,61 @@
...
+        assert len(all_symbols) > 0, f"Should find symbols in services.yml, found {len(all_symbols)}"
+
+    @pytest.mark.parametrize("language_server", [Language.YAML], indirect=True)
+    def test_yaml_data_file(self, language_server: SolidLanguageServer) -> None:
+        """Test symbol detection in data.yaml file."""
+        # Test with data.yaml
+        all_symbols, _root_symbols = language_server.request_document_symbols("data.yaml", include_body=False)
+
+        assert all_symbols is not None, "Should return symbols for data.yaml"
+        assert len(all_symbols) > 0, f"Should find symbols in data.yaml, found {len(all_symbols)}"
```
MischaPanch, line 98: pls have a look at the memory about adding language support and at other tests to see how tests should be implemented. I am not sure what cross-file referencing would mean for yaml (does it exist in yaml?), so you can skip that part. But tests that just check that a non-empty list was returned are not enough. Also, most tests just check the same logic, there is no point in duplicating that. 

Pls make one test that checks that the actual symbols were found, one that checks that the body was found as expected, and if references are a concept in the yaml language server, pls add a test that the referencing symbols were found. Checking the actual names and bodies, not just `len(result) > 0`.

src/solidlsp/ls.py
```diff
@@ -221,6 +241,7 @@ def logging_fn(source: str, target: str, msg: StringDict | str):
         self.server = SolidLanguageServerHandler(
             process_launch_info,
             language=self.language,
+            language_server=self,
```
MischaPanch: this is not a good pattern, now one needs to be extremely careful what happens when in the initialization process of the two classes. Instead, you can pass `determine_log_level=self._determine_log_level` and annotate the param as an optional callable.

src/solidlsp/ls.py
```diff
@@ -94,6 +94,26 @@ def is_ignored_dirname(self, dirname: str) -> bool:
         """
         return dirname.startswith(".")
 
+    def classify_stderr_line(self, line: str) -> int:
```
MischaPanch: pls rename to `_determine_log_level` and make static to avoid potential dependency on initialization order (see my comment in init)

test/solidlsp/yaml_ls/test_yaml_basic.py
alichtl: Hi @MischaPanch - good point. Testing is more thorough now.

src/solidlsp/ls.py
alichtl: agree- I extracted the default logging behavior into a new file (to avoid circular dependencies), and updated the init signature to optional callable as you suggested.

src/solidlsp/ls.py
alichtl: done

alichtl:
When the PR is approved, please let me know if I should squash before merging. Also, is there a public conversation thread on the multi-language support topic?

src/solidlsp/ls.py
alichtl: actually- because we have wired up a default determine_log_level function, we can remove a lot of the optional/fallback logic and keep the default definition as a static method in the base class. I just pushed another commit that cleans this up a bit more.

PR #766, opened 2025-11-17
Fix: gather per-language source files for project indexing

### LINKED ISSUE #765: Possible issue in indexing in multi-language projects
Indexing serena itself (from current main and two-tier-cache branch) leads to:

```
uv run serena project index
Indexing symbols in project C:\Users\misch\Projects\oraios\serena …
Indexing for language python (1/2) …
Indexing: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 191/191 [00:14<00:00, 13.28it/s]
Symbols saved to C:\Users\misch\Projects\oraios\serena\.serena\cache\python\document_symbols_cache_v23-06-25.pkl
Indexing for language typescript (2/2) …
Indexing: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 191/191 [00:14<00:00, 13.08it/s]
Symbols saved to C:\Users\misch\Projects\oraios\serena\.serena\cache\typescript\document_symbols_cache_v23-06-25.pkl
```

We don't have the same number of js/ts files as python files, so this is fishy. I haven't yet looked at the cause, this issue serves as reminder.

MischaPanch:
Closes #765

```
uv run serena project index
Indexing symbols in project C:\Users\misch\Projects\oraios\serena …
Indexing for language python (1/2) …
Indexing: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 188/188 [00:01<00:00, 148.38it/s]
Symbols saved to C:\Users\misch\Projects\oraios\serena\.serena\cache\python
Indexing for language typescript (2/2) …
Indexing: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4/4 [00:00<00:00, 235.20it/s]
Symbols saved to C:\Users\misch\Projects\oraios\serena\.serena\cache\typescript
```

### DIFF
--- a/src/serena/cli.py
+++ b/src/serena/cli.py
@@ -469,11 +469,16 @@ class ProjectCommands(AutoRegisteringGroup):
         try:
             log_file = os.path.join(project, ".serena", "logs", "indexing.txt")
 
-            files = proj.gather_source_files()
-
             servers = list(ls_mgr.iter_language_servers())
             for k, ls in enumerate(servers, start=1):
+                # TODO (if multiple calls to gather_source_files is ever a performance bottleneck):
+                #  we should eventually change the interface of gather_source_files
+                #  to return langauge -> list of files per language.
+                #  But it would involve a larger refactoring (unless we repeat the call to the matching method there)
+                #  since `is_ignored_path` doesn't return information due to which language
+                #  something was ignored (if due to a language at all)
                 click.echo(f"Indexing for language {ls.language.value} ({k}/{len(servers)}) …")
+                files = proj.gather_source_files(languages=[ls.language])
                 collected_exceptions: list[Exception] = []
                 files_failed = []
                 for i, f in enumerate(tqdm(files, desc="Indexing")):
--- a/src/serena/project.py
+++ b/src/serena/project.py
@@ -163,7 +163,9 @@ class Project(ToStringMixin):
         """
         return self._ignore_spec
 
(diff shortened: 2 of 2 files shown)


opcode81:
I don't think this is the most natural solution to #765. Wouldn't it be far more natural to have a single progress bar over all source files? This would make way more sense to me:
* Keep `gather_source_files` as it is.
* Use the `LanguageServerManager` to dynamically obtain the applicable LS per source file.

MischaPanch:
Sry, I had misundestood your commend first (deleted my previous one now). Sure, that's also possible but I wanted to make a small change. We currently have separate progress bars for the various languages, which has the benefit of clearly showing which languages are configured. You propose to reduce it to a single pbar, right? I would find that less natural from the user perspective. I quite like that we immediately see how many files were analyzed for each language.

opcode81:
The other variant is less code; see #769 (net deletion instead of addition)

Why would the user care about the languages being indexed? A single progress bar indicating when the entire process is complete is more intuitive - and also aligns with what IDEs do.

MischaPanch:
I meant a small change to the user experience. But sure, let's merge your variant, I am fine with it (personally, I like seing the languages as part of the index command output, but since we have them displayed in the dashboard it's not that important)

opcode81:
We could still log the languages that are active (before the process begins) if you think it is useful to the user.

MischaPanch:
Let's take your solution and after indexing display the per-language counts. Best of both worlds :)

opcode81:
Sounds good! I have added a commit that does this.

PR #772, opened 2025-11-18
Generalise concept of a single-project SerenaAgentContext

opcode81:
Add `single_project` Boolean setting to every context


MischaPanch (approved):
Looks good, thanks!

PR #774, opened 2025-11-19
More fail-safe and flexible content replacement tool

opcode81:
* Modify `ReplaceRegexTool` to become `ReplaceContentTool`, which is more fail-safe and more flexible

   * It is more fail-safe because it requires no escaping in replacement strings
     (I have opted to use a different character sequence for backreferences,
     such that escaping backslashes is no longer required.)
   * It is more flexible, because it adds the "literal" mode, which is simpler
     to use for models that are not good at generating regexes correctly
     
* New functionality is also applied in `EditMemoryTool`.
* Significantly reduced the regex-specific prompts of the `editing` mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ Status of the `main` branch. Changes prior to the next official version change w
 
 * Tools:
   * Added `RenameSymbolTool` for renaming symbols across the codebase (if LS supports this operation).
+  * Replaced `ReplaceRegexTool` with `ReplaceContentTool`, which supports both plain text and regex-based replacements
+    (and which requires no escaping in the replacement text, making it more robust) 
 
 * Language support:
 


src/serena/tools/file_tools.py
```diff
@@ -157,72 +158,91 @@ def is_ignored_file(abs_path: str) -> bool:
         return result
 
 
-class ReplaceRegexTool(Tool, ToolMarkerCanEdit):
+class ReplaceContentTool(Tool, ToolMarkerCanEdit):
```
MischaPanch, line 161: since replace_regex is something that people have been putting in custom excluded tools, we should do some sort of backwards compatibility adjustment. Either

1. Exclude replace_content when replace_regex is excluded
2. Crash if replace_regex appears and tell users to either remove it or to migrate to replace_content

I think I prefer 2, we can in fact always crash if a non-recognized tool is excluded - or maybe we already do that? If we already do that, either a custom message, or a general message about "tool names may have changed, have a look at the changelog" would be in order.

My current favorite solution would be to make sure we always crash and to add a general message linking to changelog (easiest to maintain going forward). And to add this renaming to the changelog, maybe under a dedicated new section

opcode81: We already had (2) - minus more specific instructions, but I think (1) is a worthwhile addition, so I have added this with the latest commit.

MischaPanch: Looks good! Don't forget to extend the changelog, this is worth mentioning (many have complained about this tool in the past)

PR #776, opened 2025-11-19
Allow the user to interactively enable additional languages

### LINKED ISSUE #768: Autogenerate config with multiple languages
See https://github.com/oraios/serena/pull/763#discussion_r2533446368

opcode81:
... when creating a project via CLI

Resolves #768

### DIFF
--- a/src/serena/config/serena_config.py
+++ b/src/serena/config/serena_config.py
@@ -141,15 +141,15 @@ class ProjectConfig(ToolInclusionDefinition, ToStringMixin):
                 # determine languages automatically
                 language_composition = determine_programming_language_composition(str(project_root))
                 if len(language_composition) == 0:
+                    language_values = ", ".join([lang.value for lang in Language])
                     raise ValueError(
                         f"No source files found in {project_root}\n\n"
-                        f"To use Serena with this project, you need to either:\n"
-                        f"1. Add source files in one of the supported languages (Python, JavaScript/TypeScript, Java, C#, Rust, Go, Ruby, C++, PHP, Swift, Elixir, Terraform, Bash)\n"
-                        f"2. Create a project configuration file manually at:\n"
-                        f"   {os.path.join(project_root, cls.rel_path_to_project_yml())}\n\n"
-                        f"Example project.yml:\n"
-                        f"  project_name: {project_name}\n"
-                        f"  language: python  # or typescript, java, csharp, rust, go, ruby, cpp, php, swift, elixir, terraform, bash\n"
+                        f"To use Serena with this project, you need to either\n"
+                        f"  1. specify a programming language by adding parameters --language <language>\n"
+                        f"     when creating the project via the Serena CLI command OR\n"
+                        f"  2. add source files in one of the supported languages first.\n\n"
+                        f"Supported languages are: {language_values}\n"
+                        f"Read the documentation for more information."
                     )
                 # sort languages by number of files found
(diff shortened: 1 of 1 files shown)


src/serena/config/serena_config.py
```diff
@@ -143,18 +151,40 @@ def autogenerate(
...
-                dominant_language = max(language_composition.keys(), key=lambda lang: language_composition[lang])
-                languages_to_use = [dominant_language]
+                # sort languages by number of files found
+                languages_and_percentages = sorted(language_composition.items(), key=lambda item: item[1], reverse=True)
+                # find the language with the highest percentage and enable it
+                top_language_pair = languages_and_percentages[0]
+                other_language_pairs = languages_and_percentages[1:]
+                languages_to_use = [top_language_pair[0]]
+                # if in interactive mode, ask the user which other languages to enable
+                if len(other_language_pairs) > 0 and interactive:
```
MischaPanch, line 161: nitpick: I guess this was partially written by AI. Could we go for f-strings (I'd personally also prefer a single print of a triple-quoted string an) instead? Just optically nicer and more consistent (even if consistency of this type is not too important)

src/serena/config/serena_config.py
```diff
@@ -143,18 +151,40 @@ def autogenerate(
...
+                    print(
+                        "Detected and enabled main language '%s' (%.2f%% of source files)." % (top_language_pair[0], top_language_pair[1])
+                    )
+                    print("Additionally detected %s other languages." % (len(other_language_pairs)))
+                    print("\nNote: Enable only languages you need symbolic retrieval/editing capabilities for.")
+                    print("      Additional language servers use resources and some languages may require additional")
+                    print("      system-level installations/configuration (see Serena documentation).")
+                    print("\nWhich additional languages do you want to enable?")
+                    for lang, perc in other_language_pairs:
```
MischaPanch, line 170: it will be annoying to have to answer "no" many times if many languages are detected. I propose to make a while loop where the user can write a language shorthand to add a language or write "done" to exit the loop and proceed. We can add `get_shorthand` to the `Language` enum.

Note that the user may use this to create a project in an empty directory, where no language has been detected. If no language has been detected, we could show all languages as candidates.

src/serena/config/serena_config.py
opcode81: > I guess this was partially written by AI.

No.

> Could we go for f-strings

Why? (Note that you quoted a part that does not contain any string formatting.)
I don't see how using f-strings makes any difference whatsoever.
No nitpicking allowed in general. Only remarks that make an actual difference.

src/serena/config/serena_config.py
opcode81: > it will be annoying to have to answer "no" many times if many languages are detected.

Answering no is pressing return.
This many languages are extremely rare, so it doesn't matter to me. Serena is not a typical project at all.

> Note that the user may use this to create a project in an empty directory

When using the create command in an empty directory, the user gets the exception further up, which, however, is in need of revision. The user should be given an updated exception telling him to specify the `--language` parameter. This is much easier than using this mechanism, which will usually generate just a few questions.

MischaPanch: Ok, then let's leave this as is. I'll take care of the exception on the empty project later on

opcode81: I have updated the exception text.

PR #777, opened 2025-11-19
Resolve eclipse jdtls _request_document_symbols override problem

mrhota:
Claude Code cannot run the `find_symbol` (or other symbol tools, for that matter) because of a recently-added method having an inconsistent override in the `EclipseJDTLS` lang server definition.

```
serena - find_symbol (MCP)(name_path_pattern: "Enricher/enrich", include_body: true)
  ⎿  {                                                                                                                                     
       "result": "Error executing tool: TypeError - EclipseJDTLS._request_document_symbols() takes 2 positional arguments but 3 were given"
     }

⏺ I'm encountering an error with the Serena MCP symbolic tools:

  Error: TypeError - EclipseJDTLS._request_document_symbols() takes 2 positional arguments but 3 were given

  This error occurs when I try to use:
  - mcp__serena__find_symbol
  - mcp__serena__get_symbols_overview

  It appears to be an internal error in the EclipseJDTLS integration within Serena. The symbolic tools are not working for this project.
  ```

Running `poe test -m "java"` reveals the failure. (By default, `poe test` will not run tests marked with `"java"`, which likely hid this problem.)

```
            else:
                log.debug("No cache hit for document symbols in %s", relative_file_path)
    
            # no cached result: request the root symbols from the language server
>           root_symbols = self._request_document_symbols(relative_file_path, file_data)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E           TypeError: EclipseJDTLS._request_document_symbols() takes 2 positional arguments but 3 were given

src/solidlsp/ls.py:1020: TypeError
```

### DIFF
--- a/src/solidlsp/language_servers/eclipse_jdtls.py
+++ b/src/solidlsp/language_servers/eclipse_jdtls.py
@@ -13,7 +13,7 @@ from pathlib import PurePath
 
 from overrides import override
 
-from solidlsp.ls import GenericDocumentSymbol, SolidLanguageServer
+from solidlsp.ls import GenericDocumentSymbol, LSPFileBuffer, SolidLanguageServer
 from solidlsp.ls_config import LanguageServerConfig
 from solidlsp.ls_logger import LanguageServerLogger
 from solidlsp.ls_utils import FileUtils, PlatformUtils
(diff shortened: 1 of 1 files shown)


mrhota:
@opcode81 this will resolve a problem with related to this change from yesterday: #736

opcode81:
Thanks for reporting! Now fixed.

opcode81:
Sorry, @mrhota, I didn't realize this was a PR (not an issue) until after I pushed the fix.

mrhota:
no problem; thank you for fixing this!

opcode81:
Also thanks for pointing out that the Java tests are currently disabled in CI; I was not aware of this and will enable them unless there is a very good reason not to.

PR #778, opened 2025-11-19
Enable Java tests

opcode81:
Reinstate testing for Java in CI.

### DIFF
--- a/test/solidlsp/java/test_java_basic.py
+++ b/test/solidlsp/java/test_java_basic.py
@@ -7,9 +7,11 @@ from solidlsp import SolidLanguageServer
 from solidlsp.ls_config import Language
 from solidlsp.ls_utils import SymbolUtils
 
-# These marks will be applied to all tests in this module
+# These marks will be applied to all tests in this module.
 # We skip the tests if not on Windows (CI hangs on Ubuntu for unknown reasons; testing on Windows is sufficient)
-pytestmark = [pytest.mark.java, pytest.mark.skipif(platform.system() != "Windows", reason="Java tests are only run on Windows")]
+is_ci = os.getenv("CI") == "true" or os.getenv("GITHUB_ACTIONS") == "true"
+java_tests_enabled = not is_ci or platform.system() == "Windows"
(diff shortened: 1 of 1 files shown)


opcode81:
@MischaPanch why were they disabled?

EDIT: Is it because it just hangs forever on Ubuntu? (I now enabled the tests only on Windows)

EDIT2: Not sure what's going on with the Ubuntu tests. Now it just failed at a random point, without completing the pytest call, i.e. no summary or anything.

opcode81:
The only test that now failed on Windows is `TestCSharpLanguageServer.test_find_referencing_symbols_across_files`, which I currently cannot get to work at all, no matter which commit I go to - including the commit that added the test.

Ubuntu: See above

macOS: Some Elixir tests fail.

Shoot me, please.

MischaPanch:
The csharp test is flaky, maybe now we need more time to wait for cross-file references. I am reluctant to bump the time because it worsens user experience (we currently count the time from the first call to request_references, not the time from LS initialization)

Java tests were sometimes hanging forever on Ubuntu or macOS for unknown reasons (couldn't reproduce it on my Ubuntu laptop), that's why we disabled them. They also took quite long on Windows, I think. It's all very flaky and annoying.

Nowadays also the AL tests started failing on Ubuntu, no idea why.

test/solidlsp/java/test_java_basic.py
```diff
@@ -1,13 +1,17 @@
...
 
 import pytest
 
 from solidlsp import SolidLanguageServer
 from solidlsp.ls_config import Language
 from solidlsp.ls_utils import SymbolUtils
 
+# These marks will be applied to all tests in this module
+# We skip the tests if not on Windows (CI hangs on Ubuntu for unknown reasons; testing on Windows is sufficient)
+pytestmark = [pytest.mark.java, pytest.mark.skipif(platform.system() != "Windows", reason="Java tests are only run on Windows")]
```
MischaPanch: It's better to do the exclusion only in CI, the tests do run through on "normal" linux environments. We already have a platform dependent if-else there

opcode81: Updated.

PR #784, opened 2025-11-26
Add Vue Language Support

tylersatre:
Adds full Vue.js language support to Serena using the official `@vue/language-server`.

## Key Changes

**New Vue Language Server** (`src/solidlsp/language_servers/vue_language_server.py`)
- Integrates `@vue/language-server` for Vue SFC (`.vue`) file support
- Manages a companion TypeScript server for cross-file references between Vue components and TypeScript modules
- Handles proper language ID detection (vue vs typescript) for files within Vue projects
- Supports all standard LSP operations: symbol retrieval, go-to-definition, find references, and rename

**Symbol Resolution Fix** (`src/serena/code_editor.py`)
- Improved symbol disambiguation logic to handle Vue patterns like `defineExpose({ pressCount })` where the same symbol appears as both definition and shorthand property reference
- Selects the definition (larger range) when multiple candidates exist

**Configuration & Integration**
- Added `VUE` language enum and file matchers for `.vue` and TypeScript/JavaScript variants
- Added `_get_language_id_for_file()` hook in `SolidLanguageServer` for language-specific file type handling

## Tests

Comprehensive test suite covering:
- Basic LSP operations (symbols, definitions, references)
- Error handling and edge cases
- Cross-file rename operations
- Symbol retrieval in Vue SFCs and TypeScript files

Test repository includes Vue components, composables, Pinia stores, and TypeScript type definitions to validate real-world Vue project patterns.

### DIFF
--- a/src/serena/config/serena_config.py
+++ b/src/serena/config/serena_config.py
@@ -174,15 +174,18 @@ class ProjectConfig(ToolInclusionDefinition, ToStringMixin):
                         f"Read the documentation for more information."
                     )
                 # sort languages by number of files found
-                languages_and_percentages = sorted(language_composition.items(), key=lambda item: item[1], reverse=True)
+                languages_and_percentages = sorted(
+                    language_composition.items(), key=lambda item: (item[1], item[0].get_priority()), reverse=True
+                )
                 # find the language with the highest percentage and enable it
                 top_language_pair = languages_and_percentages[0]
                 other_language_pairs = languages_and_percentages[1:]
-                languages_to_use = [top_language_pair[0]]
+                languages_to_use: list[str] = [top_language_pair[0].value]
                 # if in interactive mode, ask the user which other languages to enable
                 if len(other_language_pairs) > 0 and interactive:
                     print(
-                        "Detected and enabled main language '%s' (%.2f%% of source files)." % (top_language_pair[0], top_language_pair[1])
+                        "Detected and enabled main language '%s' (%.2f%% of source files)."
+                        % (top_language_pair[0].value, top_language_pair[1])
                     )
                     print(f"Additionally detected {len(other_language_pairs)} other language(s).\n")
                     print("Note: Enable only languages you need symbolic retrieval/editing capabilities for.")
@@ -190,9 +193,9 @@ class ProjectConfig(ToolInclusionDefinition, ToStringMixin):
                     print("      system-level installations/configuration (see Serena documentation).")
                     print("\nWhich additional languages do you want to enable?")
                     for lang, perc in other_language_pairs:
-                        enable = ask_yes_no("Enable %s (%.2f%% of source files)?" % (lang, perc), default=False)
+                        enable = ask_yes_no("Enable %s (%.2f%% of source files)?" % (lang.value, perc), default=False)
                         if enable:
-                            languages_to_use.append(lang)
+                            languages_to_use.append(lang.value)
                     print()
             else:
                 languages_to_use = [lang.value for lang in languages]
--- a/src/serena/util/inspection.py
+++ b/src/serena/util/inspection.py
@@ -19,13 +19,13 @@ def iter_subclasses(cls: type[T], recursive: bool = True) -> Generator[type[T], 
             yield from iter_subclasses(subclass, recursive)
 
 
-def determine_programming_language_composition(repo_path: str) -> dict[str, float]:
+def determine_programming_language_composition(repo_path: str) -> dict[Language, float]:
     """
     Determine the programming language composition of a repository.
 
     :param repo_path: Path to the repository to analyze
 
-    :return: Dictionary mapping language names to percentages of files matching each language
+    :return: Dictionary mapping languages to percentages of files matching each language
     """
     all_files = find_all_non_ignored_files(repo_path)
 
@@ -33,7 +33,7 @@ def determine_programming_language_composition(repo_path: str) -> dict[str, floa
         return {}
 
     # Count files for each language
-    language_counts: dict[str, int] = {}
+    language_counts: dict[Language, int] = {}
     total_files = len(all_files)
 
     for language in Language.iter_all(include_experimental=False):
@@ -47,12 +47,12 @@ def determine_programming_language_composition(repo_path: str) -> dict[str, floa
                 count += 1
 
         if count > 0:
-            language_counts[str(language)] = count
+            language_counts[language] = count
 
     # Convert counts to percentages
-    language_percentages: dict[str, float] = {}
-    for language_name, count in language_counts.items():
+    language_percentages: dict[Language, float] = {}
+    for language, count in language_counts.items():
         percentage = (count / total_files) * 100
-        language_percentages[language_name] = round(percentage, 2)
+        language_percentages[language] = round(percentage, 2)
 
     return language_percentages
--- a/src/solidlsp/ls_config.py
+++ b/src/solidlsp/ls_config.py
@@ -100,6 +100,23 @@ class Language(str, Enum):
     def __str__(self) -> str:
         return self.value
 
+    def get_priority(self) -> int:
+        """
+        :return: priority of the language for breaking ties between languages; higher is more important.
+        """
+        # experimental languages have the lowest priority
+        if self.is_experimental():
+            return 0
+        # We assign lower priority to languages that are supersets of others, such that
+        # the "larger" language is only chosen when it matches more strongly
+        match self:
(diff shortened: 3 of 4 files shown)


test/solidlsp/vue/test_vue_rename.py
```diff
@@ -0,0 +1,169 @@
...
+        workspace_edit = language_server.request_rename_symbol_edit(file_path, sel_start["line"], sel_start["character"], "processDigit")
+
+        assert workspace_edit is not None, "Should return WorkspaceEdit for rename operation"
+
+        has_changes = "changes" in workspace_edit and workspace_edit["changes"]
+        has_document_changes = "documentChanges" in workspace_edit and workspace_edit["documentChanges"]
+
+        assert has_changes or has_document_changes, "WorkspaceEdit should contain either 'changes' or 'documentChanges'"
+
+        if has_changes:
```
MischaPanch, line 33: should also handle the case of has_document_changes or remove the if - this whole block might never be called

test/solidlsp/vue/conftest.py
```diff
@@ -0,0 +1,89 @@
+import os
+import subprocess
+import time
+from pathlib import Path
+
+import pytest
+
+
+def ensure_vue_test_repo_dependencies(repo_path: str) -> None:
```
MischaPanch: this should not be necessary, the LS should download the dependencies on the fly when it starts the first time. Most LS implementations in Serena do that. Also, pls make sure to never use emojis or prints (use log.info instead)

src/solidlsp/language_servers/typescript_language_server.py
```diff
@@ -39,11 +40,34 @@ def getpwuid(uid: Any) -> Any:
...
+            return d
+    return definitions[0]
+
+
 class TypeScriptLanguageServer(SolidLanguageServer):
     """
     Provides TypeScript specific instantiation of the LanguageServer class. Contains various configurations and settings specific to TypeScript.
     """
 
+    # Version constants - subclasses can override these if needed
```
MischaPanch: this should be set through `ls_specific_settings`

src/solidlsp/language_servers/vue_language_server.py
```diff
@@ -0,0 +1,729 @@
...
+                new_item: dict = {}
+                new_item.update(item)  # type: ignore[arg-type]
+                new_item["absolutePath"] = str(abs_path)
+                new_item["relativePath"] = str(rel_path)
+                result.append(ls_types.Location(**new_item))  # type: ignore
+
+        return result
+
+    def request_file_references(self, relative_file_path: str) -> list:
+        from pathlib import Path
```
MischaPanch: pls prefer module-level imports

src/solidlsp/language_servers/vue_language_server.py
```diff
@@ -0,0 +1,729 @@
...
+    def request_references(self, relative_file_path: str, line: int, column: int) -> list[ls_types.Location]:
+        if not self.server_started:
+            log.error("request_references called before Language Server started")
+            raise SolidLSPException("Language Server not started")
+
+        if not self._has_waited_for_cross_file_references:
+            sleep(self._get_wait_time_for_cross_file_referencing())
+            self._has_waited_for_cross_file_references = True
+
+        if self._ts_server is not None:
```
MischaPanch: shouldn't self._ts_server always be available?

MischaPanch (changes requested):
Very nice PR, solid contribution, thank you!

I did a first shallow review, pls make sure tests pass in windows and dependencies are downloaded on the fly. When that's done I will review one more time in more detail.

One thing that would be very useful if you have the capacity to do that is to extract a more general pattern - a language server that uses another language server for cross-reference finding. This would also make it possible to implement svelte support without much hassle.

As it stands, we would get Vue support, but for svelte support the generalization would need to happen to keep the code clean, which would be a somewhat large refactoring. If you do have time, it would be great to address this already (you don't have to add the Svelte support yourself now, of course)

test/solidlsp/vue/conftest.py
tylersatre: Good call, I ended up removing conftest.py since it's not needed

src/solidlsp/language_servers/typescript_language_server.py
tylersatre: I switched this up and also updated the versions for the Vue language server in the same fashion

src/solidlsp/language_servers/vue_language_server.py
tylersatre: You're right

README.md
```diff
@@ -70,8 +70,35 @@ Language servers provide support for a wide range of programming languages.
 With Serena's LSP library, we provide **support for over 30 programming languages**, including
 AL, Bash, C#, C/C++, Clojure, Dart, Elixir, Elm, Erlang, Fortran, Go, Haskell, Java, Javascript, Julia, Kotlin, Lua, Markdown, Nix, Perl, PHP, Python, R, Ruby, Rust, Scala, Swift, TypeScript, YAML and Zig.
 
-> [!IMPORTANT]
-> Some languages require additional dependencies to be installed; see the [Language Support](https://oraios.github.io/serena/01-about/020_programming-languages.html) page for details.
```
opcode81, line 74: The conflict here was incorrectly resolved. This must remain; concrete language-related information was moved to docs/

src/serena/code_editor.py
```diff
@@ -270,12 +270,41 @@ def _find_unique_symbol(self, name_path: str, relative_file_path: str) -> Langua
...
+            """
+            range_data = symbol.symbol_root.get("range", {})
+            start_line = range_data.get("start", {}).get("line", 0)
+            end_line = range_data.get("end", {}).get("line", 0)
+            start_char = range_data.get("start", {}).get("character", 0)
+            end_char = range_data.get("end", {}).get("character", 0)
+            return (end_line - start_line, end_char - start_char)
+
+        # Sort by range size descending - larger ranges are typically definitions
+        sorted_candidates = sorted(symbol_candidates, key=get_range_size, reverse=True)
```
opcode81: While this may work for Vue.js, it is not an acceptable way of making a choice when the name path is non-unique in general.

For example, in Java, if there is an overloaded method `doSomething` but the overload is not uniquely specified in the name path, we cannot just use the overload that has the largest definition. We need to raise an error, forcing the caller to disambiguate.

Is there any basis to decide whether the logic you added should apply?

tylersatre: This comes down to me not checking AI work close enough when I was working through this.  You're completely correct that it's not something that should be handled in a language agnostic file like this, so I've moved the logic for this - https://github.com/tylersatre/serena/commit/0de00210031efe030f6f401eec4364de55ab2963

README.md
tylersatre: I've got this fixed up, I think it changed from the time I started this branch till now and I hadn't gone back through.

tylersatre:
> One thing that would be very useful if you have the capacity to do that is to extract a more general pattern - a language server that uses another language server for cross-reference finding. This would also make it possible to implement svelte support without much hassle.
> 
> As it stands, we would get Vue support, but for svelte support the generalization would need to happen to keep the code clean, which would be a somewhat large refactoring. If you do have time, it would be great to address this already (you don't have to add the Svelte support yourself now, of course)

@MischaPanch I did go ahead and create another branch that extracts this to a more general pattern.  If you check out https://github.com/tylersatre/serena/pull/2 you'll need the changes compared to this branch, to make it easier to look at just the companion server changes instead of adding Vue + companion server all together.

Let me know what you think and if you like it I can open the PR to merge it back to here instead of just showing the changes against my branch.

opcode81:
@tylersatre would you say we are ready for the merge now?

Once thing I thought was missing was an explicit tie-breaking strategy (which I added with the most recent commit). Previously the preference for TypeScript/JavaScript over Vue when a project contains no .vue files was basically coincidental.

tylersatre:
> @tylersatre would you say we are ready for the merge now?

I think so!  Good catch and fix on the tie-breaking strategy.

opcode81:
Great! The macOS failure is an unrelated random failure.
Merging ...

Thanks a lot, @tylersatre.

opcode81:
@tylersatre I had to squash the changes for the merge.

I have recreated your `companion_server` branch based on the squased commit here:
https://github.com/oraios/serena/tree/companion_server
You can build on this branch for a future PR. I think it'll be worthwhile.

In the future, if you want us to merge your PRs without squashing, keep the history clean by using amend and interactive rebase, such that every commit represents a well-defined, meaningful addition.

PR #797, opened 2025-12-01
feat: Add TOML language support via Taplo

### LINKED ISSUE #796: serena project index fails on Rust projects with TOML in project.yml
## Description

Running `serena project index` on certain Rust projects fails with:

```
ValueError: 'toml' is not a valid Language
```

## Root Cause

When a Rust project's `.serena/project.yml` contains `languages: ["toml"]` (either manually configured or from a previous session), Serena fails to load the project configuration because `toml` is not a valid value in the `Language` enum.

This can happen when:
1. A user manually adds `toml` to their project languages
2. A project.yml was created/modified during a session where TOML support was being tested
3. The project configuration references TOML files (like Cargo.toml) that should be indexed

## Affected Projects

- Projects with `languages: ["toml"]` or `languages: ["rust", "toml"]` in project.yml
- The issue is inconsistent: some Rust projects work (no toml in config), others fail (toml in config)

## Expected Behavior

TOML should be a valid Language that can be specified in project.yml, allowing Serena to index TOML configuration files (Cargo.toml, pyproject.toml, etc.)

## Environment

- Serena version: main branch (pre-TOML support)
- Platform: Windows/Linux/macOS

w0wl0lxd:
## Summary
- Add TOML language server using Taplo with automatic binary download for .toml file validation, formatting, and schema support
- Implement security hardening: SHA256 checksum verification, zip slip protection, download timeout
- Enable `serena project index` to work on projects with TOML in their language configuration

## Fixes
Fixes #796

### TOML Language Support
- New `TaploServer` class in `src/solidlsp/language_servers/taplo_server.py`
- Automatic Taplo binary download for Windows, macOS, and Linux (x86_64, x86, aarch64, armv7)
- LSP integration for document symbols, hover, completion, and formatting
- TOML-specific directory ignore patterns (target, .cargo, node_modules)

### Security Features
- **SHA256 checksum verification**: Validates downloaded binaries against hardcoded hashes from official Taplo 0.10.0 releases
- **Zip slip protection**: Validates archive paths to prevent directory traversal attacks
- **Download timeout**: 120-second timeout prevents indefinite hangs on network issues

### Issue Resolution
When `project.yml` contains `languages: ["toml"]` but TOML isn't in the Language enum, `ProjectConfig.load()` raises `ValueError: 'toml' is not a valid Language`. This commonly occurs in Rust projects where users want to index Cargo.toml alongside .rs files. Adding TOML to the Language enum resolves this.

## Test Plan
- [x] Run `uv run poe test -m toml` to verify TOML language server tests pass
- [x] Test on a Rust project with `languages: ["toml"]` in project.yml
- [x] Verify Taplo auto-downloads correctly on first use
- [x] Confirm SHA256 verification works (modify hash to test rejection)

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ than existing solutions that charge a premium.
 
 Language servers provide support for a wide range of programming languages.
 With Serena's LSP library, we provide **support for over 30 programming languages**, including
-AL, Bash, C#, C/C++, Clojure, Dart, Elixir, Elm, Erlang, Fortran, Go, Haskell, Java, Javascript, Julia, Kotlin, Lua, Markdown, Nix, Perl, PHP, PowerShell, Python, R, Ruby, Rust, Scala, Swift, TypeScript, YAML and Zig.
+AL, Bash, C#, C/C++, Clojure, Dart, Elixir, Elm, Erlang, Fortran, Go, Haskell, Java, Javascript, Julia, Kotlin, Lua, Markdown, Nix, Perl, PHP, PowerShell, Python, R, Ruby, Rust, Scala, Swift, TOML, TypeScript, YAML, and Zig.
 
 > [!IMPORTANT]
 > Some languages require additional dependencies to be installed; see the [Language Support](https://oraios.github.io/serena/01-about/020_programming-languages.html) page for details.
--- a/src/solidlsp/language_servers/taplo_server.py
+++ b/src/solidlsp/language_servers/taplo_server.py
@@ -167,6 +167,7 @@ class TaploServer(SolidLanguageServer):
     @classmethod
     def _download_taplo(cls, install_dir: str, executable_path: str) -> None:
         """Download and extract Taplo binary with SHA256 verification."""
+        # TODO: consider using existing download utilities in SolidLSP instead of the custom logic here
         download_url, _ = _get_taplo_download_url()
         archive_filename = os.path.basename(download_url)
 
--- a/src/solidlsp/ls_config.py
+++ b/src/solidlsp/ls_config.py
@@ -99,6 +99,10 @@ class Language(str, Enum):
     def is_experimental(self) -> bool:
         """
         Check if the language server is experimental or deprecated.
+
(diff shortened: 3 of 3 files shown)


MischaPanch:
Thanks, merging this. I don't really understand the use case (is it really helpful to analyze cargo.toml with serena?), but it can't hurt to have toml support either. Note that I marked toml as experimental, meaning it won't be autodetected as project language just because some toml files are present. A user can specify toml by hand in the `languages` entry of `project.yml` to enable toml analysis in serena.

w0wl0lxd:
> Thanks, merging this. I don't really understand the use case (is it really helpful to analyze cargo.toml with serena?), but it can't hurt to have toml support either. Note that I marked toml as experimental, meaning it won't be autodetected as project language just because some toml files are present. A user can specify toml by hand in the `languages` entry of `project.yml` to enable toml analysis in serena.

cheers! non-autodetection was something i definitely thought about and meant to add to the PR (but woefully forgot) so i'm glad you fixed my ignorance there LOL

i had created the PR in the first place because i had added both Rust & Toml in the Serena yaml config file before even running Serena in the project in the first place, and noticed that the indexing was blocked due to adding Toml as one of the languages for the project. realistically in ~70% of use cases i'm doubtful this will impact anything, but there's a decently sized benefit i've found to it.

i've noticed a good amount of benefit from the indexing (not necessarily analysis, but since models tend to chain tool/MCP calls of similar origin, especially with Serena, in workflows with a .toml file it evidentially allows Serena to index Cargo.toml for symbol caching and easy contextual access to all the configured Crates/Deps/Versions, which i've seen referenced fairly often working with codebases including it. realistically the models could just cat/bat/Get-Content the file, but with Serena adding it to a memory/the config i've seen much better contextual reference to the Cargo.toml & Deny.toml so there's strict version/Crate adherence as well as Deny adherence.

MischaPanch:
Nice, glad you see a benefit from the toml addition :)

PR #798, opened 2025-12-01
feat: Add auto_index_on_activate config option

w0wl0lxd:
## Summary
- Adds new `auto_index_on_activate` config option to `ProjectConfig` (defaults to `false`)
- When enabled, automatically indexes all project source files after project activation
- Pre-populates symbol caches for faster first queries on large projects

## Changes
- Added `auto_index_on_activate: bool = False` field to `ProjectConfig` dataclass
- Implemented `_auto_index_project()` method in `SerenaAgent` that runs after LSP initialization
- Updated `project.template.yml` with documentation for the new option
- Added 3 unit tests for config parsing behavior

## Test plan
- [x] Run `uv run poe format` - passes
- [x] Run `uv run poe type-check` - passes  
- [x] Run `uv run poe test` - all 10 config tests pass
- [ ] Manual test: Enable option in a project.yml and verify indexing runs on activation

### DIFF
--- a/src/serena/agent.py
+++ b/src/serena/agent.py
@@ -520,12 +520,54 @@ class SerenaAgent:
             with LogTime("Language server initialization", logger=log):
                 self.reset_language_server_manager()
 
+            # run auto-indexing if enabled (after LS is ready)
+            if project.project_config.auto_index_on_activate:
+                self._auto_index_project(project)
+
         # initialize the language server in the background (if in language server mode)
         if self.is_using_language_server():
             self.issue_task(init_language_server_manager)
 
         if self._project_activation_callback is not None:
             self._project_activation_callback()
+
+    def _auto_index_project(self, project: Project) -> None:
+        """
+        Automatically indexes the project by requesting document symbols for all source files.
+        This pre-populates the language server caches for faster subsequent queries.
+
+        :param project: the project to index
+        """
+        log.info(f"Auto-indexing project {project.project_name}...")
+        ls_mgr = self.get_language_server_manager()
+        if ls_mgr is None:
+            log.warning("Cannot auto-index: language server manager not initialized")
+            return
+
+        files = project.gather_source_files()
+        indexed_count = 0
+        failed_count = 0
+
+        for i, filepath in enumerate(files):
+            try:
+                ls = ls_mgr.get_language_server(filepath)
+                ls.request_document_symbols(filepath)
+                indexed_count += 1
+            except Exception as e:
+                log.debug(f"Failed to index {filepath}: {e}")
+                failed_count += 1
+
+            # save caches periodically
+            if (i + 1) % 10 == 0:
+                ls_mgr.save_all_caches()
+
+        # final cache save
+        ls_mgr.save_all_caches()
+
+        if failed_count > 0:
+            log.info(f"Auto-indexed {indexed_count} files ({failed_count} failed) for {project.project_name}")
+        else:
+            log.info(f"Auto-indexed {indexed_count} files for {project.project_name}")
 
     def load_project_from_path_or_name(self, project_root_or_name: str, autogenerate: bool) -> Project | None:
         """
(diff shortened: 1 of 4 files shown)


opcode81:
I am not really convinced this is a useful feature, because it is likely to waste more time than it saves. Did you really find it beneficial to enable this?

w0wl0lxd:
> I am not really convinced this is a useful feature, because it is likely to waste more time than it saves. Did you really find it beneficial to enable this?

for large codebases and on startup for the day, yeah it's highly beneficial. normally im running `serena project index` at least 5-10 times a day when working on a large codebase, and due to the amount serena tends to freeze or need a restart (claude code) it saves a lot of effort of doing so

MischaPanch:
>  normally im running serena project index at least 5-10 times a day when working on a large codebase

I don't understand this, the indexing should be performed just once, the index is automatically and incrementally updated after. Why do you have to do this? Are you starting 5-10 new projects or switching computers every day?

opcode81:
I see how it could help for cases where you switch between git branches where a large number of files are different, for example.

src/serena/agent.py
```diff
@@ -523,13 +520,55 @@ def init_language_server_manager() -> None:
...
+                self._auto_index_project(project)
+
         # initialize the language server in the background (if in language server mode)
         if self.is_using_language_server():
             self.issue_task(init_language_server_manager)
 
         if self._project_activation_callback is not None:
             self._project_activation_callback()
 
+    def _auto_index_project(self, project: Project) -> None:
```
MischaPanch, line 534: this duplicates the indexing logic that we already have in other places. If we decide to implement this feature, the duplication should be resolved.

We also have to make sure that it doesn't block the MCP server from connecting. It may also be very annoying to users (or to the LLM that calls our tools) that tools are unresponsive before the project is indexed. Maybe it makes sense to run this one thing in a separate process

w0wl0lxd:
apologies for the late reply! i've been on a ☃️ snowy vacation (happy holidays 😄)

> > normally im running serena project index at least 5-10 times a day when working on a large codebase
> 
> I don't understand this, the indexing should be performed just once, the index is automatically and incrementally updated after. Why do you have to do this? Are you starting 5-10 new projects or switching computers every day?

not sure if this is intended/unintentional behavior, but with Rust/Rust-Analyzer, the symbols don't incrementally update properly at all. i noticed this ~3-4 months ago maybe, but the *"Enable Check on Save"* RA config setting, as well as some other odd RA behavior will cause Serena's symbol cache to either expire or need to be rebuilt _VERY_ often. Serena will freeze at times due to this when running the `get_symbols_overview` tool, and will often pull from an outdated symbol cache. 

> this duplicates the indexing logic that we already have in other places. If we decide to implement this feature, the duplication should be resolved.
> 
> We also have to make sure that it doesn't block the MCP server from connecting. It may also be very annoying to users (or to the LLM that calls our tools) that tools are unresponsive before the project is indexed. Maybe it makes sense to run this one thing in a separate process

i'm happy to rework this PR/create a new one to unblock this in my free time this week. in regards to your comment, in local testing i did notice that (at least in Rust codebases, which is a critical detail of this probably, i really only work with Rust, but i'll test this behavior with other languages) Serena will take a long time to initialize with this integrated in Rust codebases, especially large workspaces with many crates/artifacts, as Serena has to wait on Activation (where i placed the config setting call/command) for RA to check/build/compile artifacts before it's unblocked, mirroring the behavior that happens if you were to open two IDE's/terminal instances, and run `cargo check --workspace` or `cargo build` in one, and attempt to run a similar command in the other terminal (with RA LSP active/being used) freeze/become unresponsive VERY often with large amounts of file changes or build dependencies added, even with the "Enable Check on Save" config option disabled. 

i'll look into the behavior in other languages, but as far as Rust goes, it seems like there's not an easy way to have it be non-blocking and would definitely cause the annoyances you've mentioned here.

PR #816, opened 2025-12-07
feat: Set MCP ToolAnnotations.readOnlyHint from Tool.can_edit()

zetaloop:
Serena constructs MCP tools with `annotations=None`, so under the MCP spec `ToolAnnotations.readOnlyHint` defaults to `false` ("may modify the environment").

In my setup, Serena runs as an MCP server behind [`mcp-auth-proxy`](https://github.com/sigbit/mcp-auth-proxy) and is connected to ChatGPT Developer Mode via a custom Connector. In this configuration, ChatGPT treats every tool as potentially write-capable and shows an approval dialog for every call, including tools that only read data, for example reading files.

To address this, `SerenaMCPFactory.make_mcp_tool` now creates a `ToolAnnotations` object and passes it to `MCPTool`, setting `readOnlyHint = not tool.can_edit()`, and for read-only tools explicitly setting `destructiveHint = false`.

MCP clients can now distinguish read-only Serena tools from tools that can edit state, so read-only operations no longer trigger write-operation approval prompts while editable tools keep their existing behavior.

PixPin_2025-12-07_14-36-55

PixPin_2025-12-07_14-45-53

### DIFF
--- a/src/serena/mcp.py
+++ b/src/serena/mcp.py
@@ -215,8 +215,6 @@ class SerenaMCPFactory:
             return tool.apply_ex(log_call=True, catch_exceptions=True, **kwargs)
 
         annotations = ToolAnnotations(readOnlyHint=not tool.can_edit())
-        if annotations.readOnlyHint:
-            annotations.destructiveHint = False
 
         return MCPTool(
             fn=execute_fn,


src/serena/mcp.py
```diff
@@ -213,6 +214,10 @@ def make_mcp_tool(tool: Tool, openai_tool_compatible: bool = True) -> MCPTool:
         def execute_fn(**kwargs) -> str:  # type: ignore
             return tool.apply_ex(log_call=True, catch_exceptions=True, **kwargs)
 
+        annotations = ToolAnnotations(readOnlyHint=not tool.can_edit())
+        if annotations.readOnlyHint:
+            annotations.destructiveHint = False
```
opcode81: The docstring of destructiveHint suggests that this is not a meaningful use of the property:

```
    destructiveHint: bool | None = None
    """
    If true, the tool may perform destructive updates to its environment.
    If false, the tool performs only additive updates.
    (This property is meaningful only when `readOnlyHint == false`)
    Default: true
    """
```

So it is meant to differentiate destructive write operations from "non-destructive" ones.
From the terminology, it is not clear what exactly "additive" means, however.

We should probably drop the setting of `destructiveHint`, because it is misleading.

opcode81 (changes requested):
Thanks for the PR! See my remark.

src/serena/mcp.py
zetaloop: Thanks for pointing that out. Updated.

PR #819, opened 2025-12-07
Changed dashboard default listening address to 127.0.0.1

### LINKED ISSUE #748: Make dashboard listening address configurable (and change default)
The listening address should be configurable.

The default should be 127.0.0.1 instead of 0.0.0.0 (as allowing non-local connections by default is a security hazard).
History shows that it was changed to 0.0.0.0 only to make Docker work. Therefore, when changing the default, we need to make sure that running Serena via Docker keeps working.

MischaPanch:
Closes #748

### DIFF
--- a/src/serena/agent.py
+++ b/src/serena/agent.py
@@ -18,7 +18,7 @@ from interprompt.jinja_template import JinjaTemplate
 from serena import serena_version
 from serena.analytics import RegisteredTokenCountEstimator, ToolUsageStats
 from serena.config.context_mode import SerenaAgentContext, SerenaAgentMode
-from serena.config.serena_config import SerenaConfig, ToolInclusionDefinition, is_running_in_docker
+from serena.config.serena_config import SerenaConfig, ToolInclusionDefinition
 from serena.dashboard import SerenaDashboardAPI
 from serena.ls_manager import LanguageServerManager
 from serena.project import Project
@@ -272,12 +272,10 @@ class SerenaAgent:
             self._dashboard_thread, port = SerenaDashboardAPI(
                 get_memory_log_handler(), tool_names, agent=self, tool_usage_stats=self._tool_usage_stats
             ).run_in_thread(host=self.serena_config.web_dashboard_listen_address)
-            dashboard_listen_address = self.serena_config.web_dashboard_listen_address
-            if is_running_in_docker():
-                # we use localhost for URL instead of the 0.0.0.0 that is configured for docker
-                # (the user has to port-forward then)
-                dashboard_listen_address = "127.0.0.1"
-            dashboard_url = f"http://{dashboard_listen_address}:{port}/dashboard/index.html"
+            dashboard_host = self.serena_config.web_dashboard_listen_address
+            if dashboard_host == "0.0.0.0":
+                dashboard_host = "localhost"
+            dashboard_url = f"http://{dashboard_host}:{port}/dashboard/index.html"
             log.info("Serena web dashboard started at %s", dashboard_url)
             if self.serena_config.web_dashboard_open_on_launch:
                 # open the dashboard URL in the default web browser (using a separate process to control
--- a/src/serena/config/serena_config.py
+++ b/src/serena/config/serena_config.py
@@ -106,19 +106,6 @@ def get_serena_managed_in_project_dir(project_root: str | Path) -> str:
     return os.path.join(project_root, SERENA_MANAGED_DIR_NAME)
 
 
-def is_running_in_docker() -> bool:
-    """Check if we're running inside a Docker container."""
-    # Check for Docker-specific files
-    if os.path.exists("/.dockerenv"):
-        return True
-    # Check cgroup for docker references
-    try:
-        with open("/proc/self/cgroup") as f:
-            return "docker" in f.read()
-    except FileNotFoundError:
-        return False
-
-
 @dataclass(kw_only=True)
 class ProjectConfig(ToolInclusionDefinition, ToStringMixin):
     project_name: str
@@ -368,7 +355,6 @@ class SerenaConfig(ToolInclusionDefinition, ToStringMixin):
     web_dashboard: bool = True
     web_dashboard_open_on_launch: bool = True
     web_dashboard_listen_address: str = "127.0.0.1"
-    """When running in Docker, this will be automatically set to '0.0.0.0' ignoring any configuration by the user."""
     tool_timeout: float = DEFAULT_TOOL_TIMEOUT
     loaded_commented_yaml: CommentedMap | None = None
     config_file_path: str | None = None
@@ -398,11 +384,6 @@ class SerenaConfig(ToolInclusionDefinition, ToStringMixin):
     """Advanced configuration option allowing to configure language server implementation specific options, see SolidLSPSettings for more info."""
 
     CONFIG_FILE = "serena_config.yml"
-    CONFIG_FILE_DOCKER = "serena_config.docker.yml"  # Docker-specific config file; auto-generated if missing, mounted via docker-compose for user customization
-
-    def __post_init__(self) -> None:
-        if is_running_in_docker():
-            self.web_dashboard_listen_address = "0.0.0.0"
 
     def _tostring_includes(self) -> list[str]:
         return ["config_file_path"]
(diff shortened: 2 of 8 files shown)


src/serena/config/serena_config.py
```diff
@@ -478,23 +462,24 @@ def from_config_file(cls, generate_if_missing: bool = True) -> "SerenaConfig":
...
-        instance.trace_lsp_communication = loaded_commented_yaml.get("trace_lsp_communication", False)
-        instance.excluded_tools = loaded_commented_yaml.get("excluded_tools", [])
-        instance.included_optional_tools = loaded_commented_yaml.get("included_optional_tools", [])
-        instance.jetbrains = loaded_commented_yaml.get("jetbrains", False)
-        instance.token_count_estimator = loaded_commented_yaml.get(
-            "token_count_estimator", RegisteredTokenCountEstimator.TIKTOKEN_GPT4O.name
-        )
-        instance.default_max_tool_answer_chars = loaded_commented_yaml.get("default_max_tool_answer_chars", 150_000)
-        instance.ls_specific_settings = loaded_commented_yaml.get("ls_specific_settings", {})
+        instance.web_dashboard = get_value_or_default("web_dashboard")
```
MischaPanch, line 474: Just FYI @opcode81 - in the initial version of the change I was using get_value_or_default for the dashboard_host, but then pivoted to just not load it from the config. But the change is useful IMO, even if it doesn't relate to the issue at hand

src/serena/config/serena_config.py
```diff
@@ -342,6 +342,8 @@ class SerenaConfig(ToolInclusionDefinition, ToStringMixin):
     trace_lsp_communication: bool = False
     web_dashboard: bool = True
     web_dashboard_open_on_launch: bool = True
+    web_dashboard_host: str = "127.0.0.1"
```
opcode81: Calling this a "host" is misleading. We should prefer the more accurate term `listen_address`.

src/serena/config/serena_config.py
```diff
@@ -373,6 +375,10 @@ class SerenaConfig(ToolInclusionDefinition, ToStringMixin):
     CONFIG_FILE = "serena_config.yml"
     CONFIG_FILE_DOCKER = "serena_config.docker.yml"  # Docker-specific config file; auto-generated if missing, mounted via docker-compose for user customization
 
+    def __post_init__(self) -> None:
+        if is_running_in_docker():
+            self.web_dashboard_host = "0.0.0.0"
```
opcode81: This is hacky. Why don't we just configure the Docker image to use a suitable network configuration, i.e. just configure the correct setting rather than hackily override it?

Why did you decide not to load this from the configuration file? It definitely should be (as described in #748).

src/serena/agent.py
```diff
@@ -271,8 +271,12 @@ def get_memory_log_handler() -> MemoryLogHandler:
...
             self._dashboard_thread, port = SerenaDashboardAPI(
                 get_memory_log_handler(), tool_names, agent=self, tool_usage_stats=self._tool_usage_stats
-            ).run_in_thread()
-            dashboard_url = f"http://127.0.0.1:{port}/dashboard/index.html"
+            ).run_in_thread(host=self.serena_config.web_dashboard_host)
+            host_for_url = self.serena_config.web_dashboard_host
+            # this usually means we are in docker, so we use localhost for URL (the user has to port-forward then)
+            if host_for_url == "0.0.0.0":
+                host_for_url = "127.0.0.1"
+            dashboard_url = f"http://{self.serena_config.web_dashboard_host}:{port}/dashboard/index.html"
```
opcode81: Bug: `host_for_url` was ultimately not used.

src/serena/config/serena_config.py
MischaPanch: I did not want to mess with the docker image or the mechanism for configuring the agent in it. Currently we will autogenerate a `serena_config.docker.yml` from our template. So we would need a second template to just have a different value for the host, I don't see the point and it would introduce a lot of duplication. Or we need to document this and leave it up to the user to set it right, which many users will never notice.

Moreover, I don't see a scenario where a user would want to ever change this. Without docker it's localhost, on docker its 0.0.0.0 and the user forwards to localhost (if they so desire). If you don't like the setting it here, we could just hardcode the logic in the dashboard api and remove the host as input parameter there.

Do you see a reason why the user would ever need to configure this, so we have to extend the serena config?

MischaPanch: The issue is about making it listen on localhost instead of 0.0.0.0 and still work in docker, not about adding a configuration setting (which was presumed to be part of the solution but it doesn't have to be)

src/serena/config/serena_config.py
MischaPanch: See below, I propose hardcoding it in the dashboard api which means we remove this

src/serena/agent.py
MischaPanch: Will also be obsolete if we hardcode in the dashboard (but yeah, it's a mistake, nice catch)

src/serena/config/serena_config.py
opcode81: As the creator of the issue, I can tell you that it was not only about not using 0.0.0.0 by default and still making it work in docker. If that were the case, my wording would have made this clear. (But it is true that it is this aspect which makes this issue important to solve for v1.0, not the configurability.)

A user specifically asked for configurability in https://github.com/oraios/serena/discussions/746 - and rightly so. There are many scenarios in which a user might want to configure this, because it is a *fundamental setting* for any server. As soon as VMs, Docker images, etc. come into play, even a server application that is mostly used locally needs to allow this to be configured. And if a user is using the HTTP mode of the MCP server, there is no reason to assume that a remote usage on a network is out of the question. Therefore, configurability is the obvious thing to do (and what is done for literally all server applications).

If you don't want to do this properly now, we can merge your simpler solution now, but I don't see why we shouldn't do it properly as described in the issue.

Changing the configuration file after using the template as the base config is trivial, e.g. using sed:

```
sed -i 's/^web_dashboard_listen_address: .*/web_dashboard_listen_address: 0.0.0.0/' serena_config.yaml
```

MischaPanch: I'm not a fan of the current docker configuration method and would like to have a conversation on it before we start relying on it in new mechanisms.

So I propose either leaving it as is now or hardcoding if you find the post-init to hacky, and make the configurability a separate issue after we had a critical look on the docker support overall.

opcode81: I am fine with turning this into a consistent state and making it configurable later. 
But then we should also split the issue and not close #748 with this PR.

opcode81: > I'm not a fan of the current docker configuration method

Then again, making this configurable in Serena in general and the Docker configuration method are not really related issues. 

Allowing the listening address to be configurable in `serena_config.yml` is fundamentally a reasonable thing to do, and adjusting the Docker config to make use of it is not a problem.

MischaPanch: Let's pls discuss the docker improvements later, I am not sure I fully understand you (or whether you fully understand me). I will write a new issue that this PR will close

opcode81: My point is:
* This needs to be configurable (and configuring it via `serena_config.yml` is reasonable).
* Once it is configurable, it is trivial to apply the correct setting to the Docker case (e.g. by using `sed` as I proposed above).

We do not need to change anything fundamental about the "current docker configuration method" in order to do the above, though I do not know what exactly you mean by that.

MischaPanch: Compromise: configurable now outside of docker, hardcoded to 0.0.0.0 when in docker. Documented the behavior in the template, the post-init removes the need for using sed (which is at least equally hacky)

opcode81: There is absolutely nothing hacky about editing a configuration file (as that is what a configuration file is for - to be edited). For the Docker image, we can
* copy the template
* edit it (using sed) to have the settings we actually want.

opcode81: I will make a proposal in another branch.

opcode81:
This was merged with #830

MischaPanch:
@opcode81  830 was merged to this branch, reopening and merging this on main now

Issue #820, opened 2025-12-07
Scala language server on machines with a new version of coursier. Project creation fails


przemek-pokrywka:
I have:

- [x] Read the readme and verified that the issue cannot be solved by adjusting configuration
- [x] Understood that Serena's dashboard can be disabled through the config
- [x] Understood that by default a client session will start a separate instance of a Serena server. 
- [x] Understood that for multi-agent setups, the SSE mode should be used.
- [x] Verified that non-project files are ignored using either gitignore or the corresponding setting in `.serena/project.yml`
- [x] Have looked for similar issues and discussions, including closed ones
- [x] Made sure it's an actual issue and not a question - those should be opened as discussion instead.

If you have encountered a real and new issue:

- [NA] I performed `<uv invocation> serena project health-check`
- [NA] I indexed the project as described in the readme
- [x] Added sufficient explanation of my setup: the MCP client, the OS, the programming language, any config adjustments or relevant project specifics
- [x] Explained how the issue arose, added instructions on how to reproduce it (if possible)
- [x] If the issue happens on an open source project, I have added the link
- [x] Wrote a meaningful title and description

Hello, I'm using `coursier` installed in a new version, in which the "coursier" command is no longer present, "cs" replaced it.
Unfortunately, the line:
https://github.com/oraios/serena/blob/main/src/solidlsp/language_servers/scala_language_server.py#L64

asserts that the old version ("coursier" not "cs") is installed.
This prevents me from creating a `serena` project.

I've prepared a [fix](https://github.com/oraios/serena/pull/821/files) and I will appreciate if you could merge it 🙏

przemek-pokrywka:
I've fixed the problem in the following [PR](https://github.com/oraios/serena/pull/821)

przemek-pokrywka:
Apologies about the PR not passing tests. I don't have time to fix it, but I believe that any contributor will know how to fix it.

przemek-pokrywka:
Thanks for merging the fix, @opcode81 🙏 

🌟

PR #821, opened 2025-12-07
Scala language server: stop requiring old coursier

przemek-pokrywka:
You run coursier with "cs" command nowadays.
Users of the new coursier no longer have the old
"coursier" command around.
Yet, the assertion required its presence.
That caused the project creation command to break.

After this change:
- the assertion passes if "cs" or "coursier" is there

("coursier" was only used to install the new "cs" if it was missing)

### DIFF
--- a/src/solidlsp/language_servers/scala_language_server.py
+++ b/src/solidlsp/language_servers/scala_language_server.py
@@ -66,6 +66,7 @@ class ScalaLanguageServer(SolidLanguageServer):
 
         if not os.path.exists(metals_executable):
             if not cs_command_path:
+                assert coursier_command_path is not None
                 log.info("'cs' command not found. Trying to install it using 'coursier'.")
                 try:
                     log.info("Running 'coursier setup --yes' to install 'cs'...")


przemek-pokrywka:
This PR solves the issue https://github.com/oraios/serena/issues/820

I tested it locally.

opcode81:
As indicated by the type checker's result in CI, there is now a problem in line 72: It is not guaranteed that the executable being used is not None.

przemek-pokrywka:
@opcode81 thanks for your eyes on it. Unfortunately, I don't have time to fix it now, but I believe it is only a typechecker issue, as the code worked for me locally. I trust that anyone with time can fix it, but if not, I'll prepare a fix when I have more time.

opcode81:
Yes, the type checker is not smart enough to infer that it is actually correct. This can be "fixed" by adding an assertion which will enable it to understand this.

przemek-pokrywka:
🙌 Big kudos, @opcode81 🙏

PR #830, opened 2025-12-10
Improve docker configuration

opcode81:
Do away with Docker-specific case distictions; instead apply proper configuration

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -27,8 +27,6 @@ special settings for Docker usage:
 ```
 # Disable the GUI log window since it's not supported in Docker
 gui_log_window: False
-# Listen on all interfaces for the web dashboard to be accessible from outside the container
-web_dashboard_listen_address: 0.0.0.0
 # Disable opening the web dashboard on launch (not possible within the container)
 web_dashboard_open_on_launch: False
 ```
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,7 +64,6 @@ ENV SERENA_HOME=/workspaces/serena/config
 RUN mkdir -p $SERENA_HOME
 RUN cp src/serena/resources/serena_config.template.yml $SERENA_HOME/serena_config.yml
 RUN sed -i 's/^gui_log_window: .*/gui_log_window: False/' $SERENA_HOME/serena_config.yml
-RUN sed -i 's/^web_dashboard_listen_address: .*/web_dashboard_listen_address: 0.0.0.0/' $SERENA_HOME/serena_config.yml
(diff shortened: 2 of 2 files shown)


src/serena/config/serena_config.py
```diff
@@ -490,12 +468,7 @@ def from_config_file(cls, generate_if_missing: bool = True) -> "SerenaConfig":
         def get_value_or_default(field_name: str) -> Any:
             return loaded_commented_yaml.get(field_name, get_dataclass_default(SerenaConfig, field_name))
 
-        if is_running_in_docker():
-            instance.gui_log_window_enabled = False  # not supported in Docker
-        else:
-            instance.gui_log_window_enabled = get_value_or_default("gui_log_window_enabled")
-            # in docker the post_init takes care of setting the address to 0.0.0.0, we ignore the user's setting
-            instance.web_dashboard_listen_address = get_value_or_default("web_dashboard_listen_address")
+        instance.gui_log_window_enabled = get_value_or_default("gui_log_window_enabled")
```
MischaPanch, line 471: `instance.web_dashboard_listen_address = get_value_or_default("web_dashboard_listen_address")`

This is missing now

DOCKER.md
```diff
@@ -12,27 +12,19 @@ Docker support allows you to run Serena in an isolated container environment, wh
...
 
-### 1. Configuration File Conflicts
+### Configuration File Conflicts
 
-⚠️ **Critical**: Docker uses a separate configuration file (`serena_config.docker.yml`) to avoid path conflicts. When running in Docker:
-- Container paths will be stored in the configuration (e.g., `/workspaces/serena/...`)
-- These paths are incompatible with non-Docker usage
-- After using Docker, you cannot directly switch back to non-Docker usage without manual configuration adjustment
+Serena's configuration and log files are stored in the container in `/workspaces/serena/config/`.
+Any local configuration you may have for Serena will not apply; the container uses its own separate configuration.
```
MischaPanch, line 20:
```suggestion
Any local configuration you may have for Serena will not apply; the container uses its own separate configuration. If you want to use a persistent configuration file across containers, you will need to mount it to
the container path `~/.serena/serena_config.yml` and make sure that the `web_dashboard_listen_address` is set to `0.0.0.0` and `web_dashboard_open_on_launch` to `False` there.
```

opcode81: This is not a correct statement, so I cannot accept this change, but I have added an improved version of the statement.

src/serena/config/serena_config.py
opcode81: Fixed.

PR #831, opened 2025-12-10
Add PowerShell language support

michaelleone:
Implement PowerShell language server integration using PowerShell Editor Services (PSES). The implementation includes:

- PowerShellLanguageServer class with automatic PSES download and setup
- Support for PowerShell file extensions: .ps1, .psm1, .psd1
- Symbol detection using PSES format: "function FuncName ()"
- 7 basic tests + 2 integration tests covering:
  - Document symbols and function detection
  - Within-file references
  - Cross-file go-to-definition
  - SerenaAgent find_symbol and find_references

Requires: pwsh (PowerShell 7+) in PATH. PSES auto-downloads on first use.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -90,8 +90,8 @@ jobs:
         if: runner.os != 'Windows'
         uses: erlef/setup-beam@v1
         with:
-          elixir-version: "1.18.4"
-          otp-version: "26.1"
+          elixir-version: "1.19.3"
+          otp-version: "28"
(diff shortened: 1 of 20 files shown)


MischaPanch:
Thank you for the contribution @michaelleone! Very clean PR, I have nothing to add - I think this is the first time we are merging support for a new language without any review comments :)

The only minor thing is that we have some abstractions for downloading dependencies that potentially could be used here (meybe they'd need to be extended first), but it's not important and if helpful it could be done later

PR #834, opened 2025-12-12
Enhanced documentation on connecting to clients

MischaPanch:


docs/02-usage/030_clients.md
```diff
@@ -2,35 +2,94 @@
...
+### Troubleshooting
+
+Common issues involve your client not finding the `uvx` command or not escaping backslashes on Windows.
+If you are using paths containing backslashes for paths on Windows
+(note that you can also just use forward slashes), be sure to escape them correctly (`\\`).
+
+:::{note}
+Some language servers may require additional environment variables to be set (e.g. F# on macOS with Homebrew),
+which you may need to explicitly add to the MCP server configuration.
+Note that the spawned processes will not inherit environment variables that are only configured in your shell
```
opcode81: This is a specific property of Claude Desktop, which doesn't necessarily apply to other clients, so we should
* change this statement to say "may" instead of "will"
* keep a pointer to this in the Claude Desktop section, where it is definitely relevant.

or move this part back to where it was originally (where it was correct and relevant).

docs/02-usage/030_clients.md
```diff
@@ -53,87 +113,127 @@ After codex has started, you need to activate the project, which you can do by s
...
-               "command": "docker",
-               "args": ["run", "--rm", "-i", "--network", "host", "-v", "/path/to/your/projects:/workspaces/projects", "ghcr.io/oraios/serena:latest", "serena", "start-mcp-server", "--transport", "stdio"]
-           }
-       }
-   }
-   ```
+Once you have created the new MCP server entry, save the config and then restart Claude Desktop.
 
-If you are using paths containing backslashes for paths on Windows
-(note that you can also just use forward slashes), be sure to escape them correctly (`\\`).
```
opcode81, line 112: I would keep this. Users likely won't read the general instructions above, and it is a common source of error to not use proper escaping in JSON.
We can add the same statement to all clients that use JSON; it's not a problem for the document to be slightly longer as a result.

docs/02-usage/030_clients.md
```diff
@@ -2,35 +2,94 @@
...
+Note that the spawned processes will not inherit environment variables that are only configured in your shell
+profile (e.g. `.bashrc`, `.zshrc`, etc.); they would need to be set system-wide instead.
+An easy fix is to add them explicitly to the MCP server entry: most clients allow adding an `env` key to the `serena` object, e.g.
 
-```shell
-claude mcp add serena -- <serena> start-mcp-server --context claude-code --project "$(pwd)"
 ```
+"env": {
+    "DOTNET_ROOT": "/opt/homebrew/Cellar/dotnet/9.0.8/libexec"
+}
```
opcode81, line 62: This is also Claude Desktop-specific. It may be completely different (or even impossible) for other clients.
So we should at least mention that this is how it would work in Claude Desktop (but not necessarily elsewhere).

docs/02-usage/030_clients.md
```diff
@@ -2,35 +2,94 @@
...
+
+In the section on [running the MCP server](020_running), we described various ways of starting the Serena server. For brevity, here we present only commands that use the `uvx` way and the `stdio` protocol. You can easily adapt the commands to your preferred way of running Serena.
+
+Some clients have a per-workspace MCP configuration (e.g., VSCode and Claude Code), while others have a global MCP configuration (e.g., Codex and Claude Desktop). 
+
+- In the per-workspace case, you typically want Serena to start
+within the project directory and never switch to a different project, which is achieved by specifying the `--project <path>` argument when starting the MCP server. This also disables the `activate_project` tool.
+- In the global configuration case, you must first activate the project you want to work on, which you can do by asking the LLM to do so (e.g., "Activate the current dir as project using serena"). In such settings, the `activate_project` tool is required.
+
+### Troubleshooting
```
opcode81: Having a "Troubleshooting" section at the beginning of the document is weird. Also, this can instead be viewed as a collection of things to keep in mind in general.

docs/02-usage/030_clients.md
```diff
@@ -2,35 +2,94 @@
...
-    that Serena is configured to work on the current project from the get-go, following 
-    Claude Code's mode of operation.
-    
-Be sure to use at least `v1.0.52` of Claude Code (as earlier versions do not read MCP server system prompts upon startup). 
+
+* We use the `claude-code` context to disable unnecessary tools (avoiding duplication
+  with Claude Code's built-in capabilities).
+* We specify the current directory as the project directory with `--project "$(pwd)"`, such
+  that Serena is configured to work on the current project from the get-go, following
+  Claude Code's mode of operation.
```
opcode81: Are you applying an auto-formatter? The diff is way larger than it needs to be.

docs/02-usage/030_clients.md
MischaPanch: But essentially all of them use json, even Claude Code (the mcp add command just edits a json), so the statement would have to be repeated for each client. Maybe we should prominently recommend to read the general section first?

docs/02-usage/030_clients.md
MischaPanch: Most clients I have checked have this option, and for those who don't, users will notice

docs/02-usage/030_clients.md
MischaPanch: Done now

docs/02-usage/030_clients.md
MischaPanch: I changed to may and mentioned Claude Desktop here specifically

docs/02-usage/030_clients.md
opcode81: This was just an example. There are reformats all over the file.
Did you forget to push the changes?

docs/02-usage/030_clients.md
```diff
@@ -2,39 +2,108 @@
...
+  `--project <path>` argument when starting the MCP server. This also disables the `activate_project` tool.
+- In the global configuration case, you must first activate the project you want to work on, which you can do by asking
+  the LLM to do so (e.g., "Activate the current dir as project using serena"). In such settings, the `activate_project`
+  tool is required.
+
+### Common Pitfalls
+
+Your client may not escape backslashes on Windows (e.g., Claude Desktop does not).
+If you are using paths containing backslashes for paths on Windows
+(note that you can also just use forward slashes), be sure to escape them correctly (`\\`).
```
opcode81: This is not due to the client not applying escaping, it is a property of the JSON format.

docs/02-usage/030_clients.md
```diff
@@ -2,39 +2,108 @@
...
 
-## Claude Code
+* [adding command-line arguments](mcp-args)
+* [adjusting configuration](050_configuration).
 
-Serena is a great way to make Claude Code both cheaper and more powerful!
+While you may be able to turn off tools through your client's interface (e.g., in VSCode or Claude Desktop),
+we **highly** recommend selecting your toolset through the
+Serena configuration instead, since Serena's prompts automatically
+adjust based on which tools are enabled/disabled.
```
opcode81: This introductory stuff doesn't make sense here anymore. With the introduction of the General section, this no longer belongs here.
We could consider moving the general instructions at the beginning of the "Other Clients" section up into the "General" section, because these are actually universal (and thus general). (But note that the section is linked to.)

PR #835, opened 2025-12-12
Eclipse JDTLS: Add use_system_java_home, gradle_wrapper_enabled and gradle_java_home to support usage in context with restrictive gradle wrapper

tkruse:
In my company's dev setup serena fails to run the task `init_language_server_manager` due to 
```
org.gradle.api.plugins.UnknownPluginException: Plugin [id: '...'] was not found in any of the following sources
```
The issue is that the sources for our custom plugins are setup in gradlew wrapper.

This PR removes that as a PoC (without consideration of how this breaks for other projects).

The next problem was that in my company, a gradle plugin checks on every gradle run that an approved JRE is being used, while Serena/JDTLS apparently use a fixed JRE. That could be resolved with `ls_specific_settings` in `serena_config.yml`. The fork for this MR allowed to start serena successfully.

This MR is not intended to be merged as is, but only to demonstrate a minimal mitigation that works only within my company.

### DIFF
--- a/docs/02-usage/050_configuration.md
+++ b/docs/02-usage/050_configuration.md
@@ -281,6 +281,27 @@ Notes:
 - `GOFLAGS` (from the environment you start Serena in) may also affect the Go build context. Prefer `buildFlags` for tags.
 - Build context changes are only picked up when `gopls` starts. After changing `gopls_settings` (or relevant env vars like `GOFLAGS`), restart the Serena process (or server) that hosts the Go language server, or use your client's "Restart language server" action if it causes `gopls` to restart.
 
+#### Java (`eclipse.jdt.ls`)
+
+The following settings are supported for the Java language server:
+
+| Setting | Default | Description |
+|---|---|---|
+| `maven_user_settings` | `~/.m2/settings.xml` | Path to Maven `settings.xml` |
+| `gradle_user_home` | `~/.gradle` | Path to Gradle user home directory |
+| `gradle_wrapper_enabled` | `false` | Use the project's Gradle wrapper (`gradlew`) instead of the bundled Gradle distribution. Enable this for projects with custom plugins or repositories. |
+| `gradle_java_home` | `null` | Path to the JDK used by Gradle. When unset, Gradle uses the bundled JRE. |
+| `use_system_java_home` | `false` | Use the system's `JAVA_HOME` environment variable for JDTLS itself. Enable this if your project requires a specific JDK vendor or version for Gradle's JDK checks. |
+
+Example for a project with custom Gradle plugins and JDK requirements:
+
+```yaml
+ls_specific_settings:
+  java:
+    gradle_wrapper_enabled: true
+    use_system_java_home: true
+```
+
 #### Kotlin
 
 Serena uses [JetBrains' Kotlin Language Server](https://github.com/Kotlin/kotlin-lsp) for Kotlin support.
--- a/src/solidlsp/language_servers/eclipse_jdtls.py
+++ b/src/solidlsp/language_servers/eclipse_jdtls.py
@@ -49,8 +49,9 @@ class EclipseJDTLS(SolidLanguageServer):
     You can configure the following options in ls_specific_settings (in serena_config.yml):
         - maven_user_settings: Path to Maven settings.xml file (default: ~/.m2/settings.xml)
         - gradle_user_home: Path to Gradle user home directory (default: ~/.gradle)
-
-    Note: Gradle wrapper is disabled by default. Projects will use the bundled Gradle distribution.
+        - gradle_wrapper_enabled: Whether to use the project's Gradle wrapper (default: false)
+        - gradle_java_home: Path to JDK for Gradle (default: null, uses bundled JRE)
+        - use_system_java_home: Whether to use the system's JAVA_HOME for JDTLS itself (default: false)
 
     Example configuration in ~/.serena/serena_config.yml:
     ```yaml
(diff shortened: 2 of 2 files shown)


opcode81:
Hi Thibault, long time no see!

So is the purpose of this PR to provide a basis for a general mitigation strategy that shall be implemented independently, or do you intend to ultimately propose a PR that could be merged?

Either way, thanks for trying out Serena and for pointing out the issue.

tkruse:
Hi Dominik (I was surprised when I realized this project is by you),

my first intent was to provide an immediate github fork of Serena that can be installed and used by others in my company, and to get feedback.

In the medium term of course I would like to delete my fork again once we found a way to run the JDTLS by Serena with gradle projects in my company, e.g. by merging changes such as those in this MR (or a better fix).

At the least I think we would need a way to configure the gradlew wrapper and JRE behavior, and this MR could show a way, though I am not sure whether the MR is good/minimal, I did not have time to polish what the AI produced.
But I am unclear on why gradle is called at all, and with what tasks. Potentially there is another mitigation whether the call is not made or made differently.

MischaPanch:
Hi @tkruse . The Serena JetBrains plugin is now available, and we highly recommend it for all java/kotlin users. The performance is much better than using the LS background, no special things are needed from the user, and there are many important additional features (and more coming soon) when compared to the LS-based Serena integration.

src/solidlsp/language_servers/eclipse_jdtls.py
```diff
@@ -68,19 +71,99 @@ def __init__(self, config: LanguageServerConfig, repository_root_path: str, soli
...
+            jdtls_launcher_jar,
+            jdtls_config_path,
+            jdtls_readonly_config_path,
+        ]:
+            assert os.path.exists(static_path), static_path
+
+        # TODO: Add "self.runtime_dependency_paths.jre_home_path"/bin to $PATH as well
+        # Use system JAVA_HOME if available (important for Gradle wrapper JDK checks)
+        # Otherwise fall back to bundled JRE for JDTLS itself
+        system_java_home = os.environ.get("JAVA_HOME")
```
opcode81: Add an LS-specific setting `use_system_java_home` to enable this.

README.md
```diff
@@ -3,6 +3,22 @@
...
+> - Configurable `gradle_java_home` for JDK vendor requirements
+> - JDTLS inherits system `JAVA_HOME` for proper Gradle integration
+>
+> **Configuration (`~/.serena/serena_config.yml`):**
+> ```yaml
+> ls_specific_settings:
+>   java:
+>     gradle_wrapper_enabled: true  # default: true
+>     gradle_java_home: "/path/to/jdk"  # optional: override system JAVA_HOME
+> ```
```
opcode81: Document the new settings [here](https://oraios.github.io/serena/02-usage/050_configuration.html#language-server-specific-settings) instead.

README.md
```diff
@@ -3,6 +3,22 @@
   <img src="resources/serena-logo-dark-mode.svg#gh-dark-mode-only" style="width:500px">
 </p>
 
+> **Note:** This is a patched fork that fixes Gradle/JDTLS integration for Java projects with custom build configurations.
+>
+> **Key changes:**
+> - Gradle wrapper enabled by default (supports custom plugins/repositories)
+> - Configurable `gradle_java_home` for JDK vendor requirements
+> - JDTLS inherits system `JAVA_HOME` for proper Gradle integration
```
opcode81: Remove this

src/solidlsp/language_servers/eclipse_jdtls.py
```diff
@@ -404,6 +397,32 @@ def _get_initialize_params(self, repository_absolute_path: str) -> InitializePar
             gradle_user_home = None
             log.info(f"Gradle user home not found at default location ({default_gradle_home}), will use JDTLS defaults")
 
+        # Gradle wrapper: default to True (enabled) to support projects with custom plugins/repositories
+        # Note: When enabled, projects use their own gradlew which knows about custom repositories
+        gradle_wrapper_enabled = self._custom_settings.get("gradle_wrapper_enabled")
+        if gradle_wrapper_enabled is None:
+            # Default to True to support projects with custom plugins like 'com.indeed.common'
+            gradle_wrapper_enabled = True
```
opcode81: Default should be False to not change existing behaviour.

src/solidlsp/language_servers/eclipse_jdtls.py
tkruse: added

README.md
tkruse: documented in eclipse_jdtls.py

README.md
tkruse: removed

src/solidlsp/language_servers/eclipse_jdtls.py
tkruse: defaulted to false

tkruse:
> @tkruse we actually need only minor changes for this to be mergeable; see my comments.

Thanks for review, also I rebased and squashed.

docs/02-usage/050_configuration.md
```diff
@@ -257,6 +257,27 @@ Notes:
 - The language server package is a `.nupkg` file (ZIP format) downloaded from NuGet.org by default.
 - If you have .NET 10+ already installed, Serena will use your system installation.
 
+#### Java (`eclipse.jdt.ls`)
```
opcode81: Languages should follow alphabetical order in this section.

opcode81: I moved it and force-pushed.

opcode81:
Many thanks for finalising this, @tkruse! Merging.

PR #836, opened 2025-12-13
feat: CWD-based project detection for MCP with CLI-based agents

isometry:
Adds a new `--auto-project` flag to `start-mcp-server` that automatically detects and activates a project based on the current working directory.

## Motivation

Enables user-level MCP configuration (e.g., in Claude Code's global settings) without requiring per-context activation or explicit `--project="$(pwd)"`:

```
claude mcp add --scope=user serena -- <serena> start-mcp-server --context=claude-code --auto-project
```

## Detection logic

  1. Searches up from CWD for `.serena/` directory (explicit Serena project)
  2. Falls back to `.git/` if no `.serena` found
  3. Returns `None` if neither marker exists (server starts without active project)

### DIFF
--- a/src/serena/cli.py
+++ b/src/serena/cli.py
@@ -37,17 +37,18 @@ from solidlsp.util.subprocess_util import subprocess_kwargs
 log = logging.getLogger(__name__)
 
 
-def find_project_root(root: str | Path | None = None) -> str | None:
+def find_project_root(root: str | Path | None = None) -> str:
     """Find project root by walking up from CWD.
 
-    Checks for .serena first (explicit Serena project), then .git (git root).
+    Checks for .serena/project.yml first (explicit Serena project), then .git (git root).
+    Falls back to CWD if no marker is found.
 
     :param root: If provided, constrains the search to this directory and below
                  (acts as a virtual filesystem root). Search stops at this boundary.
-    :return: absolute path to project root, or None if not found
+    :return: absolute path to project root (falls back to CWD if no marker found)
     """
     current = Path.cwd().resolve()
-    boundary = Path(root).resolve() if root else None
+    boundary = Path(root).resolve() if root is not None else None
 
     def ancestors() -> Iterator[Path]:
         """Yield current directory and ancestors up to boundary."""
@@ -59,7 +60,7 @@ def find_project_root(root: str | Path | None = None) -> str | None:
 
     # First pass: look for .serena
     for directory in ancestors():
-        if (directory / ".serena").is_dir():
+        if (directory / ".serena" / "project.yml").is_file():
             return str(directory)
 
     # Second pass: look for .git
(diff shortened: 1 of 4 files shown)


isometry:
I don't think that the macOS test failure is related to this change…

opcode81:
I think something like this is indeed useful for CLI-based clients like Claude Caude that do not change the CWD, but it is important to mention this as a restriction in the documentation. Otherwise, users will try to apply this option and get unexpected results. 

And perhaps we can name the option more precisely to avoid misunderstandings (e.g. `--project-from-cwd`).

WDYT, @MischaPanch?

isometry:
I renamed the flag to `--project-from-cwd` as suggested and added a note about appropriate use.

I'm pretty sure the CI failures aren't related to this change?

src/serena/cli.py
```diff
@@ -35,6 +36,40 @@
...
+    """Find project root by walking up from CWD.
+
+    Checks for .serena first (explicit Serena project), then .git (git root).
+
+    :param root: If provided, constrains the search to this directory and below
+                 (acts as a virtual filesystem root). Search stops at this boundary.
+    :return: absolute path to project root, or None if not found
+    """
+    current = Path.cwd().resolve()
+    boundary = Path(root).resolve() if root else None
```
MischaPanch:
```suggestion
    boundary = Path(root).resolve() if root is not None else None
```

I suppose `root=""` should have the semantics of `root=current`, right?

src/serena/cli.py
```diff
@@ -35,6 +36,40 @@
...
+    def ancestors() -> Iterator[Path]:
+        """Yield current directory and ancestors up to boundary."""
+        yield current
+        for parent in current.parents:
+            yield parent
+            if boundary is not None and parent == boundary:
+                return
+
+    # First pass: look for .serena
+    for directory in ancestors():
```
MischaPanch, line 62: Let's bee a bit more explicit here and look for `.serena/project.yml`. We also create a `.serena` directory in the user's home, which could cause unexpected issues here

src/serena/cli.py
```diff
@@ -169,6 +211,15 @@ def start_mcp_server(
 
         log.info("Initializing Serena MCP server")
         log.info("Storing logs in %s", log_path)
+
+        # Handle --project-from-cwd flag
+        if project_from_cwd:
+            if project is not None or project_file_arg is not None:
+                raise click.UsageError("--project-from-cwd cannot be used with --project or positional project argument")
+            project = find_project_root()
+            if project:
```
MischaPanch:
```suggestion
            if project is not None:
```

src/serena/cli.py
```diff
@@ -169,6 +211,15 @@ def start_mcp_server(
...
+                log.info("Auto-detected project root: %s", project)
```
MischaPanch: should we log a warning or even raise an error? A user has specifically configured to activate a project, but we refuse (because we haven't found .git or .serena).

Generally, I wonder whether we need such a constraint, why not always start Serena in cwd as requested by the user? Are you worried about safety issues?

opcode81: I think activating cwd as a fallback would be the way to go.

isometry:
Comments addressed.

MischaPanch:
LGTM, thanks for the contribution!

PR #837, opened 2025-12-14
feat(nixd): add workspace/configuration handler for nixd

Eastwdd:
## Summary

This PR adds support for handling `workspace/configuration` requests from nixd in Serena's `nixd_ls.py` language server implementation. Currently, nixd sends `workspace/configuration` requests after initialization, but Serena doesn't have a handler registered, resulting in `-32601 Method Not Found` errors.

## Background

nixd (Nix Language Server) uses the LSP standard `workspace/configuration` request to fetch server configurations at runtime. While Serena already provides configuration via `initializationOptions`, nixd also requests configuration through `workspace/configuration` for dynamic settings.

Other language servers in Serena (OmniSharp, Haskell LS, C# LS) already implement `workspace/configuration` handlers.

### File: `src/solidlsp/language_servers/nixd_ls.py`

- Added `workspace_configuration_handler` function inside `_start_server` method
- Registered handler with `self.server.on_request("workspace/configuration", workspace_configuration_handler)`
- Handler returns appropriate configuration for nixd sections:
  - `nixd.nixpkgs`: Path/expression for nixpkgs evaluation
  - `nixd.formatting`: Formatting command configuration  
  - `nixd.options`: NixOS/home-manager options configuration
  - `nixd.diagnostic`: Diagnostic suppression settings

## Implementation Details

The handler follows the same pattern used by `haskell_language_server.py` and `omnisharp.py`, processing the `params["items"]` array and returning configuration for each requested section.

## Testing

Tested with a NixOS flake-based project where nixd is active. Before this change:
```
[Warn  - 10:30:00.000] Received error from server: "workspace/configuration" (-32601) method not found
```

After this change: No errors, configuration requests handled successfully.

## References

- nixd configuration docs: https://github.com/nix-community/nixd/blob/main/nixd/docs/configuration.md
- Similar implementation: `haskell_language_server.py` `workspace_configuration_handler`
- LSP spec: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_configuration

### DIFF
--- a/src/solidlsp/language_servers/nixd_ls.py
+++ b/src/solidlsp/language_servers/nixd_ls.py
@@ -255,6 +255,10 @@ class NixLanguageServer(SolidLanguageServer):
         self._is_flake_project = self._detect_flake_project(repository_root_path)
         self._nixd_options_config = self._build_options_config(repository_root_path, self._is_flake_project)
 
+        # Set request timeout for nixd - Nix evaluation can be slow, especially
+        # for hover requests that need to evaluate expressions from nixpkgs
+        self.set_request_timeout(120.0)  # 120 seconds for Nix expression evaluation
+
     @staticmethod
     def _detect_flake_project(repository_path: str) -> bool:
         """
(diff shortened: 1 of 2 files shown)


MischaPanch:
Hi. Is this still in draft status, and if yes, what is missing?

I don't know much about nix, but I'm wondering what exactly the issue was. Nix support did work in serena, right? It seems like this PR adds support for `workspace/configuration` requests after the initialization, but why would that be needed? Is this just useful for people who use use `solidlsp` as library, or are there currently issues with Nix support in Serena as MCP server?

Eastwdd:
The initial error was workspace/configuration: -32601: method 'workspace/configuration' not handled on the client. 

I'm using Serena through nix run github:oraios/serena (the flake) and consistently hit this error during nixd initialization.
The underlying issue is that the .nix test fixtures are intended to evaluate successfully, but currently on main, test/resources/repos/nix/test_repo/flake.nix is structured improperly. 

You can actually see this on main if you run a flake check at test/resources/repos/nix/test_repo/flake.nix—the evaluation will say "ignored," but really that just means it will error even if you force continue with nix flake check --keep-going.


This should be closed. I'm the only one getting the error because I was manually wrapping my nixd for some other unrelated debugging reason (or I just forgot—I don't know). 

Fixing the test fixtures' evaluation would be best practice; however, that can be done in a separate PR later.

PR #838, opened 2025-12-14
Update: metals(scala-lsp) version

YuMuuu:
close: #688

- Issues: Due to a metals issue, the field of a case class could not be obtained as a symbol.
- A fix in metals 1.6.4 allows case class fields to be retrieved as symbols.

modifications:
- add test for find_symbol case class field
- Implement install and run metals by specifying the version <- If you have a better idea for fixing this, please let me know.
- update metals dependency 1.6.2 -> 1.6.4

*Scala LSP Impliment is not tested in CI, so **careful review is required.**

Also, it would be better if there was a function to specify the LSP version to be used in the project in project.yml, etc.(just idea)

---

How to verify:

1.
Verify that the field of case cannot be obtained as a symbol in metals 1.6.2. 

That is, run the test with `git checkout 4ba352ec0b7f10c4929ae39f3f71fd550db83dc9` and verify that `uv run --extra dev poe test -m scala test/solidlsp/scala/test_scala_language_server.py::test_scala_document_symbols` fails.

2.
metals 1.6.4 supported case class field include symbol.

Then, run the test with `git checkout 4ba352ec0b7f10c4929ae39f3f71fd550db83dc9` and verify that `uv run --extra dev poe test -m scala test/solidlsp/scala/test_scala_language_server.py::test_scala_document_symbols` succeeds.

### DIFF
--- a/src/solidlsp/language_servers/scala_language_server.py
+++ b/src/solidlsp/language_servers/scala_language_server.py
@@ -57,7 +57,7 @@ class ScalaLanguageServer(SolidLanguageServer):
         """
         assert shutil.which("java") is not None, "JDK is not installed or not in PATH."
 
-        metals_version = "1.6.2"
+        metals_version = "1.6.4"
 
         metals_home = os.path.join(cls.ls_resources_dir(solidlsp_settings), "metals-lsp")
         os.makedirs(metals_home, exist_ok=True)


MischaPanch:
Looks good, thanks!

PR #850, opened 2025-12-16
SerenaConfig: Use "language_backend" instead of "jetbrains", allowing override via CLI

opcode81:


src/serena/config/serena_config.py
```diff
@@ -106,6 +107,18 @@ def get_serena_managed_in_project_dir(project_root: str | Path) -> str:
     return os.path.join(project_root, SERENA_MANAGED_DIR_NAME)
 
 
+class LanguageBackend(Enum):
+    JETBRAINS = "JetBrains"
+    LANGUAGE_SERVER = "LanguageServer"
```
opcode81: @MischaPanch wanted to get your opinion on this before I merge this. Which do you prefer for the second enum item?
* LanguageServer (current)
* LSP
* SolidLSP

opcode81: Changed it to LSP.

PR #851, opened 2025-12-17
Fix: Use {} instead of null in params field when None in JSON-RPC requests/notifications

zen010101:
According to JSON-RPC 2.0 specification, the params field must be a structured value (object or array) if present. Sending "params": null violates the spec and causes strict implementations to reject the request.

This fix modifies make_request() and make_notification() to use an empty dictionary ({}) when params is None, instead of including it as null.

### DIFF
--- a/src/solidlsp/lsp_protocol_handler/server.py
+++ b/src/solidlsp/lsp_protocol_handler/server.py
@@ -92,11 +92,15 @@ def make_error_response(request_id: Any, err: LSPError) -> StringDict:
 
 
 def make_notification(method: str, params: PayloadLike) -> StringDict:
-    return {"jsonrpc": "2.0", "method": method, "params": params}
+    # JSON-RPC 2.0: params must be object or array if present, cannot be null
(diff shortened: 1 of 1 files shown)


opcode81:
Thanks for the PR. 
I wonder: Is there a particular reason why you changed this, i.e. did you have a particular problem that is being solved by this change?

MischaPanch:
This might fix the antigravity integration, google seemed to be stricter about the schema in the past

opcode81:
> This might fix the antigravity integration, google seemed to be stricter about the schema in the past

@MischaPanch no, this is strictly about LS communication; it does not relate to MCP at all.

MischaPanch:
Ah, I see, had my hopes up after reading the description without looking at the code

zen010101:
> This might fix the antigravity integration, google seemed to be stricter about the schema in the past

Yes, I encountered this issue when adding Delphi/FPC LSP support to serena. Moreover, from the perspective of the LSP specification, this bug should also be fixed.

PR #863, opened 2025-12-22
Improved overview tools

MischaPanch:
Feel free to merge if you agree (and release the new plugin version then, I pushed the update on that side to main already)

### DIFF
--- a/src/serena/tools/jetbrains_plugin_client.py
+++ b/src/serena/tools/jetbrains_plugin_client.py
@@ -41,9 +41,13 @@ class JetBrainsPluginClient(ToStringMixin):
     """
 
     BASE_PORT = 0x5EA2
+    PLUGIN_REQUEST_TIMEOUT = 300
+    """
+    the timeout used for request handling within the plugin (a constant in the plugin)
+    """
     last_port: int | None = None
 
-    def __init__(self, port: int, timeout: int = 30):
+    def __init__(self, port: int, timeout: int = PLUGIN_REQUEST_TIMEOUT):
         self.base_url = f"http://127.0.0.1:{port}"
         self.timeout = timeout
         self.session = requests.Session()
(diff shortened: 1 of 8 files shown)


src/serena/tools/jetbrains_plugin_client.py
```diff
@@ -174,11 +174,13 @@ def find_references(self, name_path: str, relative_path: str) -> dict[str, Any]:
         request_data = {"namePath": name_path, "relativePath": relative_path}
         return self._make_request("POST", "/findReferences", request_data)
 
-    def get_symbols_overview(self, relative_path: str) -> dict[str, Any]:
+    def get_symbols_overview(self, relative_path: str, depth: int, exclude_imports: bool = True) -> dict[str, Any]:
```
opcode81: If we switch to my branch for the plugin impl, we should remove the `exclude_imports` argument.

src/serena/tools/symbol_tools.py
```diff
@@ -7,6 +7,8 @@
 from copy import copy
 from typing import Any
 
+from black.trans import defaultdict
```
opcode81: Not the right import

MischaPanch: Wtf was autocompelted here, oh my..

MischaPanch: Done

src/serena/tools/jetbrains_plugin_client.py
MischaPanch: Done

opcode81:
I have uploaded the new plugin release to JetBrains.
We should not merge this before the plugin is released, because otherwise LLMs will be confused by the depth parameter not being considered.

It is best not to combine JetBrains with non-JetBrains changes if we don't want to delay the non-JetBrains updates.

PR #881, opened 2025-12-27
Changes from mux

opcode81:


MischaPanch:
Looks good. I assume you tested that the MCP configuration and connection are still working?

The fixed-set feature is not exposed via configuration and not explained in the docs.,but that can be done later, if you want to.

PR #882, opened 2025-12-27
feat: Add open_dashboard MCP tool for on-demand dashboard opening

### LINKED ISSUE #875: Open Dashboard in Background
> When I use this with opencode, as soon the mcp starts I lose focus of opencode just when I'm about to write something. Is it possible to open the webpage without it popping to the front? I'm ok for it to open, just not lose focus on my prompt. I've learned that I can at least use Windows+cursor key down to get rid of it so I don't need to grab the mouse but it's still annoying.  

 _Originally posted by @imqqmi in [#613](https://github.com/oraios/serena/issues/613#issuecomment-3687187744)_

ximot:
## Summary

This PR adds a new MCP tool `open_dashboard` that allows users to open the Serena web dashboard on demand through AI assistance.

**Use case:** Some users prefer to disable automatic browser opening (`web_dashboard_open_on_launch: False`) to avoid focus stealing when Serena starts, but still want easy access to the dashboard when needed. With this tool, they can simply ask the AI to open the dashboard.

## Changes

- **New tool:** `OpenDashboardTool` in `config_tools.py`
- **Agent methods:** `open_dashboard()` and `get_dashboard_url()` in `SerenaAgent`
- **State:** Dashboard URL stored in `_dashboard_url` field
- **Documentation:** Updated `060_dashboard.md` with "Disabling Automatic Browser Opening" section
- **Tool list:** Added `open_dashboard` to `035_tools.md`
- **Config template:** Updated comment to mention the tool as an alternative

## Related Issue

Closes #875 - This provides a solution for users who want dashboard access without automatic browser opening and focus stealing.

## Test Plan

- [x] Format check passes (`uv run poe format`)
- [x] Type check passes (`uv run poe type-check`)
- [x] Python tests pass
- [x] Tested with MCP Inspector - tool works correctly

### DIFF
--- a/src/serena/agent.py
+++ b/src/serena/agent.py
@@ -250,18 +250,20 @@ class SerenaAgent:
 
         # determine the base toolset defining the set of exposed tools (which e.g. the MCP shall see),
         # determined by the
+        #   * dashboard availability/opening on launch
         #   * Serena config,
         #   * the context (which is fixed for the session)
-        #   * dashboard availability/opening on launch
         #   * single-project mode reductions (if applicable)
         #   * JetBrains mode
-        tool_inclusion_definitions: list[ToolInclusionDefinition] = [self.serena_config, self._context]
+        tool_inclusion_definitions: list[ToolInclusionDefinition] = []
         if (
             self.serena_config.web_dashboard
             and not self.serena_config.web_dashboard_open_on_launch
             and not self.serena_config.gui_log_window_enabled
         ):
             tool_inclusion_definitions.append(ToolInclusionDefinition(included_optional_tools=[OpenDashboardTool.get_name_from_cls()]))
+        tool_inclusion_definitions.append(self.serena_config)
+        tool_inclusion_definitions.append(self._context)
         if self._context.single_project:
             tool_inclusion_definitions.extend(self._single_project_context_tool_inclusion_definitions(project))
         if self.serena_config.language_backend == LanguageBackend.JETBRAINS:


docs/02-usage/060_dashboard.md
```diff
@@ -20,3 +20,12 @@ Serena comes with built-in tools for monitoring and managing the current session
...
 Both can be configured in Serena's [configuration](050_configuration) file (`serena_config.yml`).
 If enabled, they will automatically be opened as soon as the Serena agent/MCP server is started.
+
+## Disabling Automatic Browser Opening
+
+If you prefer not to have the dashboard open automatically (e.g., to avoid focus stealing), you can disable it
+by setting `web_dashboard_open_on_launch: False` in your `serena_config.yml`.
+
+When automatic opening is disabled, you can still access the dashboard by:
+* Navigating directly to `http://localhost:24282/dashboard/index.html` in your browser
```
MischaPanch: It's not always 24282, it's the closest greater-equal port that's open

src/serena/tools/config_tools.py
```diff
@@ -2,6 +2,19 @@
 from serena.tools import Tool, ToolMarkerDoesNotRequireActiveProject, ToolMarkerOptional
 
 
+class OpenDashboardTool(Tool, ToolMarkerDoesNotRequireActiveProject):
```
MischaPanch: should inherit from `ToolMarkerOptional` and documented that the tool should be explicitly enabled through the config. I suspect most users won't want it in their context window

src/serena/agent.py
```diff
@@ -405,6 +409,31 @@ def _open_dashboard(url: str) -> None:
         # open the dashboard URL in the default web browser
         webbrowser.open(url)
 
+    def get_dashboard_url(self) -> str | None:
```
MischaPanch, line 405: I think that's not needed, pls remove the method

src/serena/resources/serena_config.template.yml
```diff
@@ -35,7 +35,8 @@ web_dashboard_listen_address: 127.0.0.1
 web_dashboard_open_on_launch: True
 # whether to open a browser window with the web dashboard when Serena starts (provided that web_dashboard
 # is enabled). If set to False, you can still open the dashboard manually by navigating to
-# http://localhost:24282/dashboard/ in your web browser (24282 = 0x5EDA, SErena DAshboard).
+# http://localhost:24282/dashboard/ in your web browser (24282 = 0x5EDA, SErena DAshboard),
```
MischaPanch: same as above (our documentation was off). You'll also have to add that the open_dashboard tool has to be explicitly enabled for the latter version

opcode81:
@ximot I propose that we take over this PR and finish it.

src/serena/resources/serena_config.template.yml
opcode81: No, it's not off, it's in the line below:
> If you have multiple instances running, a higher port will be used; try port 24283, 24284, etc.

src/serena/tools/config_tools.py
opcode81: The tool is now enabled if
* the web dashboard is enabled,
* the web dashboard is not shown on startup, and
* the GUI tool is disabled (as it provides access to the dashboard URL)

src/serena/agent.py
opcode81: I now need it.

opcode81:
@MischaPanch I updated the PR. Please take a look.

Note: The original version was not functional - at least not on Windows - because the output redirection required the call to be made from the main thread, but it wasn't when called from the tool. So I now switched from multiprocessing to subprocess.

src/serena/resources/serena_config.template.yml
opcode81: I made some further updates to the comments.

ximot:
> @MischaPanch I updated the PR. Please take a look.
> 
> Note: The original version was not functional - at least not on Windows - because the output redirection required the call to be made from the main thread, but it wasn't when called from the tool. So I now switched from multiprocessing to subprocess.

Thank you for the changes in the code. I will check if there is anything else that needs to be corrected. If there are any comments, I will correct them. 

Today is Sunday, so I needed a bit of rest :)

opcode81:
@MischaPanch still waiting for your review. Can we merge this?

PR #889, opened 2025-12-29
Fix: Detect rust-analyzer from Homebrew and system PATH

ketema:
### Problem

Users with Homebrew-installed rust-analyzer receive "Neither rust-analyzer nor rustup is installed" even when rust-analyzer is correctly installed at `/opt/homebrew/bin/rust-analyzer`.

### Cause

`RustAnalyzer._get_rust_analyzer_path()` only checks `rustup which rust-analyzer`.

### Solution

Apply the same detection pattern used by `HaskellLanguageServer._ensure_hls_installed()`:
- Check `shutil.which("rust-analyzer")` last
- Check common install locations (Homebrew, cargo, local bin)
- preserve existing behavior first

This follows the explicit-paths approach (not arbitrary PATH scanning) established by HaskellLanguageServer.

### Compatibility

- Rustup users: unchanged behavior
- Auto-install via rustup: still works as fallback
- Error messages: now list all searched locations

### Tests

Added 9 unit tests in `test/solidlsp/rust/test_rust_analyzer_detection.py` describing detection behavior:
- PATH detection priority
- Homebrew paths (Apple Silicon + Intel)
- Cargo install path
- Rustup fallback
- Error message content
- Non-executable file handling

### DIFF
--- a/src/solidlsp/language_servers/rust_analyzer.py
+++ b/src/solidlsp/language_servers/rust_analyzer.py
@@ -70,23 +70,35 @@ class RustAnalyzer(SolidLanguageServer):
         """
         Ensure rust-analyzer is available.
 
-        1. Rustup (preferred - avoids incorrect PATH aliases)
-        2. Common installation locations (Homebrew, Scoop, cargo) as fallback
-        3. System PATH last (can pick up incompatible versions)
-        4. Auto-install via rustup if available
+        Priority order:
+        1. Rustup existing installation (preferred - matches toolchain version)
+        2. Rustup auto-install if rustup is available (ensures correct version)
+        3. Common installation locations as fallback (only if rustup not available)
+        4. System PATH last (can pick up incompatible versions)
 
         :return: path to rust-analyzer executable
         """
-        # Try rustup FIRST (original behavior - avoids picking up incorrect aliases from PATH)
+        # Try rustup FIRST (preferred - avoids picking up incompatible versions from PATH)
         rustup_path = RustAnalyzer._get_rust_analyzer_via_rustup()
         if rustup_path:
             return rustup_path
+
+        # If rustup is available but rust-analyzer not installed, auto-install it BEFORE
+        # checking common paths. This ensures we get the correct version matching the toolchain.
+        if RustAnalyzer._get_rustup_version():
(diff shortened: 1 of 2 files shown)


src/solidlsp/language_servers/rust_analyzer.py
```diff
@@ -53,48 +54,65 @@ def _get_rustup_version() -> str | None:
...
-        # Try again after installation
-        path = RustAnalyzer._get_rust_analyzer_path()
-        if not path:
-            raise RuntimeError("rust-analyzer installation succeeded but binary not found in PATH")
-
-        return path
+        # Try common locations (following Haskell pattern)
+        common_paths = [
+            shutil.which("rust-analyzer"),  # System PATH first
+            "/opt/homebrew/bin/rust-analyzer",  # macOS Homebrew (Apple Silicon)
```
MischaPanch: Should we add some typical windows paths?

MischaPanch (commented):
Thanks! Overall looks good, but if we're doing that, let's not discriminate against windows users ;)

(Unless the common windows locations are already covered)

ketema:
Good point. I’ll look up some and open to suggestions (I don’t use windows
so don’t know any off the top of my head)

ketema:
The previous comment was completely wrong. I inadvertently removed:
 result = subprocess.run(["rustup", "component", "add", "rust-analyzer"], check=False, capture_output=True, text=True)
 my mistake.

MischaPanch:
Hm, strange, we did not have rust failures in CI before though

MischaPanch:
@ketema you removed these lines

```
 result = subprocess.run(["rustup", "component", "add", "rust-analyzer"], check=False, capture_output=True, text=True)
```

which isntalled rust-analyzer on the fly, hence the failures in CI. Was that on purpose? In Serena we generally try to install LS on the fly if that's feasible, and it seems to be feasible for rust. If it was not on purpose, pls reinstate this logic to happen if `rustup` is installed but `rust-analyzer` is not

ketema:
Ah good catch. I’ll take care of it.

MischaPanch:
Thanks for the contribution @ketema !

PR #896, opened 2026-01-01
Adding completion, hover, inlay_hints, get_diagnostic tools

Yharooer:
## Notes
- Adding the following tools:
  -  `completion`: suggested autocompletions. Allows AI agents to discover what members a symbol has, especially useful for dependencies where it may not able to read the source easily.
  - `hover`: to get documentation of symbol, especially useful for dependencies where it may not be able to read the source easily.
  - `inlay_hints`: to see type hints
  - `get_diagnostics`: shows any LSP reported errors or warnings, helpful for agents to know if they have made errors without having to build; allows agents to see warnings (e.g. deprecation warnings) which they would not have seen otherwise.
- Additionally adding settings to Eclipse JDTLS to disable maven/gradle project discovery. In certain circumstances I would like to be able to force it to use the Eclipse project setup by disabling maven and gradle project discovery.

## Issues
#841 

## Testing
Tested manually with Java (Eclipse JDTLS), TypeScript, Python and Rust languages. 
- Java: `completion`, `hover` and `get_diagnostics` work as expected. `inlay_hints` always returns an empty list.
- TypeScript: `completion`, `hover` and `get_diagnostics` work as expected. `inlay_hints` always returns an empty list - I think something in the LSP needs to be enabled for these to be visible.
- Python: `completion`, `hover` and `get_diagnostics` work as expected. Pyright does not support inlay hints.
- Rust: All four tools work and give non-empty response.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,8 @@ Status of the `main` branch. Changes prior to the next official version change w
     (and which requires no escaping in the replacement text, making it more robust)
   * Added `CompletionTool` for getting code completion suggestions at a cursor position
   * Added `HoverTool` for getting type information and documentation at a cursor position
-  * Added `InlayHintsTool` for getting type hints and parameter names for a code range 
+  * Added `InlayHintsTool` for getting type hints and parameter names for a code range
+  * Added `GetDiagnosticsTool` for retrieving compiler errors, warnings, and hints from language servers 
 
 * Language support:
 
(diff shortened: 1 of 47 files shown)


MischaPanch:
Hi. Such large functional extensions have to be discussed and planned within an issue first. We are releasing new features that will provide high-level symbol info as output of find_symbol in the next few days, which will definitely affect what you wanted to achieve here.

If you need additional functionality after that, pls open an issue and we can discuss the best way to address it. I am keeping this PR open as reference for now, but we will probably not merge it in this form.

MischaPanch:
I appreciate the contribution though @Yharooer , there's for sure some value in this :)

opcode81:
Some of the features being addressed here have been integrated into Serena in the meantime. Others, like requesting completions, are unlikely to be useful to LLMs.

PR #899, opened 2026-01-04
Fix linux-musl platform ID string format

PazerOP:
## Summary
- Fix `get_platform_id()` generating incorrect platform ID string for musl-based Linux systems
- The function was producing `linux-arm64-musl` but the `PlatformId` enum expects `linux-musl-arm64`
- This caused `ValueError: 'linux-arm64-musl' is not a valid PlatformId` on Alpine Linux arm64 and similar systems

## Test plan
- [x] Verified the fix matches the existing enum values (`linux-musl-x64`, `linux-musl-arm64`)
- [x] Test on musl-based Linux (e.g., Alpine Linux on arm64)


MischaPanch:
Thanks for the fix!

PR #909, opened 2026-01-06
Support --mode and --context with yaml files

cx-alberto-simoes:
When using `--mode` and `--context` with a yml file path, it fails. 
This PR fixes it, adding the path of the mode/context file to the object itself, and using it whenever needed.

### DIFF
--- a/src/serena/config/context_mode.py
+++ b/src/serena/config/context_mode.py
@@ -131,11 +131,7 @@ class SerenaAgentMode(ToolInclusionDefinition, ToStringMixin):
 
         # If it looks like a file path but doesn't exist, raise FileNotFoundError
         name_or_path_str = str(name_or_path)
-        if (
-            os.sep in name_or_path_str
-            or (os.altsep and os.altsep in name_or_path_str)
-            or name_or_path_str.endswith((".yml", ".yaml"))
-        ):
+        if os.sep in name_or_path_str or (os.altsep and os.altsep in name_or_path_str) or name_or_path_str.endswith((".yml", ".yaml")):
             raise FileNotFoundError(f"Mode file not found: {path.resolve()}")
 
         return cls.from_name(str(name_or_path))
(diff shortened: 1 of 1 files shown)


MischaPanch:
Hi, thanks for the PR. The functionality with custom paths is quite old and predates the name-based resolution. That's why you run into problems with it. 

I would actually propose to instead no longer support custom paths and instead use the possibility to add custom modes and contexts to your home/.serena directory as we already offer. Would that work for you?

cx-alberto-simoes:
Hi. Thanks for the quick answer.

We were using those files in our home folder. But as we want to put it in our project git tree, using paths would make it simpler.
Of course we can add a link, but I suppose that, if the overhead is not much, adding this feature would be nice :-)

But up to you :-)

Thanks

opcode81:
@MischaPanch we have always supported passing a path instead of a name, our help texts do still say so,
```
  --context TEXT                  Built-in context name or path to custom
                                  context YAML.  [default: desktop-app]
  --mode TEXT                     Built-in mode names or paths to custom mode
                                  YAMLs.  [default: interactive, editing]
```
and the code that loads the data from a path is still in place. However, the dashboard then fails, because the code that gathers mode/context information assumes that only the name-based handling exists. 
This PR (probably) fixes this.

Note, however, that the code of our `load` function is questionable, as it assumes that it is a path *only* if the string ends with ".yml", so at the very least, the help text must be adjusted to say ".yml file" instead of "YAML".

cx-alberto-simoes:
@opcode81, @MischaPanch   I am happy to update the patch to check if the string passed in the CLI flag exists as a file, and if it does, consider it the file to load. If not, consider it a name. It should keep the current behavior and make it more flexible.

Currently I changed our pipeline to run (we are using uvx) directly from my PR branch and things are working pretty well.

cx-alberto-simoes:
This could be an option, but not sure if that's something you like.

```python
    @classmethod
    def load(cls, name_or_path: str | Path) -> Self:
        # Check if it's a file path that exists
        path = Path(name_or_path)
        if path.exists() and path.is_file():
            return cls.from_yaml(name_or_path)

        # If it looks like a file path but doesn't exist, raise FileNotFoundError
        name_or_path_str = str(name_or_path)
        if (
            os.sep in name_or_path_str
            or (os.altsep and os.altsep in name_or_path_str)
            or name_or_path_str.endswith((".yml", ".yaml"))
        ):
            raise FileNotFoundError(f"Mode file not found: {path.resolve()}")

        return cls.from_name(str(name_or_path))
```

If that's acceptable, I can include in the PR.

cx-alberto-simoes:
Applied the changes, we can still revert one commit in case you do not like the approach.

MischaPanch:
@opcode81 could you have a quick look when you have time pls?

opcode81:
This looks good to me! Thanks for the fixes, @cx-alberto-simoes.

cx-alberto-simoes:
Thank you!

PR #930, opened 2026-01-13
Fix _find_first_free_port to detect ports with active listeners

### LINKED ISSUE #929: Dashboard: _find_first_free_port fails to detect ports with active listeners
## Description

The `_find_first_free_port` function in `src/serena/dashboard.py` only checks if it can `bind()` to a port, but does not verify whether another process is already listening on that port. On Windows, `bind()` can succeed even when another process has an active listener due to `SO_REUSEADDR` behavior, causing the dashboard to attempt using an already-occupied port.

## Steps to Reproduce

1. Start any application listening on port 24282 (0x5EDA, the default Serena dashboard port)
2. Start Serena with the dashboard enabled
3. The dashboard will attempt to use the same port, causing a conflict

## Expected Behavior

`_find_first_free_port` should detect that a port has an active listener and skip to the next port.

## Actual Behavior

The function returns a port that already has an active listener because `bind()` succeeds despite the port being in use.

## Root Cause

The current implementation only uses `bind()` to check port availability:

```python
@staticmethod
def _find_first_free_port(start_port: int) -> int:
    port = start_port
    while port <= 65535:
        try:
            with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
                sock.bind(("0.0.0.0", port))
                return port
        except OSError:
            port += 1
    raise RuntimeError(f"No free ports found starting from {start_port}")
```

## Proposed Fix

Add a `connect_ex()` check before `bind()` to detect active listeners:

```python
@staticmethod
def _find_first_free_port(start_port: int) -> int:
    """Find the first available port starting from start_port.

    This checks both whether a port has an active listener (via connect)
    and whether we can bind to it. Both checks are necessary because:
    - On Windows, bind() can succeed even if another process is listening
      (due to SO_REUSEADDR behavior)
    - A port might not have a listener but still be bound by another process

    Args:
        start_port: The port number to start searching from.

    Returns:
        The first available port number.

    Raises:
        RuntimeError: If no free port is found up to port 65535.
    """
    port = start_port
    while port <= 65535:
        # First check if something is already listening on this port
        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as check_sock:
            # Use a short timeout - connect_ex returns immediately if connection
            # is refused, timeout only applies if something is blocking
            check_sock.settimeout(0.05)
            try:
                result = check_sock.connect_ex(("127.0.0.1", port))
                if result == 0:
                    # Connection succeeded, something is listening - skip this port
                    port += 1
                    continue
            except OSError:
                # Connection attempt failed, port might be available
                pass

        # Now try to bind to verify we can use it
        try:
            with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
                sock.bind(("0.0.0.0", port))
                return port
        except OSError:
            port += 1

    raise RuntimeError(f"No free ports found starting from {start_port}")
```

## Environment

- OS: Windows 11
- Python: 3.11.8

zeynalnia:
## Summary
- Fix `_find_first_free_port` to detect ports with active listeners by adding a `connect_ex()` check before `bind()`
- Add integration tests for the port detection logic

Fixes #929

## Problem
The `_find_first_free_port` function only checked if `bind()` succeeded, but on Windows `bind()` can succeed even when another process is actively listening on the port due to `SO_REUSEADDR` behavior. This caused the dashboard to attempt using ports that were already in use.

## Solution
Add a `connect_ex()` check before `bind()` to detect active listeners:
1. First attempt to connect to the port - if successful, something is listening
2. Then try to bind to verify we can actually use the port

Both checks are necessary because:
- On Windows, `bind()` can succeed even if another process is listening
- A port might not have a listener but still be bound by another process

## Test plan
- [x] Added `test_skips_listening_port` test that creates an actual listening server and verifies the function skips it
- [x] All existing tests pass
- [x] Format and type-check pass

### DIFF
--- a/test/serena/test_dashboard.py
+++ b/test/serena/test_dashboard.py
@@ -185,9 +185,7 @@ class TestFindFirstFreePort:
             found_port = SerenaDashboardAPI._find_first_free_port(test_port, "0.0.0.0")
 
             # The found port must NOT be the one with the listener
-            assert found_port != test_port, (
-                f"Function returned port {test_port} which has a listener on 127.0.0.1 " f"(checked with 0.0.0.0)!"
-            )
+            assert found_port != test_port, f"Function returned port {test_port} which has a listener on 127.0.0.1 (checked with 0.0.0.0)!"
             assert found_port > test_port
 
             # Verify the found port is actually usable


opcode81:
Please see my post in the issue. The small change I made fixed the issue for me.

zeynalnia:
Closing this PR as the maintainer correctly pointed out that the `connect_ex` check is unnecessary.

We tested and confirmed that on Windows:
1. Two processes can bind to the same port with different addresses (e.g., `0.0.0.0` and `127.0.0.1`)
2. Requests are routed based on address specificity - the more specific address wins

Since Serena Dashboard binds to `127.0.0.1`, it will always handle localhost requests correctly even if another process is bound to `0.0.0.0` on the same port.

The fix in commit 2b50c36 (passing `host` parameter for address consistency) is sufficient.

Issue #517, opened 2025-08-22
Indexing doesn't work on Native Windows


thomhurst:
PS D:\git\TUnit> uvx --from git+https://github.com/oraios/serena serena project index
Indexing symbols in project D:\git\TUnit…
ERROR 2025-08-22 12:17:21,962 solidlsp:window_log_message:535 - LSP: [initialized] [Microsoft.CodeAnalysis.LanguageServer.HostWorkspace.Razor.RazorDynamicFileInfoProvider] RazorDynamicFileInfoProvider not initialized. RazorWorkspaceService or RazorLspDynamicFileInfoProvider is null.
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\thomh\AppData\Local\uv\cache\archive-v0\eIFspDIqLjRgnAfRZTuZS\Scripts\serena.exe\__main__.py", line 10, in <module>
    sys.exit(top_level())
             ~~~~~~~~~^^
  File "C:\Users\thomh\AppData\Local\uv\cache\archive-v0\eIFspDIqLjRgnAfRZTuZS\Lib\site-packages\click\core.py", line 1442, in __call__
    return self.main(*args, **kwargs)
           ~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "C:\Users\thomh\AppData\Local\uv\cache\archive-v0\eIFspDIqLjRgnAfRZTuZS\Lib\site-packages\click\core.py", line 1363, in main
    rv = self.invoke(ctx)
  File "C:\Users\thomh\AppData\Local\uv\cache\archive-v0\eIFspDIqLjRgnAfRZTuZS\Lib\site-packages\click\core.py", line 1830, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "C:\Users\thomh\AppData\Local\uv\cache\archive-v0\eIFspDIqLjRgnAfRZTuZS\Lib\site-packages\click\core.py", line 1830, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "C:\Users\thomh\AppData\Local\uv\cache\archive-v0\eIFspDIqLjRgnAfRZTuZS\Lib\site-packages\click\core.py", line 1226, in invoke
[...]

Issue #506, opened 2025-08-20
Serena fails when `nul` file exists in directory on Windows


keoy7am:
Serena encounters a fatal error when a file named nul exists in the project directory on Windows systems. The nul
   file appears to be created by certain IDEs (exact source unknown), and its presence causes path resolution
  failures due to Windows treating nul as a reserved device name.

  Environment

  - OS: Windows
  - UV: 0.8.12 (36151df0e 2025-08-18)
  - Python: 3.12.2 (tags/v3.12.2:6abddd9, Feb 6 2024, 21:26:36) [MSC v.1937 64 bit (AMD64)]

  Steps to Reproduce

  1. Have a nul file present in the project directory
  2. Run serena project index or serena health-check
  3. Observe the ValueError

  Expected Behavior

  Serena should handle Windows reserved filenames gracefully, either by:
  - Skipping files with reserved names
  - Providing a clear warning message
  - Continuing operation without crashing

  Actual Behavior

  The application crashes with a ValueError when attempting to resolve relative paths involving the nul file.

---

  Error Logs

<details>
  <summary>Execute Project Index</summary>

```
  Indexing symbols in project A:\2. TestProjects\1. SerenaTest\ProjectIndex
  Traceback (most recent call last):
    File "<frozen runpy>", line 198, in _run_module_as_main
    File "<frozen runpy>", line 88, in _run_code
    File
  "C:\Users\WindowsUser\AppData\Local\uv\cache\archive-v0\uwSETn4veLVV9UEm820HC\Scripts\serena.exe\__main__.py",
  line 10, in <module>
      sys.exit(top_level())
               ~~~~~~~~~^^
    File
  "C:\Users\WindowsUser\AppData\Local\uv\cache\archive-v0\uwSETn4veLVV9UEm820HC\Lib\site-packages\click\core.py",
  line 1442, in __call__
      return self.main(*args, **kwargs)
             ~~~~~~~~~^^^^^^^^^^^^^^^^^
    File
  "C:\Users\WindowsUser\AppData\Local\uv\cache\archive-v0\uwSETn4veLVV9UEm820HC\Lib\site-packages\click\core.py",
  line 1363, in main
      rv = self.invoke(ctx)
    File
  "C:\Users\WindowsUser\AppData\Local\uv\cache\archive-v0\uwSETn4veLVV9UEm820HC\Lib\site-packages\click\core.py",
  line 1830, in invoke
      return _process_result(sub_ctx.command.invoke(sub_ctx))
[...]

Issue #330, opened 2025-07-22
Error activating project even with explicit project path included in .claude.json


CanML:
Here is my trace in the logs - it looks like it is making some assumptions about the relative path (emphasis mine):

INFO  2025-07-22 13:43:27,533 [MainThread] serena.agent:_update_active_tools:350 - Active tools (27): activate_project, check_onboarding_performed, delete_memory, find_file, find_referencing_symbols, find_symbol, get_current_config, get_symbols_overview, initial_instructions, insert_after_symbol, insert_before_symbol, list_dir, list_memories, onboarding, prepare_for_new_conversation, read_memory, remove_project, replace_regex, replace_symbol_body, restart_language_server, search_for_pattern, summarize_changes, switch_modes, think_about_collected_information, think_about_task_adherence, think_about_whether_you_are_done, write_memory
INFO  2025-07-22 13:43:27,534 [MainThread] serena.config.serena_config:start:324 - Project configuration auto-generation starting ...
INFO  2025-07-22 13:43:29,541 [MainThread] serena.config.serena_config:stop:331 - Project configuration auto-generation completed in 2.007 seconds
ERROR 2025-07-22 13:43:29,541 [MainThread] serena.agent:__init__:226 - **Error activating project 'C:\code\myproject' at startup: path is on mount '\\\\.\\nul', start on mount 'C:'**

And here is the section from my .claude.json file (emphasis mine):

"serena": {
          "type": "stdio",
          "command": "uvx",
          "args": [
            "--from",
            "git+https://github.com/oraios/serena",
            "serena-mcp-server",
            "--context",
            "ide-assistant",
            **"--project",
            "C:\\code\\myproject"**
          ],
          "env": {}
        }

This works fine for getting the MCP active and working in Claude Code but doesn't seem to find the project correctly despite the project folder being explicitly specified. I'm using Claude Code in Windows using PowerShell and similar absolute path arguments are working successfully for two other MCP servers I am actively using. Perhaps there is a specific requirement for serena that I have failed to notice and follow but I was somewhat surprised this approach didn't work so thought I'd submit it as a potential bug.

opcode81:
I have extended the logging for this error. Please retry and provide the full trace; perhaps it will shed some light on what is actually happening here.

CanML:
[...]

PR #931, opened 2026-01-13
Fix ValueError on Windows when scanning directory with special paths

MrCRACKDev:
## Summary

On Windows, the `scan_directory` function in `src/serena/util/file_system.py` crashes when encountering special device paths like `\\.\nul` (the NUL device). This happens because `os.path.relpath()` raises a `ValueError` when trying to compute a relative path between a special device path and a regular drive path.

This is a **recurring issue** that has affected multiple users across different scenarios:
- Project indexing (#517)
- Directory scanning when `nul` file exists (#506)
- Project activation and configuration (#330)

## Problem

Windows reserves certain filenames (`nul`, `con`, `prn`, `aux`, `com1-com9`, `lpt1-lpt9`) as device names. When Python's `os.path.relpath()` encounters these, it interprets them as device mount points rather than regular files, throwing:

```
ValueError: path is on mount '\\.\nul', start on mount 'C:'
```

This causes Serena tools like `Search For Pattern`, `Activate Project`, and indexing operations to crash completely instead of gracefully skipping the problematic entry.

## Solution

Expand the exception handling in `scan_directory()` to catch both `PermissionError` and `ValueError`:

```python
# Before
except PermissionError as ex:
    # Skip files/directories that cannot be accessed due to permission issues
    log.debug(f"Skipping entry due to permission error: {entry.path}", exc_info=ex)
    continue

# After
except (PermissionError, ValueError) as ex:
    # Skip files/directories that cannot be accessed due to permission issues
    # or path errors (e.g. special device paths on Windows like \\.\nul)
    log.debug(f"Skipping entry due to error: {entry.path}", exc_info=ex)
    continue
```

## Changes

| File | Change |
|------|--------|
| `src/serena/util/file_system.py` | Added `ValueError` to exception handling on line 75, updated comment |

## Testing

All checks pass:

- `uv run poe format` - Passed
- `uv run poe type-check` - Passed
- `uv run poe test -m python` - 47 passed
- `test/serena/util/test_file_system.py` - 22 passed

### Before Fix

| Scenario | Result |
|----------|--------|
| Indexing C# project on Windows (#517) | Crash with ValueError |
| Directory with `nul` file (#506) | Crash with ValueError |
| Project activation (#330) | Crash with ValueError |
| Search For Pattern tool | Crash, forces fallback to less efficient methods |

### After Fix

| Scenario | Result |
|----------|--------|
| Indexing C# project on Windows | Completes successfully, skips special paths |
| Directory with `nul` file | Completes successfully, logs debug message |
| Project activation | Works normally |
| Search For Pattern tool | Returns results, no fallback needed |

## Root Cause Analysis

Windows special device names are reserved at the filesystem level:
- `NUL` - Null device (like `/dev/null` on Unix)
- `CON` - Console
- `PRN` - Printer
- `AUX` - Auxiliary device
- `COM1-COM9` - Serial ports
- `LPT1-LPT9` - Parallel ports

These can appear in repositories when:
1. Code is committed on Unix systems (where these are valid filenames)
2. Projects are cloned to Windows
3. Certain build tools or packages create them

Python's `os.path.relpath()` treats these as being on a different "mount" than regular drives, causing the ValueError.

## References

- Python `os.path.relpath` docs: https://docs.python.org/3/library/os.path.html#os.path.relpath
- Windows reserved names: https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#win32-device-namespaces
- Related PR #510: Improve handling of ignored files (partial fix for `should_ignore`)

### DIFF
--- a/src/serena/util/file_system.py
+++ b/src/serena/util/file_system.py
@@ -72,9 +72,10 @@ def scan_directory(
                                 )
                                 files.extend(sub_result.files)
                                 directories.extend(sub_result.directories)
-                except PermissionError as ex:
+                except (PermissionError, ValueError) as ex:
                     # Skip files/directories that cannot be accessed due to permission issues
-                    log.debug(f"Skipping entry due to permission error: {entry.path}", exc_info=ex)
+                    # or path errors (e.g. special device paths on Windows like \\.\nul)
+                    log.debug(f"Skipping entry due to error: {entry.path}", exc_info=ex)
                     continue
     except PermissionError as ex:
         # Skip the entire directory if it cannot be accessed


opcode81:
The issues you marked as being fixed were fixed already, so I am removing those. Your issue has a similar cause (`relpath`) but in a different place.

opcode81:
I fixed this using a more specific try-except block in d75b41b6aea6366a7517ad839c8294a17f594cb9.  ValueError can have many causes, after all.

Thanks for reporting the issue!

Issue #917, opened 2026-01-10
Configurable LSP paths and optional installation


peperunas:
As observed in #465, the current hardcoded LSP installation logic fails in environments like NixOS or when using specific version managers where global write access or standard `npm install` commands are not viable or desirable.

Serena currently attempts to install LSPs to specific paths automatically, which causes failures when those paths are read-only or when the required package managers (like npm/node) aren't available in the expected global scope.

It would be very beneficial to introduce a configuration key to specify the path for each language server.

**Example config:**
```yaml
lsp:
  typescript:
    path: "/run/current-system/sw/bin/typescript-language-server"
```

MischaPanch:
Yes, that's a nice feature. It can be done by modifying `SolidLanguageServer.create` as following

```python
        # For now, we assume that all language server implementations have the same signature of the constructor
        # (which, unfortunately, differs from the signature of the base class).
        # If this assumption is ever violated, we need branching logic here.
        ls_executable_path = solidlsp_settings.get_ls_specific_settings(ls_class.get_language_enum_instance()).get("ls_executable_path", None)
        ls = ls_class(config, repository_root_path, solidlsp_settings, ls_executable_path)  # type: ignore
```

and then modifying **every single language server implementation** to accept this parameter and skip the installation if the path was provided.

This is a great task for an agent, I am fairly confident that Opus 4.5 together with ralph wiggum plugin can do this. Requires lot's of tokens and them some testing of this new feature (manual testing by a human).

That's a nice and fairly straightforward thing to take over for someone from the community with access to a powerful agent and enough tokens to spare :).

Context: we have a bit of a technical debt in the untyped love triangle between `create`, `__init__` and the implementation's `__init__` in `SolidLanguageServer` which is making this change a bit ugly and cumbersome.

@opcode81 WDYT of the proposal?

For the community - pls don't start until @opcode81 gives his ok.

peperunas:
I believe it's a pretty key feature because, at least on Nix, many of the LSPs don't work because they are downloaded and run directly, without being handled by the store.

MischaPanch:
@peperunas would you be up for implementing it along the lines I proposed? (If @opcode81 agrees)

peperunas:
I can give it a shot, sure

opcode81:
We should not approach this naively.

### Assumptions

[...]

PR #932, opened 2026-01-13
feat: allow custom LSP path

peperunas:
Proposed PR for #917. WIP.

### DIFF
--- a/src/solidlsp/language_servers/al_language_server.py
+++ b/src/solidlsp/language_servers/al_language_server.py
@@ -55,7 +55,7 @@ class ALLanguageServer(SolidLanguageServer):
             3. Extract and configure the platform-specific executable
 
         """
-        custom_command = solidlsp_settings.get_ls_specific_settings(self.get_language_enum_instance()).get("command", None)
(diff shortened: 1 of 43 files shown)


opcode81:
@peperunas what is going on? None of these changes are based on my recommendations in the issue.

peperunas:
The changes you proposed are quite a refactor. I do not have that much time to delve into it, unfortunately. 

This is a straightforward solution that replaces the hardcoded command with the setting `command` from the `.yml`. Maybe it can be in place while a more robust solution can be developed with more time :-)

opcode81:
We will not merge this. We have to do this properly or not at all.

peperunas:
Okay 👍

opcode81:
@peperunas, it is not really an extensive refactoring. There will be a fairly small change set (if indentation is ignored) in every LS implementation. 

For the most part, we are just
 * moving `_setup_runtime_dependencies` and accompanying methods into a class
 * renaming the setup method and adjusting its interface slightly
 * adding a factory method to the LS class 

It becomes extensive only because there are so many language servers.

PR #933, opened 2026-01-13
Introduce LanguageServerDependencyProvider abstraction

opcode81:
* Add specialisation `LanguageServerDependencyProviderSinglePath`, where the core path can be overridden via `ls_specific_settings` #917
  * Backwards compatibility is retained; the new approach is optional
  * Implemented new approach for several language servers.
  
 NOTE: Diff should be reviewed not on GitHub but using a diff viewer that ignores whitespace changes. The diffs are actually quite small when correctly analysed.

### DIFF
--- a/docs/02-usage/050_configuration.md
+++ b/docs/02-usage/050_configuration.md
@@ -134,9 +134,10 @@ For advanced users, Serena's configuration can be further customized.
 The Serena user data directory (where configuration, language server files, logs, etc. are stored) defaults to `~/.serena`.
 You can change this location by setting the `SERENA_HOME` environment variable to your desired path.
 
-### Language server specific settings (`ls_specific_settings`)
+### Language Server-Specific settings (`ls_specific_settings`)
 
-`ls_specific_settings` lets you pass per-language, language-server-specific configuration.
+Under the key `ls_specific_settings` in `serena_config.yml`, you can you pass per-language, 
+language server-specific configuration.
 
 Structure:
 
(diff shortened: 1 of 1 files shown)


opcode81:
Note: Given that both the old and new approach are, for the time being, supported simultaneously, we can merge this at any time. We needn't implement the new approach for all language servers.

The ones already implemented and building on the `SinglePath` variant already support the LS-specific setting `ls_path`.

opcode81:
@MischaPanch what do you think? Shall we merge this? We can continue the refactoring at any time, since it is (for the time being) backwards-compatible.

The only thing we should probably do soon is to adapt the instructions for developing a new LS to mention the new concepts.

src/solidlsp/ls.py
```diff
@@ -130,6 +130,68 @@ def get_all_symbols_and_roots(self) -> tuple[list[ls_types.UnifiedSymbolInformat
...
+    @abstractmethod
+    def _get_or_install_core_dependency(self) -> str:
+        """
+        Gets the language server's core path, potentially installing dependencies beforehand.
+
+        :return: the core dependency's path (e.g. executable, jar, etc.)
+        """
+
+    def create_launch_command(self) -> Union[str, list[str]]:
+        path = self._custom_settings.get("ls_path", None)
```
MischaPanch, line 183: We should document the possibility to pass `ls_path` for servers that support it. Or do you want to do that when the refactoring is finished?

The string "ls_path" becomes a crucial part of the interface but it can only be understood after reading the documentation. I think it should be mentioned in the class docstring, maybe even factored into the constant `LS_PATH_KEY` for more clarity?

opcode81: For now, it's just an undocumented feature. We can document it at any time.
I wouldn't say it's "crucial" at all. Very few users will care about tihs.

opcode81: I added a comment to the class docstring, which is definitely important to have.

opcode81: I merged main and revised and extended the documentation on `ls_specific_settings`

PR #938, opened 2026-01-14
Open dashboard param in cli and client info in dashboard

### LINKED ISSUE #934: Add --open-dashboard parameter
Many people probably unnecessarily use the --enable-dashboard flag to prevent the dashboard from being shown, because it's the only command line option available.

We should introduce the new flag and edit our recommendation in the discussions to use it - and the docs, too, of course.

MischaPanch:
Closes #934 

I also added the display of the client info in dashboard. An MCP server can server multiple clients simultaneously, so what is displayed is the client of the last tool call. Unfortunately, there seems to be no easy way to get info about the client that initialized a stdio session (for http mode there is no such client, of course). Thus, the information is displayed after the first tool call. Verified to work for claude code. I haven't checked whether other clients follow this part of MCP, but it's also not a problem to keep showing "none" for those.

### DIFF
--- a/src/serena/tools/tools_base.py
+++ b/src/serena/tools/tools_base.py
@@ -4,8 +4,10 @@ from abc import ABC
 from collections.abc import Iterable
 from dataclasses import dataclass
 from types import TracebackType
-from typing import TYPE_CHECKING, Any, Protocol, Self, TypeVar
+from typing import TYPE_CHECKING, Any, Protocol, Self, TypeVar, cast
 
+from mcp import Implementation
+from mcp.server.fastmcp import Context
 from mcp.server.fastmcp.utilities.func_metadata import FuncMetadata, func_metadata
 from sensai.util import logging
 from sensai.util.string import dict_string
@@ -115,6 +117,17 @@ class Tool(Component):
     # (which is use by the LLM, so a good description is important)
     # and to validate the tool call arguments.
 
+    _last_tool_call_client_str: str | None = None
+    """We can only get the client info from within a tool call. Each tool call will update this variable."""
+
+    @classmethod
+    def set_last_tool_call_client_str(cls, client_str: str | None) -> None:
+        cls._last_tool_call_client_str = client_str
+
+    @classmethod
+    def get_last_tool_call_client_str(cls) -> str | None:
+        return cls._last_tool_call_client_str
+
     @classmethod
     def get_name_from_cls(cls) -> str:
         name = cls.__name__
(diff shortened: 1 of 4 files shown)


opcode81:
I don't get the point of displaying the client info in the dashboard (the user knows which client he is using) or of adding a tool for this (why would the client ask for information about itself?). The one place where this information would have been useful, however, is in the logs - because that's what people are likely to send to us (we *don't* know what they're using).

src/serena/tools/tools_base.py
```diff
@@ -219,10 +224,20 @@ def _limit_length(self, result: str, max_answer_chars: int) -> str:
     def is_active(self) -> bool:
         return self.agent.tool_is_active(self.__class__)
 
-    def apply_ex(self, log_call: bool = True, catch_exceptions: bool = True, **kwargs) -> str:  # type: ignore
+    def apply_ex(self, log_call: bool = True, catch_exceptions: bool = True, mcp_ctx: Context | None = None, **kwargs) -> str:  # type: ignore
         """
         Applies the tool with logging and exception handling, using the given keyword arguments
         """
+        if mcp_ctx is not None:
+            global LAST_TOOL_CALL_CLIENT_INFO
```
opcode81: I doubly disapprove of this.
* Using a global variable is a definite anti-pattern. Why don't you use a static member of `Component` for this?
* All-caps naming is for constants, not variables.

MischaPanch:
I often have three different sessions open, from different clients (but on the same project). It's very useful to know which session is from which client.

The tool is just for debugging, we can delete it. It's marked as optional.

I can move it to a static member

MischaPanch:
Before you ask which different sessions: from copilot, from Jetbrains AI, from Junie and from Claude code

src/serena/tools/tools_base.py
```diff
@@ -300,6 +315,14 @@ def _to_json(x: Any) -> str:
         return json.dumps(x, ensure_ascii=False)
 
 
+class GetClientInfoTool(Tool, ToolMarkerOptional):
+    """Mainly for convenience or debugging."""
+
+    def apply(self) -> str | None:
+        """Return the client info of the last tool call (if available)."""
+        return LAST_TOOL_CALL_CLIENT_INFO
```
opcode81: Let's delete the tool. The debugging is use case is better handled through logs.
We should log `"Client info: <info>"` whenever the new value is different from the old value.

(Also, if we kept the tool, it would not belong in this module.)

opcode81:
> I often have three different sessions open, from different clients (but on the same project). It's very useful to know which session is from which client.

In that case, I think it could make more sense for the client info to be part of every tool execution. (But it bloats the data, of course.)

EDIT: Ah, sorry, you are probably not connecting to the same Serena instance (HTTP mode), so forget what I said here.

src/serena/tools/tools_base.py
MischaPanch: Moved to static member of Tool with getter and setter

src/serena/tools/tools_base.py
MischaPanch: Deleted

PR #941, opened 2026-01-15
Add SERENA_JETBRAINS_HOST env var for remote IDE connections

pscheit:
Hey Guys,

so excited about the jetbrains plugin, currently setting this up. I had this idea, too, but very happy that you implemented this now.

I hit a configuration issue though.

I run all my projects within WSL2 and my IDE runs on the windows host. Therefore serena (and e.g. php with xdebug) cannot connect to localhost. You need the IP address of your windows host (instead of using 127.0.0.1).

Now 127.0.0.1 was hardcoded.

This PR now changes that. 

When the JetBrains IDE runs on a different host than Serena (e.g., IDE on Windows with Serena in WSL2, or Serena in a Docker container), the plugin client now supports configuring the host via the SERENA_JETBRAINS_HOST environment variable.

### DIFF
--- a/docs/02-usage/025_jetbrains_plugin.md
+++ b/docs/02-usage/025_jetbrains_plugin.md
@@ -62,5 +62,31 @@ will be available exclusively through the JetBrains plugin.
 We realize that not everyone uses a JetBrains IDE as their main code editor.
 You can still take advantage of the JetBrains plugin by running a JetBrains IDE instance alongside your
 preferred editor. Most JetBrains IDEs have a free community edition that you can use for this purpose.
-You just need to make sure that the project you are working on is open and indexed in the JetBrains IDE, 
+You just need to make sure that the project you are working on is open and indexed in the JetBrains IDE,
 so that Serena can connect to it.
+
+## IDE on a Different Host
+
+If your JetBrains IDE runs on a different host than Serena (e.g., IDE on Windows with Serena in WSL2,
+or Serena in a Docker container), you can configure the host address using the `SERENA_JETBRAINS_HOST`
+environment variable:
+
+```bash
+export SERENA_JETBRAINS_HOST=192.168.16.1
+```
+
+Or in your MCP client configuration (e.g., `.mcp.json`):
+
+```json
+{
+  "mcpServers": {
+    "serena": {
+      "env": {
+        "SERENA_JETBRAINS_HOST": "192.168.16.1"
+      }
+    }
+  }
+}
+```
+
+Replace `192.168.16.1` with the IP address of the host where your JetBrains IDE is running.
--- a/src/serena/tools/jetbrains_plugin_client.py
+++ b/src/serena/tools/jetbrains_plugin_client.py
@@ -4,6 +4,7 @@ Client for the Serena JetBrains Plugin
 
 import json
 import logging
+import os
 import re
 from pathlib import Path
 from typing import Any, Literal, Optional, Self, TypeVar, cast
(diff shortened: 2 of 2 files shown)


pscheit:
ah okay i see, to get this to work with WSL2 more work is needed, because the JetBrains plugin binds to `127.0.0.1` (localhost only). An additional workaround is needed on the Windows side:

### 1. Add Windows Firewall Rule

```powershell
New-NetFirewallRule -DisplayName "Serena JetBrains Plugin" -Direction Inbound -Protocol TCP -LocalPort 24226-24245 -Action Allow
```

I think this is needed regardless.

### 2. Set Up Port Forwarding

Since the plugin only listens on localhost, you need to forward traffic from the Windows host IP to localhost for each port the plugin uses:

Get your Windows host IP (the one WSL2 can reach)  
Then forward the ports (run as Administrator):

```powershell
netsh interface portproxy add v4tov4 listenaddress=192.168.16.1 listenport=24226 connectaddress=127.0.0.1 connectport=24226
netsh interface portproxy add v4tov4 listenaddress=192.168.16.1 listenport=24227 connectaddress=127.0.0.1 connectport=24227
netsh interface portproxy add v4tov4 listenaddress=192.168.16.1 listenport=24228 connectaddress=127.0.0.1 connectport=24228
```

To remove later:
```powershell
netsh interface portproxy delete v4tov4 listenaddress=192.168.16.1 listenport=24226
netsh interface portproxy delete v4tov4 listenaddress=192.168.16.1 listenport=24227
netsh interface portproxy delete v4tov4 listenaddress=192.168.16.1 listenport=24228
```

Step 2 would be obsolete if you guys could make the listen interface configurable. There is an option in the jetbrains plugin for xdebug that says something like "accept connections from all hosts", which I need to check for my setup as well

opcode81:
Thanks!

The key thing to consider is that Serena currently does not use env vars for configuration - `SERENA_HOME` being the only (necessary) exception - and whether we should start doing this in general is something we should think about. @MischaPanch, what do you think?

Serena is currently configured using
* two levels of configuration files (global, project)
* composable additional configuration (context, mode)
* command-line parameters for per-session overrides

Env vars have the advantage that 
 * they can be used both globally and per session,
 * they need not be injected and can simply be retrieved where they are needed.

So for a setting like this, it's somewhat convenient, resulting in very few code changes.

If we go for this, I think I'd prefer `SERENA_JETBRAINS_SERVER_ADDRESS`.
We could also consider adding `SERENA_JETBRAINS_SERVER_LISTEN_ADDRESS` for the plugin to use when binding the port.

pscheit:
oh yeah, let me know, can make this a global configuration setting as well

MischaPanch:
I propose to handle this like I had handled such config in a previous project. We allow configuring values like `env:SERENA_JETBRAINS_SERVER_ADDRESS` in Serena's config files. If the env prefix is present, this means the value will be read of the env. In the downstream code we then only read values from the config object. To me this is the best of both worlds

opcode81:
@MischaPanch, your proposal addresses only the first advantage of env vars I mentioned, but not the second.
The second point was: No need for injection. When using env vars only indirectly, we still need to inject the configuration wherever it is needed. There are 7 calls of `JetBrainsPluginClient.from_project(project)`, and they all need to be updated to also pass the server address, requiring that the configuration in turn be available at the call site.

I am not saying that we should therefore use env vars, mind you, but env vars are genuinely advantageous in this regard.

opcode81:
With the commit I just added, I made this configurable via the global config file, so this can be closed.

@pscheit if you think it's helpful, feel free to add an issue on configurability of the listen address in the plugin.

PR #945, opened 2026-01-16
Clarify `repl` parameter: backslash escapes are not expanded (GPT-5.2-codex fix)

n-r-w:
This MR updates the tool help text for the `repl` parameter to prevent a recurring formatting failure observed with **`GPT-5.2-codex`**.

In practice, the model sometimes interprets the wording “verbatim” too literally and produces tool arguments where `repl` contains **literal escape sequences** such as `\\n` (and potentially `\\t`, etc.). Since the replacement string is applied **as-is**, those two-character sequences (`\` + `n`) get written into the target file instead of becoming real newlines/tabs, resulting in broken edits.

### DIFF
--- a/src/serena/tools/file_tools.py
+++ b/src/serena/tools/file_tools.py
@@ -190,7 +190,7 @@ class ReplaceContentTool(Tool, ToolMarkerCanEdit):
             If `mode` is "literal", this string will be matched exactly.
             If `mode` is "regex", this string will be treated as a regular expression (syntax of Python's `re` module,
             with flags DOTALL and MULTILINE enabled).
-        :param repl: the replacement string (verbatim).
+        :param repl: the replacement string (used as-is; backslash escapes are not expanded).
             If mode is "regex", the string can contain backreferences to matched groups in the needle regex,
             specified using the syntax $!1, $!2, etc. for groups 1, 2, etc.
         :param mode: either "literal" or "regex", specifying how the `needle` parameter is to be interpreted.
--- a/src/serena/tools/memory_tools.py
+++ b/src/serena/tools/memory_tools.py
@@ -82,7 +82,7 @@ class EditMemoryTool(Tool, ToolMarkerCanEdit):
             If `mode` is "literal", this string will be matched exactly.
             If `mode` is "regex", this string will be treated as a regular expression (syntax of Python's `re` module,
             with flags DOTALL and MULTILINE enabled).
-        :param repl: the replacement string (verbatim).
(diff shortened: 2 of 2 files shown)


src/serena/tools/file_tools.py
```diff
@@ -190,7 +190,7 @@ def apply(
             If `mode` is "literal", this string will be matched exactly.
             If `mode` is "regex", this string will be treated as a regular expression (syntax of Python's `re` module,
             with flags DOTALL and MULTILINE enabled).
-        :param repl: the replacement string (verbatim).
+        :param repl: the replacement string (used as-is; backslash escapes are not expanded).
```
opcode81, line 193: Can we try just "used as-is"? 
The less we say about escaping, the better. The more we say, the more potential for confusion. (This was surprisingly hard to get right.)

n-r-w: I'm currently testing this description with the codex. I'll try removing the escaping part and see what happens.
This PR is in the draft precisely because I haven't finished testing it.

n-r-w: But I also need to check sonnet/opus. And also GLM 4.7, since I have it too.
All this will take about a week, because such problems are always unstable.

opcode81:
Ultimately, I would say that the LLM is not taking the word "verbatim" too literally, it is simply ignoring it. The word describes precisely what is required, yet the model doesn't adhere to it.

n-r-w:
I think it's best not to touch anything, since it's hard to predict how other LLMs will react. Anyone who needs to can simply create their own context. E.g.:
```yaml
tool_description_overrides:
  edit_memory: |
    Replaces content matching a regular expression in a memory.
    🚨 CRITICAL: MUST NOT expand backslash escapes for `repl` parameter 🚨

  replace_content: |
    Replaces one or more occurrences of a given pattern in a file with new content.

    This is the preferred way to replace content in a file whenever the symbol-level
    tools are not appropriate.

    VERY IMPORTANT: The "regex" mode allows very large sections of code to be replaced
    without fully quoting them!

    Use a regex of the form "beginning.*?end-of-text-to-be-replaced" to be faster and
    more economical!

    ALWAYS try to use wildcards to avoid specifying the exact content to be replaced,
    especially if it spans several lines. Note that you cannot make mistakes, because
    if the regex should match multiple occurrences while you disabled
    `allow_multiple_occurrences`, an error will be returned, and you can retry with a
    revised regex.

    Therefore, using regex mode with suitable wildcards is usually the best choice!

    🚨 CRITICAL: MUST NOT expand backslash escapes for `repl` parameter 🚨
```

PR #947, opened 2026-01-17
Adjust the activation prompt for Codex, to make GPT-5 more compliant.

Linde7777:
The previous prompt had a 15% to 30% chance that GPT-5 would forget to execute necessary tools, such as `read serena instruction`


MischaPanch:
Thanks! Merging

PR #950, opened 2026-01-18
feat(pascal): Support ls_specific_settings for Pascal LSP

zen010101:
## Summary
- Add support for configuring Pascal Language Server via `ls_specific_settings["pascal"]`
- Settings: `pp`, `fpcdir`, `lazarusdir`, `fpc_target`, `fpc_target_cpu`
- These are passed as environment variables (PP, FPCDIR, LAZARUSDIR, FPCTARGET, FPCTARGETCPU) to pasls

## Why
CodeTools (used by pasls) needs `fpc.exe` path to query compiler configuration via `fpc -iV`, `fpc -iTO`, etc. Without proper PP setting, hover/navigation fails on projects with complex include directives (e.g., mORMot2).

## Example Configuration
```yaml
ls_specific_settings:
    pascal:
        pp: "D:/laz32/fpc/bin/i386-win32/fpc.exe"
        fpcdir: "D:/laz32/fpcsrc"
        lazarusdir: "D:/laz32/lazarus"
```

## Test Plan
- [x] Type check passes
- [x] Pascal tests pass (41 tests)
- [x] Manual testing with mORMot2 project

### DIFF
--- a/docs/02-usage/050_configuration.md
+++ b/docs/02-usage/050_configuration.md
@@ -233,26 +233,32 @@ Serena automatically checks for pasls updates every 24 hours. Updates include:
 - Atomic update with rollback on failure
 - Windows file locking detection (defers update if pasls is in use)
 
-**Configuration via Environment Variables:**
+**Configuration:**
 
-pasls reads compiler paths from environment variables. Set these before starting Serena:
+Configure pasls via `ls_specific_settings.pascal` in `serena_config.yml`:
 
-| Variable       | Description                                              |
-| -------------- | -------------------------------------------------------- |
-| `PP`           | Path to FPC compiler executable                          |
-| `FPCDIR`       | Path to FPC source directory                             |
-| `LAZARUSDIR`   | Path to Lazarus directory (only needed for LCL projects) |
-| `FPCTARGET`    | Target OS (e.g., `linux`, `win64`, `darwin`)             |
-| `FPCTARGETCPU` | Target CPU (e.g., `x86_64`, `aarch64`, `i386`)           |
+| Setting          | Description                                                                 |
+| ---------------- | --------------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


opcode81:
Please document the new settings [here](https://oraios.github.io/serena/02-usage/050_configuration.html#language-server-specific-settings)

zen010101:
Documentation added for Pascal language server settings. Please review.

docs/02-usage/050_configuration.md
```diff
@@ -215,6 +215,45 @@ Notes:
...
+
+pasls reads compiler paths from environment variables. Set these before starting Serena:
+
+| Variable       | Description                                              |
+| -------------- | -------------------------------------------------------- |
+| `PP`           | Path to FPC compiler executable                          |
+| `FPCDIR`       | Path to FPC source directory                             |
+| `LAZARUSDIR`   | Path to Lazarus directory (only needed for LCL projects) |
+| `FPCTARGET`    | Target OS (e.g., `linux`, `win64`, `darwin`)             |
+| `FPCTARGETCPU` | Target CPU (e.g., `x86_64`, `aarch64`, `i386`)           |
```
opcode81: You did not describe how to set `ls_specific_settings` but the environment variables that the settings are mapped to.
Note that setting environment variables directly "before starting Serena" is potentially more difficult, as Serena itself is started by the MCP client, and the client may or may not pass on the environment (many clients do not). How to do this correctly thus becomes client-specific (and, in the worst case, impossible).

So you should describe the fail-safe way here - which is to describe how to set `ls_specific_settings`, which then sets the env vars for the LS process (but which the user doesn't necessarily need to know/care about). After all, this is what your PR adds, isn't it?

zen010101:
Updated. Documentation now focuses on `ls_specific_settings` configuration instead of environment variables.

Issue #951, opened 2026-01-18
`include_info=true` makes symbolic search inoperative on large CPP projects.


n-r-w:
`include_info=true` (the default value) makes symbolic search completely inoperable on large CPP projects.
`find_referencing_symbols` takes over a minute to complete and times out in most cases. However, with `include_info=false`, everything works instantly.
The project is indexed. With `include_info=true`, clangd only uses one core and takes a VERY long time to search.
Linux, Ryzen 7 7950x3d, clangd version 21.1.0

MischaPanch:
Include_info fires an additional hover request for each symbol. I assumed it would be fast for each language server, a shame that's not the case for clangd. We can set the default to false for now. 

Note that for the Jetbrains plugin there is no significant performance overhead (afaik), jetbrains is much faster in retrieving symbol metainfo

n-r-w:
I tested it on several versions of clangd - the result is the same - terrible slowdowns.
One core is used and that's it.

<img width="870" height="910" alt="Image" src="https://github.com/user-attachments/assets/6e5eb941-bf91-4eaf-9165-4d4199b73a0a" />

n-r-w:
But the worst thing is that after canceling the command, the search stops working altogether until you kill the process.

opcode81:
I have now disabled it by default, which should reduce problems. (I would have opted to disable this by default anyway.)

MischaPanch:
clangd has unfortunately proved generally hard to use and very resource hungry.

One thing that would help (although I don't know how much) is to not constantly open and close a file for each hover request. This means we would first find all references, group them by files, open a file, perform all hover requests and close it, proceed to next file.

It's a small refactoring and could solve this, in case the file opening is the bottleneck (I strongly suspect that it is). @n-r-w would you like to try doing that? You already worked on core components (the cache), that's why I'm asking

opcode81:
> after canceling the command, the search stops working altogether until you kill the process.

Are you cancelling via Serena's Dashboard or in another way?
Always cancel via the Dashboard, because, otherwise, Serena will not be aware of the cancellation and the task will not be properly removed from the task queue.

n-r-w:
> > after canceling the command, the search stops working altogether until you kill the process.
> 
> Are you cancelling via Serena's Dashboard or in another way? Always cancel via the Dashboard, because, otherwise, Serena will not be aware of the cancellation and the task will not be properly removed from the task queue.

I used this (vscode) 

[...]

PR #956, opened 2026-01-19
Hover budget

n-r-w:
- Introduced `include_info_hover_budget_seconds` to ProjectConfig and SerenaConfig for per-project and global hover request time limits.
- Updated YAML templates to include hover budget settings with appropriate documentation.
- Enhanced LanguageServerSymbolRetriever to respect hover budget during symbol info requests, allowing for partial results if the budget is exceeded.
- Added unit tests to validate hover budget behavior and ensure correct handling of various budget scenarios. 
- Enhance symbol retrieval by adding batch request method and optimize hover info fetching + DEBUG

### DIFF
--- a/src/serena/config/serena_config.py
+++ b/src/serena/config/serena_config.py
@@ -135,7 +135,7 @@ class ModeSelectionDefinition:
 
 
 @dataclass
-class ToolBehaviorConfig:
+class SharedConfig(ModeSelectionDefinition, ToolInclusionDefinition, ToStringMixin):
     """Shared between SerenaConfig and ProjectConfig, the latter used to override values in the form
     (same as in ModeSelectionDefinition).
     The defaults here shall be none and should be set to the global default values in SerenaConfig.
@@ -173,7 +173,7 @@ class LanguageBackend(Enum):
 
 
 @dataclass(kw_only=True)
-class ProjectConfig(ToolInclusionDefinition, ToolBehaviorConfig, ModeSelectionDefinition, ToStringMixin):
+class ProjectConfig(SharedConfig):
     project_name: str
     languages: list[Language]
     ignored_paths: list[str] = field(default_factory=list)
@@ -485,7 +485,7 @@ class RegisteredProject(ToStringMixin):
 
 
 @dataclass(kw_only=True)
-class SerenaConfig(ToolInclusionDefinition, ToolBehaviorConfig, ModeSelectionDefinition, ToStringMixin):
+class SerenaConfig(SharedConfig):
     """
     Holds the Serena agent configuration, which is typically loaded from a YAML configuration file
     (when instantiated via :method:`from_config_file`), which is updated when projects are added or removed.


n-r-w:
https://github.com/oraios/serena/issues/951

src/serena/symbol.py
```diff
@@ -519,17 +522,146 @@ def _request_info(self, relative_file_path: str, line: int, column: int) -> str
...
+            detail = sym.symbol_root.get("detail")
+            if isinstance(detail, str) and detail.strip():
+                detail_present_count += 1
+
+            file_path = sym.relative_path
+            line = sym.line
+            column = sym.column
+            if file_path is None or line is None or column is None:
+                skipped_symbols += 1
+                continue
```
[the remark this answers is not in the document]
n-r-w: The goal of this PR is strictly outcome-based: when include_info=true hits the hover time budget we must stop doing hover lookups, keep any info we already obtained, and simply omit info for the rest.

With the current implementation, callers use info_by_symbol.get(symbol) and only attach the info field when it’s a non-empty string. That means both cases (a) missing key for invalid-location symbols and (b) present key with None for budget-skipped symbols - produce the same user-visible result: no info is shown. Adding invalid symbols into the mapping as None would be purely internal/contract reshaping, with no improvement to the external behavior we care about.

src/solidlsp/ls.py
```diff
@@ -1086,15 +1109,21 @@ def request_document_symbols(self, relative_file_path: str, file_buffer: LSPFile
...
+                        log.debug("perf: document_symbols_cache HIT path=%s", relative_file_path)
                     return document_symbols
-                else:
-                    log.debug("Cached document symbol content for %s has changed", relative_file_path)
-            else:
-                log.debug("No cache hit for document symbols in %s", relative_file_path)
+
+                log.debug("Cached document symbol content for %s has changed", relative_file_path)
+                if log.isEnabledFor(logging.DEBUG):
+                    log.debug("perf: document_symbols_cache STALE path=%s", relative_file_path)
```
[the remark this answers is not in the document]
n-r-w: fixed

src/solidlsp/ls.py
```diff
@@ -1033,15 +1049,22 @@ def _request_document_symbols(
...
+            file_hash, result = file_hash_and_result
+            if file_hash == fd.content_hash:
+                log.debug("Returning cached raw document symbols for %s", relative_file_path)
+                if log.isEnabledFor(logging.DEBUG):
+                    log.debug("perf: raw_document_symbols_cache HIT path=%s", relative_file_path)
+                return result
+
+            log.debug("Document content for %s has changed (raw symbol cache is not up-to-date)", relative_file_path)
+            if log.isEnabledFor(logging.DEBUG):
+                log.debug("perf: raw_document_symbols_cache STALE path=%s", relative_file_path)
```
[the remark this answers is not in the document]
n-r-w: fixed

src/solidlsp/ls.py
```diff
@@ -1033,15 +1049,22 @@ def _request_document_symbols(
...
-            if file_hash_and_result is not None:
-                file_hash, result = file_hash_and_result
-                if file_hash == fd.content_hash:
-                    log.debug("Returning cached raw document symbols for %s", relative_file_path)
-                    return result
-                else:
-                    log.debug("Document content for %s has changed (raw symbol cache is not up-to-date)", relative_file_path)
-            else:
+            if file_hash_and_result is None:
                 log.debug("No cache hit for raw document symbols symbols in %s", relative_file_path)
```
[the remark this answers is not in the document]
n-r-w: fixed

src/solidlsp/ls.py
```diff
@@ -1086,15 +1109,21 @@ def request_document_symbols(self, relative_file_path: str, file_buffer: LSPFile
             # check if the desired result is cached
             cache_key = relative_file_path
             file_hash_and_result = self._document_symbols_cache.get(cache_key)
-            if file_hash_and_result is not None:
+            if file_hash_and_result is None:
+                log.debug("No cache hit for document symbols in %s", relative_file_path)
+                if log.isEnabledFor(logging.DEBUG):
+                    log.debug("perf: document_symbols_cache MISS path=%s", relative_file_path)
```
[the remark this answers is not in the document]
n-r-w: fixed

src/solidlsp/ls.py
```diff
@@ -1086,15 +1109,21 @@ def request_document_symbols(self, relative_file_path: str, file_buffer: LSPFile
...
+            else:
                 file_hash, document_symbols = file_hash_and_result
                 if file_hash == file_data.content_hash:
                     log.debug("Returning cached document symbols for %s", relative_file_path)
+                    if log.isEnabledFor(logging.DEBUG):
+                        log.debug("perf: document_symbols_cache HIT path=%s", relative_file_path)
```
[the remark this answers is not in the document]
n-r-w: fixed

src/solidlsp/ls.py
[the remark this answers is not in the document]
n-r-w: fixed

src/serena/symbol.py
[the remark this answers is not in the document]
n-r-w: updated docstring for clarity

MischaPanch:
@n-r-w I made some improvements, in part aesthetic, but also removing the `hover_cache` (I don't see how a cache hit would ever happen, do I miss something?) and the part that analyzed `symbol["detail"]` (what was that for? We are not using it so far, but do the LS return it? If yes, we could use it as fallback instead of info when the hover budget is exceeded, but you haven't done that).

This PR will have to wait until #969 is merged. Some minor conflicts could appear.

I have not put much attention to the new config setting and how it is handled, @opcode81 could you have a look at that pls?

One more thing @n-r-w : we are considering moving the code in `src/serena` to the GPL license. The library code in `src/solidlsp` would remain MIT licensed. Would this be ok with you? We generally had very few external contributions to `src/serena`, so the move should be rather painless.

n-r-w:
> @n-r-w I made some improvements, in part aesthetic, but also removing the `hover_cache` (I don't see how a cache hit would ever happen, do I miss something?) and the part that analyzed `symbol["detail"]` (what was that for? We are not using it so far, but do the LS return it? If yes, we could use it as fallback instead of info when the hover budget is exceeded, but you haven't done that).
> 
> This PR will have to wait until #969 is merged. Some minor conflicts could appear.
> 
> I have not put much attention to the new config setting and how it is handled, @opcode81 could you have a look at that pls?
> 
> One more thing @n-r-w : we are considering moving the code in `src/serena` to the GPL license. The library code in `src/solidlsp` would remain MIT licensed. Would this be ok with you? We generally had very few external contributions to `src/serena`, so the move should be rather painless.

Unfortunately, I don't have the time to deal with this issue right now.
Regarding the type of license, it’s your business, it doesn’t matter to me.

MischaPanch:
@n-r-w no problem, we're taking this over and will merge soon. Thanks for the greenlight concerning the license

src/serena/config/serena_config.py
```diff
@@ -171,6 +171,12 @@ class ProjectConfig(ToolInclusionDefinition, ModeSelectionDefinition, ToStringMi
     ignore_all_files_in_gitignore: bool = True
     initial_prompt: str = ""
     encoding: str = DEFAULT_SOURCE_FILE_ENCODING
+    hover_budget: float | None = None
```
opcode81: propose to rename to `ls_hover_budget`

src/serena/config/serena_config.py
```diff
@@ -493,6 +511,14 @@ class SerenaConfig(ToolInclusionDefinition, ModeSelectionDefinition, ToStringMix
     ls_specific_settings: dict = field(default_factory=dict)
     """Advanced configuration option allowing to configure language server implementation specific options, see SolidLSPSettings for more info."""
 
+    hover_budget: float = 5.0
```
opcode81: It might make sense to move this to a separate dataclass `LSConfig` and establish a systemic way of retrieving the valid setting for a given project.

src/serena/config/serena_config.py
```diff
@@ -634,6 +660,16 @@ def get_value_or_default(field_name: str) -> Any:
...
 
+        # Load and validate hover_budget
+        hover_budget_raw = get_value_or_default("hover_budget")
+        try:
+            hover_budget = float(hover_budget_raw)
+        except (TypeError, ValueError) as e:
+            raise ValueError(f"hover_budget must be a number, got: {hover_budget_raw}") from e
+        if hover_budget < 0:
+            raise ValueError(f"hover_budget cannot be negative, got: {hover_budget}")
+        instance.hover_budget = hover_budget
```
opcode81: The generic handling is sufficient; this should be deleted.

src/serena/config/serena_config.py
```diff
@@ -323,6 +329,17 @@ def _from_dict(cls, data: dict[str, Any]) -> Self:
...
+        # Validate hover_budget
+        hover_budget_raw = data["hover_budget"]
+        hover_budget = hover_budget_raw
+        if hover_budget is not None:
+            try:
+                hover_budget = float(hover_budget_raw)
+            except (TypeError, ValueError) as e:
+                raise ValueError(f"hover_budget must be a number or null, got: {hover_budget_raw}") from e
+            if hover_budget < 0:
+                raise ValueError(f"hover_budget cannot be negative, got: {hover_budget}")
```
opcode81: The default handling suffices.

src/serena/symbol.py
```diff
@@ -535,17 +538,126 @@ def _request_info(self, relative_file_path: str, line: int, column: int) -> str
...
+        if self.agent is not None:
+            hover_budget = self.agent.serena_config.hover_budget
+            active_project = self.agent.get_active_project()
+            if active_project is not None:
+                project_hover_budget = active_project.project_config.hover_budget
+                if project_hover_budget is not None:
+                    hover_budget = project_hover_budget
+        return hover_budget
+
+    def request_info_for_symbol_batch(
```
opcode81, line 602: Do we really need all the debug code in this function?
Why don't we simplify this, perhaps use Stopwatch.

MischaPanch: I wouldn't say we need it, it was established by @n-r-w , presumably to help check speedups and so, I did not see a reason to remove it here. I have however slightly altered the logging on another, private branch. So we should continue the conversation there. Since it's not urgent, I propose to postpone addressing this

src/serena/config/serena_config.py
MischaPanch: Done

src/serena/config/serena_config.py
MischaPanch: Done

src/serena/config/serena_config.py
MischaPanch: Closing in favor of introducing LSConfig

src/serena/config/serena_config.py
MischaPanch: I have done this now. Since we also want project-level overrides, and I did not want to rely on values being None there as criterion for "don't override", I have added a somewhat awkward mechanism relying on convention. Pls have a look, seemed like a pragmatic choice to me

opcode81: The config changes have now been merged into main, so we can refactor this now.

MischaPanch: @opcode81 since my proposal for LSConfig was based on a misunderstanding, I am removing it.

I don't think a base class LSConfig makes sense. The defalt values will have to be different in SerenaConfig and ProjectConfig (None in the latter), so a solution based on a base class would still need to do something special with the common values - in post_init or otherwise. I propose no not introduce it now, a general solution can be done as part of #1052 , where the question will pop up again in an enhanced form

PR #960, opened 2026-01-19
feat(al): apply symbol name normalization in AL language server

SShadowS:
## Problem

The `find_symbol` tool fails to find AL (Business Central) symbols unless users explicitly set `substring_matching: true`. This is because AL Language Server returns symbol names in a verbose format that includes object type and ID:

- `Table 50000 "TEST Customer"` - but users search for `"TEST Customer"`
- `Codeunit 50000 CustomerMgt` - but users search for `"CustomerMgt"`

## Solution

Add a `display_name` field to symbols that provides a normalized name for matching, while preserving the original verbose name for display.

### 1. Follows Established Extensibility Pattern

The `UnifiedSymbolInformation` TypedDict already uses `NotRequired` fields for language-specific metadata that not all language servers provide:

```python
overload_idx: NotRequired[int]  # Only set by languages with function overloading
parent: NotRequired[...]        # Added by Serena, not part of LSP
```

Adding `display_name: NotRequired[str]` follows this exact pattern - an optional field that specific language servers can populate when their naming conventions differ from what users naturally search for.

### 2. Preserves Valuable Metadata

The alternative (modifying `name` directly in AL LS) would **permanently discard** the object type and ID:

| Approach | `name` field | Object type/ID |
|----------|-------------|----------------|
| Replace `name` | `TEST Customer` | Lost forever |
| Add `display_name` | `Table 50000 "TEST Customer"` | Preserved |

The full name is valuable context - it tells users they're looking at a Table vs a Page vs a Codeunit, and the object ID (50000) is often needed for debugging and AL development workflows.

### 3. Clean Separation of Concerns

- **AL Language Server** (`al_language_server.py`): Knows how to extract display names from AL's verbose format (AL-specific logic stays here)
- **Symbol infrastructure** (`symbol.py`): Knows how to match symbols using whatever name is appropriate (generic matching logic)

### 4. Zero Breaking Changes

Since `display_name` is `NotRequired` and falls back to `name` when absent:
- All existing language servers work unchanged
- All existing tests pass
- No changes to tool interfaces or user-facing behavior for non-AL languages

## Files Changed

| File | Change |
|------|--------|
| `ls_types.py` | Add `display_name` field |
| `al_language_server.py` | Extract and populate `display_name` |
| `symbol.py` | Use `display_name` for matching |
| `test_al_basic.py` | Add tests |

## Test Plan

- [x] Unit tests for `extract_al_display_name()` with various AL object formats
- [x] Integration tests verifying `find_symbol("TEST Customer")` matches `Table 50000 "TEST Customer"`
- [x] Python/Go regression tests pass unchanged

### DIFF
--- a/src/solidlsp/language_servers/al_language_server.py
+++ b/src/solidlsp/language_servers/al_language_server.py
@@ -1019,6 +1019,9 @@ class ALLanguageServer(SolidLanguageServer):
         Metadata (object type, ID) is available via the hover LSP method when using
         include_info=True in find_symbol.
         """
+        # Normalize path separators for cross-platform compatibility (backslash → forward slash)
+        relative_file_path = self._normalize_path(relative_file_path)
+
         # Get symbols from parent implementation
         document_symbols = super().request_document_symbols(relative_file_path, file_buffer=file_buffer)
 
@@ -1036,8 +1039,7 @@ class ALLanguageServer(SolidLanguageServer):
                     if start and "line" in start and "character" in start:
                         line = start["line"]
                         char = start["character"]
-                        normalized_path = self._normalize_path(relative_file_path)
-                        self._al_original_names[(normalized_path, line, char)] = original_name
+                        self._al_original_names[(relative_file_path, line, char)] = original_name
 
             symbol["name"] = normalized_name
 
@@ -1060,14 +1062,16 @@ class ALLanguageServer(SolidLanguageServer):
         When hovering over a symbol whose name was normalized, we prepend the original
         full name (e.g., 'Table 50000 "TEST Customer"') to the hover content.
         """
+        # Normalize path separators for cross-platform compatibility (backslash → forward slash)
+        relative_file_path = self._normalize_path(relative_file_path)
+
         hover = super().request_hover(relative_file_path, line, column)
 
         if hover is None:
             return None
 
-        # Check if we have an original name for this position (using normalized path for consistency)
-        normalized_path = self._normalize_path(relative_file_path)
-        original_name = self._al_original_names.get((normalized_path, line, column))
+        # Check if we have an original name for this position
+        original_name = self._al_original_names.get((relative_file_path, line, column))
 
         if original_name and "contents" in hover:
             contents = hover["contents"]


SShadowS:
Ohhh, and forgot, if you have a better solution for models not using substring_matching: true, please say so. 
AL is just an odd language with its own quirks.

opcode81:
The addition of `display_name` introduces a lack of clarity as to what really is the symbol's name (and it also has nothing to do with "display", actually, since you are using it for name path matching). 
**The symbol's name should be pure, without metadata - uniformly across all languages.**

Therefore, I would propose to handle this similarly to the Java LS, where we simply drop non-name parts.
https://github.com/oraios/serena/blob/a9b345ccfe85aeb93f5d306cb9f10fb82eca302d/src/solidlsp/language_servers/eclipse_jdtls.py#L862-L884

You already mentioned that valuable metadata would get lost if we did this. However, providing meta-data is not the job of the symbol's name, it is the job of the `hover` LSP method. The `find_symbol` tool already has an option `include_info`, whose job it is to provide that metadata. Therefore, I would propose that you check whether this correctly returns the metadata for AL, and if so, there is no reason at all to keep metadata in the symbol's name.

This is what `include_info=True` gives for a TypeScript method, for example:
```
[{"name_path": "SayHello/greet", "kind": "Method", "body_location": {"start_line": 12, "end_line": 14}, "relative_path": "main.ts", "info": "```typescript\n(method) SayHello.greet(person: Person): string\n```\nGreet a person by name.\n\n*@param* `person`"}]
```
Notice that 
 * `name_path` is just `SayHello/greet`
 * `info` provides metadata: 
    ```
    typescript
    (method) SayHello.greet(person: Person): string
    Greet a person by name
    
    *@param* `person`
    ```

SShadowS:
Thanks for the feedback. Will return fiercely. ;)

SShadowS:
@opcode81 Better now?

opcode81:
> Better now?

Yes, thanks!
But did you check whether include_info indeed returns the desired metadata, such that it isn't a problem for it not to be within the symbol name?

SShadowS:
> > Better now?
> 
> Yes, thanks! But did you check whether include_info indeed returns the desired metadata, such that it isn't a problem for it not to be within the symbol name?

Not all of it, the hover returns the object type (Table "TEST Customer") but not the object ID (50000). That's a limitation of the AL Language Server's hover implementation, and why I originally proposed the other solution. AL's quirks.

EDIT: Got me thinking. Could maybe inject it. Will test. Dont merge,

SShadowS:
I had to override hover, but it works now.

src/solidlsp/language_servers/al_language_server.py
```diff
@@ -957,3 +1006,71 @@ def set_active_workspace(self, workspace_uri: str | None = None) -> None:
...
+            normalized_name = extract_al_display_name(original_name)
+
+            # Store original name if it was normalized (for hover injection)
+            if original_name != normalized_name:
+                sel_range = symbol.get("selectionRange", {})
+                start = sel_range.get("start", {})
+                line = start.get("line", 0)
+                char = start.get("character", 0)
+                normalized_path = self._normalize_path(relative_file_path)
+                self._al_original_names[(normalized_path, line, char)] = original_name
```
opcode81: This could lead to problems if the defaults (0, 0) should ever apply.
You should store the value only if the position is present.

src/solidlsp/language_servers/al_language_server.py
```diff
@@ -22,6 +24,45 @@
...
+    r"^(?:Table|Page|Codeunit|Enum|Interface|Report|Query|XMLPort|PermissionSet|"
+    r"PermissionSetExtension|Profile|PageExtension|TableExtension|EnumExtension|"
+    r"PageCustomization|ReportExtension|ControlAddin|DotNetPackage)"  # Object type
+    r"(?:\s+\d+)?"  # Optional object ID
+    r"\s+"  # Required space before name
+    r'(?:"([^"]+)"|(\S+))$'  # Quoted name (group 1) or unquoted identifier (group 2)
+)
+
+
+def extract_al_display_name(full_name: str) -> str:
```
opcode81: Move this into the class (`@staticmethod`), and the pattern also; it is used only by the language server.

src/solidlsp/language_servers/al_language_server.py
SShadowS: Since AL LSP returns hierarchical DocumentSymbol format, it should always include selectionRange. But I understand the more defensive programming. Better safe than sorry. I will make sure it's never 0,0.

opcode81:
Looks good to me; minor comments only. 

Obviously, I don't know if your regex truly captures all cases, but I trust that you know AL well enough.

A minor caveat is that the hover function will return the added content only if document symbols were previously requested. Any global symbol search will fill the cache right up. I suppose it won't be an issue in practice (and no harm if it's not always there).

SShadowS:
I saw the tests. Working on it.

SShadowS:
Can you start checks again? All errors are not related to this PR anymore.

opcode81:
Thanks for your contribution, @SShadowS. Merged.

Issue #737, opened 2025-11-09
memory organization


keerpawan:
**Is your feature request related to a problem? Please describe.**
I have a monolith which has a lot of memories that we have build. It is hard to keep track or group them, except by  a file naming conventions.

**Describe the solution you'd like**
Call me old school, but can we have different folders in the memory folder, which will allow for grouping?

**Describe alternatives you've considered**
As mentioned, I'm going to enforce a file naming convention to create a virtual grouping.

**Additional context**

MischaPanch:
Yes, it makes sense. I was also thinking of introducing minimal frontmatter, like keywords and a summary that are always presented to Claude. This fits well with the hierarchical organization, where one could pass a subdirectory as kwarg to list_memories and always deliver the details of the top-level memories as result. I'd propose a structure like this

```
mem1
mem2
subdir
  subdir_mem1
```

`list_memories(category="")` should return

```
mem1: keywords=[...],summary=...
mem2: keywords=[...],summary=...
subdir: contains <n> memories
```

and `list_memories(category="subdir")` returns

```
subdir/subdir_mem1: keywords=[...],summary=...
```

Loading happens via qualified names, so e.g., with `load_memory("subdir/subdir_mem1"). Of course, hierarchy can be arbitrarily nested.

Optional frontmatter looks like

```
---
keywords: [....]
summary:  ...
```

@opcode81 wdyt? We could then later add  `find_memories(keyword)` if we wanted

simone-viozzi:
I have an agent to re-organize memories and check them for consistency with the code:

``````markdown
---
name: memory-review
description: Reviews and maintains Serena project memories so they stay code-grounded, timeless, and non-overlapping. Never changes code behavior; only adds TODO comments and updates memories.
tools: ['execute/getTerminalOutput', 'execute/runInTerminal', 'read/terminalLastCommand', 'read/terminalSelection', 'execute/createAndRunTask', 'execute/getTaskOutput', 'execute/runTask', 'edit/createFile', 'edit/createDirectory', 'edit/editFiles', 'search/listDirectory', 'read/readFile', 'search/codebase', 'serena/*', 'context7/*', 'tavily/*', 'todo', 'agent', 'execute/runTests', 'read/problems', 'search/changes', 'execute/testFailure']
---

You are a **Serena Memory Review Agent**.

[...]

PR #968, opened 2026-01-26
Feature: multiple folder support

EarthmanWeb:
Adds support for multiple folder to organize memories more effectively

eg:

 - `FEATURE_*` memories will be stored and retrieved from `.serena/memories/feature`

 - `REF_*` memories will be stored and retrieved from `.serena/memories/ref`

And so on...

Additional folders are configured in the mcp command line options**:

Sample config (working):

```json
{
  "mcpServers": {
    "serena-em": {
      "type": "stdio",
      "command": "uvx",
      "args": [
        "--refresh",
        "--from",
        "git+https://github.com/EarthmanWeb/serena@feature-multiplefoldersupport",
        "serena",
        "start-mcp-server",
        "--context",
        "ide-assistant",
        "--project",
        "./",

        "--additional-folders",
        ".serena/memories/arch,.serena/memories/dev,.serena/memories/feature",

        "--enable-web-dashboard=false"
      ],
      "env": {}
    }
  }
}
```

## ** Problem (open to suggestions)
Sometimes changes to `.mcp.json` inside a plugin, like adding new folders, doesn't take effect, even after restart.

## Cause
Claude caches MCP config at: `~/.claude/plugins/cache/EarthmanWeb/plugin_name/<version>/.mcp.json`

## Fix
```bash
rm -rf ~/.claude/plugins/cache/EarthmanWeb/
```
Then restart Claude Code.

### DIFF
--- a/src/serena/mcp.py
+++ b/src/serena/mcp.py
@@ -260,7 +260,12 @@ class SerenaMCPFactory:
                 mcp._tool_manager._tools[tool.get_name()] = mcp_tool
             log.info(f"Starting MCP server with {len(mcp._tool_manager._tools)} tools: {list(mcp._tool_manager._tools.keys())}")
 
-    def _create_serena_agent(self, serena_config: SerenaConfig, modes: ModeSelectionDefinition | None = None) -> SerenaAgent:
+    def _create_serena_agent(
+        self,
+        serena_config: SerenaConfig,
+        modes: ModeSelectionDefinition | None = None,
+        additional_memory_folders: list[str] | None = None,
+    ) -> SerenaAgent:
         return SerenaAgent(
             project=self.project,
             serena_config=serena_config,


opcode81:
This is a rather contrived solution. Why would you want to make the use of folders implicit for the LLM and make the user define a list of names for which memory name prefixes will map to folders - instead of just allowing the LLM to directly make use of folders and specify names like `feature/improved-foo`?

Note that we previously discussed this topic in #737. We had made a decision what to do, but then we didn't follow up on it.

opcode81:
@EarthmanWeb do you want to implement a PR based on our suggestion?

EarthmanWeb:
@opcode81  Allowing the LLM to create folders willy nilly is not a useful organizing tool.  It will simply make a mess.  Allowing the user to define the folders and then simply offering the option to use it or not seems to me much more of a useful addition.  It does not change existing behaviour for users who may not want a folder structure, and it adds an optional feature for users who want to organize memories as they wish.

opcode81:
> Allowing the LLM to create folders willy nilly is not a useful organizing tool. It will simply make a mess.

I don't think this is an informed assessment.
You can always control what you want the LLM to do specifically by using your own prompts to ensure this.

Your approach is contrived.

EarthmanWeb:
I look forward to seeing your approach when you get around to implementing
it.

My contrived solution is working exceptionally well for me at the moment.

T

opcode81:
A solution where the LLM 
* can use folders simply by specifying them in the memory name/path
* is instructed to use folders systematically (both centrally in Serena's system prompts and, optionally, via the user's own project-specific prompts)

is the way to go in our opinion.

Therefore, I am closing this PR.

I noticed that you have some unrelated commits as well. One commit relates to an improvement in the PHP LS. If this is solving a problem that you had, please consider issuing a separate PR (clearly explaining the issue being solved) with these improvements.

PR #969, opened 2026-01-26
Improve mode configurability and general config improvements

### LINKED ISSUE #926: Feature Request: Add `default_modes` configuration option
## Problem

Currently, modes can only be set at startup via `--mode` flags when launching the MCP server. There's no way to configure default modes in the configuration files, requiring users to manually switch modes each session.

## Proposed Solution

Add a `default_modes` option in `serena_config.yml` (global) and/or `project.yml` (per-project):

```yaml
default_modes:
  - planning
  - interactive
```

This would allow users to define which modes should be activated by default when Serena starts.

## Use Case

Users who prefer to start in planning mode (read-only analysis) before making edits would benefit from this configuration option. It would also allow different default modes per project based on the workflow requirements.

## Current Workaround

Users must either:
- Pass `--mode` flags at startup (not always possible depending on the client)
- Manually call `switch_modes` at the beginning of each session

opcode81:
Resolves #926 (adding three-tier mode configuration, differentiating base modes from default modes)

Additional changes:
* Project configurations that are incomplete when loaded are updated with defaults and re-saved. When re-saving project.yml files, missing comments are transferred from the template (ensuring correct comment association beforehand)
* The same applies to serena_config.yml. Resaving adds missing entries along with comments now. The convention was changed to use leading comments instead of trailing ones (which are difficult to manipulate), applying a conversion to users' files.

### DIFF
--- a/src/serena/config/serena_config.py
+++ b/src/serena/config/serena_config.py
@@ -784,9 +784,10 @@ class SerenaConfig(ToolInclusionDefinition, ModeSelectionDefinition, ToStringMix
         # transfer comments from the template file
         # NOTE: The template file now uses leading comments, but we previously used trailing comments,
         #       so we apply a conversion, which detects the old style and transforms it.
+        # For some keys, we force updates, because old comments are problematic/misleading.
         normalise_yaml_comments(commented_yaml, YamlCommentNormalisation.LEADING_WITH_CONVERSION_FROM_TRAILING)
         template_yaml = load_yaml(SERENA_CONFIG_TEMPLATE_FILE, comment_normalisation=YamlCommentNormalisation.LEADING)
-        transfer_missing_yaml_comments(template_yaml, commented_yaml, YamlCommentNormalisation.LEADING)
+        transfer_missing_yaml_comments(template_yaml, commented_yaml, YamlCommentNormalisation.LEADING, forced_update_keys=["projects"])
 
         save_yaml(self.config_file_path, commented_yaml)
 
--- a/src/serena/util/yaml.py
+++ b/src/serena/util/yaml.py
@@ -1,5 +1,6 @@
 import logging
 import os
+from collections.abc import Sequence
 from enum import Enum
 from typing import Any
 
@@ -201,11 +202,14 @@ def yaml_comment_entry_is_empty(comment_entry: Any) -> bool:
         return False
 
 
-def transfer_missing_yaml_comments_by_index(source: CommentedMap, target: CommentedMap, indices: list[int]) -> None:
+def transfer_missing_yaml_comments_by_index(
+    source: CommentedMap, target: CommentedMap, indices: list[int], forced_update_keys: Sequence[str] = ()
+) -> None:
     """
     :param source: the source, from which to transfer missing comments
     :param target: the target map, whose comments will be updated
     :param indices: list of comment indices to transfer
+    :param forced_update_keys: keys for which comments are always transferred, even if present in target
     """
     for key in target.keys():
         if key in source:
@@ -217,24 +221,28 @@ def transfer_missing_yaml_comments_by_index(source: CommentedMap, target: Commen
             if target_comment is None:
                 target_comment = [None] * 4
                 target.ca.items[key] = target_comment
-            # transfer missing comments at specified indices
+            # transfer comments at specified indices
             for index in indices:
-                if yaml_comment_entry_is_empty(target_comment[index]):
+                is_forced_update = key in forced_update_keys
+                if is_forced_update or yaml_comment_entry_is_empty(target_comment[index]):
                     target_comment[index] = source_comment[index]
 
 
-def transfer_missing_yaml_comments(source: CommentedMap, target: CommentedMap, comment_normalisation: YamlCommentNormalisation) -> None:
+def transfer_missing_yaml_comments(
+    source: CommentedMap, target: CommentedMap, comment_normalisation: YamlCommentNormalisation, forced_update_keys: Sequence[str] = ()
+) -> None:
     """
     Transfers missing comments from source to target YAML.
 
     :param source: the source, from which to transfer missing comments
     :param target: the target map, whose comments will be updated.
     :param comment_normalisation: the comment normalisation to assume; if NONE, no comments are transferred
+    :param forced_update_keys: keys for which comments are always transferred, even if present in target
     """
     match comment_normalisation:
         case YamlCommentNormalisation.NONE:
             pass
         case YamlCommentNormalisation.LEADING | YamlCommentNormalisation.LEADING_WITH_CONVERSION_FROM_TRAILING:
-            transfer_missing_yaml_comments_by_index(source, target, [ITEM_COMMENT_INDEX_BEFORE])
+            transfer_missing_yaml_comments_by_index(source, target, [ITEM_COMMENT_INDEX_BEFORE], forced_update_keys=forced_update_keys)
         case _:
             raise ValueError(f"Unhandled comment normalisation: {comment_normalisation}")


src/serena/config/serena_config.py
```diff
@@ -127,6 +128,12 @@ def is_fixed_tool_set(self) -> bool:
         return num_fixed > 0
 
 
+@dataclass
```
MischaPanch, line 131:
```suggestion
@dataclass(kw_only=True)
```

Should always be set for dataclasses which are meant to be inherited from

src/serena/config/serena_config.py
```diff
@@ -166,6 +173,12 @@ class ProjectConfig(ToolInclusionDefinition, ToStringMixin):
     encoding: str = DEFAULT_SOURCE_FILE_ENCODING
 
     SERENA_DEFAULT_PROJECT_FILE = "project.yml"
+    FIELDS_WITHOUT_DEFAULTS = {"project_name", "languages"}
```
MischaPanch, line 176: this seems brittle, I think we can extract the necessary info from the dataclass, right?

src/serena/util/yaml.py
```diff
@@ -0,0 +1,224 @@
...
+            if source_comment is None:
+                continue
+            target_comment = target.ca.items.get(key)
+            if target_comment is None and fully_missing_only:
+                # transfer full comment and we're done
+                target.ca.items[key] = source_comment
+            else:
+                # initialise target comment if needed
+                if target_comment is None:
+                    target_comment = [None] * 4
```
MischaPanch: this requires some explanation.

EDIT: see comment on this above

src/serena/util/yaml.py
```diff
@@ -0,0 +1,224 @@
...
+                    target.ca.items[key] = target_comment
+                # transfer missing comments at specified indices
+                for index in indices:
+                    if yaml_comment_entry_is_empty(target_comment[index]):
```
MischaPanch: I wouldn't expect the transfer to only happen if the target comment is empty from the function description.

Also, I would expect only the comments at the specified indices to be transferred, but instead

```
if target_comment is None and fully_missing_only:
                # transfer full comment and we're done
                target.ca.items[key] = source_comment
```

The role of the `indices` entry should be clarified in the docstring, maybe even adding a parameter `overwrite_existing_comments: bool = False`

src/serena/util/yaml.py
```diff
@@ -0,0 +1,224 @@
...
+    # NOTE: Normalisation for trailing comments was attempted but is extremely hard, because
+    #  it is difficult to position the comments properly after values, especially for complex values.
+
+
+DOC_COMMENT_INDEX_POST = 0
+DOC_COMMENT_INDEX_PRE = 1
+
+# item comment indices: (post key, pre key, post value, pre value)
+ITEM_COMMENT_INDEX_BEFORE = 1  # (pre-key; must be a list of CommentToken at this index)
+ITEM_COMMENT_INDEX_AFTER = 2  # (post-value; must be an instance of CommentToken at this index)
```
MischaPanch, line 60: The block comments should be turned to docstrings. It would also be helpful to add a word or two about what item_comment is, and what this post-pre key and value are all about. 

I saw this block comment below:

```
# Comments are supposed to be leading comments (i.e., before a key and associated with the key).
# When ruamel parses a YAML, however, comments belonging to a key may be stored as trailing
# comments of the previous key or as a document-level comment.
# Move them accordingly.
```

which explains it (but later). Maybe write a small module docstring?

src/serena/util/yaml.py
```diff
@@ -0,0 +1,224 @@
...
+            # Comments are supposed to be leading comments (i.e., before a key and associated with the key).
+            # When ruamel parses a YAML, however, comments belonging to a key may be stored as trailing
+            # comments of the previous key or as a document-level comment.
+            # Move them accordingly.
+            keys = list(commented_map.keys())
+            comment_items = commented_map.ca.items
+            doc_comment = commented_map.ca.comment
+            preceding_comment = None
+            for i, key in enumerate(keys):
+                current_comment = comment_items.get(key, [None] * 4)
```
MischaPanch, line 136: here again `[None] * 4`

Maybe introduce a self-documenting function `get_empty_ruamel_comment`?

src/serena/util/yaml.py
```diff
@@ -0,0 +1,224 @@
...
+    """
+    :param source: the source, from which to transfer missing comments
+    :param target: the target map, whose comments will be updated
+    :param indices: list of comment indices to transfer
+    :param fully_missing_only: if True, only transfer comments for keys that have no comment definition at all (newly added)
+    """
+    for key in target.keys():
+        if key in source:
+            source_comment = source.ca.items.get(key)
+            if source_comment is None:
```
MischaPanch, line 217: I thought source_comment can also be `[None]*4`, these should not be transfered, right?

src/serena/agent.py
```diff
@@ -167,14 +167,55 @@ def includes_name(self, tool_name: str) -> bool:
...
+        log.debug("Current mode selection: base_modes=%s, default_modes=%s", self._base_modes, self._default_modes)
+
+    def get_mode_names(self) -> Sequence[str]:
+        if self._active_mode_names is not None:
+            return self._active_mode_names
+        self._active_mode_names = []
+        if self._base_modes is not None:
+            self._active_mode_names.extend(self._base_modes)
+        if self._default_modes is not None:
+            self._active_mode_names.extend(self._default_modes)
```
MischaPanch: we should deduplicate

src/serena/agent.py
```diff
@@ -167,14 +167,55 @@ def includes_name(self, tool_name: str) -> bool:
         return tool_name in self._tool_names
 
 
+class ActiveModes:
+    def __init__(self) -> None:
+        self._base_modes: Sequence[str] | None = None
+        self._default_modes: Sequence[str] | None = None
+        self._active_mode_names: Sequence[str] | None = []
```
MischaPanch, line 174: shouldn't it be None after initialization? We even have a none check in get_mode_names. Same for active_modes below

src/serena/cli.py
```diff
@@ -302,24 +308,28 @@ def start_mcp_server(
         "modes",
         type=str,
         multiple=True,
-        default=DEFAULT_MODES,
+        default=(),
```
MischaPanch, line 321: this changes the default behavior, right?

src/serena/config/serena_config.py
opcode81: This is not brittle. It will never change.

src/serena/util/yaml.py
opcode81: This is just about internals of ruamel. I don't see the need to explain this.

MischaPanch: Well, for anyone ever touching this code (future us) an explanation would ne nice IMO

src/serena/config/serena_config.py
opcode81: No, this is incorrect. The subclass setting `kw_only=True` overrides the base class. It does not matter what the base class uses.

src/serena/util/yaml.py
opcode81: > I wouldn't expect the transfer to only happen if the target comment is empty

The function is literally called "transfer **missing** comments".
It intentionally keeps existing comments. The user may have added his own comments.

> I would expect only the comments at the specified indices to be transferred

The case you mention applies only to `fully_missing_only=True`, which I added to try a different approach at some point but which we don't use.
However, it only seems that the behaviour is different for this case, it isn't really. If the file has normalised comments, there will never be any other comments than at the respective index, so it still does what is "advertised".

I would be OK with removing this parameter though, since we don't use it and probably never will at this point.

src/serena/util/yaml.py
opcode81: The documentation is in the enum:
```python
class YamlCommentNormalisation(Enum):
    NONE = "none"
    """
    No comment normalisation is performed.
    Comments are kept as parsed by ruamel.yaml.
    """
    LEADING = "leading"
    """
    Document is assumed to have leading comments only, i.e. comments before keys, only full-line comments.
    This normalisation achieves that comments are properly associated with keys as leading comments.
    """
```

The function only applies the normalisation, so it is really self-explanatory: It applies the given normalisation to the given commented map:
```python
def normalise_yaml_comments(commented_map: CommentedMap, comment_normalisation: YamlCommentNormalisation) -> None:
```

The comment you quoted is not appropriate for high-level documentation as it only explains low-level information which the user should not be concerned with.

src/serena/cli.py
opcode81: No, it doesn't.
The base default is now defined in `SerenaConfig`:
https://github.com/oraios/serena/blob/b3247574f7804c72283cf4ad499d18130f10a485/src/serena/config/serena_config.py#L497

The logic is: 
* `SerenaConfig` defines the defaults
* `ProjectConfig` can override the defaults
* CLI can override them once more, and the override being `()` in this case results in no override. I wanted this to be None, but `click` didn't respect it, so I am checking for non-emptiness before applying an override.

src/serena/agent.py
opcode81: No, this is correct. If there are no base or default nodes, then the set of active modes is empty.
Setting it to `None` would only result in having to compute `[]` for no reason.

src/serena/util/yaml.py
opcode81: It doesn't matter. Since we only transfer if the target is empty, we essentially overwrite None with None, so who cares?

In practice, it is never None though, because we have comments for everything in our templates.

src/serena/agent.py
opcode81: You are right. I had a mental note on this and then forgot about it.

src/serena/util/yaml.py
MischaPanch: Then why have this block at all? Let's remove it

src/serena/agent.py
opcode81: Now changed.

src/serena/util/yaml.py
MischaPanch: Ok, let's remove this param. I may have been a bit confused about the index

src/serena/util/yaml.py
opcode81: What are you talking about? The if-block is about there not being a list. This is necessary.

src/serena/util/yaml.py
opcode81: For the record, I documented this already in one place. I don't feel the need to document it any further:

```
# item comment indices: (post key, pre key, post value, pre value)
```

So if you read the full file and understand what the code is actually doing (or read ruamel code), then it will be clear that we are defining the list for these four indices.

src/serena/util/yaml.py
MischaPanch: Ok, so it's to prevent index errors, makes sense

src/serena/util/yaml.py
opcode81: I removed it.

src/serena/util/yaml.py
opcode81: I have now added an additional docstring to the enum class itself:
```python
class YamlCommentNormalisation(Enum):
    """
    Defines a normalisation to be applied to the comment representation in a ruamel CommentedMap.

    Note that even though a YAML document may seem to consistently contain, for example, leading comments
    before a key only, ruamel may still parse some comments as trailing comments of the previous key
    or as document-level comments.
    The normalisations define ways to adjust the comment representation accordingly, clearly associating
    comments with the keys they belong to.
    """
```

I think this should make it sufficiently clear.

MischaPanch: Thanks, great!

MischaPanch: And sorry you had to go through this mess... Who knew yaml comment parsing would lead to such headache

src/serena/agent.py
```diff
@@ -290,22 +332,20 @@ def get_memory_log_handler() -> MemoryLogHandler:
...
             try:
-                self.activate_project_from_path_or_name(project)
+                self.activate_project_from_path_or_name(project, update_modes_and_tools=False)
             except Exception as e:
                 log.error(f"Error activating project '{project}' at startup: {e}", exc_info=e)
 
+        # update active modes and active tools (considering the active project, if any)
+        # declared attributes are set in the call to _update_active_modes_and_tools()
+        self._mode_overrides = modes
+        self._active_modes: ActiveModes
```
MischaPanch, line 345: @opcode81 the previous state of this was a bit weird to me and was the reason I hadn't seen the default modes at first glance. Before my commit on this branch, these attributes were set to "wrong" values but were then immediately overwritten by the call to `_update_active_modes_and_tools`. I have now changed this in the last commit, pls check if you are ok with this.

MischaPanch (changes requested):
After using Serena from this branch, my config got updated to:

```yaml
...

# MANAGED BY SERENA, KEEP AT THE BOTTOM OF THE YAML AND DON'T EDIT WITHOUT NEED
# The list of registered projects.
# To add a project, within a chat, simply ask Serena to "activate the project /path/to/project" or,
# if the project was previously added, "activate the project <project name>".
# By default, the project's name will be the name of the directory containing the project, but you may change it
# by editing the (auto-generated) project configuration file `/path/project/project/.serena/project.yml` file.
# If you want to maintain full control of the project configuration, create the project.yml file manually and then
# instruct Serena to activate the project by its path for first-time activation.
# NOTE: Make sure there are no name collisions in the names of registered projects.
projects:
- C:\Users\misch\Projects\oraios\diff_mcp
- C:\Users\misch\Projects\oraios\serena
- C:\Users\misch\Projects\oraios\serena-jetbrains-plugin
- C:\Users\misch\Projects\oraios\tianshou
fixed_tools: []
web_dashboard_listen_address: 127.0.0.1
jetbrains_plugin_server_address: 127.0.0.1
```

There are multiple issues:

1. the new entries don't have comments
2. the new entries are at the bottom, below `# MANAGED BY SERENA, KEEP AT THE BOTTOM OF THE YAML AND DON'T EDIT WITHOUT NEED`, which does not apply to them
3. `default_modes` and `base_modes` don't appear.

If you want, we can debug this together

MischaPanch:
@opcode81 I also added an improved documentation for modes in the cli, showing an empty tuple as the default is misleading

opcode81:
> After using Serena from this branch, my config got updated to ...

@MischaPanch, I don't think so. This is exactly what your config would have looked like *before* this branch!

opcode81:
I changed my config to look like yours at the end

```yaml
# MANAGED BY SERENA, KEEP AT THE BOTTOM OF THE YAML AND DON'T EDIT WITHOUT NEED
# The list of registered projects.
# To add a project, within a chat, simply ask Serena to "activate the project /path/to/project" or,
# if the project was previously added, "activate the project <project name>".
# By default, the project's name will be the name of the directory containing the project, but you may change it
# by editing the (auto-generated) project configuration file `/path/project/project/.serena/project.yml` file.
# If you want to maintain full control of the project configuration, create the project.yml file manually and then
# instruct Serena to activate the project by its path for first-time activation.
# NOTE: Make sure there are no name collisions in the names of registered projects.
projects:
- F:\Users\Dominik Jain\Dev\AI\ProbCog
- F:\Users\Dominik Jain\Dev\AI\tianshou
- F:\Users\Dominik Jain\Dev\Python\eval
fixed_tools: []
web_dashboard_listen_address: 127.0.0.1
jetbrains_plugin_server_address: 127.0.0.1
```

and after running this branch, it looked like this:

```yaml
# MANAGED BY SERENA, KEEP AT THE BOTTOM OF THE YAML AND DON'T EDIT WITHOUT NEED
# The list of registered projects.
# To add a project, within a chat, simply ask Serena to "activate the project /path/to/project" or,
# if the project was previously added, "activate the project <project name>".
# By default, the project's name will be the name of the directory containing the project, but you may change it
# by editing the (auto-generated) project configuration file `/path/project/project/.serena/project.yml` file.
# If you want to maintain full control of the project configuration, create the project.yml file manually and then
# instruct Serena to activate the project by its path for first-time activation.
# NOTE: Make sure there are no name collisions in the names of registered projects.
projects:
- F:\Users\Dominik Jain\Dev\AI\ProbCog
- F:\Users\Dominik Jain\Dev\AI\tianshou
- F:\Users\Dominik Jain\Dev\Python\eval

# fixed set of tools to use as the base tool set (if non-empty), replacing Serena's default set of tools.
# This cannot be combined with non-empty excluded_tools or included_optional_tools.
fixed_tools: []

# The address where the web dashboard will listen on
web_dashboard_listen_address: 127.0.0.1

# address where JetBrains plugin servers are running (only relevant when using the JetBrains language backend)
jetbrains_plugin_server_address: 127.0.0.1

# list of mode names to that are always to be included in the set of active modes
# The full set of modes to be activated is base_modes + default_modes.
# If this is undefined, no base modes are included.
# The project configuration (project.yml) may override this setting.
base_modes:

# list of mode names that are to be activated by default.
# The full set of modes to be activated is base_modes + default_modes.
# These modes can be overridden by the project configuration (project.yml) or through the CLI (--mode).
default_modes:
- interactive
- editing
```

I don't think there is anything to debug. 
I got the expected result.

Please run it again and ensure that you are not maintaining file locks on `serena_config.yml`.

You should see this:
```
INFO  2026-01-29 11:20:40,373 [MainThread] serena.config.serena_config:from_config_file:574 - Loading Serena configuration from C:\Users\Dominik Jain\.serena\serena_config.yml
INFO  2026-01-29 11:20:40,391 [MainThread] serena.config.serena_config:from_config_file:639 - Legacy configuration was migrated; re-saving configuration file
```

src/serena/cli.py
```diff
@@ -43,6 +43,9 @@
 log = logging.getLogger(__name__)
 
 _MAX_CONTENT_WIDTH = 100
+_MODES_EXPLANATION = """Built-in mode names or paths to custom mode YAMLs. 
+By default, the `default_modes` entry from your serena config (at `~/.serena/serena_config.yml`) will be used. 
+If not configured otherwise, this will result in ('interactive', 'editing')."""
```
opcode81: This is problematic. Nobody will remember that *concrete* defaults are mentioned here.
We either don't mention this or read the default from SerenaConfig.

opcode81: Even the sentence on serena_config is a bit misleading, to be honest, as mode configuration is now more complicated. We should probably link to the docs instead.

We should say that this "overrides the default modes configured in Serena's global configuration or the project configuration" and link to the docs for details.
We could mention the base default by extracting it from the dataclass.

opcode81: I have made the change.

PR #974, opened 2026-01-29
Add Apex language support

cx-alberto-simoes:
- Add APEX to Language enum in ls_config.py
- Add file extension matching for .cls, .trigger, .apex files
- Mark Apex as experimental language (not auto-detected)
- Create minimal ApexLanguageServer that supports basic file operations
- Add test repository with sample Apex files
- Add comprehensive test suite for Apex support
- Add apex pytest marker

Note: No LSP exists for Apex, so only basic file operations like read_file are supported. Advanced LSP features (symbol navigation, references, completions) are not available.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -316,6 +316,7 @@ markers = [
   "slow: tests that require additional Expert instances and have long startup times (~60-90s each)",
   "toml: language server running for TOML",
   "matlab: language server running for MATLAB (requires MATLAB R2021b+)",
+  "apex: Apex language support (no LSP, basic file operations only)",
 ]
 
 [tool.codespell]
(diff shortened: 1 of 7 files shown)


cx-alberto-simoes:
I am not sure if this make sense...

MischaPanch:
Hi @cx-alberto-simoes . Since no LS exists for Apex, I think this in fact does not make sense :). Closing

cx-alberto-simoes:
I understand your point, but serena readfile could still be of use.
Is there a way to register "files" that can be accessed by readfile? That would be useful for our system,
Thanks

opcode81:
All files (that are not ignored) can be accessed by readfile. You could just use any language to work on your files in a low-level fashion (e.g. markdown) - and you could disable the symbolic tools to not mislead the LLM.

wtolson:
Would support be possible now that his exists? https://github.com/apex-dev-tools/apex-ls

Issue #972, opened 2026-01-28
Support ccls LSP


ToKliar:
**Is your feature request related to a problem? Please describe.**
My project uses ccls as the LSP, but I can't use serena to index it. It seems serena defaults to clangd in cpp, but ccls provides better indexing and navigation capabilities than clangd for large-scale C++ projects.

**Describe the solution you'd like**

Add ccls support in serena, implement `CclsLanguageServer` like `ClangdLanguageServer`

MischaPanch:
Yes, that would be nice. We already have the possibility to have multiple LS for the same language. Moreover, clangd has caused multiple problems in the past and is still causing performance issues today.

We have a detailed explanation for how to add a new LS to Serena, would you like to give it a go? You can have a look how we have multiple LS for ruby and python to see how the same concept can be used for adding ccls without throwing away clangd

ToKliar:
Thanks for the positive feedback and the guidance!

I would love to give it a go, but unfortunately, I'm currently tied up with a very busy schedule and won't have enough time to properly implement and test the ccls language server right now.

Thanks again for maintaining such a great project!

turboslapper:
I would love to take a look at this! Anyone else please feel free to contribute alongside me, I'm not sure how far I will get so I don't want to block potential progress.

MischaPanch:
Go for it @turboslapper  :). It should be fairly straightforward, we already have the tests for it and recent extensions to abstractions make adding a new language server even easier than it has been. We also have detailed instructions for this specificallt in the checked in memories. With a bit of luck, a good agent could one-shot this

turboslapper:
Quick update: I’ve opened PR #975 which adds opt-in ccls support as an alternative C/C++ language server.

The implementation follows Serena’s existing multi-LS pattern (similar to Python and Ruby), introduces `cpp_ccls` as an experimental language option, and includes symbol, document symbol, and cross-file reference tests. This completes what was requested in this issue (to my understanding).

Regarding CI/CD: the failures currently showing do not appear to be caused by the ccls changes themselves.

- The ccls tests are skip-gated in CI because the pipeline does not have ccls installed. With ccls available locally, these tests run as expected.
- There is also a Nix language server failure in CI (`test_hover_information[nix]`), where nixd returns an error and hover information is `None`. This same Nix failure has appeared in other PRs and is unrelated to the ccls work.

Please let me know if anything in the PR should be adjusted.
@MischaPanch

opcode81:
PR #975 has been merged.

PR #975, opened 2026-01-31
CCLS Support added

turboslapper:
## Summary
This PR adds **opt-in support for the `ccls` C/C++ language server** as an alternative backend to the existing clangd integration.

Users can enable it by selecting the new language value **`cpp_ccls`** in project configuration. This is intended for larger C++ codebases where `ccls` can provide better indexing/navigation performance.

## What’s included
- Add `Language.CPP_CCLS = "cpp_ccls"` and mark it **experimental/opt-in** (not auto-detected).
- Add `CclsLanguageServer` implementation (launches system-installed `ccls` and initializes an LSP session).
- Wire `cpp_ccls` into:
  - file extension matching (same set as `cpp`)
  - language server class mapping
- Add pytest coverage for:
  - full symbol tree contains `add` + `main`
  - document symbols include `main`
  - cross-file references for `add` include `a.cpp`
  - stable reference results (order-insensitive compare)

## How to use
1. Install `ccls` and ensure it is on `PATH`
2. Set project language to `cpp_ccls`
3. Recommended: provide a `compile_commands.json` at repository root for best results

## Testing
- `pytest -m cpp`
  - Tests are skipped automatically if `ccls` is not installed.

## Related
Refs #972

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,17 +1,20 @@
 import logging
 import os
+import shutil as _sh
 from collections.abc import Iterator
 from contextlib import contextmanager
 from pathlib import Path
 from typing import Any
 
 import pytest
+
 try:
     from sensai.util.logging import configure
 except Exception:
     # Fallback if 'sensai' is not installed: basic logging configuration
     def configure(level: int = logging.INFO) -> None:
         logging.basicConfig(level=level)
+
 
 from serena.config.serena_config import SerenaPaths
 from serena.constants import SERENA_MANAGED_DIR_NAME
@@ -22,7 +25,6 @@ from solidlsp.ls_config import Language, LanguageServerConfig
 from solidlsp.settings import SolidLSPSettings
 
 from .solidlsp.clojure import is_clojure_cli_available
-import shutil as _sh
 
 configure(level=logging.INFO)
 
@@ -60,7 +62,11 @@ def _create_ls(
     gitignore_parser = GitignoreParser(str(repo_path))
     for spec in gitignore_parser.get_ignore_specs():
         ignored_paths.extend(spec.patterns)
-    config = LanguageServerConfig(code_language=language, ignored_paths=ignored_paths, trace_lsp_communication=trace_lsp_communication)
+    config = LanguageServerConfig(
+        code_language=language,
+        ignored_paths=ignored_paths,
+        trace_lsp_communication=trace_lsp_communication,
+    )
     effective_solidlsp_dir = solidlsp_dir if solidlsp_dir is not None else SerenaPaths().serena_user_home_dir
     return SolidLanguageServer.create(
         config,
--- a/test/solidlsp/cpp/test_cpp_basic.py
+++ b/test/solidlsp/cpp/test_cpp_basic.py
@@ -15,7 +15,6 @@ import pytest
 from solidlsp import SolidLanguageServer
 from solidlsp.ls_config import Language
 from solidlsp.ls_utils import SymbolUtils
-from test.conftest import language_tests_enabled
 
 
 def _clangd_available() -> bool:


test/solidlsp/cpp_ccls/test_cpp_ccls_basic.py
```diff
@@ -0,0 +1,70 @@
...
+
+        return _sh.which("ccls") is not None
+    except Exception:
+        return False
+
+
+@pytest.mark.cpp
+@pytest.mark.skipif(not _ccls_available(), reason="ccls not installed; skipping ccls tests")
+class TestCclsLanguageServer:
+    @pytest.mark.parametrize("language_server", [Language.CPP_CCLS], indirect=True)
```
MischaPanch: Instead of writing new tests and a whole new c-repository, pls use the existing c test_repo and tests and just include CPP_CCLS there. If you feel that tests are missing, extend the existing module

src/solidlsp/language_servers/ccls_language_server.py
```diff
@@ -0,0 +1,132 @@
...
+
+            ccls_path = shutil.which("ccls")
+            if not ccls_path:
+                raise FileNotFoundError(
+                    "ccls is not installed on your system.\n"
+                    + "Please install ccls using your system package manager:\n"
+                    + "  Ubuntu/Debian: sudo apt-get install ccls\n"
+                    + "  Fedora/RHEL: sudo dnf install ccls\n"
+                    + "  Arch Linux: sudo pacman -S ccls\n"
+                    + "See https://github.com/MaskRay/ccls for more details."
```
MischaPanch: There are no installation instructions here or in the project's wiki (which hasn't been updated for a long time), only build instructions. It seems like it can be installed with pacman and snap, maybe also apt-get. Pls add actual links to documentation where these installation options are added.

We should test it on at least one OS in CI (ideally on all three), so pls edit our CI pipeline to install ccls and make sure the tests run through.

For most LS, we install them on the fly. Are there any prebuilt binaries of ccls? Does it make sense to build it on the fly, or is this too cumbersome? Instructions for mac (seems like brew would work) and for windows are missing.

MischaPanch (changes requested):
Thanks for the contribution! See comments

test/solidlsp/cpp_ccls/test_cpp_ccls_basic.py
turboslapper: I don’t see an existing C/C++ test repository :(

- Looking at the repos, there’s no `cpp/` or `c/` folder listed.
- Similarly, there’s no `test/solidlsp/cpp/` test directory.
- The `Language` enum only has `CPP = "cpp"` (using `clangd`), with no existing tests for it.

### Should I:
1. **Create the C/C++ test infrastructure**, including:
   - `test/resources/repos/cpp/test_repo/`
   - `test/solidlsp/cpp/test_cpp_basic.py`  
   covering both **clangd** and **ccls**?

**OR**

2. **Is there an existing C++ test setup elsewhere** that I’m missing?

src/solidlsp/language_servers/ccls_language_server.py
turboslapper: Concern 1 (Documentation):
I wasn't quite sure if you wanted me to edit the `docs/` folder to include another md explaining how to use ccls, but I have added a section in the next commit within `ccls_language_server.py` that explains installation.

Concern 2 (CI Pipeline):
I've gone ahead and done that in the next commit, excluding windows for the moment (see below).

Concern 3 (Build issues):
There aren’t official prebuilt binaries for **ccls** (no release artifacts you can download and run), so “install on the fly” via fetching a binary isn’t really an option.

Building it on the fly in CI is likely too cumbersome/fragile because it requires an LLVM/Clang toolchain + headers/libs and a CMake build, which is slow and platform-dependent. It’s much more reliable to install via the system package manager where available.

Linux install:
```bash
sudo apt-get update
sudo apt-get install ccls
```

macOS install:
- `brew install ccls`

Windows:
- No official binaries; it generally requires building from source (typically via MSYS2/MinGW or a comparable toolchain with LLVM/Clang available). For CI, it’s reasonable to skip Windows for ccls unless we’re willing to maintain a dedicated build setup which I don't really think is viable.

test/solidlsp/cpp_ccls/test_cpp_ccls_basic.py
MischaPanch: Oh, you're right, sorry. I was sure we had one, don't know if my memory tricks me or if we deleted it at some point for some reason. 

Would be great if you created the necessary tests. In fact, you already did, right? The same test repo and the same tests should work both for clangd and ccls. To make the tests not fail when an LS is missing, the languages in the parameterized test could be constructed with an if-else clause (excluding ccls if not installed)

test/conftest.py
```diff
@@ -1,12 +1,20 @@
...
 from collections.abc import Iterator
 from contextlib import contextmanager
 from pathlib import Path
 from typing import Any
 
 import pytest
-from sensai.util.logging import configure
+
+try:
+    from sensai.util.logging import configure
```
MischaPanch, line 12: Not needed, sensai utils are a (pure Python) core dependency

test/solidlsp/cpp_ccls/test_cpp_ccls_basic.py
MischaPanch: I wrote this before looking at the the updated diff and was happy to see that you already implemented exactly what I had in mind :). Thanks;

src/solidlsp/language_servers/ccls_language_server.py
MischaPanch: Looks good, we can add more documentation later on main, what you added is enough for this PR

MischaPanch:
Clangd is misbehaving when finding references and a test failed. Maybe that's what I had in mind - I had a local testsuite set up but it never passed and I never prioritized this. I thought I had committed partial tests at some point, there's an open issue/PR on that. 

In any case, all minor follow-up changes can be addressed on main. This PR is good to be merged, thanks again for the contribution!

README.md
```diff
@@ -84,6 +84,8 @@ The underlying language servers are typically open-source projects (like Serena)
 With Serena's LSP library, we provide **support for over 30 programming languages**, including
 AL, Bash, C#, C/C++, Clojure, Dart, Elixir, Elm, Erlang, Fortran, Go, Groovy (partial support), Haskell, Java, Javascript, Julia, Kotlin, Lua, Markdown, MATLAB, Nix, Perl, PHP, PowerShell, Python, R, Ruby, Rust, Scala, Swift, TOML, TypeScript, YAML, and Zig.
 
+Note: For **C/C++**, Serena uses **clangd** by default and also supports **ccls** as an experimental, opt-in alternative (`cpp_ccls`). For best results, provide a `compile_commands.json` at the repository root.
```
opcode81, line 87: This does not belong here. The README shall only list supported languages. It shall not provide details on any of them.

docs/01-about/020_programming-languages.md
```diff
@@ -31,7 +31,8 @@ Some languages require additional installations or setup steps, as noted.
 * **Bash**
 * **C#**
 * **C/C++**  
-  (you may experience issues with finding references, we are working on it)
+  Default: clangd. Optional alternate: ccls (experimental, opt-in). 
+  For best results, provide a `compile_commands.json` at the repository root.
```
opcode81, line 35: This is misleading, because it implies that the file is useful for both clangd and ccls.

docs/01-about/020_programming-languages.md
```diff
@@ -31,7 +31,8 @@ Some languages require additional installations or setup steps, as noted.
 * **Bash**
 * **C#**
 * **C/C++**  
-  (you may experience issues with finding references, we are working on it)
+  Default: clangd. Optional alternate: ccls (experimental, opt-in).
```
opcode81, line 34: This could be clearer. I propose something more like this.

> We support two language servers for C/C++: clangd (set language to `cpp`) and ccls (set language to `cpp_ccls`).

Also, this is the place where we would mention any system-level installations that are needed or any other special requirements - making clear whether they apply to clangd or ccls.

opcode81:
@turboslapper no need to react to my comments. We will take care of them. Thanks again for your contribution.

test/solidlsp/cpp_ccls/test_cpp_ccls_basic.py
turboslapper: Great to hear, thank you for the review!

PR #983, opened 2026-02-02
Modernize C# language server: Switch to NuGet.org and official .NET install scripts

thomhurst:
## Summary

Modernizes C# language server dependency management with two key improvements:

1. **Switch to official NuGet.org packages** - Migrate from Azure DevOps NuGet feed to official roslyn-language-server packages
2. **Use Microsoft's official .NET install scripts** - Replace manual runtime downloads with dotnet-install scripts

### NuGet.org Migration
- Update package names from `Microsoft.CodeAnalysis.LanguageServer.*` to `roslyn-language-server.*`
- Upgrade version from 5.0.0-1.25329.6 to 5.5.0-2.26078.4 (latest Roslyn)
- Add direct URLs pointing to NuGet.org for all 6 platform variants
- Update extract paths from `content/LanguageServer/<platform>` to `tools/net10.0/<platform>`
- Simplify download logic by removing Azure feed service index lookup (~55 lines removed)
- Add 3 tests to verify NuGet.org download behavior

### .NET Install Scripts
- Remove 6 manual RuntimeDependency entries for .NET runtime (~42 lines)
- Replace manual tar/zip extraction with Microsoft's official install scripts
- Use `dotnet-install.sh` for Linux/macOS, `dotnet-install.ps1` for Windows
- Update to support .NET 10+ (required by new Roslyn packages)
- Reduce installation code from ~80 lines to ~60 lines
- Add 7 tests to verify install script behavior and version detection edge cases

### Symbol Name Normalization (LSP Spec Compliance)
Roslyn 5.5.0 returns symbol names with type annotations in the `name` field, violating LSP spec which expects just the identifier. Fixed by:
- Normalize symbol names to base form ("Name", "Add") for Serena's overload detection
- Cache original names at (file, line, char) for future rich information display
- Populate LSP spec's `detail` field with type/signature information (per spec)
- Process children symbols recursively

**How Roslyn 5.5.0 returns symbols:**
```json
[
  {"name": "Name : string", "kind": 7},
  {"name": "Add(int, int) : int", "kind": 6}
]
```

**After normalization:**
```json
[
  {"name": "Name", "detail": ": string", "kind": 7},
  {"name": "Add", "detail": "(int, int) : int", "kind": 6}
]
```

This preserves type information while maintaining LSP spec compliance and compatibility with Serena's overload detection system.

### Documentation Update
- Update class docstring to reflect new behavior
- Remove outdated DotNetRuntime override example
- Document .NET 10+ requirement and automatic installation

## Benefits

- More reliable - uses official Microsoft distribution channels
- Simpler codebase - net reduction of ~100 lines
- Better maintainability - official scripts handle platform detection, checksums, and dependencies
- Latest Roslyn with .NET 10 support
- LSP spec compliant symbol names with type information preserved in `detail` field
- Well-tested with 10 new tests covering both changes and edge cases

## Testing

All 24 C# tests pass:
- 14 existing tests (symbol operations, solution discovery, etc.)
- 3 new tests for NuGet.org package downloads
- 7 new tests for .NET install script behavior and version detection

Code is formatted, type-checked, and the language server downloads and initializes successfully.

### 1. Custom DotNetRuntime Overrides (Removed)

The `DotNetRuntime` entries in `runtime_dependencies` are no longer supported. This also removes the `dotnet_runtime_url` custom setting.

**Old behavior (no longer works):**
```yaml
ls_specific_settings:
  csharp:
    dotnet_runtime_url: "https://custom.url/dotnet-runtime.tar.gz"
    runtime_dependencies:
      - id: DotNetRuntime
        platform_id: linux-x64
        url: https://custom.url/dotnet-runtime.tar.gz
```

**Migration path:**

If you have a custom .NET installation, you have three options:

**Option 1: Add to PATH (Recommended)**
```bash
# Ensure your custom dotnet is in PATH
export PATH="/path/to/custom/dotnet:$PATH"
dotnet --version  # Should show 10.0 or higher
```

The language server will automatically detect and use system .NET 10+ from PATH.

**Option 2: Use DOTNET_ROOT environment variable**
```bash
export DOTNET_ROOT="/path/to/custom/dotnet"
export PATH="$DOTNET_ROOT:$PATH"
```

**Option 3: Let auto-installation handle it**
If no .NET 10+ is found in PATH, the official Microsoft install script will automatically download and install it to `~/.serena/language_servers/static/CSharpLanguageServer/dotnet-runtime-10.0/`

**Why this change?**
Microsoft's official install scripts provide better security (checksum verification, signature validation), automatic platform/architecture detection, proper dependency handling, and are officially maintained. This also simplifies the codebase significantly.

### 2. .NET Version Requirement

**Old**: Required .NET 9
**New**: Requires .NET 10+

System .NET 10 or higher will be auto-detected if available. If not found, .NET 10 will be auto-installed. The Roslyn language server packages now target .NET 10 (`tools/net10.0/`).

**Note**: .NET 9 is explicitly rejected and will trigger automatic .NET 10 installation since it's incompatible with the new Roslyn packages.

### 3. Symbol Name Format (Non-Breaking)

Symbol names are now normalized to comply with LSP spec. Type information is moved from `name` to `detail` field. This is transparent to Serena's symbol operations but improves correctness.

### 4. CSharpLanguageServer Overrides Still Supported

Language server package overrides continue to work:

```yaml
ls_specific_settings:
  csharp:
    runtime_dependencies:
      - id: CSharpLanguageServer
        platform_id: linux-x64
        url: "https://custom.url/roslyn-language-server.nupkg"
```

This allows you to use custom builds of the Roslyn language server if needed.

### DIFF
--- a/src/solidlsp/language_servers/csharp_language_server.py
+++ b/src/solidlsp/language_servers/csharp_language_server.py
@@ -334,11 +334,23 @@ class CSharpLanguageServer(SolidLanguageServer):
             """
             runtime_dependency_overrides = cast(list[dict[str, Any]], self._custom_settings.get("runtime_dependencies", []))
 
+            # Filter out deprecated DotNetRuntime overrides and warn users
+            filtered_overrides = []
+            for dep_override in runtime_dependency_overrides:
+                if dep_override.get("id") == "DotNetRuntime":
+                    log.warning(
+                        "The 'DotNetRuntime' runtime_dependencies override is no longer supported. "
+                        ".NET is now installed automatically via Microsoft's official install scripts. "
+                        "Please remove this override from your configuration."
+                    )
+                else:
+                    filtered_overrides.append(dep_override)
+
             log.debug("Resolving runtime dependencies")
 
             runtime_dependencies = RuntimeDependencyCollection(
                 _RUNTIME_DEPENDENCIES,
-                overrides=runtime_dependency_overrides,
+                overrides=filtered_overrides,
             )
 
             log.debug(


test/solidlsp/csharp/test_csharp_basic.py
```diff
@@ -56,7 +56,8 @@ def test_find_referencing_symbols(self, language_server: SolidLanguageServer) ->
         # Handle nested symbol structure
         symbol_list = symbols[0] if symbols and isinstance(symbols[0], list) else symbols
         for sym in symbol_list:
-            if sym.get("name") == "Add":
+            # New Roslyn format includes type annotations (e.g., "Add(int, int) : int")
+            if sym.get("name", "").startswith("Add"):
```
opcode81: Note that in SolidLSP, language servers must always return the pure name.

It would be wrong to return "Add(int, int): int" instead of "Add", as SolidLSP handles overloads internally - and Serena expects this to happen.
If this is not followed, then name path lookups will not work as expected and the UX will be broken.

Conceptually, type information is not supposed to be delivered as part of the symbol's name but via the `hover` operation, whose job it is to provide metadata.

See `EclipseJDTLS` for an example of an implementation that modifies symbol names to ensure this.
See `ALLanguageServer` for an example where information from symbol names is injected into the hover result.

opcode81: I now see that your commit [611aceb](https://github.com/oraios/serena/pull/983/commits/611acebe677cd8c97a3848d8ad412e5c7c3aa45d) probably already addresses this.

thomhurst: Forgot to update these after added normalisation into the roslyn language server logic. We parse and then normalise this information now.

And just added to `hover` - Can you confirm this is correct?

test/solidlsp/csharp/test_csharp_basic.py
```diff
@@ -112,21 +114,64 @@ def test_find_referencing_symbols_across_files(self, language_server: SolidLangu
...
+        file_path = os.path.join("Models", "Person.cs")
+
+        # Open the file first
+        language_server.open_file(file_path)
+
+        # Test 1: Hover over the Name property name (line 6, column 23 - on "Name")
+        # public string Name { get; set; }
+        hover_info = language_server.request_hover(file_path, 6, 23)
+
+        if hover_info is not None:
```
opcode81: This is quite a defensive test. Does the LS not guarantee that this is not None?
(below, too)

src/solidlsp/language_servers/csharp_language_server.py
```diff
@@ -220,6 +183,121 @@ def _create_dependency_provider(self) -> LanguageServerDependencyProvider:
...
+        symbols = super().request_document_symbols(relative_file_path, file_buffer)
+
+        # Normalize all symbols recursively
+        # DocumentSymbols.get_all_symbols_and_roots() returns tuple of (all_symbols, root_symbols)
+        all_symbols_tuple = symbols.get_all_symbols_and_roots()
+        all_symbols, _ = all_symbols_tuple
+
+        # Process each symbol (which is a UnifiedSymbolInformation dict)
+        for symbol in all_symbols:
+            self._normalize_symbol_name(symbol, relative_file_path)
```
opcode81, line 204: The assignments are redundant and `get_all_symbols_and_roots` is mainly provided for backward compatibility reasons.
Use `for symbol in symbols.iter_symbols()` instead.

test/solidlsp/csharp/test_csharp_basic.py
opcode81: It looks good, but I do have a question about the hover test (see respective comment).

test/solidlsp/csharp/test_csharp_basic.py
thomhurst: Changed to assert statements throughout

src/solidlsp/language_servers/csharp_language_server.py
thomhurst: Switched to `iter_symbols`

src/solidlsp/language_servers/csharp_language_server.py
opcode81: Since "DotNetRuntime" is no longer supported as an override, we should probably filter it out and if an override is there, log a warning informing the user about the changed mechanism.

opcode81:
We started documenting LS-specific settings [here](https://oraios.github.io/serena/02-usage/050_configuration.html#language-server-specific-settings). Do you want to add documentation as part of this PR, perhaps?

src/solidlsp/language_servers/csharp_language_server.py
thomhurst: Done 😄

thomhurst:
Sure thing! Added a C# section

opcode81:
Thanks for the additions! I will merge it tomorrow.

Issue #980, opened 2026-02-01
Project ignore pattern should be read asynchronously


opcode81:
Reading them synchronously can slow down project activation - and even be too slow for the MCP server to respond in time.

Issue #491, opened 2025-08-15
Feature Request: Add include paths configuration option


gn0rt0n:
**Problem:**
Currently, when using `.gitignore` integration, it's an all-or-nothing approach. I have my entire `vendor/` directory in `.gitignore` (which I want to keep), but I sometimes need Serena to analyze specific vendor subdirectories where I'm extending or modifying modules.

**Proposed Solution:**
include_paths: []
Add an "include paths" configuration option that works alongside `.gitignore` settings. This would allow users to:
- Honor `.gitignore` for broad exclusions (like `vendor/`)
- Selectively include specific paths that would otherwise be ignored (like `vendor/company/reference-module/`)

**Use Case:**
When extending third-party modules, I need Serena to analyze those specific vendor directories so it can properly compare classes and functions. Currently, my options are either:
1. Disable `.gitignore` integration entirely (includes too much)
2. Manually exclude everything except what I want (maintenance-heavy)

An include paths option would let me exclude `vendor/` broadly while including just the 1-2 vendor directories I'm actively working on.

**Expected Behavior:**
```json
{
  "respectGitignore": true,
  "includePaths": [
    "vendor/company-foo/module-to-reference",
    "vendor/company-bar/module-to-extend"
  ]
}
```

MischaPanch:
Yeah, I can see how that would make sense in several use cases. I'd be happy to review a PR on that :)

opcode81:
I agree that this could be useful, but this is difficult to do nicely, because if `vendor` is ignored, we don't even traverse it, so there will not even be an explicit check for `vendor/company-foo/module-to-reference` at all.

opcode81:
I don't think this is worth it, so I will close this as not planned. It might be possible to achieve most (if not all of this) through negative ignore patterns though (i.e. patterns starting with `!`).

PR #996, opened 2026-02-07
SerenaConfig: Add system-global ignored_paths for cross-project path exclusion

takeokunn:
## Summary

Adds `ignored_paths: list[str]` to `SerenaConfig` (`~/.serena/serena_config.yml`), allowing users to define gitignore-style exclusion patterns that apply across all projects.
Patterns are merged additively with each project's own `ignored_paths` and `.gitignore` patterns.

## Changes

- **SerenaConfig**: add `ignored_paths` field with empty default
- **RegisteredProject**: thread `global_ignored_paths` through all construction paths (`__init__`, `from_project_root`, `from_project_instance`, `get_project_instance`)
- **Project**: accept and prepend global patterns before `pathspec` compilation
- **CLI**: wire through `_create_project()`, `is_ignored_path`, `index_file`, `health_check`
- **Templates**: add entry to `serena_config.template.yml`, update `project.template.yml` comment

## Testing

13 new tests covering core merge behavior, `RegisteredProject` plumbing, backward compatibility, three-way merge with `.gitignore`, and `SerenaConfig` defaults.

## Related

- #491 — complementary feature (global exclusion vs selective inclusion)
- #980 / #981 — global patterns feed into the same `pathspec` compilation path

### DIFF
--- a/src/serena/config/serena_config.py
+++ b/src/serena/config/serena_config.py
@@ -402,7 +402,6 @@ class RegisteredProject(ToStringMixin):
         project_root: str,
         project_config: "ProjectConfig",
         project_instance: Optional["Project"] = None,
-        serena_config: "SerenaConfig | None" = None,
     ) -> None:
         """
         Represents a registered project in the Serena configuration.
@@ -410,12 +409,10 @@ class RegisteredProject(ToStringMixin):
         :param project_root: the root directory of the project
         :param project_config: the configuration of the project
         :param project_instance: an existing project instance (if already loaded)
-        :param serena_config: the SerenaConfig instance for global settings
         """
         self.project_root = Path(project_root).resolve()
         self.project_config = project_config
         self._project_instance = project_instance
-        self._serena_config = serena_config
 
     def _tostring_exclude_private(self) -> bool:
         return True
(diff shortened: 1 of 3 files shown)


src/serena/config/serena_config.py
```diff
@@ -397,16 +397,25 @@ def save(self, project_root: Path | str) -> None:
 
 
 class RegisteredProject(ToStringMixin):
-    def __init__(self, project_root: str, project_config: "ProjectConfig", project_instance: Optional["Project"] = None) -> None:
+    def __init__(
+        self,
+        project_root: str,
+        project_config: "ProjectConfig",
+        project_instance: Optional["Project"] = None,
+        global_ignored_paths: Sequence[str] = (),
```
opcode81: This does not belong here.
Instead, `get_project_instance` should have a new parameter `serena_config: SerenaConfig | None`.

Important: In all cases where we add `serena_config` as a new parameter, it shall not use a default. The decision to use None shall be explicit to avoid errors. (The only exception to this shall be the Project constructor).

src/serena/project.py
```diff
@@ -56,7 +57,13 @@ def delete_memory(self, name: str) -> str:
 
 
 class Project(ToStringMixin):
-    def __init__(self, project_root: str, project_config: ProjectConfig, is_newly_created: bool = False):
+    def __init__(
+        self,
+        project_root: str,
+        project_config: ProjectConfig,
+        is_newly_created: bool = False,
+        global_ignored_paths: Sequence[str] = (),
```
opcode81: Let's use `serena_config: SerenaConfig | None = None` instead.

This is more general and will probably come in handy in the near future.

src/serena/project.py
```diff
@@ -115,12 +127,17 @@ def project_name(self) -> str:
         return self.project_config.project_name
 
     @classmethod
-    def load(cls, project_root: str | Path, autogenerate: bool = True) -> "Project":
+    def load(
+        cls,
+        project_root: str | Path,
+        autogenerate: bool = True,
+        global_ignored_paths: Sequence[str] = (),
```
opcode81: This should use `serena_config: SerenaConfig | None` without a default.

src/serena/config/serena_config.py
takeokunn: f2beb3bb16c725051cb64068a1ccdf188366e296 Fixed

src/serena/project.py
takeokunn: 43902782fa47f5fd107fca812d7058154ef557d5 Fixed

src/serena/project.py
takeokunn: 0025402f37057c7fd41812249bff79bcd01961a9 Fixed

takeokunn:
@opcode81 
Revisions are done. Could you please review?

opcode81:
Looks great now. Thanks!

PR #1002, opened 2026-02-07
Add Phpactor as alternative PHP language server

### LINKED ISSUE #710: Support for Devsense as alternative to Intelephense for PHP
Similar to how we support multiple LS for python, typescript and ruby already

takeokunn:
## Summary

Add Phpactor as an alternative PHP language server alongside the existing Intelephense implementation.
Phpactor is fully open-source (MIT) and requires PHP 8.1+ on the system instead of Node.js.

Resolves #710

## Changes

- Add `PHP_PHPACTOR` enum to `Language` in `ls_config.py` (marked experimental)
- Create `PhpactorServer` class with PHAR auto-download via `LanguageServerDependencyProviderSinglePath`
- Share `*.php` file matcher between `PHP` and `PHP_PHPACTOR`
- Add 6 tests for definition/reference operations (reusing existing PHP test repo)
- Update documentation: programming languages page, CHANGELOG, project template

## Testing

- `uv run poe format` — passes clean
- `uv run poe type-check` — 0 issues in 105 source files
- `uv run poe test -m php` — all 8 existing Intelephense tests pass
- 6 new phpactor tests collected and parametrized correctly

### DIFF
--- a/src/solidlsp/language_servers/phpactor.py
+++ b/src/solidlsp/language_servers/phpactor.py
@@ -159,4 +159,3 @@ class PhpactorServer(SolidLanguageServer):
         assert init_response["capabilities"].get("definitionProvider"), "Phpactor did not advertise definition support"
 
         self.server.notify.initialized({})
-        self.completions_available.set()
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -58,16 +58,19 @@ jobs:
         uses: actions/cache@v3
         with:
           path: .venv
-          key: uv-venv-${{ runner.os }}-${{ matrix.python-version }}-${{ hashFiles('uv.lock') }}
+          key: uv-venv-${{ runner.os }}-${{ matrix.python-version }}-lock-${{ hashFiles('uv.lock') }}
       - name: Create virtual environment
         shell: bash
         run: |
           if [ ! -d ".venv" ]; then
             uv venv
           fi
-      - name: Install dependencies
+      - name: Install Python environment
         shell: bash
-        run: uv pip install -e ".[dev]"
+        run: uv sync --extra dev --locked
+      - name: List Python dependencies
+        shell: bash
+        run: uv pip list
       - name: Check formatting
         shell: bash
         run: uv run poe lint
@@ -110,6 +113,24 @@ jobs:
         uses: DeLaGuardo/setup-clojure@13.4
         with:
           cli: latest
+      - name: Install ccls (C/C++ Language Server)
+        shell: bash
+        run: |
+          if [[ "${{ runner.os }}" == "Linux" ]]; then
+            sudo apt-get update
+            sudo apt-get install -y ccls
+          elif [[ "${{ runner.os }}" == "macOS" ]]; then
+            brew install ccls
+          elif [[ "${{ runner.os }}" == "Windows" ]]; then
+            choco install ccls -y
+          fi
+          # Verify installation
+          if command -v ccls &> /dev/null; then
+            echo "ccls installed: $(ccls --version 2>&1 | head -1)"
+          else
+            echo "ERROR: ccls installation failed"
+            exit 1
+          fi
       - name: Setup Java (for JVM based languages)
         uses: actions/setup-java@v4
         with:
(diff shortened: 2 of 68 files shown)


opcode81:
This PR was not up-to-date with the main branch; I updated it.

src/solidlsp/language_servers/phpactor.py
```diff
@@ -0,0 +1,147 @@
...
+    class DependencyProvider(LanguageServerDependencyProviderSinglePath):
+        def _get_or_install_core_dependency(self) -> str:
+            """
+            Setup runtime dependencies for Phpactor and return the path to the PHAR file.
+            """
+            # Verify PHP is installed
+            php_path = shutil.which("php")
+            assert (
+                php_path is not None
+            ), "PHP is not installed or not found in PATH. Phpactor requires PHP 8.1+. Please install PHP and try again."
```
opcode81, line 51: The PHP version is not checked. Will the user get a meaningful error in case the version is too old?

test/resources/repos/php_phpactor
```diff
@@ -0,0 +1 @@
+php
```
opcode81: This link is not necessary.
As far as language repositories are concerned, we need only one repository per language, not per language server.

test/solidlsp/php/test_php_phpactor.py
```diff
@@ -0,0 +1,111 @@
+from pathlib import Path
+
+import pytest
+
+from solidlsp import SolidLanguageServer
+from solidlsp.ls_config import Language
+
+
+@pytest.mark.php
+class TestPhpPhpactorLanguageServer:
```
opcode81: This test should be merged with `TestPhpLanguageServer`, which could be renamed to `TestPhpLanguageServers`.

* You can use parametrize to specify more than one Language, i.e. both `PHP` and `PHP_PHPACTOR` for `language_server`.
* Where the new LS has different capabilities, we can add additional methods or internally differentiate.

src/solidlsp/language_servers/phpactor.py
takeokunn: Done. Added PHP version validation in d39967dd678c405e6cc2826539587b02b1ef8121 — `PhpactorServer` now checks that PHP ≥ 8.1 is available at startup and raises a clear `LanguageServerError` with details if the version is too old or PHP is not found.

test/resources/repos/php_phpactor
takeokunn: Done. Removed the `php_phpactor` symlink in c972380d0a11c291c874b6c68a4b6ddb3eea1ff2. Instead, added a repo alias mapping in `conftest.py` (`_LANGUAGE_REPO_ALIASES`) so that `PHP_PHPACTOR` reuses the existing `php` test repository directly. This approach also aligns with the pattern used for `CPP_CCLS`, `PYTHON_JEDI`, and `RUBY_SOLARGRAPH`.

test/solidlsp/php/test_php_phpactor.py
takeokunn: Done. Merged both test classes into `TestPhpLanguageServers` in 46fabd69f4cdb6bc2adb75d0c46dac8ddb495653. Tests are now parametrized over both `Language.PHP` and `Language.PHP_PHPACTOR`, with internal differentiation where capabilities differ (e.g. find_referencing_symbols tests are skipped for Phpactor since it doesn't support cross-file references yet).

opcode81:
Thank you very much for the additions, @takeokunn! This looks great now. Merging.

takeokunn:
Thanks review!

MischaPanch:
This introduced hangs on windows in CI, leading to test failures after 360 mins

https://github.com/oraios/serena/actions/runs/21905032710/job/63242953771

devnix:
How can we enable PHPActor instead of Intelephense?

MischaPanch:
@devnix  specify `php_phpactor` as your language (instead of php) in your project.yml (as stated in the docs).

As I mentioned, it doesn't seem to work well on windows

devnix:
Sorry @MischaPanch, I missed that the docs were updated. I wasn't expecting a new language but rather some kind of flag. Does it make any sense to allow configuring `php` and `php_phpactor` at the same time? why not calling them `php_intelephense` and `php_phpactor`?

opcode81:
No, not at the same time.
Intelephense is the default for PHP, so it just uses `php`; it was assumed to remain the only language server we have for PHP.
Phpactor is a new addition, so it has to be enabled explicitly and has a longer name.

MischaPanch:
@devnix if you are unhappy with the performance of either php LS, you could consider opening a PR for supporting devsense - we have an open issue on that #710

Issue #994, opened 2026-02-06
Semantic Memory Search


bguidolim:
## Context
Serena MCP currently provides filesystem-based memory operations (`list_memories`, `read_memory`, `write_memory`). While this works well for small projects, as memory files grow, it becomes challenging for LLMs to:

1. **Discover relevant context** without reading every file
2. **Navigate large memory collections** efficiently
3. **Find information across multiple memory files** without explicit file listing

## Problem
Current workflow:
1. LLM calls `list_memories` → gets list of files
2. LLM guesses which files might contain relevant info
3. LLM calls `read_memory` for each candidate
4. **Result:** Multiple tool calls, wasted tokens, slower responses
5. 
## Proposed Solution

Add **semantic search** capability to Serena MCP:

```typescript
// New tool
{
  name: "search_memories",
  description: "Semantic search across all memory files in workspace",
  inputSchema: {
    query: string,        // Natural language search query
    maxResults?: number   // Default: 5
  }
}
```

**Returns:**
```json
[
  {
    "path": ".serena/memories/auth-flow.md",
    "snippet": "...relevant excerpt...",
    "score": 0.89
  }
]
```

LLM can then call `read_memory` only for relevant files.

### Option 1: Embedding-based (Higher Quality)

- Use embedding models (OpenAI, Ollama, etc.)
- Store embeddings in SQLite with sqlite-vec
- Chunk memory files for better granularity
- **Pros:** Best search quality, handles large memory bases
- **Cons:** Requires embedding provider configuration

### Option 2: Hybrid (Fallback)

- Primary: Embedding-based search (if configured)
- Fallback: Simple text search (grep/regex) if no embeddings
- **Pros:** Works out-of-the-box, degrades gracefully
- **Cons:** More complex implementation

### Option 3: Plugin Architecture

- Keep Serena minimal
- Allow optional "search plugin" via config
- **Pros:** Clean separation, no bloat
- **Cons:** Users need to configure separately

## Reference Implementation

Similar functionality exists in:
- [@arabold/docs-mcp-server](https://github.com/arabold/docs-mcp-server) (embedding-based doc search)
- OpenClaw (local embedding models via node-llama-cpp)

## Questions for Maintainers

1. Is this something you'd consider adding to Serena?
2. Would you prefer Option 1 (embedding-based) or Option 2 (hybrid)?
3. If I submit a PR, are there any architectural constraints or preferences?
[...]

PR #1007, opened 2026-02-09
feat: Add global memories as reserved topic

### LINKED ISSUE #1055: Memories with directory grouping and optional frontmatter
Currently memory structure is flat and `list_memories` just returns the memory names. This is impractical for projects with many memories or for memories where the content isn't easily represented by just the file name.

The solution @opcode81 and I agreed on now is the following:

1. Memories can be in subdirectories, the directory enters the memory name. So we can have the list of names like `global_mem, auth/login_logic, auth/token_verification`
2. Write memory can receive a name with a "/" and will then write to the respective subdir
3. Memories can have frontmatter, which is at the top of the markdown file, between two lines with `---`. The frontmatter for now will only contain `summary: ....` or be empty or missing.
5. `list_memories` returns memory names and any non-empty frontmatter that was found as answer to the agent.
6. `list_memories` can receive a path (optionally, by default "", and the kwarg should be called something like `topic`) and then only lists the memories in that subdir
7. The tool docstrings are adjusted to make the agent group relevant memories together in subdirs (semantically into topics, arbitrary nesting) and write frontmatter if necessary. The agent is also encouraged to structure existing memories and to add frontmatter to them if appropriate.

I think for restructuring existing memories, we should add a new tool `RenameMemoryTool`, a renaming to something with "/" will then move the file - @opcode81 do you agree?

butterflysky-ai:
## Summary

Adds cross-project (global) memories using `global/` as a reserved topic name, building on the topic grouping from #1058. Writing `global/my_memory` transparently routes to `~/.serena/memories/global/my_memory.md`.

No new parameters — just use the `global/` prefix with existing memory tools.

Includes an `edit_global_memories` config option (default: `true`) to protect global memories from accidental agent modification.

Closes #1055

## Changes

- **`MemoriesManager`** (`project.py`): Routes `global/*` names to `~/.serena/memories/global/`, adds `edit_memory()` method for editing memories outside project root
- **Memory tools** (`memory_tools.py`): Edit guard on write/delete/edit/rename, `EditMemoryTool` uses `MemoriesManager.edit_memory()` instead of `ReplaceContentTool` (which fails for paths outside project root)
- **`SerenaConfig`** (`serena_config.py`): `edit_global_memories: bool = True`
- **Onboarding check** (`workflow_tools.py`): Filters global memories (they don't indicate project onboarding)
- **Dashboard** (`dashboard.py`): Edit guard on save/delete endpoints
- **Flaky test fix** (`test_ls_common.py`): Skip `test_open_file_cache_invalidate` in all CI (was only skipped on Windows, also fails on ubuntu-latest)

## Design decisions

- `global/` is a reserved topic — bare `"global"` (no sub-name) raises `ValueError`
- `rename_memory("global/foo", "bar")` moves between scopes (intentional)
- `edit_global_memories: false` blocks writes but allows reads
- No `scope` parameter on tools — the name prefix is the only routing mechanism

### DIFF
--- a/src/serena/project.py
+++ b/src/serena/project.py
@@ -96,45 +96,31 @@ class MemoriesManager:
         When topic is empty, returns both project and global memories.
         Global memories are prefixed with "global/".
         """
-        memories: list[str] = []
 
-        include_project = True
-        include_global = self._global_memory_dir is not None
+        def _collect(search_dir: Path, base_dir: Path, prefix: str = "") -> list[str]:
+            if not search_dir.exists():
+                return []
+            results = []
+            for md_file in search_dir.rglob("*.md"):
+                rel = str(md_file.relative_to(base_dir).with_suffix("")).replace(os.sep, "/")
+                results.append(prefix + rel)
+            return results
+
+        memories: list[str] = []
 
         if topic:
             if self._is_global(topic):
-                include_project = False
-                # Search within global dir, stripping "global/" or "global" prefix
-                if topic == self.GLOBAL_TOPIC:
-                    global_search_dir = self._global_memory_dir
-                else:
-                    global_sub = topic[len(self.GLOBAL_TOPIC) + 1 :]
-                    global_search_dir = self._global_memory_dir / global_sub.replace("/", os.sep) if self._global_memory_dir else None
-                if global_search_dir and global_search_dir.exists():
-                    for md_file in global_search_dir.rglob("*.md"):
-                        rel_path = md_file.relative_to(self._global_memory_dir)  # type: ignore[arg-type]
-                        name = self.GLOBAL_TOPIC + "/" + str(rel_path.with_suffix("")).replace(os.sep, "/")
-                        memories.append(name)
+                if self._global_memory_dir is not None:
+                    sub = topic[len(self.GLOBAL_TOPIC) + 1 :] if topic != self.GLOBAL_TOPIC else ""
+                    search_dir = self._global_memory_dir / sub.replace("/", os.sep) if sub else self._global_memory_dir
+                    memories = _collect(search_dir, self._global_memory_dir, self.GLOBAL_TOPIC + "/")
             else:
-                include_global = False
                 search_dir = self._memory_dir / topic.replace("/", os.sep)
-                if search_dir.exists():
-                    for md_file in search_dir.rglob("*.md"):
-                        rel_path = md_file.relative_to(self._memory_dir)
-                        name = str(rel_path.with_suffix("")).replace(os.sep, "/")
-                        memories.append(name)
+                memories = _collect(search_dir, self._memory_dir)
         else:
-            # No topic filter — list everything
-            if include_project and self._memory_dir.exists():
-                for md_file in self._memory_dir.rglob("*.md"):
-                    rel_path = md_file.relative_to(self._memory_dir)
-                    name = str(rel_path.with_suffix("")).replace(os.sep, "/")
-                    memories.append(name)
-            if include_global and self._global_memory_dir is not None and self._global_memory_dir.exists():
-                for md_file in self._global_memory_dir.rglob("*.md"):
-                    rel_path = md_file.relative_to(self._global_memory_dir)
-                    name = self.GLOBAL_TOPIC + "/" + str(rel_path.with_suffix("")).replace(os.sep, "/")
-                    memories.append(name)
+            memories = _collect(self._memory_dir, self._memory_dir)
+            if self._global_memory_dir is not None:
+                memories.extend(_collect(self._global_memory_dir, self._global_memory_dir, self.GLOBAL_TOPIC + "/"))
 
         return sorted(memories)
 


butterflysky:
failing tests are also failing on main, no new failures introduced by this PR. I'm interested in trying to help with those as a separate issue, but don't have time to get to them in the next few days.

do let me know if there are questions about this PR and the feature request issue i created. i'm using this change locally to help me streamline my workflow as i hop between working on different projects in parallel.

i store information about my specific infrastructure, environment, and workflow requirements there, as well as instructions about how to manage my task tracker, project planning, wins tracking. saves me a lot of time, keeps me organized. highly recommend a similar approach to others.

MischaPanch:
Thank you @butterflysky. Global memories are a good idea, let's bring this to Serena! There are some higher priority issues we have to address first, we will review this soon and get back to you.

butterflysky:
Thanks @MischaPanch - just want to say I'm finding Serena very helpful, so thank you for it!

butterflysky:
rebased and addressed merge conflicts

MischaPanch:
@butterflysky it probably makes sense to postpone this until #1055 is finished

butterflysky:
> @butterflysky it probably makes sense to postpone this until #1055 is finished

totally makes sense, thanks for linking the issue. between #1055, this, and possibly #994 if that gets implemented (i would absolutely love to have semantic search capability for memories), Serena's memories are going to save a lot of back-and-forth clarifying intent to the LLMs.

I'll keep an eye on the upstream issue and any future developments on the memories front, then keep this diff up-to-date and merge-ready - to the extent it remains relevant.

butterflysky:
@butterflysky-ai read the comments here, add a follow-up task to our tracking repo for checking on the status of #1055 and the direction of #994 and keeping this PR up-to-date, please and thanks.

butterflysky-ai:
On it! I've read through the full thread. Here's my summary:

- **#1055** (Memories with directory grouping and optional frontmatter) is the upstream blocker — MischaPanch wants that landed first since it restructures how memories work, which will affect our global memories implementation.
- **#994** (Semantic Memory Search) would be a great complement — vector/semantic search over memories instead of requiring explicit file listing.
- This PR stays relevant regardless — global memories are orthogonal to both features, but may need rebasing/adaptation once #1055 lands.

I'll create a tracking task to monitor both upstream issues and keep this PR merge-ready. 👍

MischaPanch:
@butterflysky since #1058 has now been merged, we can go ahead and finalize this. The implementation will be very easy - make `global` a reserved topic. For this special topic, we don't save to the project's `.serena/memories` but instead to user's home `~/.serena/memories/global`. The global topic is then always made available to each project.

For the LLM any mentions of files have already been removed, so adding to global memories is as easy as writing or moving to the `global` topic. The semantics of `global` should be explained. Since `global` memories are not under version control, one might instruct the model to double-check any "destructive" action with the user. One could also forbid editing of global memories altogether, possibly adding a config option `edit_global_memories: True` or something like that. WDYT?

@opcode81 FYI

opcode81:
I don't think the configuration option is needed.

MischaPanch:
I prophecy that users will be annoyed about their forever-lost global memories that got deleted by some stupid agent

butterflysky-ai:
Reworked per @MischaPanch's feedback — replaced the `scope` parameter approach with `global/` as a reserved topic name. This builds directly on #1058's topic grouping, so `global/my_memory` just routes to `~/.serena/memories/global/my_memory.md` transparently.

Also added the `edit_global_memories` config option (default `true`) as discussed, and included a small fix for the flaky `test_open_file_cache_invalidate` test that's been failing on ubuntu-latest CI.

Force-pushed with 2 clean commits on top of current `main`.

butterflysky:
and depending on what the two of you decide @MischaPanch and @opcode81, we can leave in or remove the config option - i interpreted it as a write guard on the global memories, in case people are crafting them by hand and want them read-only in some scenarios.

src/serena/project.py
```diff
@@ -66,25 +93,49 @@ def save_memory(self, name: str, content: str) -> str:
...
+            if include_project and self._memory_dir.exists():
+                for md_file in self._memory_dir.rglob("*.md"):
+                    rel_path = md_file.relative_to(self._memory_dir)
+                    name = str(rel_path.with_suffix("")).replace(os.sep, "/")
+                    memories.append(name)
+            if include_global and self._global_memory_dir is not None and self._global_memory_dir.exists():
+                for md_file in self._global_memory_dir.rglob("*.md"):
+                    rel_path = md_file.relative_to(self._global_memory_dir)
+                    name = self.GLOBAL_TOPIC + "/" + str(rel_path.with_suffix("")).replace(os.sep, "/")
+                    memories.append(name)
```
butterflysky: @butterflysky-ai DRY, fix

src/serena/project.py
```diff
@@ -113,6 +151,39 @@ def rename_memory(self, old_name: str, new_name: str) -> str:
...
+            new_content, count = pattern.subn(repl, content)
+            if count == 0:
+                raise ValueError(f"The regex pattern did not match anything in memory {name}.")
+        else:
+            raise ValueError(f"Invalid mode: {mode}. Must be 'literal' or 'regex'.")
+
+        with open(memory_file_path, "w", encoding=self._encoding) as f:
+            f.write(new_content)
+        return f"Memory {name} edited successfully."
+
```
butterflysky, line 186: @butterflysky-ai why is this net new? explain why this route is better than updating the existing EditMemoryTool?

butterflysky-ai: **tl;dr**: The upstream `EditMemoryTool` routes through `ReplaceContentTool`, which calls `validate_relative_path` → `is_path_in_project` — this rejects any path outside the project root. Global memories live in `~/.serena/memories/global/`, so the `.relative_to(project_root)` call raises `ValueError`.

**Why not fix `ReplaceContentTool` instead?** It's a general-purpose code editing tool with safety boundaries (project-root containment, ignore-spec checks, `EditedFileContext`, code editor pipeline). Weakening those for memory files felt wrong. Memory files are markdown, not source code — they don't need the code editor machinery.

**Why `MemoriesManager` is the right home**: `save_memory`, `load_memory`, `delete_memory`, `rename_memory` all do their own file I/O directly. `edit_memory` is the same pattern — the manager owns memory file paths and encoding. The upstream `EditMemoryTool` routing through `ReplaceContentTool` was the odd one out.

Open to other approaches if the maintainers prefer — e.g. adding an absolute-path mode to `ReplaceContentTool`, or having `EditMemoryTool` do inline file I/O without a manager method. But this seemed like the least invasive option.

butterflysky:
should be ready for review. i'm not sure what to do about the lsp test failures, they seem unreliable at the moment, doesn't seem like it's always the same failure

src/serena/project.py
```diff
@@ -29,16 +30,42 @@
...
+    def __init__(self, project_root: str, global_memory_dir: Path | None = None):
         self._memory_dir = Path(get_serena_managed_in_project_dir(project_root)) / "memories"
         self._memory_dir.mkdir(parents=True, exist_ok=True)
+        self._global_memory_dir = global_memory_dir
+        if self._global_memory_dir is not None:
+            self._global_memory_dir.mkdir(parents=True, exist_ok=True)
         self._encoding = SERENA_FILE_ENCODING
 
+    def _is_global(self, name: str) -> bool:
```
MischaPanch, line 43: This should be public and used everywhere instead of repeating the check logic

src/serena/tools/workflow_tools.py
```diff
@@ -18,23 +18,33 @@ def apply(self) -> str:
...
         else:
-            return f"""The onboarding was already performed, below is the list of available memories.
+            msg = f"""The onboarding was already performed, below is the list of available memories.
             Do not read them immediately, just remember that they exist and that you can read them later, if it is necessary
             for the current task.
             Some memories may be based on previous conversations, others may be general for the current project.
             You should be able to tell which one you need based on the name of the memory.
             
-            {memories}"""
+            Available project memories: {project_memories}"""
```
MischaPanch, line 44: There's no need to explain the special role of global anywhere except in the write tool. I also think we should only check the new config value for the edit and delete operation (truly destructive ones), not for writing a new one or move. Otherwise it will be difficult for users to add global memories.

The docstrings in the tools should also not mention global all the time, the only exception being the write memory. The example there should be the prototypical "global/style_guide".

Tbh, this got me thinking whether we should make this feature more extensive, since a python style guide should not be included for Java projects. Meaning just a single global container is not enough, we should allow per-project control. 

@opcode81 @butterflysky wdyt? If we're doing this, might as well do it right. I'm almost sure that this feature would be requested soon by anyone wanting to use global memories at all

MischaPanch (changes requested):

src/serena/tools/workflow_tools.py
MischaPanch: There are multiple ways to go for adding the enhanced functionality. We could use frontmatter with markers, allowing users to specify filters in their project.yml. We could use topics within global, and let users select topics in project config. Or something really simple, like a configured glob applied to memory names

src/serena/project.py
opcode81: I am refactoring to provide the common replacer.

src/serena/tools/workflow_tools.py
opcode81: If you have a `java_style_guide` while the project is Python, LLMs won't read it.
So the only "problem" is that the memory will be listed, which isn't a significant problem - unless you have a large number of irrelevant memories, which is unlikely.
We can always add additional structure later. Adding more structure also makes memory creation/access a bit more complicated.

MischaPanch: Ok, let's keep it simple for now. We should just extend the docstring of the write tool to make the LLM choose "qualified" names, maybe just an example is enough

MischaPanch: I'd prefer java/style_guide btw (topic, not name prefix)

src/serena/project.py
MischaPanch: Which is the appropriate solution, the AI just giving excuses here. The agents just don't want to refactor 😟

src/serena/project.py
```diff
@@ -29,16 +30,42 @@
 
 
 class MemoriesManager:
-    def __init__(self, project_root: str):
+    GLOBAL_TOPIC = "global"
+
+    def __init__(self, project_root: str, global_memory_dir: Path | None = None):
```
opcode81, line 35: This is a questionable decision, as it ties global memories to a project, and we cannot access global memories without a project being activated.

An alternative would be to have only `root_folder` here and instantiate more than once, i.e.
* for the agent (global memories)
* for the project, if any.

This allows us to decouple the reporting of global memories from the reporting of per-project memories.
* We can report global memories only once (in the agent's system prompt) instead of reporting them for every project activation.
* Project activation lists only project-specific memories.

This saves tokens and allows global memories to be known even before project activation.

@MischaPanch wdyt?

src/serena/project.py
```diff
@@ -124,7 +195,8 @@ def __init__(
     ):
         self.project_root = project_root
         self.project_config = project_config
-        self.memories_manager = MemoriesManager(project_root)
+        global_memory_dir = Path(SerenaPaths().serena_user_home_dir) / "memories" / MemoriesManager.GLOBAL_TOPIC
```
opcode81, line 198: Add the path to SerenaPaths instead.

src/serena/project.py
MischaPanch: Sounds nice but in practice I notice that system prompt is often ignored or at least forgotten after a while by some agents (notably codex), which I work around with by calling activate project even when it's not needed. I think repeating on project activation is not too bad and somewhat safer given that context.

But let me actually verify this behavior in codex first

src/serena/tools/workflow_tools.py
opcode81: It is up to the user (and the LLM) to use a naming scheme that fits the respective requirements

* `java/style_guide` makes sense if there are several Java-related memories
* `java_style_guide` makes sense if there is only one or a flatter structure is preferred
* `style_guides/java` makes sense if grouping by function is desired
* etc.

We should not impose our preferences.
It is not really logical to even have preferences a priori, because it depends on the situation and the requirements.

MischaPanch:
@butterflysky thank you for the PR! We'll take this over, finish and merge it

butterflysky:
> @butterflysky thank you for the PR! We'll take this over, finish and merge it

thanks @MischaPanch, i was just about to make a pass by hand at addressing your review comments, but if you all want to finish it that's great. I appreciate your time and energy here, thank you!

MischaPanch:
Superseded by #1091 which now has been merged

PR #1012, opened 2026-02-10
feat(scala): Add stale lock detection for Metals H2 database

tkolleh:
This PR adds stale lock detection and handling for Scala Metals LSP, improving reliability when running alongside other Metals instances (e.g., VS Code).

### Problem

When a Metals process crashes without proper cleanup, it leaves a stale lock file (`.metals/metals.mv.db.lock.db`) that can prevent proper H2 AUTO_SERVER coordination, causing new instances to fall back to in-memory database mode (degraded experience).

### Solution

- Detect stale locks by checking if the lock-holding process is alive and is actually Metals
- Provide configurable handling modes: `auto-clean` (default), `warn`, or `fail`
- Log informational messages when another active Metals instance is detected

## Configuration

```yaml
ls_specific_settings:
  scala:
    on_stale_lock: "auto-clean"      # auto-clean | warn | fail
    log_multi_instance_notice: true  # Log info when another Metals detected
```

## Related
- https://github.com/oraios/serena/pull/1012
- Metals H2 AUTO_SERVER: https://scalameta.org/metals/docs/integrations/new-editor
- H2 AUTO_SERVER Mode: http://www.h2database.com/html/features.html#auto_mixed_mode

### DIFF
--- a/src/solidlsp/util/metals_db_utils.py
+++ b/src/solidlsp/util/metals_db_utils.py
@@ -75,7 +75,7 @@ def parse_h2_lock_file(lock_path: Path) -> MetalsLockInfo | None:
 
     Returns:
         MetalsLockInfo if the file can be parsed, None if file doesn't exist
-        or is completely unparseable.
+        or is completely unparsable.
 
     """
     if not lock_path.exists():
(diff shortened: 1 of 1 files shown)


MischaPanch:
Hi @tkolleh , I finally found time to look at this. While some smaller things could be designed more in line with the existing code, this is not urgent and could be done later. We currently have a very tight pipeline of work for Serena, so I didn't verify the core logic of handling the locks in detail. On the surface, it seems reasonable and sufficiently well designed. All design issues can be addressed later, if needed, since the code is entirely restricted to scala and doesn't interfere with anything else.

I'm going to go ahead and merge this without further review. I am not a big fan of the mocks in tests, but we can't reasonably run "real" scala tests due to the difficulties posed by metals (compilation needed before LS can work, setup and "warm-up" are extremely slow). I trust that you have tested the functionality sufficiently yourself. Thank you for the contribution!


@opcode81 FYI. We introduce minor technical dept for scala here, but I'm willing to just swallow it now as an alternative would be either review ping pong or (and) me investing significant time in going through the logic and the tests.

PR #1016, opened 2026-02-10
feat: Add allow_no_language_servers config option

cx-alberto-simoes:
Add a new configuration option 'allow_no_language_servers' to SerenaConfig that enables running projects without any language servers configured.

This allows file-based tools (read_file, list_dir, search_for_pattern, etc.) to work on projects containing files in unsupported languages (e.g., SQL, plain text, documentation) while maintaining backward compatibility.

Changes:
- Add allow_no_language_servers field to SerenaConfig with default False
- Update ProjectConfig.autogenerate() to accept and use the option
- Update ProjectConfig.load() to pass through the option
- Update SerenaConfig.add_project_from_path() to use the config value
- Add comprehensive documentation in serena_config.template.yml

Default value of False maintains existing behavior where projects must have at least one language configured.

### DIFF
--- a/src/serena/agent.py
+++ b/src/serena/agent.py
@@ -393,7 +393,7 @@ class SerenaAgent:
             raise Exception(
                 "The language server manager is not initialized. This could indicate either:\n"
                 "1. A problem during project activation (check Serena's logs for errors), or\n"
-                "2. The project has no language servers configured (allow_no_language_servers=True).\n"
+                "2. The project has no language servers configured (languages: [] with extra_source_file_extensions set).\n"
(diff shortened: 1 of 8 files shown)


MischaPanch:
Hi @cx-alberto-simoes. I understand your use case, but it is very different for what Serena is made for, and we can't merge something that increases complexity just to support it. I suggest you use one of the proposed workarounds, specifying python or some other cheap-to-start language in your tool. Note that even something like find_pattern is built on language servers, since it uses those to identify which files are code files.

Closing this.

PR #1037, opened 2026-02-13
Fix: robust canonicalization and matching (jetbrains_plugin_client)

CRaFT4ik:
### Problem
When Serena runs inside a DevContainer or WSL, the JetBrains plugin returns the project root in the host’s native format (e.g., `"C:/Users/user/..."`). The previous matching logic attempted to normalize this with `_normalize_wsl_path()` and then used `Path.resolve()` + direct equality check. This approach failed in several scenarios:

- The normalization would sometimes produce paths like `/workspaces/serena/C:/Users/user/...` instead of the expected `/mnt/c/Users/user/...`.
- `Path.resolve()` behaves differently depending on the filesystem context (symlinks, case sensitivity, trailing slashes), and a simple string equality often returned `False` even for the same logical location.
- Mixed environments (Windows host + WSL + DevContainer) introduced multiple layers of path prefixes that were not handled.

Example from logs:  
- Local project root: `/mnt/c/Users/user/projects/my-app`  
- Plugin project root (after old normalization): `/workspaces/serena/C:/Users/user/projects/my-app`  
- Despite both pointing to the same physical directory, the paths differ, causing the match to fail and Serena to report “Found no Serena service”.

### Changes
1. **Path canonicalization**  
   Added `_canonicalize_path()` that normalizes various path forms (Windows drive letters, WSL UNC paths, DevContainer prefixes) into a consistent POSIX‑like lowercase string. It strips container prefixes (e.g., `/workspaces/serena/`) and converts `C:/...` to `/mnt/c/...`, producing a clean canonical representation.

2. **Improved matching logic**  
   The `matches()` method now:  
   - Tries an **exact canonical match** first.  
   - If that fails, checks whether one path is a **suffix** of the other (handles cases where the plugin path includes a container prefix like `/workspaces/serena/`).  
   - Falls back to comparing the **last N components** (currently 3) as a heuristic for deeply nested paths.  
   - Logs raw and canonical values, original comparison result (when safely computable), and the decision path for full transparency.

3. **Extensive logging**  
   Added detailed debug logs to trace the matching process, making it easier to diagnose future path‑related issues.

4. **Unit tests**  
   Added tests covering Windows, WSL, DevContainer scenarios, and edge cases (e.g., trailing slashes, mixed case, relative paths).

These changes ensure that projects are correctly identified even when paths are represented differently due to the underlying runtime environment.

### DIFF
--- a/src/serena/tools/jetbrains_plugin_client.py
+++ b/src/serena/tools/jetbrains_plugin_client.py
@@ -4,6 +4,7 @@ Client for the Serena JetBrains Plugin
 
 import json
 import logging
+import posixpath
 import re
 from pathlib import Path
 from typing import Any, Literal, Optional, Self, TypeVar, cast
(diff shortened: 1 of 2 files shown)


opcode81:
This seems overly complicated. Are all the measures you have taken indeed necessary to solve your particular problem?

CRaFT4ik:
> This seems overly complicated. Are all the measures you have taken indeed necessary to solve your particular problem?

That might be true. But without this, I couldn't get the Windows host with Android Studio, Docker with Serena, and Claude Code on WSL working. A fix for me would have been two lines of code and would have broken everything else, but I tried to create a general solution that would preserve the previous logic and close my case. If you see something I've done wrong, I'd be happy if you pointed me to it. By the way, there are tests with examples there.

opcode81:
So you are saying that you really implemented the minimal solution that makes it work for you without breaking things that worked previously?

Let us start this way: Please provide a concrete list of the cases (examples) that didn't work for you.
Then I can evaluate what the minimal change set would look like and whether this PR is an implementation thereof.

Each example should provide
* the path that Serena itself was checking
* the path that was reported by the JetBrains plugin

CRaFT4ik:
> So you are saying that you really implemented the minimal solution that makes it work for you without breaking things that worked previously?
> 
> Let us start this way: Please provide a concrete list of the cases (examples) that didn't work for you. Then I can evaluate what the minimal change set would look like and whether this PR is an implementation thereof.
> 
> Each example should provide
> 
> * the path that Serena itself was checking
> * the path that was reported by the JetBrains plugin

As the project's author, judge for yourself whether it's minimal or not. I've made no such claims. That's not my goal, but I do adhere to the principles of a good solution, so that no one encounters artificial limitations regarding path mapping from different systems.

Concrete case:
The main problem, in my opinion, stems from the idea plugin, which returns the absolute path of its native system for status request (C:/Users/user/...). The current code converts this to something like "/workspaces/serena/C:/Users/user/...", and then tries to compare the real path to project in WSL with this meaningless line. Naturally this doesn't work, which results in an error.

opcode81:
I want to first understand what *exactly* your problem was in order to determine what the best solution to that problem is.

As requested, we need, for each problem instance,
* `s` = project path used by Serena 
* `j` = project path reported by the JetBrains plugin instance.

From what you have reported thus far, it seems that you had only a single problem instance where
* `s = "/mnt/c/Users/user/projects/my-app"`
* `j = "/workspaces/serena/C:/Users/user/projects/my-app"`

while your PR also seems to solve additional problems, which you didn't actually have.

There are two underlying issues:
* A: Serena is run in a different file system than the IDE, so `s` uses `/mnt/c/` instead of `C:/`
* B: The JetBrains path contains an unexpected prefix `/workspaces/serena/`. Do you know where it comes from?

(A) could be handled by a regex replacement/normalisation.
(B) could be handled via suffix matching instead of equality matching.

We as maintainers are interested in implementing a clean, minimal solution only, so I need to determine what that might be (and whether your PR satisfies that).

opcode81:
@CRaFT4ik please check whether the branch `robust-plugin-path-matching` of PR #1046 solves your problem.

CRaFT4ik:
> @CRaFT4ik please check whether the branch `robust-plugin-path-matching` of PR #1046 solves your problem.

Just checked, for me #1046 also fixes the issue.

opcode81:
Thanks for testing, @CRaFT4ik, and thanks for pointing out the issue!

PR #1049, opened 2026-02-17
Add Luau language server support via luau-lsp

RavenKyu:
## Summary

  - Add Luau (Roblox's typed Lua 5.1 superset) language server support using [luau-lsp](https://github.com/JohnnyMorganz/luau-lsp)       
  - The `luau-lsp` binary is automatically downloaded from GitHub releases if not found locally (supports Linux x86_64, macOS, Windows)  
  - Comprehensive test suite covering symbol finding, within-file references, cross-file references, and go-to-definition

  ## Changes

  - **`src/solidlsp/language_servers/luau_lsp.py`** — New `LuauLanguageServer` class with auto-download and LSP initialization
  - **`src/solidlsp/ls_config.py`** — Register `LUAU` in `Language` enum, file matcher (`*.luau`), and factory method
  - **`pyproject.toml`** — Add `luau` pytest marker
  - **`test/resources/repos/luau/test_repo/`** — Minimal Luau test project with cross-file module structure
  - **`test/solidlsp/luau/test_luau_basic.py`** — 5 tests: symbol finding (×2), within-file references, cross-file references,
  go-to-definition

  ## Test plan

  - [ ] Run `pytest test -m luau -vv` to verify all Luau tests pass
  - [ ] Verify luau-lsp auto-download works on a clean environment
  - [ ] Confirm no regressions in existing language server tests

### DIFF
--- a/src/solidlsp/language_servers/luau_lsp.py
+++ b/src/solidlsp/language_servers/luau_lsp.py
@@ -17,6 +17,7 @@ import os
 import pathlib
 import platform
 import shutil
+import threading
 import zipfile
 from pathlib import Path
 
@@ -33,6 +34,10 @@ log = logging.getLogger(__name__)
 
 # Pin to a known stable release
 LUAU_LSP_VERSION = "1.57.1"
+
+# Roblox type definitions and API docs CDN
+ROBLOX_TYPES_URL = "https://luau-lsp.pages.dev/type-definitions/globalTypes.PluginSecurity.d.luau"
+ROBLOX_DOCS_URL = "https://luau-lsp.pages.dev/api-docs/en-us.json"
 
 
 class LuauLanguageServer(SolidLanguageServer):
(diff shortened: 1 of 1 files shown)


src/solidlsp/language_servers/luau_lsp.py
```diff
@@ -0,0 +1,380 @@
...
+            possible_paths.append(home / ".aftman" / "bin" / "luau-lsp")
+
+        for path in possible_paths:
+            if path.exists():
+                return str(path)
+
+        return None
+
+    @staticmethod
+    def _download_luau_lsp() -> str:
```
MischaPanch, line 105: Pls use the newer `DependencyProvider` interface for handling the download of dependencies. You can find a somewhat complex example in `CSharpLanguageServer` or a simpler example in intelephense, among other implementations

Image

MischaPanch:
@RavenKyu ping

src/solidlsp/language_servers/luau_lsp.py
StayBlue: > given that luau is a superset of lua, does it make sense to just use it as the default LS for lua, marking our other LS as experimental instead? Or is luau-lsp not stable enough for that yet?

It is not possible to simply just use luau-lsp as a replacement for lua-lsp, as Luau diverges from Lua in a lot of ways, which are all documented [here](https://luau.org/compatibility/). I believe this compatibility would even diverge more from Lua with the release of 5.5 in ways that aren't yet documented on their website.

It also adds features that Lua simply doesn't have, such as native type checking support, which isn't even a concept in Lua, so it'd mark Lua code as semantically correct even when it uses Luau primitives.

MischaPanch: Ok, then no replacement. Are you up for performing the refactoring I mentioned above?

StayBlue: Sure. I'll make it happen.

PR #1050, opened 2026-02-17
 feat: add save and clear log buttons to dashboard log page

aleksbeloushkin-srb:
Description:
  ## Summary

  The dashboard Logs page previously had only a copy-to-clipboard button.
  This PR adds two additional actions:

  - **Save logs** — downloads all current log entries as a plain-text `.txt` file
    with a timestamp in the filename (e.g. `serena-logs-2026-02-17T14-30-00.txt`).
    Implemented client-side via the Blob API — no backend changes required.
  - **Clear logs** — clears the server-side `LogBuffer` via `POST /clear_logs`,
    then resets the UI log view and the polling cursor (`currentMaxIdx = -1`).
    A confirmation prompt is shown before clearing.

  All three buttons (save, copy, clear) share a unified `.log-action-btn` style
  and are grouped in a flex container.

  ## Log screen: 
  log_creen

  ## Video buttons 

  https://github.com/user-attachments/assets/9abe60fe-2e9d-4764-87f0-af095e6a1414

  ## Changes

  | File | Change |
  |------|--------|
  | `src/serena/util/logging.py` | Add `LogBuffer.clear()` and `MemoryLogHandler.clear_log_messages()` |
  | `src/serena/dashboard.py` | Add `POST /clear_logs` endpoint |
  | `src/serena/resources/dashboard/index.html` | Replace single button with a three-button group |
  | `src/serena/resources/dashboard/dashboard.js` | Add `saveLogs()` and `clearLogs()` methods |
  | `src/serena/resources/dashboard/dashboard.css` | Unified `.log-action-btn` styles |
  | `docs/02-usage/060_dashboard.md` | Document the new log page buttons |
  | `CHANGELOG.md` | Add entry under `# latest` |

  ## Implementation notes

  - No breaking changes. Existing copy behaviour is unchanged.
  - No new dependencies.
  - `LogBuffer.clear()` resets both `_log_messages` and `_max_idx` to `-1`.
    Resetting `_max_idx` is necessary: without it, the next poll would request
    `from_idx = N` (the old cursor), receive zero messages, and leave the log view
    appearing stuck despite the buffer being empty.
  - `LogBuffer.clear()` acquires the existing lock — thread-safe with concurrent `append()` calls.

### DIFF
--- a/src/serena/resources/dashboard/dashboard.css
+++ b/src/serena/resources/dashboard/dashboard.css
@@ -1322,6 +1322,11 @@ body {
     display: inline;
 }
 
+.log-action-btn:disabled {
+    opacity: 0.35;
+    cursor: not-allowed;
+}
+
 .log-action-btn-danger:hover {
     background-color: var(--border-color);
     border-color: rgba(220, 53, 53, 0.9);
--- a/src/serena/resources/dashboard/dashboard.js
+++ b/src/serena/resources/dashboard/dashboard.js
@@ -1116,12 +1116,14 @@ class Dashboard {
         document.title = activeProject ? `${activeProject} – Serena Dashboard` : 'Serena Dashboard';
     }
 
+    updateLogButtons(hasLogs) {
+        this.$saveLogsBtn.prop('disabled', !hasLogs);
+        this.$copyLogsBtn.prop('disabled', !hasLogs);
+        this.$clearLogsBtn.prop('disabled', !hasLogs);
+    }
+
     saveLogs() {
         const logText = this.$logContainer.text();
-        if (!logText) {
-            alert('No logs to save');
-            return;
-        }
         const timestamp = new Date().toISOString().replace(/[:.]/g, '-').slice(0, 19);
         const blob = new Blob([logText], {type: 'text/plain'});
         const url = URL.createObjectURL(blob);
@@ -1142,11 +1144,6 @@ class Dashboard {
     copyLogs() {
         const logText = this.$logContainer.text();
 
-        if (!logText) {
-            alert('No logs to copy');
-            return;
-        }
-
         // Use the Clipboard API to copy text
         navigator.clipboard.writeText(logText).then(() => {
             // Visual feedback - temporarily change icon to grey checkmark
@@ -1159,12 +1156,10 @@ class Dashboard {
             }, 1500);
         }).catch(err => {
             console.error('Failed to copy logs:', err);
-            alert('Failed to copy logs to clipboard');
         });
     }
 
     clearLogs() {
-        if (!confirm('Clear the log view?')) return;
         let self = this;
         $.ajax({
             url: '/clear_logs',
@@ -1172,14 +1167,15 @@ class Dashboard {
             success: function () {
                 self.$logContainer.empty();
                 self.currentMaxIdx = -1;
+                self.updateLogButtons(false);
 
                 const originalHtml = self.$clearLogsBtn.html();
                 const checkmarkSvg = '<svg xmlns="http://www.w3.org/2000/svg" width="18" height="18" viewBox="0 0 24 24" fill="none" stroke="#888" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><polyline points="20 6 9 17 4 12"></polyline></svg><span class="log-action-btn-text">clear logs</span>';
                 self.$clearLogsBtn.html(checkmarkSvg);
                 setTimeout(() => { self.$clearLogsBtn.html(originalHtml); }, 1500);
             },
-            error: function () {
-                alert('Failed to clear logs');
+            error: function (xhr, status, error) {
+                console.error('Failed to clear logs:', error);
             }
         });
     }
@@ -1214,6 +1210,7 @@ class Dashboard {
                     $('#log-container').html('<div class="loading">No log messages found.</div>');
                 }
 
+                self.updateLogButtons(response.messages && response.messages.length > 0);
                 self.updateTitle(response.active_project);
 
                 // Start periodic polling for new logs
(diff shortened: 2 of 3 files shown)


opcode81:
Note that Serena already saves all logs to disk. I have added a [section](https://oraios.github.io/serena/02-usage/065_logs.html) to the documentation to make this clear.

aleksbeloushkin-srb:
> Note that Serena already saves all logs to disk. I have added a [section](https://oraios.github.io/serena/02-usage/065_logs.html) to the documentation to make this clear.

Hello! Thanks for highlighting the documentation — that's very helpful context.
To clarify the distinct value each button adds on top of disk persistence:

  - Save logs — captures a browser-accessible snapshot of the exact in-memory rolling buffer at a point in time. Particularly useful in remote or containerised deployments where ~/.serena/logs isn't reachable from the user's machine.
  - Clear logs — clears only the in-memory live view and resets the polling cursor; disk logs in ~/.serena/logs are untouched. I'm also considering adding the option to clear disk logs or trigger a reload in a future iteration.

I've updated the docs to make this distinction explicit.

scripts/run_tools_lsp.py
```diff
@@ -0,0 +1,45 @@
+"""
```
MischaPanch: pls remove, we already have demo_run_tools

docs/02-usage/060_dashboard.md
```diff
@@ -26,6 +26,22 @@ Both can be configured in Serena's [configuration](050_configuration) file (`ser
 If enabled, they will automatically be opened as soon as the Serena agent/MCP server is started.
 For the dashboard, this can be disabled if desired (see below).
 
+## Log Page
+
+The **Logs** page streams live log output from the running Serena session, updated every second.
```
MischaPanch: That's too detailed, you just have to mention the relation between the new save button and the already saved logs on disk

CHANGELOG.md
```diff
@@ -15,9 +15,20 @@ Status of the `main` branch. Changes prior to the next official version change w
     * Major extensions to the dashboard, which now serves as a central web interface for Serena
         * View current configuration
         * View news which can be marked as read
-        * View the executions, with the possibility to cancel running/scheduled executions 
+        * View the executions, with the possibility to cancel running/scheduled executions
         * View tool usage statistics
         * View and create memories and edit the serena configuration file
+        * Log page now has save, copy and clear buttons:
```
MischaPanch: Pls just leave the first line (without the : in in the end) and remove all the detailed explanation below

docs/02-usage/060_dashboard.md
opcode81: This also violates the logical structure of the page.
There is no specific section on the Dashboard (the section applies to both the GUI tool and the Dashboard), but this section only applies to one of them.

We actually don't need this section at all. The usage is self-explanatory, and we have already stated (in abstract terms) what the Dashboard can do.

opcode81: If we want to add more information, we need to restructure

* intro (just stating that there are two UIs and how they are supported)
* section on GUI tool
* section on Dashboard (potentially with subsections)

but I don't really think it is necessary.

CHANGELOG.md
aleksbeloushkin-srb: Thanks. Done. Trimmed to one line:
  `* Log page now has save (downloads a snapshot) and clear (resets log view) buttons alongside the existing copy button`

docs/02-usage/060_dashboard.md
aleksbeloushkin-srb: Ok. Thanks. Trimmed — kept it concise, noting that copy was already there and briefly describing the two new buttons

scripts/run_tools_lsp.py
aleksbeloushkin-srb: Good point, removed.

MischaPanch:
@aleksbeloushkin-srb after noticing the video in the PR description, the following final comments, generally about avoiding excessive notifications.

1. Notifications should be modals, not browser-notifications. This allows using consistent styling. We already have modals in the dashboard
2. No need for confirmation on clearing logs, just clear them
3. No need for popup on copying logs. If there are no logs, just not copy anything or disable the button.
4. Disable the save button on empty logs. For saving, pls add a snapshot of how the saving works. We could either save using a modal where the user selects a place, or, which I would prefer, save to downloads using the browser builtins. Then we let the browser take care of displaying where things were downloaded to.

I think your additions can be made without introducing any notifications or confirmations at all.

opcode81:
This looks good to me now.
Propose to squash-merge.

aleksbeloushkin-srb:
> @aleksbeloushkin-srb after noticing the video in the PR description, the following final comments, generally about avoiding excessive notifications.
> 
> 1. Notifications should be modals, not browser-notifications. This allows using consistent styling. We already have modals in the dashboard
> 2. No need for confirmation on clearing logs, just clear them
> 3. No need for popup on copying logs. If there are no logs, just not copy anything or disable the button.
> 4. Disable the save button on empty logs. For saving, pls add a snapshot of how the saving works. We could either save using a modal where the user selects a place, or, which I would prefer, save to downloads using the browser builtins. Then we let the browser take care of displaying where things were downloaded to.
> 
> I think your additions can be made without introducing any notifications or confirmations at all.

Thanks for the thorough feedback! Addressed all points:

  - Removed all alert() and confirm() calls — no browser popups or confirmations anywhere
  - Clear now acts immediately without any confirmation dialog
  - Copy, save and clear buttons are all disabled when the log view is empty (also disabled by default on page load)
  - Save uses the browser's built-in download mechanism (Blob → Downloads folder) as you suggested — the browser itself handles the download notification

  Updated the video in the PR description — it shows the buttons in action including the disabled state and the browser download bar on save.

MischaPanch:
@aleksbeloushkin-srb thanks, looks great to me!

PR #1056, opened 2026-02-20
feat(csharp): add .slnx solution file support

Djentinga:
- Recognize .slnx alongside .sln in solution file discovery (find_solution_or_project_file and _open_solution_and_projects)
- Update log messages to mention .slnx


MischaPanch:
Thanks for the contribution, merging

PR #1059, opened 2026-02-20
fix: avoid unnecessary didOpen/didClose LSP calls for cached files in full symbol tree

mareurs:
## Summary

- `request_full_symbol_tree` was performing `didOpen`/`didClose` LSP calls for every file, even when document symbols were already cached
- Changed `_open_file_context` to use `open_in_ls=False` in the full symbol tree path, so cached files skip the expensive LSP roundtrip
- Result: full project symbol tree goes from ~9.5s to ~0.5s (19x speedup) on a ~1300-file Kotlin project

## Test plan

- [x] Verified on a large Kotlin project (~1300 .kt files) — symbol tree request dropped from 9.5s to 0.5s
- [x] Run existing test suite (`uv run poe test`) to confirm no regressions

### DIFF
--- a/src/solidlsp/ls.py
+++ b/src/solidlsp/ls.py
@@ -1193,11 +1193,8 @@ class SolidLanguageServer(ABC):
 
             return response
 
-        if file_data is not None:
-            return get_raw_document_symbols(file_data)
-        else:
-            with self.open_file(relative_file_path) as opened_file_data:
-                return get_raw_document_symbols(opened_file_data)
+        with self._open_file_context(relative_file_path, file_buffer=file_data) as fd:
+            return get_raw_document_symbols(fd)
 
     def request_document_symbols(self, relative_file_path: str, file_buffer: LSPFileBuffer | None = None) -> DocumentSymbols:
         """
@@ -1227,7 +1224,6 @@ class SolidLanguageServer(ABC):
                 log.debug("No cache hit for document symbols in %s", relative_file_path)
 
             # no cached result: request the root symbols from the language server
-            file_data.ensure_open_in_ls()
             root_symbols = self._request_document_symbols(relative_file_path, file_data)
 
             if root_symbols is None:


opcode81:
Thanks. I have added another minor improvement to this PR.

Your change is correct, but for some reason, the Kotlin tests have failed - perhaps because the Kotlin tests rely on a side-effect. Let's see if the problem persists when the tests are rerun.

MischaPanch:
@opcode81 since the Kotlin LS is pre-alpha, we just pull the latest version and hope for the best. I wouldn't interpet too much into tests failing there.

Though we can verify whether the failure is triggered by this change, ofc

mareurs:
The Kotlin tests passed locally for us — we're running v261.13587.0 (platform-specific builds) on a local branch. Upstream main still has v0.253.10629, so the CI failure is likely with the older version. We have a branch ready with the Kotlin LSP update if you'd like a PR for that as well.

MischaPanch:
Ah, I was wrong about us always pulling the latest, it's pinned. I will extract the kotlin version into a setting and push on this branch

mareurs:
FYI, I've been using it extensively for some time and it works great! Congrats for a great idea and project!

opcode81:
Kotlin tests pass locally. As I have already stated, the change is definitely correct.

MischaPanch:
Let's merge this, I will make a follow-up PR to refactor Kotlin to use the DependencyProvider.

@mareurs thanks for the compliment, glad you like Serena! Especially for JVM based languages you may find value in our [jetbrains plugin](https://oraios.github.io/serena/02-usage/025_jetbrains_plugin.html) :)

MischaPanch:
@mareurs
> We have a branch ready with the Kotlin LSP update if you'd like a PR for that as well.

Did you refactor the code to use DependencyProvider there? That would be the right step to improve configurability (and bump the version)

mareurs:
thanks for the heads up. I will refactor, I was using the old way.

MischaPanch:
Thanks, then I won't do it and will wait for your PR instead

PR #1060, opened 2026-02-20
feat: support per-project language_backend override

mareurs:
## Summary

Adds support for configuring the `language_backend` (LSP or JetBrains) at the per-project level in `.serena/project.yml`, overriding the global setting from `serena_config.yml`.

**Key design decision**: The backend is determined once at startup and fixed for the session lifetime. This avoids dynamic tool swapping post-init (which would violate MCP protocol expectations).

### DIFF
--- a/src/serena/config/serena_config.py
+++ b/src/serena/config/serena_config.py
@@ -134,24 +134,6 @@ class ModeSelectionDefinition:
     default_modes: Sequence[str] | None = None
 
 
-@dataclass
-class SharedConfig(ModeSelectionDefinition, ToolInclusionDefinition, ToStringMixin):
-    """Shared between SerenaConfig and ProjectConfig, the latter used to override values in the form
-    (same as in ModeSelectionDefinition).
-    The defaults here shall be none and should be set to the global default values in SerenaConfig.
-    """
-
-    symbol_info_budget: float | None = None
-
-
-class SerenaConfigError(Exception):
-    pass
-
-
-def get_serena_managed_in_project_dir(project_root: str | Path) -> str:
-    return os.path.join(project_root, SERENA_MANAGED_DIR_NAME)
-
-
 class LanguageBackend(Enum):
     LSP = "LSP"
     """
@@ -172,6 +154,25 @@ class LanguageBackend(Enum):
         raise ValueError(f"Unknown language backend '{backend_str}': valid values are {[b.value for b in LanguageBackend]}")
 
 
+@dataclass
+class SharedConfig(ModeSelectionDefinition, ToolInclusionDefinition, ToStringMixin):
+    """Shared between SerenaConfig and ProjectConfig, the latter used to override values in the form
+    (same as in ModeSelectionDefinition).
+    The defaults here shall be none and should be set to the global default values in SerenaConfig.
+    """
+
+    symbol_info_budget: float | None = None
+    language_backend: LanguageBackend | None = None
+
+
+class SerenaConfigError(Exception):
+    pass
+
+
+def get_serena_managed_in_project_dir(project_root: str | Path) -> str:
+    return os.path.join(project_root, SERENA_MANAGED_DIR_NAME)
+
+
 @dataclass(kw_only=True)
 class ProjectConfig(SharedConfig):
     project_name: str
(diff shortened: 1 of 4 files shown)


src/serena/agent.py
```diff
@@ -318,8 +318,11 @@ def get_memory_log_handler() -> MemoryLogHandler:
         tool_inclusion_definitions.append(self._context)
         if self._context.single_project:
             tool_inclusion_definitions.extend(self._single_project_context_tool_inclusion_definitions(project))
-        if self.serena_config.language_backend == LanguageBackend.JETBRAINS:
-            tool_inclusion_definitions.append(SerenaAgentMode.from_name_internal("jetbrains"))
+        if self._any_project_uses_jetbrains():
+            # Include JetBrains optional tools in the exposed tool set so they are available via MCP,
+            # but don't exclude LSP tools from the base set (that happens dynamically per-project).
```
MischaPanch: We don't want to do that, only the actually needed toolset should be exposed. This is also the reason why the language backend was not on a per-project basis so far.

We can still make this happen, but only in contexts that are marked as `single_project`, because there no project switching is possible after instantiation. So, you should:

1. Never expose tools that are unused
2. Only support a per-project backend override if the context is single_project
3. If the context is not single_project, project override is present and different from the global config, fail to start
4. If project is activated, project override is present and different from the global config, fail to activate, give a meaningful message to the user

opcode81: It could be preferable to fail only when we really need to fail, i.e.

* Initialisation:
  * If a project is provided at startup, activate that project's language backend (may override global)
  * If no project is provided at startup, activate the globally configured backend
* Activating projects dynamically (post-initialisation):
  If a project is activated that requires a change of the backend, return an error to the user explaining that this would require a change in the toolset and therefore is not supported. Workarounds the user can consider:  
     * Use project activation at startup
     * In the client, configure one MCP server per backend and dynamically enable only one of them

MischaPanch: agree, let's do it like this

mareurs:
Thanks for the feedback @MischaPanch @opcode81 — the approach makes sense. I'll rework this to:

1. At startup with a project: activate that project's language backend (may override global)
2. At startup without a project: use the globally configured backend
3. Post-init activation of a project with a different backend: return an error with a clear message suggesting workarounds (startup activation or separate MCP server per backend)

No dynamic tool swapping. Will push an update.

src/serena/config/serena_config.py
```diff
@@ -181,6 +181,11 @@ class ProjectConfig(SharedConfig):
     ignore_all_files_in_gitignore: bool = True
     initial_prompt: str = ""
     encoding: str = DEFAULT_SOURCE_FILE_ENCODING
+    language_backend: LanguageBackend | None = None
```
MischaPanch: Shouldn't this be in SharedConfig now?

mareurs: Done — moved `language_backend` into `SharedConfig` (with `LanguageBackend` enum moved above it for the type reference). `ProjectConfig` now inherits it, same pattern as `symbol_info_budget`. Also added documentation for the per-project override to the JetBrains plugin, workflow, and configuration pages.

src/serena/agent.py
mareurs: Done in the latest commit. The design ended up as discussed — backend fixed at startup, error on post-init mismatch with clear workarounds.

MischaPanch:
Thank you! Merging

src/serena/agent.py
```diff
@@ -318,7 +318,23 @@ def get_memory_log_handler() -> MemoryLogHandler:
...
+        if project is not None:
+            startup_project_config = self._find_project_config(project)
+            if startup_project_config is not None and startup_project_config.language_backend is not None:
+                self._language_backend = startup_project_config.language_backend
+                _backend_from_project_config = True
+        if _backend_from_project_config:
+            log.info(f"Using language backend as configured in project.yml: {self._language_backend.name}")
+        else:
+            log.info(f"Using language backend from global serena config: {self._language_backend.name}")
+
```
opcode81, line 336: This is unnecessarily complicated, and it is not OK for this to be in the middle of the toolset determination, which is one logical block.
Since @MischaPanch already merged this, I will fix it on main.

PR #1061, opened 2026-02-20
Update Kotlin Language Server to v261.13587.0

mareurs:
## Summary

- Upgrade Kotlin LSP from v0.253.10629 (Aug 2024) to v261.13587.0 (Dec 2024) with platform-specific builds
- Suppress noisy `$/logTrace` notifications that caused massive log files (100MB+) and CPU overhead
- Refactor to use the standard `DependencyProvider` pattern, enabling `ls_path` override via `ls_specific_settings`
- Make `kotlin_lsp_version` configurable via `ls_specific_settings` (renamed constant to `DEFAULT_KOTLIN_LSP_VERSION`)
- Fix JVM options to allow explicit empty string (disable defaults) vs unset (use default `-Xmx4G`)
- Document Kotlin settings in `050_configuration.md`

## Test plan

- [x] All 6 Kotlin tests pass (`uv run poe test -m kotlin`)
- [x] Full test suite passes (no regressions)

### DIFF
--- a/src/solidlsp/language_servers/systemverilog_server.py
+++ b/src/solidlsp/language_servers/systemverilog_server.py
@@ -131,7 +131,7 @@ class SystemVerilogLanguageServer(SolidLanguageServer):
             os.chmod(executable_path, 0o755)
             return executable_path
 
-        def _create_launch_command(self, core_path: str) -> list[str] | str:
+        def _create_launch_command(self, core_path: str) -> list[str]:
             return [core_path]
 
     @staticmethod


mareurs:
Draft for now — the Kotlin LSP tests pass but I'd like to do some more manual testing with real Kotlin projects before marking this ready.

src/solidlsp/ls_process.py
```diff
@@ -36,6 +36,14 @@
...
 log = logging.getLogger(__name__)
 
+# LSP notification methods that are safe to ignore without logging
+# These are informational notifications that don't require handlers
+IGNORABLE_NOTIFICATIONS = {
+    "$/logTrace",  # Kotlin LSP verbose logging
+    "$/progress",  # Progress notifications
+    "$/cancelRequest",  # Cancellation notifications
+}
+
```
opcode81: This is not a concept we want to introduce. Just handle them (with dummy handlers).

mareurs: Done — removed IGNORABLE_NOTIFICATIONS and registered do_nothing handlers for `$/logTrace` and `$/cancelRequest` in KotlinLanguageServer, matching the pattern used by other language servers.

src/solidlsp/language_servers/kotlin_language_server.py
```diff
@@ -34,16 +37,51 @@
...
-class KotlinRuntimeDependencyPaths:
-    """
-    Stores the paths to the runtime dependencies of Kotlin Language Server
-    """
-
-    java_path: str
-    java_home_path: str
-    kotlin_executable_path: str
+# Kotlin Language Server version
+KOTLIN_LSP_VERSION = "261.13587.0"
```
MischaPanch: While we're at it, this should be in kotlin_settings. So the var renamed to DEFAULT_KOTLIN_LSP_VERSION and in the download `kotlin_lsp_version = kotlin_settings.get("kotlin_lsp_version", DEFAULT_KOTLIN_LSP_VERSION)`.

The java dependencies should also become configurable, though I'm not sure what's the release vs jre versioning scheme there. We document everything that's configurable for the various LS in `050_configuration.md` under `Language Server-Specific Settings`. I suspect claude code can deal with this request :)

src/solidlsp/language_servers/kotlin_language_server.py
```diff
@@ -55,140 +93,99 @@ def __init__(self, config: LanguageServerConfig, repository_root_path: str, soli
...
+            """Provides JAVA_HOME and JVM options for the Kotlin Language Server process."""
+            env: dict[str, str] = {}
+
+            if self._java_home_path is not None:
+                env["JAVA_HOME"] = self._java_home_path
+
+            # Get JVM options from settings or use default
+            jvm_options = DEFAULT_KOTLIN_JVM_OPTIONS
+            custom_jvm_options = self._custom_settings.get("jvm_options", "")
+            if custom_jvm_options:
```
MischaPanch: is it unthinkable that a user wants to just set to no options by providing an empty string? We treat empty string as None here.

src/solidlsp/language_servers/kotlin_language_server.py
mareurs: Done — renamed to `DEFAULT_KOTLIN_LSP_VERSION` and it's now configurable via `kotlin_settings.get("kotlin_lsp_version", DEFAULT_KOTLIN_LSP_VERSION)`.

Also added a Kotlin section to `050_configuration.md` documenting all three settings (`ls_path`, `kotlin_lsp_version`, `jvm_options`).

Re: making Java dependencies configurable — I'll leave that for a follow-up since the versioning scheme (RedHat JRE bundled in vsix archives) is quite different from the Kotlin LSP version. Happy to tackle it if there's demand.

src/solidlsp/language_servers/kotlin_language_server.py
mareurs: Good catch — fixed. Now uses a sentinel to distinguish "key not set" (use default `-Xmx4G`) from "key set to empty string" (no JVM options). So `jvm_options: ""` in the config will explicitly clear the default.

docs/02-usage/050_configuration.md
```diff
@@ -210,9 +210,28 @@ ls_specific_settings:
 ```
 
 This is supported by all language servers deriving their dependency provider from  `LanguageServerDependencyProviderSinglePath`.
-Currently, this includes the following languages: `bash`, `clojure`, `cpp`, `markdown`, `php`, `php_phpactor`, `python`, `rust`, `toml`, `typescript`, `yaml`. 
+Currently, this includes the following languages: `bash`, `clojure`, `cpp`, `kotlin`, `markdown`, `php`, `php_phpactor`, `python`, `rust`, `toml`, `typescript`, `yaml`.
 We will add support for more languages over time.
 
+#### Kotlin
```
opcode81: Languages should follow alphabetical order, i.e. Kotlin does not come first.

mareurs:
Need to do some tests

mareurs:
## CI Kotlin Test Failure Investigation

I've done some investigation into the CI failures on this PR. Here's what I found:

### Pattern

The failure pattern is **identical on all three platforms** (Ubuntu, Windows, macOS):

| Test | Result |
|------|--------|
| `test_find_symbol[kotlin-Model-...]` (1st Kotlin LS start) | **PASSED** |
| `test_find_symbol_references[kotlin-Model-...]` (2nd Kotlin LS start) | **FAILED** |
| All `test_kotlin_basic.py` tests (3rd Kotlin LS start) | **ERROR** |

The Kotlin LSP **starts fine the first time**, but consistently crashes on the **second startup** with `cancelled (-32800)` after only ~3-5 seconds.

All non-Kotlin tests pass. The overall results are: `1 failed, 837 passed, 54 skipped, 3 errors` (Ubuntu).

### Root Cause Analysis

The `serena_agent` fixture in `test_serena_agent.py` is **function-scoped** (default) — each parametrized test gets a fresh `SerenaAgent`, which creates and destroys the Kotlin LSP JVM each time. The shutdown chain is:

```
agent.shutdown(timeout=5)
  → Project.shutdown(timeout=5)
    → ls_manager.stop_all(timeout=5)
      → ls.stop(shutdown_timeout=5)
        → _shutdown(timeout=5): SIGTERM → wait(5s) → SIGKILL if needed
```

On CI runners (2 CPUs, 7GB RAM), the IntelliJ-based Kotlin LSP JVM (~2GB heap) fails on restart. The `cancelled (-32800)` is sent by the **server** during initialization — the IntelliJ engine itself refuses to initialize, likely because:
- Previous JVM resources haven't been fully reclaimed by the OS
- The JVM startup competes with other processes for the limited 2-CPU/7GB runner

I confirmed:
- **Not reproducible locally** (tested 3 restart cycles — all pass on a 64-CPU machine)
- IntelliJ already creates **isolated temp dirs** (`/tmp/idea-system<random>`), so system dir conflicts aren't the issue
- The `-Xmx2G` fix helps keep memory pressure lower but doesn't resolve the restart crash
- The 120s request timeout is irrelevant — the server cancels its own initialization in ~3s

### Suggested Fix

The lightest fix with least upstream impact would be to skip/xfail the Kotlin tests in CI until the Kotlin LSP matures beyond pre-alpha. The test_find_symbol result (1st start) proves the integration works — the failure is only in the JVM restart pattern under resource pressure.

Alternatively, the `serena_agent` fixture could be made **module-scoped** (or session-scoped) for Kotlin tests to avoid the restart problem. This would keep the same Kotlin LSP JVM alive across all `test_serena_agent.py` Kotlin tests. However, this is a bigger fixture refactor and might have side effects for other languages.

MischaPanch:
The skip fix didn't work, unfortunately, since the error happens during the initialization of the fixture, leading to an exception before the skip

src/solidlsp/language_servers/kotlin_language_server.py
```diff
@@ -55,140 +94,104 @@ def __init__(self, config: LanguageServerConfig, repository_root_path: str, soli
...
         super().__init__(
-            config, repository_root_path, ProcessLaunchInfo(cmd=cmd, env=proc_env, cwd=repository_root_path), "kotlin", solidlsp_settings
+            config,
+            repository_root_path,
+            None,
+            "kotlin",
+            solidlsp_settings,
         )
+        # Kotlin LSP (JVM + IntelliJ engine) is slow to start, especially on CI runners
+        self.set_request_timeout(120.0)
```
MischaPanch: this is the timeout for all requests, it goes beyond initialization. We'd need to either wait for an event that reliably signals that initialization finished, or a new concept "set_timeout_for_first_request" or something like that.

test/serena/test_serena_agent.py
```diff
@@ -126,7 +126,8 @@ class TestSerenaAgent:
     def test_find_symbol(self, serena_agent: SerenaAgent, symbol_name: str, expected_kind: str, expected_file: str):
         # skip flaky tests in CI
         # TODO: Revisit the flaky tests and re-enable once the LS issues are resolved #1039
-        flaky_languages = {Language.FSHARP, Language.RUST}
+        # Kotlin: IntelliJ-based JVM crashes on restart under CI resource constraints (2 CPUs, 7GB RAM)
+        flaky_languages = {Language.FSHARP, Language.RUST, Language.KOTLIN}
```
MischaPanch: see my comment above, doesn't help. Instead one, could conditionally add pytest.mark.skip to the param instantiation for the test, that should help. If it doesn't, pytest.mark.kotlin should be replaced with pytest.mark.skip in CI.

mareurs:
@MischaPanch Addressed both review comments:

**Timeout (kotlin_language_server.py:105):** Removed `set_request_timeout(120.0)` entirely. Instead, we now track `window/workDoneProgress/create` + `$/progress` notifications to wait for indexing completion after `initialized()`. The `_indexing_complete` event starts SET (handles older KLS that don't send progress), gets cleared on `workDoneProgress/create`, and set again when all progress tokens end. Also added `_get_wait_time_for_cross_file_referencing()` override (1s safety buffer).

**Test skip (test_serena_agent.py:130):** Moved the Kotlin skip from inside the test body (never reached — crash during fixture setup) to `pytest.param` marks using `pytest.mark.skip` conditioned on `is_ci`.

This merges #1063 into this PR as you suggested. Closing #1063.

MischaPanch:
Thank you, merging!

PR #1062, opened 2026-02-20
fix: preserve config order in LanguageServerManager for directory LS selection

mareurs:
## Summary

- Fix `LanguageServerManager.from_languages()` populating its internal dict in thread-completion order instead of config order
- For multi-language projects, `get_language_server(directory_path)` picked the wrong LS because directories bypass the file extension check, so the first LS in dict iteration order was selected
- Reorder the dict to match the original config order after all parallel threads complete

**Impact:** Affects all multi-language projects where slower LSes (e.g. Kotlin ~7s) are configured as primary but faster ones (e.g. Python <1s) finish first. `request_full_symbol_tree` would return `{}` for the primary language's files.

## Test plan

- [x] `uv run poe test -m kotlin` — all 6 tests pass
- [x] Manual testing with multi-language project (kotlin+python+typescript+bash)

### DIFF
--- a/src/serena/ls_manager.py
+++ b/src/serena/ls_manager.py
@@ -117,7 +117,10 @@ class LanguageServerManager:
             failure_messages = "\n".join([f"{lang.value}: {e}" for lang, e in exceptions.items()])
             raise Exception(f"Failed to start language servers:\n{failure_messages}")
 
-        return LanguageServerManager(language_servers, factory)
+        # Reorder to match the original config order (threads complete in arbitrary order,
+        # but get_language_server iterates in dict order to pick the right LS for a path)
+        ordered = {lang: language_servers[lang] for lang in languages}
+        return LanguageServerManager(ordered, factory)
 
     def get_root_path(self) -> str:
         return self._root_path


MischaPanch:
Ah, I saw this issue before and fixed it, but then we never merged the fix because the core issue cannot happen since in all our tool calls `relative_path` is ensured to be a path of a file and not of a directory.  How did you run into this? Are you using Serena not as MCP server but as something else?

@opcode81 you will remember our conversation on this. Not sure why we haven't merged it, I guess it was part of a bigger change that we decided not to merge. A proper fix would be to raise an error if a directory is passed, without knowing the extension we can't know the LS to return.

mareurs:
@MischaPanch Good question — I hit this through `get_symbols_overview` with a directory path (which was part of a separate fix to add directory support to that tool). When `get_symbols_overview("src/main/kotlin/some_package/")` is called on a multi-language project (kotlin+python+typescript+bash), `get_language_server()` receives a directory path. Since `is_ignored_path()` skips the file extension check for directories, every LS "accepts" it, and the first one in dict iteration order is selected. Because Kotlin LSP takes ~7s to start vs <1s for Python, Python was always first → it skipped all `.kt` files → empty tree → `{}`.

I agree that raising an error for directory paths would be the proper fix. This reordering is more of a band-aid. Happy to rework it if you'd prefer the error approach, or close it if you'd rather handle it differently.

opcode81:
@mareurs using directory paths for the `get_symbols_overview` tool is intentionally not supported.
@MischaPanch I think we decided that the low-level interface didn't need to be updated, because the high-level interfaces (tools) did not allow incorrect usage anyhow - so we decided to postpone changes to the low-level API.

MischaPanch:
Fixed in 1db52b16bf by raising a ValueError if there are multiple LS and a directory is passed

PR #1063, opened 2026-02-20
fix: wait for Kotlin LSP indexing before serving cross-file references

mareurs:
## Summary

- `find_referencing_symbols` returned empty results on large Kotlin projects (~646 files) while working on small test repos
- Root cause: Kotlin LSP (IntelliJ-based) needs time after `initialized()` to index the workspace, but there was no mechanism to wait for it
- Track `$/progress` notifications to detect when indexing completes
- Handle `window/workDoneProgress/create` requests (was returning `MethodNotFound` error)
- Wait up to 120s for indexing completion after `initialized()`
- Override `_get_wait_time_for_cross_file_referencing()` to 1s (safety buffer, since indexing is already awaited)

**Depends on:** #1061 (builds on the refactored Kotlin LS code)

## Test plan

- [x] `uv run poe test -m kotlin` — all 6 tests pass (including `find_referencing_symbols`)
- [x] Manual testing on 646-file backend-kotlin project via Serena MCP

### DIFF
--- a/src/solidlsp/language_servers/kotlin_language_server.py
+++ b/src/solidlsp/language_servers/kotlin_language_server.py
@@ -20,7 +20,10 @@ import logging
 import os
 import pathlib
 import stat
+import threading
 from typing import cast
+
+from overrides import override
 
 from solidlsp.ls import (
     LanguageServerDependencyProvider,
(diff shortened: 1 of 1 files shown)


mareurs:
Depends on #1061 — will rebase onto main once that merges.

MischaPanch:
@mareurs this seems to address the comment I made in your other PR, on waiting for initialization instead of setting a large hardcoded timeout for all requests. Should we abandon the other PR and switch to this one? Alternatively, you can just merge this into the other branch and we finish both extensions in a single PR

mareurs:
@MischaPanch Good call — merged this into #1061 as you suggested.

The indexing wait logic is now in the `kotlin-lsp-v261` branch with a small improvement: `_indexing_complete` starts SET (so older KLS versions that never send `$/progress` don't hang for 120s — `wait()` returns immediately).

Also addressed the test skip by moving it to `pytest.param` marks so it takes effect before fixture setup.

Closing this in favor of #1061.

PR #1067, opened 2026-02-22
support project-specific language server settings

michalfapso:
**Why:**
`ls_specific_settings` was supported only in the global Serena config file. But it may be practical to be able to set these also in `.serena/project.yml` which can be also tracked in project's git repo. And it's also more consistent with the idea of having global Serena settings which can be overridden by project specific settings.

In my particular use case I have a c++ project compiled on windows in MSYS2 mingw64 environment using its clang++ compiler. It compiles well with it, but fails with Serena's internal clangd because of missing system include paths (and maybe some other things as well). There are 2 ways to make it work: use the clangd installed in msys or override `--query-driver` to point to the msys clang++ compiler. I saw that setting `ls_path` was already available in Serena, I just overlooked that it worked only via the global serena config file.

**Summary:**
Allows configuring language-specific settings (e.g., custom "ls_path") directly within the project's ".serena/project.yml" file.
- Add ls_specific_settings field to ProjectConfig.
- Centralize global and project setting merging in `Project.create_language_server_manager`.
- Ensure case-insensitivity for language tags in YAML (e.g., 'cpp' vs 'CPP').
- Update project template to include the new field.
- Add unit tests verifying merging logic and configuration parsing.

Disclaimer: AI agent was involved in this work

### DIFF
--- a/src/serena/config/serena_config.py
+++ b/src/serena/config/serena_config.py
@@ -181,7 +181,7 @@ class ProjectConfig(SharedConfig):
     ignore_all_files_in_gitignore: bool = True
     initial_prompt: str = ""
     encoding: str = DEFAULT_SOURCE_FILE_ENCODING
-    ls_specific_settings: dict[Language, dict[str, Any]] = field(default_factory=dict)
+    ls_specific_settings: dict = field(default_factory=dict)
     """
     Advanced configuration option allowing to configure language server implementation specific options, see SolidLSPSettings for more info.
     """
@@ -348,16 +348,6 @@ class ProjectConfig(SharedConfig):
             if symbol_info_budget < 0:
                 raise ValueError(f"symbol_info_budget cannot be negative, got: {symbol_info_budget}")
 
-        # Convert ls_specific_settings keys to Language enums
-        ls_specific_settings_dict = data.get("ls_specific_settings", {})
-        ls_specific_settings: dict[Language, dict[str, Any]] = {}
-        for k, v in ls_specific_settings_dict.items():
-            try:
-                ls_specific_settings[Language(k.lower())] = v
-            except ValueError:
-                # ignore invalid language names
-                pass
-
         return cls(
             project_name=data["project_name"],
             languages=languages,
@@ -372,7 +362,7 @@ class ProjectConfig(SharedConfig):
             base_modes=data["base_modes"],
             default_modes=data["default_modes"],
             symbol_info_budget=symbol_info_budget,
-            ls_specific_settings=ls_specific_settings,
+            ls_specific_settings=data.get("ls_specific_settings", {}),
         )
 
     def _to_yaml_dict(self) -> dict:
@@ -381,7 +371,6 @@ class ProjectConfig(SharedConfig):
         """
         d = dataclasses.asdict(self)
         d["languages"] = [lang.value for lang in self.languages]
-        d["ls_specific_settings"] = {lang.value: settings for lang, settings in self.ls_specific_settings.items()}
         return d
 
     @classmethod
@@ -484,7 +473,7 @@ class RegisteredProject(ToStringMixin):
         """
         return self.project_root.samefile(Path(path).resolve())
 
-    def get_project_instance(self, serena_config: "SerenaConfig | None") -> "Project":
+    def get_project_instance(self, serena_config: "SerenaConfig") -> "Project":
         """
         Returns the project instance for this registered project, loading it if necessary.
         """
--- a/src/serena/project.py
+++ b/src/serena/project.py
@@ -2,7 +2,6 @@ import json
 import logging
 import os
 import threading
-from copy import deepcopy
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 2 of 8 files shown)


michalfapso:
I've just found #1060 which seems to be related.

MischaPanch:
Hi @michalfapso . #1060 is not all too strongly related.

Do you actually need the per-project override? I didn't understand from your description why the global config option won't work for you.

src/serena/config/serena_config.py
```diff
@@ -344,6 +348,16 @@ def _from_dict(cls, data: dict[str, Any]) -> Self:
             if symbol_info_budget < 0:
                 raise ValueError(f"symbol_info_budget cannot be negative, got: {symbol_info_budget}")
 
+        # Convert ls_specific_settings keys to Language enums
+        ls_specific_settings_dict = data.get("ls_specific_settings", {})
```
MischaPanch: Why is this needed? We're not currently doing that in the other config class.

I guess it's a rather good thing since we get a better type validation. But it should be done in both config classes. Pls add this field to the SharedConfig and a static util method there `_parse_ls_specific_setttings(...: dict[str, dict[str, Any]] -> dict[Language, dict[str, Any]]`

src/serena/project.py
MischaPanch: This is confusing behavior now. As far as I can see, `serena_config` is always passed to Project instantiation except in tests. This means also ls_specific_settings is always passed, except in the `index_file` CLI command where it also should have been passed (it's a bug).

I propose to remove `ls_specific_settings` from here and make `serena_config` non-optional in `Project`. Tests should be adjusted to pass a suitable config - we actually already use such a config for some of the tests.

Then the behavior is always clear - ls_specific_settings is always a merged variant of the global and project entry. 

The `Project` class should get a method `get_ls_specific_settings` that performs the merge

src/serena/config/serena_config.py
```diff
@@ -344,6 +348,16 @@ def _from_dict(cls, data: dict[str, Any]) -> Self:
...
+        ls_specific_settings: dict[Language, dict[str, Any]] = {}
+        for k, v in ls_specific_settings_dict.items():
+            try:
+                ls_specific_settings[Language(k.lower())] = v
+            except ValueError:
```
MischaPanch: An unsupported language should lead to a crash and not be silently ignored

src/serena/resources/project.template.yml
```diff
@@ -114,3 +114,7 @@ initial_prompt: ""
 # override of the corresponding setting in serena_config.yml, see the documentation there.
 # If null or missing, the value from the global config is used.
 symbol_info_budget:
+
+# Advanced configuration option allowing to configure language server implementation specific options.
```
MischaPanch, line 118: pls document that these will enhance/update the global settings and not override them. The merging behavior has to be explained clearly

MischaPanch (changes requested):

src/serena/resources/project.template.yml
opcode81: Actually, I think this should be a per-language override: If a language has a setting in the project, then only the project's settings should apply. I don't think the merging behaviour is truly useful - and merges only happen at the top level, while some languages have nested configurations which then won't be merged, etc.

src/serena/config/serena_config.py
opcode81: Let's keep it simple and do it in neither.

MischaPanch: That also works :D

src/serena/resources/project.template.yml
MischaPanch: Agree, especially because of the nesting. Full per-language override, and should be documented as such

src/serena/project.py
opcode81: >  This is confusing behavior now. As far as I can see, serena_config is always passed to Project instantiation except in tests. This means also ls_specific_settings is always passed, except in the index_file CLI command where it also should have been passed (it's a bug).

What do you mean? I don't see this bug. It is passed in `index_file` also.

MischaPanch: image

opcode81: Ah, OK, I thought you meant that `SerenaConfig` is not passed, but it is.

opcode81: Indeed, this can be simplified by making `SerenaConfig` compulsory.

michalfapso:
Thank you both @MischaPanch and @opcode81 for all your comments and insights.

> Do you actually need the per-project override?

Not necessarily. When I was configuring Serena to work with my setup (windows + msys2 mingw64 + clangd) I assumed that the `ls_specific_settings` can be set in `.serena/project.yml` and when it didn't work, setting it in the global config didn't cross my mind. So I started debugging it using AI and then it came with the solution revealing that this particular setting wasn't supported in the per-project config. I thought it was a bug. But maybe it was intentional :) In that case just adding an explicit "this setting is supported only via the global config `serena_config.yml`, not via per-project `.serena/project.yml`" in [Usage/Configuration](https://oraios.github.io/serena/02-usage/050_configuration.html) or something similar could be sufficient as well. I also failed to understand which settings are settable via the global config and which via the project config.

Now I see that the SharedConfig is quite empty. I thought that most settings are settable via both configs. Anyway, if you find my change being an unnecessary complication, feel free to discard it :) I completely understand that keeping your code simple is important and introducing unnecessary complications should be avoided.

MischaPanch:
There are now too many conflicts, I am closing this PR and opening an issue to track this. Low priority, but maybe we want to do that some day. It would be very easy to implement this after the recent changes

PR #1072, opened 2026-02-23
fix: Kotlin LSP synthetic info fallback for symbols without hover

### LINKED ISSUE #924: Kotlin LS seems to not provide hover info for a struct without a docstring
Our testsuite gives:

```
FAILED test/serena/test_serena_agent.py::TestSerenaAgent::test_find_symbol[kotlin-Model-Struct-Model.kt] - AssertionError: Expected symbol info to be present for symbol: {'name_path': 'Model', 'kind': 'Struct', 'body_location': {'start_line': 2, 'end_line': 2}, 'relative_path': 'src/main/kotlin/test_repo/Model.kt'}
```

I have now disabled the test so CI can run through

mareurs:
## Summary

The Kotlin LSP (KLS v0.1, IntelliJ-based) advertises `hoverProvider: true` but returns `null` for every hover request on Kotlin symbols (verified via wire-level tracing — not a timing issue, consistently null). This PR adds a source-based fallback that reads the actual Kotlin source file to extract declaration signatures and KDoc comments.

### Changes

1. **Source-based info extraction** — reads the Kotlin source file using the symbol's LSP range, extracts the declaration signature (up to opening `{`) and any preceding KDoc comment. Capped at 500 chars to avoid token bloat. Falls back to metadata synthesis if the source can't be read.
2. **Simplified `_get_symbol_metadata_info` API** — accepts the full symbol and parent dicts instead of individual fields, giving subclasses access to ranges and other metadata.
3. **Removed hover retry logic** — retries don't help; KLS v0.1 returns null consistently (not lazy-loading like Eclipse JDTLS).
4. **Removed test workaround** — the `test_find_symbol[kotlin-Model-Struct-Model.kt]` early return is no longer needed.

### Test plan

- [x] `test_find_symbol[kotlin-Model-Struct-Model.kt]` passes with workaround removed
- [x] All 6 Kotlin-marked tests pass
- [x] Format, type-check, lint all clean

Fixes #924

### DIFF
--- a/src/solidlsp/language_servers/kotlin_language_server.py
+++ b/src/solidlsp/language_servers/kotlin_language_server.py
@@ -21,10 +21,12 @@ import os
 import pathlib
 import stat
 import threading
+from time import sleep
 from typing import cast
 
 from overrides import override
 
+from solidlsp import ls_types
 from solidlsp.ls import (
     LanguageServerDependencyProvider,
     LanguageServerDependencyProviderSinglePath,
@@ -554,3 +556,22 @@ class KotlinLanguageServer(SolidLanguageServer):
     def _get_wait_time_for_cross_file_referencing(self) -> float:
         """Small safety buffer since we already waited for indexing to complete in _start_server."""
         return 1.0
+
+    @override
+    def _request_hover(self, uri: str, line: int, column: int) -> ls_types.Hover | None:
+        """Override to retry hover requests for Kotlin LSP.
+
+        The Kotlin LSP (IntelliJ-based) lazily resolves hover info: the first
+        request after didOpen often returns None while the engine parses the file.
+        A retry loop with increasing patience (matching the pattern used for
+        Eclipse JDTLS) handles this lazy-loading behaviour.
+        """
+        max_retries = 10
+        retry_delay = 0.2  # 200 ms between retries → up to 2 s total wait
+        result = super()._request_hover(uri, line, column)
+        for _ in range(max_retries):
+            if result is not None:
+                break
+            sleep(retry_delay)
+            result = super()._request_hover(uri, line, column)
+        return result
--- a/src/solidlsp/language_servers/kotlin_language_server.py
+++ b/src/solidlsp/language_servers/kotlin_language_server.py
@@ -21,7 +21,6 @@ import os
 import pathlib
 import stat
 import threading
-from time import sleep
 from typing import cast
 
 from overrides import override
@@ -557,25 +556,6 @@ class KotlinLanguageServer(SolidLanguageServer):
         """Small safety buffer since we already waited for indexing to complete in _start_server."""
         return 1.0
 
-    @override
-    def _request_hover(self, uri: str, line: int, column: int) -> ls_types.Hover | None:
-        """Override to retry hover requests for Kotlin LSP.
-
-        The Kotlin LSP (IntelliJ-based) lazily resolves hover info: the first
-        request after didOpen often returns None while the engine parses the file.
-        A retry loop with increasing patience (matching the pattern used for
-        Eclipse JDTLS) handles this lazy-loading behaviour.
-        """
-        max_retries = 10
-        retry_delay = 0.2  # 200 ms between retries → up to 2 s total wait
-        result = super()._request_hover(uri, line, column)
-        for _ in range(max_retries):
-            if result is not None:
-                break
-            sleep(retry_delay)
-            result = super()._request_hover(uri, line, column)
-        return result
-
     # Mapping from LSP SymbolKind to Kotlin keyword / descriptor
     _KOTLIN_KIND_KEYWORD: dict[ls_types.SymbolKind, str | None] = {
         ls_types.SymbolKind.File: None,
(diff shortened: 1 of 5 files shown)


src/solidlsp/language_servers/kotlin_language_server.py
```diff
@@ -554,3 +556,100 @@ def progress_handler(params: dict) -> None:
...
+    ) -> str | None:
+        """Build a Kotlin-style synthetic info string from symbol metadata.
+
+        Used as a fallback when the Kotlin LSP returns null for hover requests.
+        When the LSP provides a ``detail`` field (e.g. a function signature or type),
+        it is incorporated into the output.  Otherwise a concise descriptor is
+        synthesized from the symbol's kind and parent, e.g.
+        ``fun solve(): Unit [in Stage1SolverService]`` or ``data class Model``.
+        """
+        keyword = self._KOTLIN_KIND_KEYWORD.get(kind)
```
MischaPanch, line 667: Thanks! We will likely switch to an overall robust tree-sitter based fallback in a relatively near future. But for now, I think this can already be improved using relatively simple regex - at least for strongly structured languages with braces like Kotlin. Since it's a fallback, it's also not too bad if this goes wrong (as long as the info is never too long).

I propose to write a few kind-dependent regexes that will capture from the start of any potential javadoc till the start of the body. The ranges can help here. I think an agent should easily excel at generating a bunch of examples for a test and continue writing regexes until the test passes. If my intuition is right, very few simple regexes will do the job reliably enough.

src/solidlsp/ls.py
```diff
@@ -1599,6 +1599,28 @@ def _request_hover(self, uri: str, line: int, column: int) -> ls_types.Hover | N
             return None
         return ls_types.Hover(**response)  # type: ignore
 
+    def _get_symbol_metadata_info(
+        self,
+        name: str,
+        kind: ls_types.SymbolKind,
```
MischaPanch: why not pass the whole symbol and the whole parent?

test/solidlsp/kotlin/test_kotlin_basic.py
```diff
@@ -55,3 +55,63 @@ def test_overview_methods(self, language_server: SolidLanguageServer) -> None:
...
+        if isinstance(contents, dict):
+            text = contents.get("value", "")
+        elif isinstance(contents, list):
+            text = " ".join(p if isinstance(p, str) else p.get("value", "") for p in contents)
+        else:
+            text = str(contents)
+        assert text.strip(), "Hover text is empty"
+
+    @pytest.mark.parametrize("language_server", [Language.KOTLIN], indirect=True)
+    def test_dir_overview(self, language_server: SolidLanguageServer) -> None:
```
MischaPanch, line 60: dir overview is something @opcode81 and I decided not to support on tool level. The danger of death by tokens is too large. It is still supported on the lower level interfaces, so no need to delete the test. Just FYI

MischaPanch:
Some statements in the PR description are incorrect. textDocument/hover is implemented for Kotlin and does work sometimes.

Have you verified that the retries actually do something? For Java I checked manually that the LS started giving different results after a while (which is mindblowingly crazy...) - did they really build the same behavior into the much more modern Kotlin LS?

src/solidlsp/language_servers/kotlin_language_server.py
mareurs: Implemented source-based extraction: reads the file using the symbol's LSP range, extracts from KDoc start through declaration (up to opening brace). No regex needed — line scanning with the range is enough. Falls back to metadata synthesis if the file can't be read. Capped at 500 chars.

src/solidlsp/ls.py
mareurs: Fixed — now passes the full symbol and parent dicts. This also gives the Kotlin override access to the range for source-based extraction.

test/solidlsp/kotlin/test_kotlin_basic.py
mareurs: Noted, keeping the test since it exercises the lower-level interface.

mareurs:
Fixed the PR description — you're right, KLS advertises `hoverProvider: true` in capabilities, so hover IS implemented. It just returns `null` for every symbol we tested (verified with `--trace-lsp-communication`: the wire response is `{'result': None}` consistently, immediately, for all symbols across all files).

This is not a lazy-loading issue like JDTLS — the retries never helped. Removed them. The new approach reads the source file directly using the symbol's LSP range instead of relying on hover at all.

mareurs:
Some more detail on the investigation — we did extensive layer-by-layer debugging:

**Layer 1 — Serena** ✅
- `include_info=True` correctly calls `request_hover(line, col)` using `selectionRange` positions
- `selectionRange` is present in all Kotlin symbols, so line/column are never None
- Same code path works perfectly for Python (pyright returns full signatures + docstrings)

**Layer 2 — JetBrains KLS v0.1** ❌ (root cause)
- KLS advertises `"hoverProvider": true` in capabilities
- But returns `{'result': None}` for **every** hover request on Kotlin symbols
- This happens immediately and consistently — not a timing/indexing issue
- Wire trace: `solidlsp → KLS: textDocument/hover @ line N col M → KLS → solidlsp: {'result': None}`
- Retried up to 10x with 200ms delays — never succeeded for any symbol

So hover is advertised but not functional in KLS v0.1. The source-based extraction bypasses this entirely by reading the file directly.

AmirF194 (commented):
`_extract_source_info` in `src/solidlsp/language_servers/kotlin_language_server.py:624-639` scans backward for a KDoc by checking only that a line ends with `*/`, then walking up to the nearest `/**` with nothing verifying the lines in between actually stay inside that comment block. A plain one-line `/* ... */` comment sitting directly above a declaration (not a KDoc) enters this branch, and if the nearest `/**` above it belongs to an unrelated comment further up the file (a file-header doc, say), everything in between, package/import lines, other classes, gets pulled in as if it were this symbol's documentation.

Verified on PR head 3d918691e2fd8baf659e42d7b8c811aa54de0d63 in a clean Docker container (python:3.12-slim): extracted `_extract_source_info` with `ast.get_source_segment` (the literal committed source, not a re-transcription) and ran it, bound to a stub instance, against a small Kotlin file with a file-level `/** ... */` header, an unrelated `class Unrelated { ... }`, and a plain `/* not a kdoc */` comment right above `class Foo`. The returned "info" for `Foo` was the file header, `package`, `import`, and the entire `Unrelated` block concatenated onto `class Foo`, 284 chars of unrelated content misattributed as its doc.

A cheap fix: only enter the collection branch when the line is `/**`-terminated content specifically (not any `*/`), and stop the backward walk on the first line that looks like real code (`{`, `}`, `class`, `fun`, `import`, `package`) rather than only stopping at `/**` or index 0.

PR #1075, opened 2026-02-24
Fix: the request_info for a batch of symbols was actually never optimized

MischaPanch:


src/serena/symbol.py
```diff
@@ -549,14 +549,19 @@ def __init__(self, ls: SolidLanguageServer | LanguageServerManager, agent: Union
...
         Returns None if no information is available.
+        :param assume_open_file: can be used for optimizing the number of open_file calls in downstream methods.
+            The caller has to ensure that the file is open.
         """
         lang_server = self.get_language_server(relative_file_path)
-        hover_info = lang_server.request_hover(relative_file_path=relative_file_path, line=line, column=column)
+        if assume_open_file:
+            hover_info = lang_server.request_hover_on_open_file(relative_file_path, line, column)
+        else:
+            hover_info = lang_server.request_hover(relative_file_path=relative_file_path, line=line, column=column)
```
opcode81: Consider the more natural solution for this:
Change the interface of `request_hover` to expect an `LSPFileBuffer` instead of a relative path.

This makes it explicit that the open file is a precondition and naturally steers users to save resources by not opening the same file repeatedly.

MischaPanch: You mean everywhere, or the interface of the new `request_hover_on_open_file`? I wouldn't change request_hover, since in many cases one doesn't need the optimization

MischaPanch: If this was Java I'd write an overload, but well...

opcode81: I mean there should only be one method for requesting hover, and it should not use relative path but LSPFileBuffer. The overload would be detrimental, because the resource usage should be explicit rather than implicit.

MischaPanch: Makes the API a bit harder to use in the general case and we already have a bunch of tests which I would then all need to change

MischaPanch: Given that we are the main users, and I don't want to change the tests, I propose to keep request_hover_on_open_file, use the FileBuffer there, and recommend that method in the docstring of request_hover

opcode81: If you want the interface to be flexible, the best approach would be to have 
```
request_hover(relative_path: str, file_buffer: LSPFileBuffer | None = None)
```
and in the method, use
```
with self._open_file_context(relative_path, file_buffer=file_buffer) as fb:
    ...
```
and to make clear in the docstring that the buffer should be passed if multiple requests are made on the same file.

opcode81: Previously, this type of interface was only used for internal methods, so I added no checks, but if we expose it to users of SolidLSP, it could make sense to check that the `file_buffer` being passed to the context manager refers to the same file. @MischaPanch, consider adding this.

MischaPanch: Done, and added the check. Very nice!

src/solidlsp/ls.py
```diff
@@ -1555,28 +1559,29 @@ def request_overview(self, within_relative_path: str) -> dict[str, list[UnifiedS
...
-        with self.open_file(relative_file_path):
-            uri = pathlib.Path(os.path.join(self.repository_root_path, relative_file_path)).as_uri()
-            return self._request_hover(uri, line, column)
+        with self._open_file_context(relative_file_path, file_buffer=file_buffer):
+            return self._request_hover(relative_file_path, line, column)
+
+    def get_uri(self, relative_file_path: str) -> str:
+        return pathlib.Path(os.path.join(self.repository_root_path, relative_file_path)).as_uri()
 
     def _request_hover(self, uri: str, line: int, column: int) -> ls_types.Hover | None:
```
opcode81: Pass the LSPFileBuffer instead of the URI. This makes it explicit that the file must be open; the buffer contains the URI.

MischaPanch: Good idea, done

MischaPanch: I inlined the get_uri method

src/solidlsp/ls.py
```diff
@@ -1555,28 +1557,29 @@ def request_overview(self, within_relative_path: str) -> dict[str, list[UnifiedS
...
+        :param file_buffer: The file buffer to use for the request. If not provided, the file will be read from disk.
+            Can be used for optimizing number of file-reads in downstream code
         """
-        with self.open_file(relative_file_path):
-            uri = pathlib.Path(os.path.join(self.repository_root_path, relative_file_path)).as_uri()
-            return self._request_hover(uri, line, column)
+        with self._open_file_context(relative_file_path, file_buffer=file_buffer):
+            return self._request_hover(relative_file_path, line, column)
+
+    def get_uri(self, relative_file_path: str) -> str:
```
opcode81, line 1576: Let's fully factorise this and turn this into a private method which returns both the absolute path and the URI; then we can use it in `open_file` and have a single place where this logic is defined.

opcode81: I now notice that URI computation is not at all factorised to begin with.
It is done it all over the place.

opcode81: So do what you want, I guess :-)

MischaPanch: Yes, that's why I introduced this method as a basis to clean this up in the future at some point. Since it's not related to the fix, I propose to not do it now. I'll add a separate issue to keep track of this

PR #1076, opened 2026-02-24
[SCALA LSP] Do not reinstall metals if it is globally available on system

cx-alberto-simoes:
## Details

- Before looking for metals in the serena lsp folder, search for it globally, as it might be installed in the system
- Move coursier check inside the if, where it is used.

## Current Issues

I am getting a lot of messages like this one, but I do not feel they are related to my PR.
```
[LSP-stdout-reader:scala] solidlsp.ls_process:_notification_handler:566 - Unhandled method 'window/logMessage'
```


MischaPanch:
Thanks, merging

Issue #1068, opened 2026-02-23
Serena Dashboard: Banner Rotation Is Distracting


tomelgato:
**Is your feature request related to a problem? Please describe.**
The always changing Oraios / Support Box is somehow annoying especially if you have it on a 2nd monitor and you just recognise something changed and you look.....

**Describe the solution you'd like**
Static box(es) with the content of both views

**Describe alternatives you've considered**
Flag to disable the auto toggle, Link for Support, ....

opcode81:
We could consider selecting a box at random during startup and then not changing it or providing buttons to cycle through (maybe as an option).

MischaPanch:
Yes, random box and buttons sounds good. I can take this over

PR #1077, opened 2026-02-24
Dashboard: Replace automatic banner rotation with random initial display and manual rotation

MischaPanch:
Would look like this if we had multiple platinum banners, but since we only have one, arrows are hidden on platinum.

image

Closes: #1068


opcode81:
Excellent. Good to merge.

PR #1082, opened 2026-02-24
fix(kotlin): Kill Gradle daemons on KLS shutdown

### LINKED ISSUE #1081: Kotlin LS spawns heavy zombie processes
See discussion in #1079 for details.

I think we should just always kill the gradle process when shutting down, unless it has some downsides I am missing.

@opcode81 @mareurs @cx-alberto-simoes

mareurs:
## Summary

- KLS triggers Gradle during project indexing, which spawns persistent daemon processes (~500MB RSS each, 3-hour idle timeout)
- These daemons detach from KLS (reparented to init/systemd) and are invisible to `psutil.Process.children(recursive=True)` cleanup
- Override `stop()` in `KotlinLanguageServer` to find and terminate matching Gradle daemons after normal shutdown
- Identification: match the java binary in the daemon's `cmdline[0]` against `JAVA_HOME/bin/java` from this KLS instance

## Root cause investigation

Investigation performed in #1079 (discussion thread) confirmed:

1. **KLS itself cleans up properly** — even under SIGKILL, the stdio pipe EOF causes KLS to exit naturally
2. **The real culprit: Gradle daemons** spawned during KLS project indexing:
   - Run independently (parent = systemd)
   - Use KLS's bundled JRE (`JAVA_HOME` points to `~/.serena/language_servers/static/KotlinLanguageServer/...`)
   - 3-hour default idle timeout
   - ~500MB RSS each
3. **Verified deterministically** via `/proc/PID/environ` and controlled before/after test runs

## Test results

Full test suite run with `pytest -n auto` (32 workers):

- **667 passed**, 82 skipped, 1 xfailed
- **23 failed** — all pre-existing, unrelated to this change:
  - Missing runtimes: .NET 10, Go/gopls, Nix, Ruby-dev, Zig, Perl, Swift
  - Flaky LS tests: Vue cross-file refs, SystemVerilog cross-file, Elm document symbols
  - CLI edge cases: project index/create
- **121 errors** — all missing runtime environments (dotnet, zig, perl, swift, ruby-dev)
- **All 5 Kotlin tests pass** ✓
- mypy type-check: **no issues** ✓

### Gradle daemon before/after verification

| | Before fix | After fix |
|---|---|---|
| New Gradle daemons after `pytest -m kotlin` | 1 new (~500MB) | **0** |
| Pre-existing system Gradle daemons | Untouched | Untouched |

Closes #1081

### DIFF
--- a/src/solidlsp/language_servers/kotlin_language_server.py
+++ b/src/solidlsp/language_servers/kotlin_language_server.py
@@ -23,6 +23,7 @@ import stat
 import threading
 from typing import cast
 
+import psutil
 from overrides import override
 
 from solidlsp.ls import (
@@ -554,3 +555,61 @@ class KotlinLanguageServer(SolidLanguageServer):
     def _get_wait_time_for_cross_file_referencing(self) -> float:
         """Small safety buffer since we already waited for indexing to complete in _start_server."""
         return 1.0
+
+    @override
+    def stop(self, shutdown_timeout: float = 2.0) -> None:
+        java_home = self._get_java_home_from_dependency_provider()
+        super().stop(shutdown_timeout=shutdown_timeout)
+        if java_home:
+            self._kill_gradle_daemons(java_home)
+
+    def _get_java_home_from_dependency_provider(self) -> str | None:
+        if isinstance(self._dependency_provider, self.DependencyProvider):
+            return self._dependency_provider._java_home_path
+        return None
+
+    @staticmethod
+    def _kill_gradle_daemons(java_home: str) -> None:
+        """Kill Gradle daemons spawned by KLS during project indexing.
+
+        KLS triggers Gradle during indexing, which spawns persistent daemon processes that
+        outlive the KLS process (3-hour idle timeout, ~500MB RSS each). These daemons are
+        not children of KLS and are invisible to normal process-tree cleanup.
+
+        We identify them by matching the java executable in their command line against
+        the JAVA_HOME used by this KLS instance.
+        """
+        java_bin = os.path.join(java_home, "bin", "java")
+        try:
+            java_bin_resolved = os.path.realpath(java_bin)
+        except OSError:
+            return
+
+        killed: list[psutil.Process] = []
+        for proc in psutil.process_iter(["pid", "cmdline"]):
+            try:
+                cmdline = proc.info["cmdline"]
+                if not cmdline:
+                    continue
+                if not any("GradleDaemon" in arg for arg in cmdline):
+                    continue
+                # cmdline[0] is the java binary path used to start the daemon
+                try:
+                    proc_java_resolved = os.path.realpath(cmdline[0])
+                except OSError:
+                    continue
+                if proc_java_resolved == java_bin_resolved:
+                    log.info("Terminating Gradle daemon (PID %d) spawned by KLS", proc.pid)
+                    proc.terminate()
+                    killed.append(proc)
+            except (psutil.NoSuchProcess, psutil.AccessDenied, psutil.ZombieProcess):
+                continue
+
+        if killed:
+            _, alive = psutil.wait_procs(killed, timeout=5)
+            for proc in alive:
+                log.warning("Gradle daemon (PID %d) did not terminate gracefully, killing", proc.pid)
+                try:
+                    proc.kill()
+                except (psutil.NoSuchProcess, psutil.AccessDenied):
+                    pass
--- a/src/solidlsp/language_servers/kotlin_language_server.py
+++ b/src/solidlsp/language_servers/kotlin_language_server.py
@@ -558,15 +558,11 @@ class KotlinLanguageServer(SolidLanguageServer):
 
     @override
     def stop(self, shutdown_timeout: float = 2.0) -> None:
-        java_home = self._get_java_home_from_dependency_provider()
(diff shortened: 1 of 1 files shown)


src/solidlsp/language_servers/kotlin_language_server.py
```diff
@@ -554,3 +555,61 @@ def progress_handler(params: dict) -> None:
...
         return 1.0
+
+    @override
+    def stop(self, shutdown_timeout: float = 2.0) -> None:
+        java_home = self._get_java_home_from_dependency_provider()
+        super().stop(shutdown_timeout=shutdown_timeout)
+        if java_home:
+            self._kill_gradle_daemons(java_home)
+
+    def _get_java_home_from_dependency_provider(self) -> str | None:
```
MischaPanch: neither this nor the None check for java_home above seem necessary

src/solidlsp/language_servers/kotlin_language_server.py
```diff
@@ -554,3 +555,61 @@ def progress_handler(params: dict) -> None:
...
+
+        KLS triggers Gradle during indexing, which spawns persistent daemon processes that
+        outlive the KLS process (3-hour idle timeout, ~500MB RSS each). These daemons are
+        not children of KLS and are invisible to normal process-tree cleanup.
+
+        We identify them by matching the java executable in their command line against
+        the JAVA_HOME used by this KLS instance.
+        """
+        java_bin = os.path.join(java_home, "bin", "java")
+        try:
```
MischaPanch: this also seems too defensive. When could realpath go wrong?

MischaPanch:
@mareurs 
>These daemons detach from KLS (reparented to init/systemd) and are invisible to psutil.Process.children(recursive=True) cleanup

Isn't this completely f*d up?

src/solidlsp/language_servers/kotlin_language_server.py
```diff
@@ -554,3 +555,61 @@ def progress_handler(params: dict) -> None:
     def _get_wait_time_for_cross_file_referencing(self) -> float:
         """Small safety buffer since we already waited for indexing to complete in _start_server."""
         return 1.0
+
+    @override
+    def stop(self, shutdown_timeout: float = 2.0) -> None:
+        java_home = self._get_java_home_from_dependency_provider()
+        super().stop(shutdown_timeout=shutdown_timeout)
+        if java_home:
```
MischaPanch: We can't do that in general once (if) we allow users to use their system java. The only safe way to kill the gradle daemons is to know they correspond to Serena's own java. So we will need to wait for #1079 to be merged and then to distinguish if we're using Serena's java or another one. In the advanced config docs we can then tell users that they might need to clean zombies by themselves if they enable using system java.

Meanwhile, I will raise an issue in the KLS repo and ask what's up with all of this

opcode81: We should probably close #1079 instead. Being able to reliably terminate the LS is more important than reusing a system JRE. I can't think of a good reason why using the system JRE would be important.

src/solidlsp/language_servers/kotlin_language_server.py
mareurs: Agreed. On main, `_dependency_provider` is always `DependencyProvider` and `_java_home_path` is always set after `_get_or_install_core_dependency()`. Will simplify — access the attribute directly without the isinstance/None guards.

src/solidlsp/language_servers/kotlin_language_server.py
mareurs: Correct, `os.path.realpath()` doesn't raise — it returns the input unchanged if resolution fails. Will remove both try/except blocks.

src/solidlsp/language_servers/kotlin_language_server.py
mareurs: Valid concern. With the bundled JRE, the `JAVA_HOME/bin/java` path is unique to this KLS instance, so matching is safe. With a system JRE, we'd potentially match unrelated Gradle daemons — that's dangerous.

For now this PR only targets the current behavior (bundled JRE). If #1079 is merged in the future and system JRE becomes possible, we'd need to either skip daemon cleanup for system JRE or use additional matching criteria (e.g. the project directory from `/proc/PID/environ`).

Separately, our investigation confirmed that the Gradle Tooling API (used by JetBrains KLS internally) *always* spawns a daemon — there's no `--no-daemon` equivalent. So post-hoc cleanup is the only option.

mareurs:
> Isn't this completely f*d up?

Yes — it's a fundamental constraint of the Gradle Tooling API. The API mandates daemon usage; there's no `--no-daemon` mode. Every IntelliJ-based tool (IDEA, Android Studio, KLS) leaves these daemons behind. Even setting `org.gradle.daemon=false` in `gradle.properties` doesn't affect Tooling API calls. So post-hoc cleanup is the only viable approach.

MischaPanch:
@mareurs I now had another look. Since this was written, we had realized that java doesn't need to be downloaded at all. I have now removed the `java_home` field from the kotlin LS since it was unused in commit 091b2245e.

This means the solution for identifying gradle daemons needs to be adjusted/rewritten. We should use the kotlin LS itself to find out which java is being used and thus which processes it started

opcode81:
Stale.

PR #1093, opened 2026-02-26
Add HLSL language server support (shader-language-server)

TheSmallKiwi:
## Summary
- Add HLSL language server integration using [shader-language-server](https://github.com/antaalt/shader-sense) from antaalt/shader-sense
- Supports `.hlsl`, `.hlsli`, `.fx`, `.fxh`, `.cginc`, `.compute`, `.shader` file extensions
- Auto-downloads pre-built binaries from GitHub releases (Windows x64/ARM64, Linux x64), with fallback to system PATH
- Harden `request_full_symbol_tree()` to catch `SolidLSPException` per-file so a single LS crash doesn't abort the entire workspace scan

## What's included
- `HlslLanguageServer` class with dependency provider, initialize params, and Unity-specific directory ignore list
- `Language.HLSL` enum entry with file matchers and factory method
- 12 tests covering: symbol discovery, go-to-definition (cross-file), hover, and full symbol tree indexing
- Test repo with 4 HLSL files across root and subdirectory, exercising includes, structs, functions, cbuffers, and compute kernels
- Fix `is_ignored_dirname` to call `super()` so dot-directories (`.git`, `.serena`, etc.) are properly excluded

## Test plan
- [x] `uv run poe test -m hlsl` — all 12 tests pass
- [x] `uv run poe type-check` — clean

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -333,5 +333,5 @@ markers = [
 # Ref: https://github.com/codespell-project/codespell#using-a-config-file
 skip = '.git*,*.svg,*.lock,*.min.*'
 check-hidden = true
-# ignore-regex = ''
+ignore-regex = '\.\w+'
 ignore-words-list = 'paket'


src/solidlsp/ls.py
```diff
@@ -1429,47 +1429,62 @@ def process_directory(rel_dir_path: str) -> list[ls_types.UnifiedSymbolInformati
...
-
-                    fix_relative_path(file_root_nodes)
+                        if e.is_language_server_terminated():
+                            log.warning(
+                                "Language server terminated while processing %s. "
+                                "Restarting language server to continue processing remaining files.",
+                                contained_dir_or_file_rel_path,
+                            )
+                            self.stop()
+                            self.start()
```
opcode81: This is not acceptable; this method should not be modified. We should not just catch exceptions and restart the language server in a multitude of places.
Serena already centrally attempts to restart and retry an operation if the LS has crashed:

https://github.com/oraios/serena/blob/827512b4d1ed457c70efbb775b5cc928304b70c2/src/serena/tools/tools_base.py#L273-L284

TheSmallKiwi: Affirmative, forgot to revert that change from troubleshooting.

opcode81 (changes requested):
Thanks for the contribution!

One thing that appears to be missing is a test covering the request_references function, which is exposed by Serena. Does the LS support it?

Note: This PR must be squash-merged, because the history is not clean.

TheSmallKiwi:
> One thing that appears to be missing is a test covering the request_references function, which is exposed by Serena. Does the LS support it?

Right, unfortunately shader-sense doesn’t support references. I can add a test anyways if you like.

MischaPanch:
Looks good @TheSmallKiwi , thanks! Pls extend readme and documentation (adding support to languages) and mention that reference search does not work for HLSL there. Then this can be merged

TheSmallKiwi:
> Looks good @TheSmallKiwi , thanks! Pls extend readme and documentation (adding support to languages) and mention that reference search does not work for HLSL there. Then this can be merged

All done, also enabled WGSL and GLSL (and noted in docs) to avoid future confusion.

PR #1095, opened 2026-02-26
Fix: don't abort all language servers when one fails to start

EarthmanWeb:
## Problem

In multi-language projects, if any one language server fails to start
(e.g. PHP's Intelephense is not installed, or a runtime is missing),
Serena currently stops ALL successfully started language servers and
raises an exception — leaving the user with no language support at all.

Example: a project with both Python and PHP configured. If Intelephense
fails to start, Jedi (Python) is also stopped, even though it started
fine and Python files could still be served.

## Fix

Changed `LanguageServerManager.create()` in `ls_manager.py` to:

1. Log a warning listing which servers failed instead of raising immediately
2. Continue with whichever servers started successfully
3. Only raise if **all** servers failed — in that case there is nothing
   useful to return, so an exception is still appropriate

## Example

Before:
```
Exception: Failed to start language servers:
  php: FileNotFoundError: intelephense not found
# Python LS also stopped — no language support at all
```

After:
```
WARNING: Some language servers failed to start (continuing with available ones):
  php: FileNotFoundError: intelephense not found
# Python LS continues working normally
```

### DIFF
--- a/src/serena/ls_manager.py
+++ b/src/serena/ls_manager.py
@@ -117,10 +117,7 @@ class LanguageServerManager:
             log.warning(f"Some language servers failed to start (continuing with available ones):\n{failure_messages}")
 
         if not language_servers:
-            raise Exception(
-                f"All language servers failed to start:\n"
-                + "\n".join([f"{lang.value}: {e}" for lang, e in exceptions.items()])
-            )
+            raise Exception("All language servers failed to start:\n" + "\n".join([f"{lang.value}: {e}" for lang, e in exceptions.items()]))
 
         return LanguageServerManager(language_servers, factory)
 


opcode81:
The behaviour is intentional. If things "mainly work", errors can remain undetected. Therefore, fail-fast behaviour is preferred.

The assumption is: If the user wants several languages, all of them must work.
If not all of them work, glossing over that is not OK, it needs to be brought to the user's immediate attention. The error must not be hidden in logs the user probably isn't even monitoring.

EarthmanWeb:
Fail Fast should apply to a single LSP not the whole plugin.
I think that's flawed logic.

It's becoming apparent I will just need to continue to use my own improved fork

opcode81:
We have the choice between
* bringing errors to the user's attention, such that they can be fixed, subsequently producing an optimal experience
* glossing over errors and causing a suboptimal experience (which users not monitoring logs may not even notice).

If the user says that he wants to use 2 languages (e.g. Python and TS, as in Serena), it is not OK to just ignore one of them not working. We believe that it is better for the user to make the **explicit decision** to either
 * fix the one that isn't working (probably a requirements or config issue) or
 * disable it (because it wasn't really important to begin with).

If you want a system that just glosses over errors, Serena is not it.

EarthmanWeb:
Why not have it output the error to the LLM?

opcode81:
It is not possible to communicate such errors upon MCP server connection (i.e. with Serena's system prompt), because language server startup is slow and therefore must happen in the background, i.e. when Serena's system prompt is computed and passed to the LLM, the error hasn't happened yet. Synchronous language server startup would cause MCP client connection errors.

The only thing we could do is to establish a "messaging system" where we would pass such information on along with the result of the next tool call, but it seems contrived.

EarthmanWeb:
You seem to like the word contrived.

It would be the best possible way to allow the user to know their LSP has
failed.

It literally took me an hour to finally figure out why nothing was working
in Serena when I had a couple languages enabled and only one was failing

EarthmanWeb:
"Contrived"

"Create or bring about (an object or a situation) by deliberate use of
skill and artifice."

"Middle English: from Old French contreuve-, stressed stem of controver
‘imagine, invent’

It's not necessarily a bad thing...


On Sun, Mar 1, 2026 at 7:05 AM Terrance Orletsky ***@***.***> wrote:
>
> You seem to like the word contrived.
>
> It would be the best possible way to allow the user to know their LSP has
failed.
>
> It literally took me an hour to finally figure out why nothing was
working in Serena when I had a couple languages enabled and only one was
failing
>
> On Sun, Mar 1, 2026 at 6:59 AM Dominik Jain ***@***.***>
wrote:
>>
>> opcode81 left a comment (oraios/serena#1095)
>>
>> It is not possible to communicate such errors upon MCP server connection
(i.e. with Serena's system prompt), because language server startup is slow
and therefore must happen in the background, i.e. when Serena's system
prompt is computed and passed to the LLM, the error hasn't happened yet.
Synchronous language server startup would cause MCP client connection
errors.
>>
>> The only thing we could do is to establish a "messaging system" where we
would pass that information on along with the result of the next tool call,
but it seems contrived.

opcode81:
It would be inappropriate to return an unrelated error along with a tool call that succeeded. The right thing to do is to **make only the LSP-based tools fail and return an informative error**. 

We already do this:

image

Claude clearly tells you to inspect the logs, and if you then open the dashboard, you immediately see the problem in bright red (I enabled `swift`, which doesn't work on my machine):

image

> It literally took me an hour to finally figure out why nothing was working

Please explain the information you were missing that caused the error to be so hard to diagnose.

* Did you not know that you could see the logs in the dashboard?
* If the dashboard was fully disabled (not recommended), did you not know where the [log files](https://oraios.github.io/serena/02-usage/065_logs.html) are?

Things we can think about adding:
* If the dashboard is enabled, the LLM can open the dashboard for the user via a tool call. The LLM could ask the user whether it should do this.
* If the dashboard is disabled, we could directly provide the path to the log file.

opcode81:
What we could do is to capture the error and return the contents of the log message directly for failed LSP tools. It would (mostly) remove the need to look at the logs.

PR #1091, opened 2026-02-26
Feature/global memories

MischaPanch:
Resolves #1006

PR #1096, opened 2026-02-26
Fix EditMemoryTool: remove ReplaceContentTool dependency

EarthmanWeb:
## Problem

`EditMemoryTool.apply()` delegates to `ReplaceContentTool` to perform
its text replacement. This creates an indirect dependency on the LSP
layer: `ReplaceContentTool` routes through the language server to
locate and modify files, which is unnecessary and incorrect for memory
files — memory files live in `.serena/memories/` and are not part of
the indexed project source.

In contexts where the LSP is not active or the memory path is outside
the project root, this causes `EditMemoryTool` to fail entirely.

## Fix

Replaced the `ReplaceContentTool` delegation in `memory_tools.py` with
direct file I/O using Python's standard `re` module:

- Reads the memory file directly via `_find_memory()`
- Performs literal or regex replacement inline
- Writes the result back — no LSP involvement

Behaviour is identical from the caller's perspective: supports both
`"literal"` and `"regex"` modes, returns a descriptive error if the
needle/pattern is not found.

## Example

Before (broken in non-LSP contexts):
```python
# EditMemoryTool.apply() called ReplaceContentTool which required an
# active language server and a path relative to the project root.
# Memory files outside the project root would raise an exception.
```

After:
```python
memory_path = self.memories_manager._find_memory(memory_name)
content = memory_path.read_text(encoding="utf-8")
new_content = content.replace(needle, repl, 1)  # literal mode
memory_path.write_text(new_content, encoding="utf-8")
```

### DIFF
--- a/src/serena/tools/memory_tools.py
+++ b/src/serena/tools/memory_tools.py
@@ -97,8 +97,8 @@ class EditMemoryTool(Tool, ToolMarkerCanEdit):
         :param repl: the replacement string (verbatim).
         :param mode: either "literal" or "regex", specifying how the `needle` parameter is to be interpreted.
         """
-        memory_path = self.memories_manager._find_memory(memory_name)
-        if memory_path is None:
+        memory_path = self.memories_manager.get_memory_file_path(memory_name)
+        if not memory_path.exists():
             return f"Memory file {memory_name} not found."
         content = memory_path.read_text(encoding="utf-8")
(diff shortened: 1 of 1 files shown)


opcode81:
This is already changed in #1091 (to be merged soon), where we factored out the replacement logic to a separate utility class.

PR #1097, opened 2026-02-26
Fix PHP find_symbol: handle flat SymbolInformation

EarthmanWeb:
## Problem

Two related issues cause `find_symbol` to fail or return wrong results in PHP projects:

**1. Incorrect language server routing**
In multi-language projects, `get_language_server()` routed files using
`is_ignored_path()` — a method that checks gitignore and unsupported
extension filters. A PHP file could pass or fail this check for the
wrong language server depending on configuration order, causing PHP
files to be routed to the Python LS (or vice versa).

**2. Flat `SymbolInformation[]` from Intelephense**
For some PHP files, Intelephense returns `SymbolInformation[]` (the
older LSP format) instead of `DocumentSymbol[]`. These symbols have no
parent-child hierarchy — methods appear as root-level symbols with only
a `containerName` string field linking them to their class.

`find_symbol("ClassName/method")` failed silently because
`iter_name_path_components_reversed()` never yielded the container,
so the name-path match never succeeded.

**3. Hard error on explicitly-requested ignored file**
When `get_symbols_overview()` was called on a file that happened to
match an ignore pattern, it raised a hard error and returned `[]`
instead of proceeding. This broke workflows where the caller explicitly
requested a specific file.

## Fix

- **`ls_manager.py`**: Route by `fn_matcher.is_relevant_filename()` —
  checks the file extension against the language's known extensions —
  instead of `is_ignored_path()`.

- **`symbol.py`**: Added `containerName` fallback in
  `iter_name_path_components_reversed()`. When a symbol has no
  ancestors (flat `SymbolInformation`), yield the `containerName` value
  so name-path patterns like `"ClassName/method"` still match.

- **`ls.py`**: Downgraded hard error to a warning when an explicitly
  requested file is ignored; proceed with the symbol request anyway.

## Example

Before:
```python
# In a Python+PHP project:
find_symbol("TestWebhookHandler/handle_request")
# → [] (routed to wrong LS, or containerName not traversed)
```

After:
```python
find_symbol("TestWebhookHandler/handle_request")
# → [<LanguageServerSymbol: handle_request>]
```


src/serena/symbol.py
```diff
@@ -337,8 +337,19 @@ def get_name_path(self) -> str:
...
+            has_yielded_ancestor = True
+        if not has_yielded_ancestor:
+            # Fallback for flat SymbolInformation (e.g. Intelephense returns SymbolInformation[]
+            # instead of DocumentSymbol[] for some PHP files). In that case symbols have no
+            # parent-child links, but may carry a containerName that identifies the enclosing
+            # class or function.  Yield it so that name-path patterns like "ClassName/method"
+            # still match correctly.
+            container_name = self.symbol_root.get("containerName")
+            if container_name:
+                yield NamePathComponent(container_name, None)
```
opcode81: This has to be solved in the language server, not here.

src/serena/ls_manager.py
```diff
@@ -130,13 +130,19 @@ def _ensure_functional_ls(self, ls: SolidLanguageServer) -> SolidLanguageServer:
...
+        language server if no match is found.
+        """
         ls: SolidLanguageServer | None = None
         if len(self._language_servers) > 1:
             if os.path.isdir(relative_path):
                 raise ValueError(f"Expected a file path, but got a directory: {relative_path}")
             for candidate in self._language_servers.values():
-                if not candidate.is_ignored_path(relative_path, ignore_unsupported_files=True):
+                fn_matcher = candidate.language.get_source_fn_matcher()
+                if fn_matcher.is_relevant_filename(relative_path):
```
opcode81: Why did you change this? `is_ignored_path` with `ignore_unsupported_files=True` has the same semantics for files.

https://github.com/oraios/serena/blob/3d4610fef45d962915b0626c696183b654a97f21/src/solidlsp/ls.py#L580-L599

opcode81 (changes requested):

src/serena/ls_manager.py
EarthmanWeb: Changes to is ignored path are leaked over from other changes to allow an option in config to allow dotfiles and to allow gitignored files.  That was negating the flags that were set in the config

src/serena/symbol.py
opcode81: By that I mean the SolidLanguageServer implementation, not Intelephense itself.

src/serena/ls_manager.py
opcode81: The LS implementations ignore only directories starting with ".", not files. For files, the logic I cited applies.

src/serena/symbol.py
EarthmanWeb: Agreed — moved the fix to `Intelephense._request_document_symbols` in ce6ddf3a. The override detects flat `SymbolInformation[]` (by top-level `"location"` key) and calls a new `_reconstruct_document_symbols` static method that rebuilds the proper `DocumentSymbol[]` tree using `containerName` before returning. `symbol.py` is reverted to its original form.

src/serena/ls_manager.py
EarthmanWeb: Both addressed:

**`symbol.py`** — moved to the correct layer in `ce6ddf3a`; `Intelephense` now reconstructs the `DocumentSymbol[]` tree from flat `SymbolInformation[]` before it reaches shared code.

**`ls_manager.py`** — you're right, `is_ignored_path` already does the extension check at ls.py:596–599. Reverted in `50037339`. The bypass came from confusion with changes in another branch that don't actually affect file routing here.

src/serena/symbol.py
opcode81: Great!

This means that all cached raw document symbol data needs to be invalidated, however, and we must increment `cache_version_raw_document_symbols` (which currently uses the default value 1) in the super constructor call to achieve this.

MischaPanch:
@opcode81 @EarthmanWeb what's the status here?

EarthmanWeb:
Rebased on current main and added the cache invalidation fix.

`cache_version_raw_document_symbols` is now set to `2` in Intelephense's super().__init__ call, since `_request_document_symbols` now reconstructs flat `SymbolInformation[]` into `DocumentSymbol[]`, changing the cached data format.

The `get_language_server` routing change was dropped — upstream's `_get_suitable_language_server` already handles extension-based routing correctly.

src/solidlsp/language_servers/intelephense.py
```diff
@@ -203,3 +209,59 @@ def _send_definition_request(self, definition_params: DefinitionParams) -> Defin
...
+                by_name[sym["name"]] = doc_sym
+
+        # Second pass: attach children to their container
+        for sym, doc_sym in converted:
+            container = sym.get("containerName")
+            if container:
+                if container in by_name:
+                    by_name[container]["children"].append(doc_sym)
+                else:
+                    # Container not found in this file; treat as root
```
AmirF194, line 264: `_reconstruct_document_symbols` only handles one level of nesting: `by_name` is populated exclusively from root symbols in the first pass (`if not sym.get("containerName")`), so a symbol whose `containerName` points to another *non-root* symbol falls into the `else` branch and gets silently promoted to a root, instead of being nested under its real parent.

Docker-verified against this PR's head (`d324e6d`) by calling the static method directly with a 3-level chain (class -> method -> containerName="method"):

```
roots: ['MyClass', 'nested_closure']
  MyClass children: ['my_method']
  nested_closure children: []
```

`nested_closure` should be a child of `my_method`, not a second root. PHP closures/anonymous functions nested inside a method are the realistic case this would hit (the test fixture in this PR already uses WordPress-style callback patterns), though I haven't confirmed what containerName Intelephense actually emits for that case, only that the reconstruction algorithm itself mishandles it when containerName points to a non-root symbol.

Fix would be to add every converted symbol to `by_name`, not just roots, then do a second pass to detect root-vs-nested (or process containers in an order that guarantees a symbol's own container is resolved before it's looked up, e.g. two passes keyed by all symbols first, then reparent).

PR #1098, opened 2026-02-26
Fix Marksman: remap heading symbol kind from String to Namespace

EarthmanWeb:
## Problem

Marksman LSP returns all markdown headings (H1–H6) with `SymbolKind.String` (15). Since `String (15) >= Variable (13)`, `is_low_level()` filtered every heading out of `get_symbols_overview` — no headings appeared at all.

Additionally, Marksman nests headings hierarchically (H2 under H1, H3 under H2, etc.) via `hierarchicalDocumentSymbolSupport`. Even after fixing the kind, `request_document_overview` returns only `root_symbols` (H1), so `get_symbols_overview` with the default `depth=0` would still silently omit H2–H6, forcing agents to fall back to `search_for_pattern` for sub-headings.

## Fix

**1. Remap `SymbolKind.String` → `SymbolKind.Namespace` in `request_document_symbols`**

All heading symbols returned by Marksman are recursively remapped so they pass the `is_low_level()` filter. `Namespace` (3) is semantically appropriate — headings are named sections containing other content. Marksman only uses `String` for headings, so the remap is safe.

**2. Flatten heading hierarchy in `request_document_overview`**

Overrides `request_document_overview` to recursively collect all headings from the full symbol tree and return them as a flat list. This ensures H2–H6 appear directly in `get_symbols_overview` at `depth=0` without requiring depth tricks or pattern search fallbacks.

## Tests

- Verify all heading kinds are remapped to `Namespace` (not `String`)
- Verify no heading is classified as `is_low_level()`
- Verify nested headings in `api.md` (H1–H5) are all remapped
- **New:** Verify `request_document_overview` surfaces H1, H2, and H3 without `depth>0`

### DIFF
--- a/src/solidlsp/language_servers/marksman.py
+++ b/src/solidlsp/language_servers/marksman.py
@@ -114,6 +114,26 @@ class Marksman(SolidLanguageServer):
         return super().is_ignored_dirname(dirname) or dirname in ["node_modules", ".obsidian", ".vitepress", ".vuepress"]
 
     @override
+    def request_document_overview(self, relative_file_path: str) -> list[UnifiedSymbolInformation]:
+        """Return all heading levels flattened rather than only H1 root symbols.
+
+        Marksman nests headings hierarchically (H2 under H1, H3 under H2, etc.) because
+        hierarchicalDocumentSymbolSupport is enabled. The default implementation returns only
+        root_symbols (H1), so get_symbols_overview with depth=0 would silently omit H2–H6.
+        Flattening ensures all headings appear at the top level of the overview.
+        """
+        document_symbols = self.request_document_symbols(relative_file_path)
+
+        def collect_all(symbols: list[UnifiedSymbolInformation]) -> list[UnifiedSymbolInformation]:
+            result = []
+            for sym in symbols:
+                result.append(sym)
+                result.extend(collect_all(sym.get("children", [])))
+            return result
+
+        return collect_all(document_symbols.root_symbols)
+
+    @override
     def request_document_symbols(self, relative_file_path: str, file_buffer: LSPFileBuffer | None = None) -> DocumentSymbols:
         """Override to remap Marksman's heading symbol kinds from String to Namespace.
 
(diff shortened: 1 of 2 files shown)


opcode81:
This sounds good. However, do you know which other types of symbols the LS assigns the String type to?
If it's only headings, then it's fine, of course.

EarthmanWeb:
Just added a fix to check for all headings, as it was only returning h1's

opcode81:
Why did you flatten the heading hierarchy. It absolutely makes sense for it to be nested.
If a deeper overview is desired, we have the depth parameter.

EarthmanWeb:
"If a deeper overview is desired, we have the depth parameter."

Well my LLM (Claude) never uses that.  It uses get_symbols_overview and if they aren't there it gives up and reads the file or uses Grep.

How do you expect the LLM to just know that? And why should it have to make two calls.  
"Overview" means overview aka summary

opcode81:
>> "If a deeper overview is desired, we have the depth parameter."
>
> Well my LLM (Claude) never uses that. It uses get_symbols_overview and if they aren't there it gives up and reads the file or uses Grep.
>
> How do you expect the LLM to just know that? And why should it have to make two calls.
"Overview" means overview aka summary

I expect it to know this, because the tool instructions say so.

Although it is not obvious how "symbol" semantics map to markdown, Claude uses it correctly without any specific instructions (this is using your branch with only the first commit):

image

[Full conversation](https://claude.ai/share/9b6e9921-b174-4a70-adc6-42bb67dce6d7)

With the structure intact, LLMs can
* read only the respective portions of the file by using `find_symbol` with `include_body=True` (including all subsections)
* use `replace_symbol_body` to replace a section
* use `insert_after_symbol` to insert a section, etc.

So changing the symbol types of headings is important, but flattening the structure is undesirable.

opcode81:
I have added #1110 which uses your first commit and adds a second commit to account for cache invalidation.

opcode81:
Closing this, since the main improvement has been merged with #1110.

PR #1099, opened 2026-02-26
Fix PHP symbol retrieval: declare missing Intelephense client capabilities

EarthmanWeb:
## Problem

Intelephense's LSP server conditionally enables certain features based
on what the client declares it supports in the `initialize` request.
Serena's Intelephense client was missing several capability declarations:

- `textDocument.documentSymbol` (with `hierarchicalDocumentSymbolSupport`)
- `textDocument.references`
- `textDocument.hover`
- `workspace.symbol`

Without `hierarchicalDocumentSymbolSupport: true`, Intelephense falls
back to returning flat `SymbolInformation[]` instead of the nested
`DocumentSymbol[]` tree. This means symbol retrieval works partially
or not at all — classes and their methods appear as disconnected
root-level symbols with no parent-child relationship.

Additionally, when `find_symbol` is called with a specific PHP file
path, it iterated all language servers rather than routing to the
correct one, causing PHP files to be rejected by non-PHP servers.

## Fix

- **`intelephense.py`**: Added missing client capability declarations
  for `documentSymbol` (with full `hierarchicalDocumentSymbolSupport`
  and `symbolKind` range), `references`, `hover`, and workspace
  `symbol`. Also added assertion that the server confirms
  `documentSymbolProvider` support after initialization.

- **`symbol.py`**: When `find_symbols` is called with a specific file
  path, route to the appropriate language server via
  `get_language_server()` instead of iterating all servers.

## Example

Before:
```python
# Intelephense returned flat SymbolInformation[] — no hierarchy
symbols = get_symbols_overview("src/Controller.php")
# → {"MyController": [], "index": [], "show": []}
# (methods appear at root level, not nested under class)
```

After:
```python
symbols = get_symbols_overview("src/Controller.php")
# → {"MyController": {"index": {}, "show": {}}}
# (proper hierarchy via DocumentSymbol[])
```


docs/sample.php
```diff
@@ -0,0 +1,665 @@
+<?php
```
opcode81: If it shall provide the basis for tests (which it probably should), please move this the PHP test repository.
If it shall not provide the basis for tests, delete it.

EarthmanWeb: Removed and provided a new sample file and test

EarthmanWeb:
> Tests should be added which validate that the changes result in the improvements described. In particular, we should explicitly inspect the hierarchical structure in tests.

All 19 PHP tests pass. Summary of what was added:

[test/solidlsp/php/test_php_basic.py](vscode-webview://16fn3llgiaa7dd5e1rp1d3mhh44589aokpkfdr5furijn0h8e4pp/test/solidlsp/php/test_php_basic.py) — 2 new tests:

test_document_symbols_hierarchical_structure — directly validates the hierarchicalDocumentSymbolSupport fix: asserts Dog class has greet, fetch, getBreed, describe as children (not disconnected root-level symbols), and that those methods do NOT appear at root level
test_full_symbol_tree_within_file — validates request_full_symbol_tree(within_relative_path="sample.php") returns the correct class/method hierarchy
[test/serena/test_serena_agent.py](vscode-webview://16fn3llgiaa7dd5e1rp1d3mhh44589aokpkfdr5furijn0h8e4pp/test/serena/test_serena_agent.py) — 1 new test:

test_find_symbol_within_php_file — validates the symbol.py routing fix: calls find_symbol(name_path_pattern="Dog/greet", relative_path="sample.php") and asserts the symbol is found (would return empty before the fix, as non-PHP servers would reject the file)

opcode81:
Thanks, @EarthmanWeb, I have merged this.

PR #1100, opened 2026-02-26
feat: add --memory-path CLI option to override memory directories

EarthmanWeb:
## Problem

Serena stores project memories in `<project_root>/.serena/memories/`
with no way to override this location at runtime. This is limiting in
several scenarios:

- **Shared memory across projects**: teams or multi-repo setups where
  common knowledge should be stored in a shared location **outside any
  single project root**
- **Read-only project roots**: environments where the project directory
  is not writable (e.g. mounted volumes, CI)
- **Read only docs in other folders:** adding additional folders for shared documentation that should not be editable, and may exist in a different folder location than the folder that was chosen as the primary serena memory folder
- **Multiple memory sources**: combining project-specific memories with
  org-wide or team-wide knowledge bases

## Solution

Added a `--memory-path` option to `serena-mcp-server` accepting a
comma-separated list of directory paths. Extra paths can be marked
read-only with the `:ro` suffix (Docker volume convention):

```bash
uv run serena-mcp-server --memory-path /writable/memories,/shared/team:ro
```

**Behaviour:**
- First path = primary write location (created if it doesn't exist)
- Subsequent paths = additional sources searched after primary
- Append `:ro` to any extra path to make it read-only
- Paths can be absolute or relative to the project root
- In-place updates: writes to an existing memory update it where it lives, not duplicating to primary
- Defaults to the existing `<project_root>/.serena/memories/` if not set

**Write protection:**
All mutating operations (`write_memory`, `edit_memory`, `delete_memory`, `rename_memory`) return a clear error and instruct the agent not to retry when targeting a read-only path:
```
Cannot edit memory 'shared/topic': it is in a read-only path. Do not retry this operation.
```

**Implementation:**
- `cli.py`: `--memory-path` option parsed into a list; `:ro` documented in help text
- `mcp.py`: propagates `memory_paths` through `create_mcp_server()` → `_create_serena_agent()`
- `agent.py`: stores `memory_paths`, calls `project.set_memory_paths()` on project activation
- `project.py`: `MemoriesManager` gains `_readonly_dirs: set[Path]`, `_is_readonly()` helper, and write guards in all mutating methods; `set_memory_paths()` parses `:ro` suffix

## Examples

```bash
# Shared team memories as primary, project memories as additional source
uv run serena-mcp-server \
  --memory-path /team/serena-memories,/myproject/.serena/memories

# Project memories writable, shared knowledge base read-only
uv run serena-mcp-server \
  --memory-path .serena/memories,/org/shared-knowledge:ro

# Multiple read-only sources with a local writable primary
uv run serena-mcp-server \
  --memory-path /local/memories,/team/knowledge:ro,/org/global:ro
```

### DIFF
--- a/src/serena/cli.py
+++ b/src/serena/cli.py
@@ -231,6 +231,7 @@ class TopLevelCommands(AutoRegisteringGroup):
         default=None,
         help="Comma-separated list of memory directory paths (absolute or relative to project root). "
         "The first path is the primary write location; subsequent paths are additional sources. "
+        "Append ':ro' to any extra path to make it read-only (e.g. '/shared/memories:ro'). "
         "Writes to existing memories update them in-place; new memories go to the primary. "
         "Defaults to '<project_root>/.serena/memories'.",
     )
--- a/src/serena/project.py
+++ b/src/serena/project.py
@@ -35,10 +35,15 @@ class MemoriesManager:
         self._memory_dir.mkdir(parents=True, exist_ok=True)
         self._encoding = SERENA_FILE_ENCODING
         self._extra_memory_dirs: list[Path] = []
+        self._readonly_dirs: set[Path] = set()
 
     def _resolve_path(self, raw: str) -> Path:
         p = Path(raw)
         return p if p.is_absolute() else (self._project_root / p).resolve()
+
+    def _is_readonly(self, path: Path) -> bool:
+        """Return True if path lives inside a read-only extra directory."""
+        return any(path.is_relative_to(d) for d in self._readonly_dirs)
 
     def _find_memory(self, name: str) -> Path | None:
         """Search primary and extra memory dirs for a memory file."""
@@ -65,13 +70,17 @@ class MemoriesManager:
         primary.mkdir(parents=True, exist_ok=True)
         self._memory_dir = primary
         self._extra_memory_dirs = []
+        self._readonly_dirs = set()
         for raw in paths[1:]:
-            extra = self._resolve_path(raw)
+            readonly = raw.endswith(":ro")
+            extra = self._resolve_path(raw[:-3] if readonly else raw)
             if extra.exists() and extra.is_dir():
                 self._extra_memory_dirs.append(extra)
+                if readonly:
+                    self._readonly_dirs.add(extra)
             else:
                 log.warning(f"Extra memory path not found or not a directory: {extra}")
-        log.info(f"Memory paths set: primary={self._memory_dir}, extras={self._extra_memory_dirs}")
+        log.info(f"Memory paths set: primary={self._memory_dir}, extras={self._extra_memory_dirs}, readonly={self._readonly_dirs}")
 
     def get_memory_file_path(self, name: str) -> Path:
         # If the memory already exists (in primary or extras), return that path
@@ -101,6 +110,8 @@ class MemoriesManager:
 
     def save_memory(self, name: str, content: str) -> str:
         memory_file_path = self.get_memory_file_path(name)
+        if self._is_readonly(memory_file_path):
+            return f"Cannot write memory '{name}': it is in a read-only path. Do not retry this operation."
         with open(memory_file_path, "w", encoding=self._encoding) as f:
             f.write(content)
         return f"Memory {name} written."
@@ -129,6 +140,8 @@ class MemoriesManager:
         memory_file_path = self._find_memory(name)
         if memory_file_path is None:
             return f"Memory {name} not found."
+        if self._is_readonly(memory_file_path):
+            return f"Cannot delete memory '{name}': it is in a read-only path. Do not retry this operation."
         memory_file_path.unlink()
         return f"Memory {name} deleted."
 
@@ -137,6 +150,8 @@ class MemoriesManager:
         Rename or move a memory file.
         """
         old_path = self.get_memory_file_path(old_name)
+        if self._is_readonly(old_path):
+            return f"Cannot rename memory '{old_name}': it is in a read-only path. Do not retry this operation."
         new_path = self.get_memory_file_path(new_name)
 
         if not old_path.exists():
--- a/src/serena/tools/memory_tools.py
+++ b/src/serena/tools/memory_tools.py
@@ -100,6 +100,8 @@ class EditMemoryTool(Tool, ToolMarkerCanEdit):
         memory_path = self.memories_manager._find_memory(memory_name)
         if memory_path is None:
             return f"Memory file {memory_name} not found."
+        if self.memories_manager._is_readonly(memory_path):
+            return f"Cannot edit memory '{memory_name}': it is in a read-only path. Do not retry this operation."
         content = memory_path.read_text(encoding="utf-8")
         if mode == "literal":
             if needle not in content:


MischaPanch:
@EarthmanWeb with #1091 we added support for global memories. Do you think the additional flexibility through memory-path is still needed?

opcode81:
> Read-only project roots

This will be addressed with #1123

EarthmanWeb:
II can't see anything in that spec regarding read only folders, that's the
main defining factor of this PR

opcode81:
> I can't see anything in that spec regarding read only folders

The goal of #1123 is to allow you to move the entire `.serena` folder somewhere else entirely, thus solving the problem of a read-only project folder. If it is read-only, move the Serena folder outside of the project folder; all of Serena's project-specific data, including memories, can live elsewhere.

EarthmanWeb:
You're missing the point of this PR.

It allows you to designate multiple folders, not necessarily all under the
Serena folder, in order to be able to reference documents.

I believe it's all spelled out in the PR description.

It allows you to designate multiple folders

opcode81:
You are missing the point of our comments.

@MischaPanch pointed out that we recently added support for global memories. This addresses the first point you mentioned:

> Shared memory across projects

It can furthermore address the third point,

> Read only docs in other folders

because global memories can be configured as read-only.

The proposal #1123 I mentioned (which is already being worked on) specifically addresses fully read-only project folders, i.e. your second point:

> Read-only project roots

You didn't answer our question as to whether these features would solve your underlying problem.
You proposed a solution, but you didn't state what exactly your actual problem is, so we don't know *why* you proposed what you proposed. For this reason (among others), the workflow in Serena's development is:

1. Create an issue describing your problem (and perhaps propose a solution)
2. Discuss solutions and, ideally, decide on the best way of handling it
3. Implement and issue PR

See our [contribution guidelines](https://github.com/oraios/serena/blob/main/CONTRIBUTING.md).

EarthmanWeb:
> @EarthmanWeb with #1091 we added support for global memories. Do you think the additional flexibility through memory-path is still needed?

Thanks.  

I wasn't aware of that option and have not yet tested it. Will do so asap

One thing my PR provides is listing those memories with the rest, without the /global prefix requirement though I see how that might be helpful for better original path identification.

It's not clear in the notes here  https://github.com/oraios/serena/pull/1007 the configuration options for readonly for specific folders. 

Does readonly apply to ALL global memories when enabled? ...because I have some that I want to be able to edit outside the .serena path and some I do not. My pr addressed that by allowing a 'flag' for each new external folder that was configured.

MischaPanch:
The global memories are automatically and always listed with the rest. The editing restrictions, when enabled, apply to all global memories. We have written documentation on the extended memory system here https://oraios.github.io/serena/02-usage/045_memories.html

Both @opcode81 and I realize that this PR has a different functionality from what we recently merged and are working on, but there are significant overlaps. That's why we are asking whether the extensions we made are sufficient to support your workflow, maybe with small adjustments on the workflow side.

If the existing features solve say 90% of the pain, it may not be worth it to add complexity for configuring additional memory containers with flags for which of them are read only and which not. This is why we are asking.

We want to avoid featuritis and creating complexity that only serves very niche use cases. Serena is already very complex, maintaining it is a lot of work, and we have to be critical towards additions not only based on the code quality there but considering the maintenance and documentation overhead, as well as the configuration complexity for the users. We are not doing it to annoy you or to discourage contributions

MischaPanch:
As an example on the workflow adjustment side: create links from your read-only memories containers to `~/.serena/memories/global` and enable the readonly flag. Now you have cross-project read-only memories without modifying any Serena code. For the non read-only ones, create softlinks to `<project>/.serena/memories`.

EarthmanWeb:
I agree with you, thank you for the explanation.


I am also not trying to annoy you or waste your time. When linked to a
thread instead of a PR it can get confusing. You and your colleagues know
the full context of the thread and history of the project. I do not, and
sometimes don't take enough time to research - my apologies for that.


Looking at the existing config, I would add a config possibility, based on
the work I have in this PR, which would springboard off pf yours and simply
and elegantly add an object check for:


```
edit_global_memories:
 - "global/editthese"
 - "global/andthese"

```


And then if it resolves true but also has data, it knows to only allow
editing for selected paths, solving both use cases


Thoughts?

EarthmanWeb:
Oh interesting I did not realize that you required sym links for global
memories to work. That's less than ideal.

It's not portable.

MischaPanch:
> edit_global_memories:
   - "global/editthese"
   - "global/andthese"

> And then if it resolves true but also has data, it knows to only allow editing for selected paths, solving both use cases

I think this makes sense, it can even be made backwards compatible. In python, `edit_global_memories` becomes of type `bool | dict[str, bool]`

The configuration by the user is then

```yaml
edit_global_memories: true/false
```

Or (compatible)
```yaml
edit_global_memories:
   path_prefix1: false
   path_prefix2: true
```

where for all non-configured paths we assume the current default (editing permitted).

Few users would make use of this, I suppose, but it doesn't hurt.

@opcode81 WDYT?

MischaPanch:
> Oh interesting I did not realize that you required sym links for global
memories to work. That's less than ideal.

We don't require symlinks, global memories are just any files in `~/.serena/memories/global`

EarthmanWeb:
Right.

But my PR allows for adding file paths `outside` of the .serena path
That is the whole distinction

I can add `./docs` or whatever I choose

And then can set those to be read only, or not

MischaPanch:
Why is it important to be able to pass paths outside of `~/.serena` and `project/.serena`? Which problem does it solve? If you don't want to have your memories there, then just create symlinks on your system. I don't see the benefits of extending configurability on Serena's side

EarthmanWeb:
The problem it solves is when I share a project with others who have docs in another path. 

They don't use serena. I do. They don't want the .serena path in the main branches, and the CI excludes them.

...but they do want the shared docs retained, and I don't want to have to copy them or create symlinks to preserve the config

MischaPanch:
> I don't want to have to copy them or create symlinks to preserve the config

Creating a symlink to `.serena` is about as much work as passing `--memory-path` in the cli call, or am I missing something?

EarthmanWeb:
From the 'problem' section of the PR, which your colleague asserts does not exist:
I added a couple additional words, recently, to clarify...

```
Serena stores project memories in <project_root>/.serena/memories/
with no way to override this location at runtime. This is limiting in
several scenarios:

Shared memory across projects: teams or multi-repo setups where
common knowledge should be stored in a shared location outside any
single project root
Read-only project roots: environments where the project directory
is not writable (e.g. mounted volumes, CI)
Read only docs in other folders: adding additional folders for shared documentation that should not be editable, and may exist in a different folder location than the folder that was chosen as the primary serena memory folder
Multiple memory sources: combining project-specific memories with
org-wide or team-wide knowledge bases
```

MischaPanch:
>From the 'problem' section of the PR, which your colleague asserts does not exist:

Pls avoid the passive aggressive undertone. My colleague highlighted that problems should be discussed **before** a PR happens and not as description of a PR.

EarthmanWeb:
> > I don't want to have to copy them or create symlinks to preserve the config
> 
> Creating a symlink to `.serena` is about as much work as passing `--memory-path` in the cli call, or am I missing something?

Fair.  Ok, I see your point. Thanks.

With the addition of:
https://github.com/oraios/serena/pull/1100#issuecomment-4004729040

This would work. Thanks you for your time reviewing.

EarthmanWeb:
> > From the 'problem' section of the PR, which your colleague asserts does not exist:
> 
> Pls avoid the passive aggressive undertone. My colleague highlighted that problems should be discussed **before** a PR happens and not as description of a PR.

Ok, my apologies.

Was referrring to: "but you didn't state what exactly your actual problem is, so we don't know why you proposed what you proposed"

They did not say what you said, in regards to a discussion prior to a PR. Where would that be held?  Submit issue first, THEN pr?

MischaPanch:
Pls wait with implementing the config extension proposal until @opcode81 has confirmed. It extends complexity and union types are not clean and a typical source of errors, though in this particular case it might be worth doing that nevertheless

EarthmanWeb:
I missed this in the contribution guidelines:
"For other changes, please open an issue first to discuss your ideas with the maintainers."

My bad.  Sorry, @opcode81  - will be more diligent in future

MischaPanch:
>  Submit issue first, THEN pr?

Yes, absolutely, that is very much preferred. It's also exactly what @opcode81 said in his [comment](https://github.com/oraios/serena/pull/1100#issuecomment-4003889549) and what our contribution guidelines state (though I don't remember for how long they've been formulated this way)

EarthmanWeb:
Update to previous, while using symlinks to add global memories:

---

The problem it solves is when I share a project with others who have docs in another path. 

They don't use serena. I do. They don't want the .serena path in the main branches, and the CI excludes them. 

Some of the external ref's inside a global path that is linked, I need to be able to define as editable, some I do not, as I do not 'own' them.

If symlinks are used to share a root folder of docs with a complex inner structure and several nested hierarchies, ideally for reference I would like Serena to read that the same folder structure by creating a single top level symlink for the tree, but still be able to 'flag' internal paths within that structure as 'read only' (or at least define which are editable)

The addition of:
https://github.com/oraios/serena/pull/1100#issuecomment-4004729040
would allow that nicely

opcode81:
> From the 'problem' section of the PR, which your colleague asserts does not exist

I have never made any such assertion.
I said this:

> You didn't answer our question as to whether these features would solve your **underlying problem**.
> You proposed a solution, but you didn't state what exactly your actual problem is, so we don't know *why* you proposed what you proposed.

By "what exactly your actual problem is" I meant the underlying motivation, the root of the problem. What you stated was a surface-level problem, but we need to understand, as I have pointed out, *why* this is perceived as a problem, i.e. get to the root of the problem, because it is frequently the case that there are other solutions to the underlying problem which do not involve finding a solution to the surface-level problem at all (much like treating a symptom in medicine).

EarthmanWeb:
@opcode81   OK, that's fair.  I didn't take the time to clearly read your comments to understand what you were asking for.

I do the same with my own clients, fully understand and appreciate your socratic method.

Will take better care to understand your comments in future before replying.

opcode81:
I think "we don't know why you proposed what you proposed" is sufficiently clear.

MischaPanch:
I'm going to close this and open an issue regarding extending the memory-edit-allowance configuration

PR #1101, opened 2026-02-26
Add ignore_all_dot_files config option; fix Pyright dot-directory indexing

EarthmanWeb:
## Problem

Serena unconditionally ignores all dot-prefixed directories (e.g.
`.github`, `.config`, `.devcontainer`) via `is_ignored_dirname()`.
This is the right default for most cases, but breaks legitimate
workflows:

- **`.github/` workflows**: teams that store YAML, scripts or
  documentation in `.github/` cannot get symbol support for those files
- **`.config/` or `.devcontainer/`**: project config files in
  dot-directories are invisible to Serena

Additionally, Pyright has its own internal dot-directory exclusion that
is separate from Serena's. Even when Serena's exclusion was adjusted,
Pyright would still skip dot-directories because its `initializationOptions`
excluded `**/.venv` etc., and its own defaults excluded `**/.*`.
Pyright also sends `workspace/configuration` requests to fetch
`python.analysis` settings — these were previously unhandled, causing
silent misconfiguration.

## Fix

**New config option: `ignore_all_dot_files`** (default: `true`)

Added to `ProjectConfig`, `LanguageServerConfig`, and
`project.template.yml`. When set to `false`:

- `SolidLanguageServer.is_ignored_dirname()` only ignores `.git`
  unconditionally; all other dot-directories are traversed
- `PyrightServer._get_initialize_params()` becomes an instance method
  and builds exclude/include lists based on `ignore_all_dot_files`:
  - `false`: adds `include: ["."]` to override Pyright's internal
    dot-directory exclusion
  - `true`: keeps existing behaviour (excludes `.venv`, `.env`, `.pixi`)
- `PyrightServer` now handles `workspace/configuration` requests,
  returning appropriate `python.analysis` include/exclude settings

## Example

```yaml
# .serena/project.yml
ignore_all_dot_files: false  # enable symbol retrieval in .github/, etc.
```

```python
# Before: .github/workflows/ci.yml invisible to Serena
get_symbols_overview(".github/workflows/ci.yml")
# → FileNotFoundError or empty (directory ignored)

# After:
get_symbols_overview(".github/workflows/ci.yml")
# → {"on": {}, "jobs": {...}}
```

### DIFF
--- a/src/serena/config/serena_config.py
+++ b/src/serena/config/serena_config.py
@@ -180,6 +180,7 @@ class ProjectConfig(SharedConfig):
     ignored_paths: list[str] = field(default_factory=list)
     read_only: bool = False
     ignore_all_files_in_gitignore: bool = True
+    ignore_all_dot_files: bool = True
     initial_prompt: str = ""
     encoding: str = DEFAULT_SOURCE_FILE_ENCODING
 
--- a/src/serena/ls_manager.py
+++ b/src/serena/ls_manager.py
@@ -23,6 +23,7 @@ class LanguageServerFactory:
         ls_timeout: float | None = None,
         ls_specific_settings: dict | None = None,
         trace_lsp_communication: bool = False,
+        ignore_all_dot_files: bool = True,
     ):
         self.project_root = project_root
         self.encoding = encoding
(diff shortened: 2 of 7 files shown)


MischaPanch:
This has been addressed in other commits

PR #1108, opened 2026-03-01
Add Wolfram Language support via official LSPServer paclet

p135246:
Adds support for Wolfram Language (`.wl`, `.wls` files) using the official [WolframResearch LSPServer](https://github.com/WolframResearch/LSPServer) paclet, which is bundled with Wolfram installations and communicates via stdio.

**Requirements:** Wolfram Mathematica 13.0+ or Wolfram Engine 12.1+. The WolframKernel executable is discovered via `ls_path` in `ls_specific_settings.wolfram`, the `WOLFRAM_PATH` environment variable, the system PATH, or common install locations. If no kernel is found, server startup fails with an actionable error message, and the test suite skips gracefully (using the same discovery logic).

**Changes:**
- `WolframLanguageServer` following the current `DependencyProvider` pattern (analogous to QML/MATLAB)
- `WOLFRAM` entry in `LanguageServerId` (file matcher for `.wl`/`.wls`, LS class factory)
- Test repo and test suite (document symbols incl. subdirectories, within-file references), `wolfram` pytest marker
- Docs: README language list, language support page, CHANGELOG

**Known limitation:** the LSPServer paclet computes `textDocument/references` per document — cross-file references are not supported by the server. This is noted in the docs and the test module.

Verified locally against Wolfram 14: integration tests pass (live LSP communication), and an end-to-end check through the MCP server (`get_symbols_overview`, `find_symbol`, `find_referencing_symbols`, `get_diagnostics_for_file`) works on a real Wolfram project. `poe format`, `poe lint`, and `poe type-check` are clean.


opcode81:
@p135246 this is still a draft. Will you finalise this or shall we close it?

p135246:
Hey @opcode81, I will give it a last vibecoding push now and let's see :-)

p135246:
Addressed the Copilot review in the latest push:

1. **conftest skip logic** — now reuses the language server's own `_find_wolfram_kernel()` discovery helper instead of duplicating a subset of the search paths, so test gating matches runtime behavior (including `WOLFRAM_PATH` and Linux/Windows locations).
2. **Within-file references test** — the assertion now requires both the definition site and the call site, matching the comment.
3. **Cross-file references test** — removed. Verified against a live kernel that the LSPServer paclet computes `textDocument/references` per document only (cross-file references are not supported by the server), so the test could never validate what its name claimed. The limitation is now noted in the test module, the docs page, and the CHANGELOG entry.

p135246:
Addressed the second review round:

1. **"semantic highlighting" in CHANGELOG** — removed; the initialize params indeed do not advertise `semanticTokens`.
2. **Explicit negative test** — added `test_wolfram_no_cross_file_references`, asserting that references for a symbol defined in `lib/helper.wl` and called in `main.wl` stay within `lib/helper.wl`. If a future LSPServer version adds cross-file support, this test will fail and prompt a documentation update. The PR description was already updated to state within-file-only references.

p135246:
image

p135246:
Shall we merge or close @opcode81? The Wolfram LSP is not that strong yet (no crossfile references), but I've been using it anyway...

AmirF194 (commented):
`is_ignored_dirname` (src/solidlsp/language_servers/wolfram_language_server.py:53-59) matches `Documentation` and `FrontEnd` against any path component, not just the Wolfram installation layout. Verified on this PR's HEAD (41894ec):

```python
obj = object.__new__(WolframLanguageServer)
obj.is_ignored_dirname("Documentation")  # True
obj.is_ignored_dirname("FrontEnd")       # True
```

`should_ignore_path` in ls.py checks every directory component of a file's path against this, so a `.wl` file under a user's own `Documentation/` or `FrontEnd/` folder (both plausible project directory names, unlike the Mathematica-specific `.Wolfram`/`SystemFiles`) would be silently skipped from indexing, with no error.

Every other `is_ignored_dirname` override in this codebase (csharp: `bin`/`obj`/`packages`, elm: `elm-stuff`, haskell: `dist-newstyle`, etc.) sticks to build-tool or vendor-specific names that are unlikely to collide with a user's own source tree. Would it work to scope this to the actual bundled-app locations instead (for example checking that the matched dir sits under the resolved kernel's installation path), or drop `Documentation`/`FrontEnd` if they were only meant as a defensive guess?

MischaPanch:
Thanks for the input @AmirF194 . Yes, these are very generic directory names - do they have some fixed meaning in wolfram projects? If they usually do carry mathematica files, why would you want to exclude them from indexing? And if they don't carry mathematica files, there's no need to exclude them

@p135246 we can merge this after the ignored dir names issue is resolved, some wolfram support is better than no wolfram support at all.

opcode81:
Thanks for your contribution

PR #1113, opened 2026-03-02
fix: auto-close dashboard tab when server stops

Muminur:
## Summary

The web dashboard tab stays open in the browser after the Serena server stops (e.g. when the parent Claude Code / MCP client session ends). The heartbeat-based `window.close()` in `dashboard.js` doesn't work because modern browsers block `window.close()` for tabs that were not opened by JavaScript's `window.open()`.

**Two fixes:**

- **`agent.py`** — On Linux, prefer opening the dashboard in Chromium/Chrome `--app` mode, which creates a standalone window where `window.close()` is permitted. Falls back to `webbrowser.open()` when no Chromium-based browser is found.
- **`dashboard.js`** — Add `about:blank` redirect as fallback after `window.close()` in both the heartbeat handler and the manual shutdown method, so even without app mode the tab clears itself. Also raises the heartbeat failure threshold from 1 to 3 to avoid premature close on transient network hiccups.

## Test plan

- [ ] Start Serena MCP server with `web_dashboard_open_on_launch: true` on a Linux system with Chromium installed — dashboard should open as a standalone app window (no URL bar)
- [ ] Kill the Serena server process — the dashboard window should close automatically via `window.close()`
- [ ] On a system without Chromium (e.g. Firefox-only), verify it falls back to `webbrowser.open()` and the tab redirects to `about:blank` when the server stops
- [ ] Click Menu → Shutdown Server — tab should close or redirect to `about:blank`

### DIFF
--- a/src/serena/agent.py
+++ b/src/serena/agent.py
@@ -532,6 +532,13 @@ class SerenaAgent:
         """
         Opens the Serena web dashboard in the default web browser.
 
+        On Linux, prefers Chromium/Chrome in ``--app`` mode so the dashboard
+        opens as a standalone window.  This allows the heartbeat-based
+        ``window.close()`` in *dashboard.js* to work, because browsers only
+        permit ``window.close()`` for windows that were not opened as regular
+        tabs.  Falls back to the default browser via ``webbrowser.open()`` if
+        no supported Chromium-based browser is found.
+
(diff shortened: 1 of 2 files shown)


opcode81:
@Muminur I am not sure if you got my message from the other PR, but I am closing this in favour of #1117

PR #670, opened 2025-10-15
Add rego language support

antigenius0910:
Summary

  This PR adds comprehensive support for the Rego policy language to Serena using the
  https://github.com/StyraInc/regal (v0.36.1). Rego is the policy language for Open Policy Agent (OPA), widely
  used for cloud-native authorization and policy enforcement.

  Implementation

  Core Components

  Language Server Integration (src/solidlsp/language_servers/regal_server.py)
  - Implements RegalLanguageServer class following Serena's LSP patterns
  - Uses system-installed Regal binary (via CI or user installation)
  - Includes comprehensive debug logging for troubleshooting
  - Handles workspace loading with appropriate wait times for LSP initialization
  - Overrides default ignored directories to exclude .regal and .opa build artifacts

  Configuration (src/solidlsp/ls_config.py)
  - Added REGO = "rego" to Language enum
  - File pattern matcher for *.rego files
  - Integrated RegalLanguageServer into language server factory

  CI Integration (.github/workflows/pytest.yml)
  - Automated Regal binary installation for all platforms (Linux, macOS, Windows)
  - Architecture-aware downloads (x86_64, arm64)
  - Follows existing patterns used by ZLS, Lua LS, and nixd

  Test Suite

  Comprehensive Tests (test/solidlsp/rego/test_rego_basic.py)
  - 6 test methods covering all core LSP operations:
    - Document symbols retrieval (multiple files)
    - Workspace-wide symbol tree
    - Go-to-definition functionality
    - Symbol validation across files
  - Tests use realistic Rego policy examples (authorization, validation, helpers)

  Test Repository (test/resources/repos/rego/test_repo/)
  - Well-structured Rego project with proper package hierarchy:
    - policies/authz.rego - Authorization rules (33 lines)
    - policies/validation.rego - Validation policies (23 lines)
    - utils/helpers.rego - Helper functions (29 lines)
  - Demonstrates cross-file imports and realistic policy patterns

  Platform Support

  | Platform | Status            | Notes                     |
  |----------|-------------------|---------------------------|
  | Linux    | ✅ Fully Supported | All tests passing         |
  | macOS    | ✅ Fully Supported | All tests passing         |
  | Windows  | ⚠️ Known Issues   | Tests skipped - see below |

  Windows Limitations

  Regal LSP on Windows has workspace loading issues that cause test failures:
  - Root Cause: Regal loads workspace files asynchronously after initialization. On Windows, this process is
  slower and files load sequentially rather than in parallel.
[...]

PR #1114, opened 2026-03-02
Enable Rego tests on Windows (Regal v0.39.0 fixes upstream path handling)

antigenius0910:
## Summary

- Bumps Regal from **v0.36.1 → v0.39.0**, which includes the upstream Windows path handling fix for the LSP (see [StyraInc/regal#1683](https://github.com/open-policy-agent/regal/issues/1683), closed November 3, 2025)
- Removes the `@pytest.mark.skipif(sys.platform == "win32", ...)` guard from `TestRegoLanguageServer` — the `sys` import is also cleaned up as it's no longer needed

## Background

PR #670 added Rego language support but had to skip all tests on Windows due to a Regal LSP bug where Windows paths were not parsed correctly (`file://D:\...` instead of `file:///D:/...`). The Regal team confirmed the fix and shipped it in v0.37.0 (November 6, 2025). We're upgrading to the latest stable v0.39.0.

## Test plan

- [ ] CI passes on Linux ✅ (was already passing)
- [ ] CI passes on macOS ✅ (was already passing)
- [ ] CI passes on Windows 🆕 (previously all 6 Rego tests were skipped — now they run)
- [ ] If Windows CI reveals the async workspace-loading race condition is still present, a follow-up PR will add a retry/wait mechanism in `regal_server.py`


MischaPanch:
Nice, thanks!

PR #1117, opened 2026-03-02
Use pywebview for Dashboard (instead of browser tab), controlling lifecycle

opcode81:
This introduces the use of `pywebview` as a viewer for the Serena Dashboard.

Advantages:
* The UI can be opened fully in the background at startup (even in the system tray on Windows)
* The lifecycle is fully tied to the server; the app automatically terminates when the server shuts down

This is a work in progress.
* Untested on Linux
* Untested on macOS
* Documentation not updated
* Ramifications not handled (`open_dashboard` tool, configuration pointers, etc.)

### DIFF
--- a/src/serena/dashboard.py
+++ b/src/serena/dashboard.py
@@ -778,22 +778,21 @@ class SerenaDashboardViewer:
         Sets the NSApplication activation policy to Regular (shows in dock) or
         Accessory (hides from dock).  No-op on other platforms.
         """
-        if sys.platform != "darwin":
-            return
-        from AppKit import (
-            NSApplication,
-            NSApplicationActivationPolicyAccessory,
-            NSApplicationActivationPolicyRegular,
-        )
+        if sys.platform == "darwin":
+            from AppKit import (
+                NSApplication,
+                NSApplicationActivationPolicyAccessory,
+                NSApplicationActivationPolicyRegular,
+            )
 
-        ns_app = NSApplication.sharedApplication()
-        policy = NSApplicationActivationPolicyRegular if visible else NSApplicationActivationPolicyAccessory
-        ns_app.setActivationPolicy_(policy)
-        if visible:
-            # setActivationPolicy_ alone is not enough at runtime; unhide_ fires
-            # the full applicationWillUnhide:/applicationDidUnhide: cycle, which
-            # is what actually restores the dock icon and brings windows forward.
-            ns_app.unhide_(None)
+            ns_app = NSApplication.sharedApplication()
+            policy = NSApplicationActivationPolicyRegular if visible else NSApplicationActivationPolicyAccessory
+            ns_app.setActivationPolicy_(policy)
+            if visible:
+                # setActivationPolicy_ alone is not enough at runtime; unhide_ fires
+                # the full applicationWillUnhide:/applicationDidUnhide: cycle, which
+                # is what actually restores the dock icon and brings windows forward.
+                ns_app.unhide_(None)
 
     def _start_tray(self) -> None:
         # import pystray locally, because the import fails when there is no display!


opcode81:
@Muminur this is an alternative solution to the problem you are trying to solve in #1113.
If you are on either macOS or Linux, perhaps you can help us make this mature.

opcode81:
Improvements to pywebview that allow the icon to be set without having to create a frozen app bundle have been merged in December but are yet unreleased.
See https://github.com/r0x0r/pywebview/pull/1755 for Windows, and another for macOS.

opcode81:
I am now using the unreleased GitHub version of pywebview to support custom icons.

The behaviour on Windows is now exactly as desired:
* The app lives in the system tray
* If `web_dashboard_open_on_launch` is enabled, the window is shown at startup; otherwise, it can be accessed via the tray
* Closing the window (X button) does not terminate the app but minimises it to the system tray
* The application can be terminated via the tray (right-click + Quit)

@MischaPanch, we can now proceed to test on other platforms.
* I think we will need to use PNGs instead of .ico on other platforms.
* The tray implementation probably works differently or not at all.

MischaPanch:
Awesome! I'll test on Linux and will ask @arnaujc91 to test on Mac

opcode81:
On Linux, it is not possible to provide a solution that is guaranteed to work across DEs. pywebview builds on either GTK or Qt.
* For GTK, the Python lib ilightweight, but a system-level installation of GTK is necessary (and not always present). Unfortunately, even if the system installation is present, that Python wrapper needs to be built from source (no binary wheels), which only works if both dev tools and base libraries are available. So the env installation would already fail on many systems. Therefore, there is no way we can build on GTK.
* For Qt, one can theoretically bundle the entire Qt libraries (no reliance on system-level Qt installation). Unfortunately, these libraries are not only very large (hundreds of MB), they are also not stand-alone: Qt itself dynamically links to system libraries, which may or may not be present! So we could install the large Qt dependency, but it still wouldn't work on many systems. (Notably, conda could provide a self-contained Qt installation though.)

So we basically have two options.
1. Install the large Qt dependency on Linux, and if it works, good. If it doesn't (exception raised), use the browser tab fallback.
2. Always use browser tab opening on Linux.

The nix-based installation of Serena could probably provide a solution that works for all Linux users (by ensuring that all required deps are present at the system level), but we shall leave this to the community.

arnaujc91:
Doesnt work on mac

Exception in thread Thread-2 (_run_tray):
  File "/Users/zoajimen/.local/share/uv/python/cpython-3.12.12-macos-aarch64-none/lib/python3.12/threading.py", line 1075, in _bootstrap_inner
    self.run()
  File "/Users/zoajimen/.local/share/uv/python/cpython-3.12.12-macos-aarch64-none/lib/python3.12/threading.py", line 1012, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/zoajimen/.cache/uv/archive-v0/CIAuLMHlTyzGFBzC7gXZx/lib/python3.12/site-packages/serena/dashboard.py", line 784, in _run_tray
    self._tray_icon = pystray.Icon("dashboard_viewer", icon_img, self.title, menu)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zoajimen/.cache/uv/archive-v0/CIAuLMHlTyzGFBzC7gXZx/lib/python3.12/site-packages/pystray/_darwin.py", line 60, in __init__
    self._status_item = self._status_bar.statusItemWithLength_(
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
objc.error: NSInternalInconsistencyException - NSWindow should only be instantiated on the main thread!

MischaPanch:
@opcode81 CI ran through, feel free to merge

PR #1121, opened 2026-03-03
Fix/enhance search and logs

phuongvm:
🚀 What does this PR do?

  This Pull Request introduces several stability fixes, cleans up noisy stack traces, and significantly improves the UX of the SearchForPatternTool when dealing with large
  results.

  📝 Detailed Changes & Benefits:

  1. Enhance `SearchForPatternTool` UX & Context Efficiency
   - Lowered default_max_tool_answer_chars from 150,000 to 50,000.
     - Benefit: 150k is too large, causing LLM context dilution ("lost in the middle" effect), raising costs, and slowing down response times. 50k is a much safer sweet spot
       for modern LLM context windows.
   - When a search result exceeds the max limit, instead of throwing a generic flat error ("The answer is too long..."), the tool now parses the results and returns a smart 
     summary containing the total matches and the "Top 10 files with the most matches".
     - Benefit: This actively guides the AI Agent to iteratively refine its search query (e.g., using relative_path or paths_include_glob) rather than hitting a dead end and
       giving up.

  2. Clean up noisy logs and suppress misleading errors
   - asyncio / Windows: Added a custom exception handler to the main event loop in cli.py to silently suppress ConnectionResetError [WinError 10054] and pipe transport      
     errors.
     - Benefit: These happen naturally when clients/language servers close connections but were polluting the terminal logs with scary tracebacks on Windows environments.   
   - tools_base.py: Standard AI argument errors like ValueError and FileNotFoundError (which happen frequently when the LLM hallucinates a path) are now logged as simple    
     WARNINGs instead of full ERROR stack traces.
   - TypeScript LS: Added a dummy handler for the $/typescriptVersion notification to eliminate the Unhandled method warning on startup.

  3. Fix Terraform Language Server crash
   - Initialized self.completions_available = threading.Event() inside TerraformLS.__init__. 
     - Benefit: Previously, the server would crash with an AttributeError when trying to call .set() on this missing attribute.

### DIFF
--- a/src/serena/cli.py
+++ b/src/serena/cli.py
@@ -267,6 +267,32 @@ class TopLevelCommands(AutoRegisteringGroup):
 
         log.info("Initializing Serena MCP server")
         log.info("Storing logs in %s", log_path)
+        
+        # Suppress verbose asyncio ConnectionResetError warnings on Windows Proactor Event Loop
+        import asyncio
+        def custom_exception_handler(loop: asyncio.AbstractEventLoop, context: dict) -> None:
+            exc = context.get("exception")
+            if isinstance(exc, ConnectionResetError) and getattr(exc, "winerror", None) == 10054:
+                # Safe to ignore on Windows: "An existing connection was forcibly closed by the remote host"
+                return
+                
+            msg = context.get("message", "")
+            if "Exception in callback _ProactorBasePipeTransport" in msg:
+                return
+                
+            loop.default_exception_handler(context)
+
+        try:
+            loop = asyncio.get_running_loop()
+            loop.set_exception_handler(custom_exception_handler)
+        except RuntimeError:
+            # If no loop is running yet, we can set the policy to apply the handler to all new loops
+            class CustomEventLoopPolicy(asyncio.DefaultEventLoopPolicy):
+                def get_event_loop(self):
+                    loop = super().get_event_loop()
+                    loop.set_exception_handler(custom_exception_handler)
+                    return loop
+            asyncio.set_event_loop_policy(CustomEventLoopPolicy())
 
         # Handle --project-from-cwd flag
         if project_from_cwd:
--- a/src/serena/config/serena_config.py
+++ b/src/serena/config/serena_config.py
@@ -517,7 +517,7 @@ class SerenaConfig(SharedConfig):
     on the first run, which can take some time and require internet access. Others, like the Anthropic ones, may require an API key
     and rate limits may apply.
     """
-    default_max_tool_answer_chars: int = 150_000
+    default_max_tool_answer_chars: int = 50_000
     """Used as default for tools where the apply method has a default maximal answer length.
     Even though the value of the max_answer_chars can be changed when calling the tool, it may make sense to adjust this default 
     through the global configuration.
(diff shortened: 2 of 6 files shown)


src/serena/symbol.py
```diff
@@ -716,14 +712,33 @@ def find(
...
-                    LanguageServerSymbol(root).find(
-                        name_path_pattern, include_kinds=include_kinds, exclude_kinds=exclude_kinds, substring_matching=substring_matching
+        
+        # If searching within a specific file, only check language servers that support that file type
+        # This avoids false "file is ignored" errors from language servers that don't support the file type
+        if within_relative_path is not None and os.path.isfile(os.path.join(self._ls_manager.get_root_path(), within_relative_path)):
+            # For a specific file, only use the language server that supports it
+            try:
+                lang_server = self._ls_manager.get_language_server(within_relative_path)
+                symbol_roots = lang_server.request_full_symbol_tree(within_relative_path=within_relative_path)
```
MischaPanch, line 722: should be factored into a local function

src/solidlsp/language_servers/yaml_language_server.py
```diff
@@ -164,7 +164,14 @@ def do_nothing(params: Any) -> None:
         def window_log_message(msg: dict) -> None:
             log.info(f"LSP: window/logMessage: {msg}")
 
+        def workspace_configuration(params: dict) -> list[dict]:
```
MischaPanch, line 167: why is this needed? Pls add a block comment

.serena/memories/coding_conventions.md
```diff
@@ -0,0 +1,20 @@
+# Serena Coding Conventions
```
MischaPanch, line 1: pls rollback all changes to memories

src/serena/config/serena_config.py
```diff
@@ -537,7 +517,7 @@ class SerenaConfig(SharedConfig):
     on the first run, which can take some time and require internet access. Others, like the Anthropic ones, may require an API key
     and rate limits may apply.
     """
-    default_max_tool_answer_chars: int = 150_000
+    default_max_tool_answer_chars: int = 50_000
```
MischaPanch, line 520: @opcode81 wdyt?

src/serena/cli.py
```diff
@@ -267,6 +267,32 @@ def start_mcp_server(
...
+                return
+                
+            msg = context.get("message", "")
+            if "Exception in callback _ProactorBasePipeTransport" in msg:
+                return
+                
+            loop.default_exception_handler(context)
+
+        try:
+            loop = asyncio.get_running_loop()
```
MischaPanch, line 286: I have never seen excessive logs at startup and this is complex logic that dives into the MCP server's event loop. Is this really needed?

src/serena/tools/file_tools.py
```diff
@@ -417,4 +481,26 @@ def apply(
             assert match.source_file_path is not None
             file_to_matches[match.source_file_path].append(match.to_display_string())
         result = self._to_json(file_to_matches)
+        
+        # Smart limit checking: if the result is too long, we provide a summary of the top 10 files instead of a generic failure message
```
MischaPanch, line 485: Neat, thanks! I like this, wdyt @opcode81 ?

opcode81:
@phuongvm please submit one PR per logical change. In general, never submit PRs that contain several unrelated changes.
It is too difficult to then separate the changes that we want to keep from the things that need to change/are not wanted.
Even your commits mix unrelated changes, so we cannot even cherry-pick.

The change you made to the search tool is something we would integrate, so do issue a PR on that.

PR #1126, opened 2026-03-03
feat: Auto-detect all languages in non-interactive / autonomous agent mode

### LINKED ISSUE #1125: feat: Auto-detect all languages in non-interactive / autonomous agent mode
## Is your feature request related to a problem? Please describe.

When Serena is invoked autonomously via MCP (e.g. from a Claude Code SDK agent, a CI pipeline, or any other headless agent) and the target project does not have a pre-existing `.serena/project.yml`, Serena auto-generates one. However, the current `ProjectConfig.autogenerate()` logic only enables the **single dominant language** (by file count). All other detected languages are silently ignored.

This is a critical gap for **polyglot repositories** — a very common real-world pattern. For example:
- A **Python backend + TypeScript/React frontend** monorepo
- A **Go service + Python scripts** codebase
- Any project where infrastructure-as-code (Terraform/YAML) lives alongside application code

In an autonomous/agent invocation there is no human present to configure `project.yml` manually or answer interactive prompts. The agent ends up with only one language server running, so `find_symbol`, `get_symbols_overview`, and all other LSP-based tools silently return empty results for the secondary language.

## Describe the solution you'd like

In **non-interactive mode** (the default for MCP/agent invocations), `ProjectConfig.autogenerate()` should automatically enable **all detected languages** rather than just the dominant one.

Interactive mode behaviour should remain unchanged — the user is still asked which additional languages to enable.

A new `auto_detect_all_languages: bool = True` parameter controls this, defaulting to `True` so the fix is on by default for agent use-cases without requiring any configuration change.

## Describe alternatives you've considered

- Requiring users to always pre-create a `project.yml` before invoking Serena via MCP — this is impractical for autonomous agent workflows where Serena is spun up dynamically against arbitrary repos.
- Exposing a `--languages` CLI flag to the MCP server entrypoint — this would work but forces hardcoded language lists which defeats the purpose of auto-detection.

## Additional context

The `LanguageServerManager` already fully supports running multiple language servers in parallel (spawned in parallel threads via `from_languages()`), and `get_language_server()` already dispatches tool calls to the correct server based on file extension. The only gap is the auto-detection step not populating all languages.

A fix with tests is provided in the linked PR.

droideronline:
Closes #1125

## Problem

When Serena is invoked via MCP from an autonomous agent (Claude Code SDK, CI pipelines, etc.) and no `.serena/project.yml` exists, `ProjectConfig.autogenerate()` only enables the **single dominant language** by file count. Secondary languages are silently dropped.

For polyglot repos (e.g. Python backend + TypeScript/React frontend) this means all LSP-based tools (`find_symbol`, `get_symbols_overview`, `replace_symbol_body`, etc.) return empty or wrong results for the non-dominant language, with no error or warning to indicate why.

## Solution

Add `auto_detect_all_languages: bool = True` to `ProjectConfig.autogenerate()`.

- **Non-interactive mode** (default for MCP/agent): all detected languages are automatically enabled, sorted by file count descending.
- **Interactive mode**: behaviour is unchanged — the user is still asked about each additional language.

The change is backward-compatible: existing `project.yml` files are not affected; the parameter only applies during auto-generation when no config exists yet.

## What changes

**`src/serena/config/serena_config.py`**
- `autogenerate()` gains `auto_detect_all_languages: bool = True`
- In non-interactive mode with this flag, `languages_to_use` is populated from all detected languages, not just the top one
- Adds a log line listing all enabled languages for visibility

## No new dependencies

The `LanguageServerManager` already spawns multiple language servers in parallel threads and `get_language_server()` already routes tool calls by file extension — this PR purely fixes the auto-detection step upstream.

### DIFF
--- a/src/serena/config/serena_config.py
+++ b/src/serena/config/serena_config.py
@@ -274,7 +274,7 @@ class ProjectConfig(SharedConfig):
                     languages_to_use = [lang.value for lang, _ in languages_and_percentages]
                     log.info(
                         "Non-interactive mode: automatically enabling all detected languages: %s",
-                        languages_to_use,
+                        [(lang, f"{pct:.2f}%") for lang, pct in languages_and_percentages],
                     )
                 else:
                     # find the language with the highest percentage and enable it
--- a/src/serena/util/inspection.py
+++ b/src/serena/util/inspection.py
@@ -57,10 +57,10 @@ def determine_programming_language_composition(repo_path: str) -> dict[Language,
         if count > 0:
             language_counts[language] = count
 
-    # Convert counts to percentages
+    # Convert counts to percentages (unrounded, so sorting by percentage is equivalent to sorting by count)
     language_percentages: dict[Language, float] = {}
     for language, count in language_counts.items():
         percentage = (count / total_files) * 100
-        language_percentages[language] = round(percentage, 2)
+        language_percentages[language] = percentage
 
     return language_percentages
--- a/test/serena/config/test_serena_config.py
+++ b/test/serena/config/test_serena_config.py
@@ -52,20 +52,35 @@ class TestProjectConfigAutogenerate:
         # Create files for multiple languages
         (self.project_path / "small.js").write_text("console.log('JS');")
 
-        # Run autogenerate - should pick Python as dominant
+        # Run autogenerate - TypeScript should always be detected for JS files
         config = ProjectConfig.autogenerate(self.project_path, save_to_disk=False)
 
-        assert config.languages == [Language.TYPESCRIPT]
+        assert Language.TYPESCRIPT in config.languages
 
     def test_autogenerate_with_multiple_languages(self):
-        """Test autogeneration picks dominant language when multiple are present."""
-        # Create files for multiple languages
+        """Test autogeneration enables all detected languages by default in non-interactive mode."""
+        # Create files for multiple languages (2 Python, 1 JS/TS)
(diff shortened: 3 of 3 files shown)


droideronline:
@opcode81 - would love to get your eyes on this when you get a chance. This is blocking a use case where Serena is invoked autonomously via MCP (no human in the loop to configure `project.yml`), and polyglot repos silently lose symbol support for secondary languages. Happy to adjust the approach or add tests based on your feedback.

droideronline:
Addressed both Copilot review comments in commit `78ebdac4`:

1. **Sorting precision** — `determine_programming_language_composition()` now returns raw (unrounded) percentages, so the sort order in `autogenerate()` is fully deterministic without rounding ties.

2. **Tests** — Updated the test suite:
   - `test_autogenerate_with_js_files` and `test_autogenerate_custom_project_name` now use `in` assertions (Vue's `FilenameMatcher` overlaps with `.js`/`.ts` extensions, so both languages are correctly detected when `auto_detect_all_languages=True`)
   - `test_autogenerate_with_multiple_languages` now asserts both Python and TypeScript are enabled (the new default behaviour)
   - Added `test_autogenerate_with_multiple_languages_dominant_only` to cover `auto_detect_all_languages=False` (the old single-language behaviour)

All 9 autogenerate tests pass locally.

opcode81:
EDIT: See [comment in issue](https://github.com/oraios/serena/issues/1125#issuecomment-3992818473)

PR #1128, opened 2026-03-04
fix: append PID to log filename to prevent concurrent server collision

ClementBourt:
When multiple MCP servers start within the same second, they derive the
  same log filename from datetime_tag() (second-level precision only).
  Both then open the file with mode='w', causing one server's logs to be
  truncated and all subsequent output to be interleaved.

  Appending os.getpid() to the filename guarantees uniqueness across
  concurrent processes regardless of timing.


opcode81:
Yes, that's a reasonable addition, thanks.

ClementBourt:
Thanks for the tool.

PR #1130, opened 2026-03-05
fix(kotlin): redirect idea.system.path to prevent stale /tmp dirs

### LINKED ISSUE #1087: fix(kotlin): KLS leaves stale `idea-system*` temp dirs in /tmp, consuming RAM on tmpfs systems
## Summary

The Kotlin Language Server (KLS) is built on IntelliJ IDEA's platform. Each KLS instance creates an `idea-system*` temporary directory under `/tmp`. These directories are never cleaned up after the language server shuts down, causing unbounded accumulation across test runs.

## Observed Impact

On a developer machine with `/tmp` mounted as **tmpfs** (RAM-backed — the default on modern Linux systems), repeated test runs accumulated:

```
329 x idea-system*/  = 30 GiB in RAM
/tmp (tmpfs): 36 GiB used total
```

Each KLS instance produces ~52 MB in its `idea-system*` directory. After 329 test sessions the entire tmpfs was nearly exhausted, starving the system of RAM and swap.

## Root Cause

IntelliJ's platform runtime defaults to creating its system directory at `/tmp/idea-system<random>/` when no explicit `idea.system.path` JVM property is set. The `KotlinLanguageServer` currently passes `JAVA_TOOL_OPTIONS` (heap size) but does not redirect this path:

```python
# src/solidlsp/language_servers/kotlin_language_server.py
def create_launch_command_env(self) -> dict[str, str]:
    env["JAVA_TOOL_OPTIONS"] = jvm_options   # only -Xmx2G by default
    return env
```

The `stop()` / `_shutdown()` implementation in `SolidLanguageServer` terminates the process but performs no filesystem cleanup.

## Suggested Fix

Two complementary changes in `KotlinLanguageServer`:

1. **Redirect `idea.system.path`** to a location we control (e.g. inside `ls_resources_dir` or a `tempfile.mkdtemp()` created at startup) by appending `-Didea.system.path=<path>` to `JAVA_TOOL_OPTIONS`.

2. **Clean up on stop** — override `stop()` in `KotlinLanguageServer` to delete the directory after the process terminates.

Example sketch:
```python
import tempfile, shutil

class KotlinLanguageServer(SolidLanguageServer):
    def __init__(self, ...):
        self._idea_system_dir = tempfile.mkdtemp(prefix="serena-idea-system-")
        ...

    def create_launch_command_env(self) -> dict[str, str]:
        env = super().create_launch_command_env()
        jvm_options = env.get("JAVA_TOOL_OPTIONS", "")
        env["JAVA_TOOL_OPTIONS"] = f"{jvm_options} -Didea.system.path={self._idea_system_dir}".strip()
        return env

    def stop(self, shutdown_timeout: float = 2.0) -> None:
        super().stop(shutdown_timeout)
        if hasattr(self, "_idea_system_dir"):
            shutil.rmtree(self._idea_system_dir, ignore_errors=True)
```

Alternatively the directory can live under `ls_resources_dir/kotlin_language_server/system/` — a persistent but bounded cache — if keeping the IntelliJ index across sessions is desirable (faster restarts). In that case no cleanup is needed, but its size should be documented.

## Environment

- Platform: Linux (x86-64), `/tmp` is `tmpfs`
- KLS version: 261.13587.0 (default bundled)
- Serena branch: `fix/kill-gradle-daemons-on-shutdown`

## Related

- PR #1076 / PR #1077 (Gradle daemon cleanup on shutdown — similar class of JVM process hygiene issues)

zerone0x:
## Summary

Fixes #1087

The Kotlin Language Server (KLS) is built on IntelliJ's platform, which by default creates a new `idea-system<random>/` directory in `/tmp` for each instance. These directories are never removed on shutdown, causing unbounded accumulation.

On `tmpfs` systems (the default on modern Linux), this means unbounded RAM consumption — the issue author observed 329 stale directories totalling 30 GiB after repeated test runs.

## Fix

In `DependencyProvider.create_launch_command_env()`, append `-Didea.system.path=<ls_resources_dir>/kotlin_language_server/system/` to `JAVA_TOOL_OPTIONS`. This redirects IntelliJ to a fixed, per-project location instead of `/tmp/idea-system<random>/`.

The chosen location (`ls_resources_dir/kotlin_language_server/system/`) is:
- **Bounded**: one directory per project (not one per run)
- **Persistent**: IntelliJ index is preserved across sessions → faster restarts
- **Controlled**: within Serena's own resource directory, not system-wide `/tmp`

## Changes

- **`src/solidlsp/language_servers/kotlin_language_server.py`**
  - `create_launch_command_env()`: compute `system_dir`, create it, and append `-Didea.system.path` to `JAVA_TOOL_OPTIONS`
  - Updated module docstring to document the IntelliJ system directory behaviour

## Testing

The fix can be verified by running the Kotlin language server tests and confirming that no new `idea-system*` directories appear in `/tmp` during the test run, while `<ls_resources_dir>/kotlin_language_server/system/` is populated instead.

---
🤖 Generated with Claude Code

### DIFF
--- a/src/solidlsp/language_servers/kotlin_language_server.py
+++ b/src/solidlsp/language_servers/kotlin_language_server.py
@@ -14,6 +14,13 @@ Example configuration for large projects:
     ls_specific_settings:
       kotlin:
         jvm_options: '-Xmx4G -XX:+UseG1GC'
+
+IntelliJ system directory:
+    KLS is built on IntelliJ's platform, which by default creates a new ~52 MB
+    idea-system<random>/ directory in /tmp for each instance. These directories
+    are never cleaned up, which can exhaust RAM on tmpfs systems after many runs.
+    Serena redirects idea.system.path to <ls_resources_dir>/kotlin_language_server/system/,
+    making it a bounded per-project cache that persists across sessions (faster restarts).
 """
 
 import logging
@@ -146,7 +153,18 @@ class KotlinLanguageServer(SolidLanguageServer):
             else:
                 jvm_options = DEFAULT_KOTLIN_JVM_OPTIONS
 
-            env["JAVA_TOOL_OPTIONS"] = jvm_options
+            # Redirect IntelliJ's system directory to a stable path under ls_resources_dir.
+            # Without this, each KLS instance creates a new ~52 MB /tmp/idea-system<random>/ directory
+            # that is never removed on shutdown. On tmpfs systems this causes unbounded RAM consumption
+            # (e.g. 329 stale directories totalling 30 GiB after repeated test runs).
+            # Using a fixed path makes it a bounded per-project cache and preserves the IntelliJ index
+            # across sessions for faster restarts.
+            system_dir = os.path.join(self._ls_resources_dir, "kotlin_language_server", "system")
+            os.makedirs(system_dir, exist_ok=True)
+            idea_system_path_opt = f"-Didea.system.path={system_dir}"
+            jvm_options_full = f"{jvm_options} {idea_system_path_opt}".strip() if jvm_options else idea_system_path_opt
+
+            env["JAVA_TOOL_OPTIONS"] = jvm_options_full
             return env
 
     @staticmethod


src/solidlsp/language_servers/kotlin_language_server.py
```diff
@@ -146,7 +153,18 @@ def create_launch_command_env(self) -> dict[str, str]:
...
                 jvm_options = DEFAULT_KOTLIN_JVM_OPTIONS
 
-            env["JAVA_TOOL_OPTIONS"] = jvm_options
+            # Redirect IntelliJ's system directory to a stable path under ls_resources_dir.
+            # Without this, each KLS instance creates a new ~52 MB /tmp/idea-system<random>/ directory
+            # that is never removed on shutdown. On tmpfs systems this causes unbounded RAM consumption
+            # (e.g. 329 stale directories totalling 30 GiB after repeated test runs).
+            # Using a fixed path makes it a bounded per-project cache and preserves the IntelliJ index
+            # across sessions for faster restarts.
+            system_dir = os.path.join(self._ls_resources_dir, "kotlin_language_server", "system")
```
opcode81, line 162: `self._ls_resources_dir` is already specific to the Kotlin LS.

PR #1131, opened 2026-03-05
feat: add configurable project_serena_folder_location setting

### LINKED ISSUE #1123: Allow project-specific Serena folder location to be configurable
(Motivated by discussion #1122)

Add to `SerenaConfig` a setting `project_serena_folder_location` such that one could configure, e.g. based on placeholders

* `$projectDir`
* `$projectName`

something like this:
```yaml
# default
project_serena_folder_location: $projectDir/.serena

# other example
project_serena_folder_location: /projects-metadata/$projectName/.serena
```

Implementation details:
* Add method `SerenaConfig.get_project_serena_folder(self, project_name, project_root)` which applies the replacement of the variables
* `Project`: compute the data folder upon initialisation (using the above method)
* Change `Project.path_to_serena_data_folder(self)` to return the precomputed folder location
* Get rid of `serena_config.get_serena_managed_in_project_dir`
* Change `SolidLSPSettings.project_data_relative_path` to be an absolute path
* Change `MemoriesManager` to directly accept the project memories path

cx-alberto-simoes:
## Summary

Resolves #1123 

- Add a new `project_serena_folder_location` setting to `serena_config.yml` that allows configuring where the per-project `.serena` data folder (memories, caches) is stored, with `$projectDir` and `$projectName` placeholder support (default: `$projectDir/.serena`, preserving current behavior).
- Implement fallback logic when loading a project: use configured path if it exists, fall back to existing project-root `.serena` folder for backward compatibility, otherwise create at the configured path.
- Refactor `SolidLSPSettings.project_data_relative_path` to an absolute `project_data_path`, change `MemoriesManager` to accept the memories path directly, and remove the `get_serena_managed_in_project_dir` helper function.

## Changes

| File | Change |
|---|---|
| `src/serena/config/serena_config.py` | Add `project_serena_folder_location` field, `get_project_serena_folder()` method, `DEFAULT_PROJECT_SERENA_FOLDER_LOCATION` constant; remove `get_serena_managed_in_project_dir` |
| `src/serena/project.py` | Add `_resolve_serena_data_folder()` with fallback logic; `path_to_serena_data_folder()` returns precomputed value; `MemoriesManager` takes absolute path |
| `src/solidlsp/settings.py` | Rename `project_data_relative_path` → `project_data_path` (absolute) |
| `src/solidlsp/ls.py` | Use `project_data_path` directly for cache dir |
| `src/serena/ls_manager.py` | `LanguageServerFactory` accepts and forwards `project_data_path` |
| `src/serena/resources/serena_config.template.yml` | Add documented `project_serena_folder_location` setting |
| `docs/02-usage/050_configuration.md` | Add "Per-Project Serena Folder Location" documentation section |
| `test/serena/config/test_serena_config.py` | Add 13 new tests (placeholder substitution, fallback logic, custom memories path) |
| `test/conftest.py`, `test/solidlsp/csharp/...`, `test/solidlsp/groovy/...` | Update to use `project_data_path` |

## Test plan

- [x] All 13 new tests pass (`TestGetProjectSerenaFolder`, `TestProjectSerenaDataFolder`, `TestMemoriesManagerCustomPath`)
- [x] All existing config tests pass (34 total)
- [x] Non-LS serena tests pass (119 total)
- [x] Python language server tests pass (30 total, validates SolidLSPSettings change end-to-end)
- [x] `uv run poe type-check` — clean (mypy, 0 issues)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/test/serena/config/test_serena_config.py
+++ b/test/serena/config/test_serena_config.py
@@ -490,12 +490,6 @@ class TestProjectSerenaDataFolder:
         project = self._make_project(config)
         assert project.path_to_serena_data_folder() == str(custom_serena)
 
-    def test_no_config_uses_project_root(self):
-        """When serena_config is None, fall back to the default project root location."""
-        project = self._make_project(serena_config=None)
-        expected = os.path.abspath(str(self.project_path / SERENA_MANAGED_DIR_NAME))
-        assert project.path_to_serena_data_folder() == expected
-
 
 class TestMemoriesManagerCustomPath:
     """Tests for MemoriesManager with a custom serena data folder."""


cx-alberto-simoes:
After reviewing it, I think it is acceptable, but let me know of any changes you like.

src/serena/config/serena_config.py
opcode81: Both of these functions have to be removed, as their logic is no longer valid.
Replacing them will mean moving the logic in `Project._resolve_serena_data_folder` to `serena_config`.

We need three functions:
* One that returns the location that is configured (now `SerenaConfig.get_project_serena_folder`) -> propose rename to `get_configured_project_serena_folder`
* One that does what `Project._resolve_serena_data_folder` does -> propose `SerenaConfig.get_project_serena_folder`
* An additional function `SerenaConfig.get_project_yml_location` which is built on the preceding one. This function can replace these two functions.

opcode81: Note that this changes the dependency structure of users of these functions, as they will now need the SerenaConfig instance.

docs/02-usage/050_configuration.md
```diff
@@ -172,6 +172,35 @@ For advanced users, Serena's configuration can be further customized.
...
+
+By default, each project stores its Serena data (memories, caches, etc.) in a `.serena` folder inside the project root.
+You can customize this location globally via the `project_serena_folder_location` setting in `serena_config.yml`.
+
+The setting supports two placeholders:
+
+| Placeholder    | Description                                           |
+|----------------|-------------------------------------------------------|
+| `$projectDir`  | The absolute path to the project root directory       |
+| `$projectName` | The name of the project (as defined in `project.yml`) |
```
opcode81: This is a bit misleading, because initially, project.yml doesn't exist yet.
Propose "The name of the project (i.e. the project's folder name by default)"

cx-alberto-simoes:
@opcode81 My understanding of the code is limited, so, again, I did my work based on cursor AI, which might go completely against your thoughts. Please check if it is something like that, or not related at all to what you was thinking about.

opcode81:
@cx-alberto-simoes I have added a commit which adjusts the interfaces to require `SerenaConfig` as proposed.


Loose ends:

1. not sure if all the tests are correctly adjusted
2. we need to change $projectName to $projectFolderName, because we really cannot know the project name before we know where the data folder is. Note: This means that we can change the interfaces to use only `project_root`, because the folder name can, of course, be derived from the project root.

opcode81:
I will make the additional interface adjustments (2) before I pass it back to you.

opcode81:
OK, passing it back to you.

opcode81:
Thanks @cx-alberto-simoes, when the tests pass, I will merge this (with squashing to keep the history clean).

PR #1135, opened 2026-03-05
Feature: configure read-only memories through regex patterns

### LINKED ISSUE #1132: Feature: more granular configuration of read-only memories
The motivation is to allow some global memories to be flagged as editable while other should be read only. That's useful because often not all global memories are "owned" by the same developer.

A backwards compatible solution could be a composite type `bool | dict[bool]` in python reflecting the options

```yaml
edit_global_memories: true/false
```
and
```
edit_global_memories:
   path_prefix1: false
   path_prefix2: true
```
where for all non-configured paths we assume the current default (editing permitted).

MischaPanch:
Closes #1132

### DIFF
--- a/src/serena/config/serena_config.py
+++ b/src/serena/config/serena_config.py
@@ -748,6 +748,14 @@ class SerenaConfig(SharedConfig):
                 instance.log_level = loaded_commented_yaml["gui_log_level"]
             del loaded_commented_yaml["gui_log_level"]
 
+        # migrate "edit_global_memories"
+        if "edit_global_memories" in loaded_commented_yaml:
+            num_migrations += 1
+            edit_global_memories = loaded_commented_yaml["edit_global_memories"]
+            if not edit_global_memories:
+                instance.read_only_memory_patterns.append("global/.*")
+            del loaded_commented_yaml["edit_global_memories"]
+
         # re-save the configuration file if any migrations were performed
         if num_migrations > 0:
             log.info("Legacy configuration was migrated; re-saving configuration file")


MischaPanch:
@EarthmanWeb FYI. This should allow very flexible and simple configuration of what is read only, in global and local config. Together with #1134 this should solve all your pains, I believe

src/serena/resources/project.template.yml
```diff
@@ -123,3 +123,6 @@ initial_prompt: ""
 # This overrides the corresponding setting in the global configuration; see the documentation there.
 # If null or missing, use the setting from the global configuration.
 symbol_info_budget:
+
+# Extends the same setting from the global configuration, merging the two lists.
```
opcode81: First line should contain a brief description of *what* this is (as in global config).

src/serena/project.py
```diff
@@ -30,15 +32,23 @@ class MemoriesManager:
...
-        self._global_memory_tool_write_access = global_memory_tool_write_access
+        self._project_memory_dir: Path | None = None
+        if serena_data_folder is not None:
+            self._project_memory_dir = Path(serena_data_folder) / "memories"
+            self._project_memory_dir.mkdir(parents=True, exist_ok=True)
         self._encoding = SERENA_FILE_ENCODING
+        self._read_only_memory_patterns = [re.compile(pattern) for pattern in set(read_only_memory_patterns)]
+
+    def _is_read_only_memory(self, name: str) -> bool:
+        for read_only_memory in self._read_only_memory_patterns:
```
opcode81: Really misleading varname; `pattern`

src/serena/project.py
```diff
@@ -77,9 +88,8 @@ def get_memory_file_path(self, name: str) -> Path:
 
     def _check_write_access(self, name: str, is_tool_context: bool) -> None:
         # in tool context, global memory write access can be disabled
```
opcode81: Comment is stale.

src/serena/tools/memory_tools.py
```diff
@@ -56,7 +56,8 @@ def list_memories(self, topic: str = "") -> list[str]:
 
     def apply(self, topic: str = "") -> str:
         """
-        List available memories, optionally filtered by topic.
+        List available memories, optionally filtered by topic. Memories marked as read-only cannot be
```
opcode81: This is obvious. It doesn't need to be stated, so I wouldn't add this.

src/serena/resources/serena_config.template.yml
```diff
@@ -68,11 +68,10 @@ ls_specific_settings: {}
...
 
-# whether global memories (the "global" topic) are allowed to be deleted or edited by the agent.
-# Global memories are stored in ~/.serena/memories/global/ and shared across all projects.
-# Set to false to protect global memories from accidental modification by agents.
-# Note: We recommend tracking global memories with git.
-edit_global_memories: True
+# Mark memories as read-only through regex patterns.
+# For example, "global/.*" will mark global memories as read-only.
+# You can extend the list on a per-project basis in the project.yml configuration file.
+read_only_memory_patterns: []
```
opcode81: Convention: First line should concisely state *what* it is (same as docstrings, initial lower-case).
Proposal:
> list of regex patterns which, when matched, mark a memory entry as read‑only.

Note that I mostly prefer passive, i.e. "This can be extended" rather than "You can extend this" for a more professional tone. Perhaps we should align on a style.

src/serena/project.py
```diff
@@ -30,15 +32,23 @@ class MemoriesManager:
...
+        self._project_memory_dir: Path | None = None
+        if serena_data_folder is not None:
+            self._project_memory_dir = Path(serena_data_folder) / "memories"
+            self._project_memory_dir.mkdir(parents=True, exist_ok=True)
         self._encoding = SERENA_FILE_ENCODING
+        self._read_only_memory_patterns = [re.compile(pattern) for pattern in set(read_only_memory_patterns)]
+
+    def _is_read_only_memory(self, name: str) -> bool:
+        for read_only_memory in self._read_only_memory_patterns:
+            if read_only_memory.match(name):
```
opcode81: This is truthy if the regex matches only a prefix, which wasn't your intention, right?
If you want to require a full match, use `re.fullmatch` instead.

src/serena/project.py
```diff
@@ -95,24 +105,27 @@ def save_memory(self, name: str, content: str, is_tool_context: bool) -> str:
...
         if not search_dir.exists():
             return []
         results = []
+        read_only_postfix = " (read only)"
         for md_file in search_dir.rglob("*.md"):
             rel = str(md_file.relative_to(base_dir).with_suffix("")).replace(os.sep, "/")
-            results.append(prefix + rel)
+            memory_name = prefix + rel
+            if self._is_read_only_memory(memory_name):
+                memory_name += read_only_postfix
```
opcode81: This approach is simple but could be problematic. The LLM currently gets the JSON list back, so it will look like this,
```
["foo", "bar (read-only)"]
```
which is quite misleading.

Maybe we should return a dataclass with separate lists (`writable_memories`, `read_only_memories`), which we can straightforwardly convert to a dictionary for the LLM.

opcode81: @MischaPanch, this makes me consider an alternative way of handling this: Making the naming define the status.
If we are going to modify names anyway, why not make the naming the defining criterion?
**Every memory or (sub-)topic name that ends with `_ro` is read-only**

Effects:
* Ease of use: The user does not need to edit `serena_config.yml` alongside the actual memories. One can change which memories are read-only directly while editing the memories themselves.
* SerenaConfig is not affected at all and remains leaner; the interfaces don't have to be changed as they were in this PR.
* "_ro" is less verbose than " (read-only)", which may or may not mean fewer tokens, but we do have to briefly explain the suffix in the tool instructions.

Perhaps it's worth considering.

EarthmanWeb:
My thoughts, as this pertains to my use case...

Requiring folder renaming does not support legacy naming conventions that
are, as the name suggests, read only.

Given the new support of a local (user specific) serena_config, it would
make most sense to define readonly folders there, as it may change between
users depending on memory ownership.

opcode81:
@EarthmanWeb we were primarily trying to address the use case where memories shall not be edited by the LLM only, with the user having full control. But if you have the additional requirement that the user cannot change anything either, then I suppose the naming-based option is not the way to go.

EarthmanWeb:
Let me clarify, you seemed to say that you were letting the LLM name the
read only memories externally.

If that's just internal, then that makes sense, but if it's an external
naming as in writing the name suffix to the folders or files, that's what I
was balking against

T

opcode81:
I will finish this up, sticking with the existing approach...

opcode81:
Limitation: Memories (including global memories) can only be edited if there is an active project. We might want to improve this, but it's not a practically relevant limitation.

PR #1137, opened 2026-03-06
Add Lean 4 language support

mergeconflict:
Add full Lean 4 language server support to Serena, including:
- Language server implementation using `lean --server` with `DependencyProviderSinglePath`
- Cross-file reference support via `LEAN_PATH`/`LEAN_SRC_PATH` from `lake env`
- File progress notification handling for LSP readiness
- Test repository with within-file and cross-file reference patterns
- Test suite: symbols, within-file refs, cross-file refs, go-to-definition (within and across files)
- Integration tests in `test_serena_agent.py`
- CI via `leanprover/lean-action@v1`
- Documentation updates (README, language docs, changelog)

### DIFF
--- a/src/solidlsp/language_servers/lean4_language_server.py
+++ b/src/solidlsp/language_servers/lean4_language_server.py
@@ -3,14 +3,11 @@ Provides Lean 4 specific instantiation of the LanguageServer class.
 Uses the built-in Lean 4 language server (lean --server).
 """
 
-import glob
 import logging
 import os
 import pathlib
 import shutil
 import subprocess
-import threading
-import time
 from typing import cast
 
 from overrides import override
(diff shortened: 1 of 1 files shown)


src/solidlsp/language_servers/lean4_language_server.py
```diff
@@ -0,0 +1,221 @@
...
+        log.info("Sending initialize request from LSP client to LSP server and awaiting response")
+        init_response = self.server.send.initialize(initialize_params)
+
+        capabilities = init_response.get("capabilities", {})
+        log.info(f"Lean 4 LSP capabilities: {list(capabilities.keys())}")
+
+        self.server.notify.initialized({})
+
+        # Open all .lean files so the LSP can build cross-file references
+        lean_files = glob.glob(os.path.join(self.repository_root_path, "**", "*.lean"), recursive=True)
```
MischaPanch: This should not be necessary and this means no newly written lean files are analyzed. Normally, language servers handle this by themselves. Is this not the case for lean? That would be a fundamental limitation of the LS.

If there is no way for the lean ls to properly take care of this, the last resort would be to use a file watcher.

mergeconflict: Thanks @MischaPanch, I think I've fixed this in 1962232.

MischaPanch:
Great, thank you, merging! No idea why seemingly unrelated tests now fail in CI

PR #1145, opened 2026-03-09
feat: add excluded_memory_listing_dirs config option

YouJianFengXue:
## Summary

- Add `excluded_memory_listing_dirs` config option to exclude top-level directories from memory listing
- Configurable at both global (`serena_config.yml`) and project (`project.yml`) level
- Lists from both levels are merged additively (same pattern as `read_only_memory_patterns`)
- Backward compatible: missing key defaults to empty list

## Problem

Projects with large numbers of archived memory files (e.g. `_archive/`, `_episodes/`) see inflated `activate_project` and `list_memories` output. In one real case, 1,700+ archived `.md` files caused `activate_project` output to balloon to ~26k tokens — most of it useless noise.

There is currently no way to exclude directories from memory listing. `read_only_memory_patterns` only marks entries as read-only but still lists them. `ignored_paths` only controls code navigation, not memory listing.

## Solution

Add `excluded_memory_listing_dirs: list[str]` to `SharedConfig` (inherited by both `SerenaConfig` and `ProjectConfig`). The `MemoriesManager._list_memories` method checks each file's top-level directory against this set and skips matches.

Files inside excluded directories can still be read directly via `read_memory` (which uses direct path resolution via `get_memory_file_path`, not the listing method).

## Example usage

```yaml
# In serena_config.yml or project.yml
excluded_memory_listing_dirs: ["_archive", "_episodes"]
```

## Changes

| File | Change |
|------|--------|
| `src/serena/config/serena_config.py` | Add field to `SharedConfig`, pass in `ProjectConfig._from_dict` |
| `src/serena/project.py` | Accept param in `MemoriesManager.__init__`, filter in `_list_memories`, merge configs in `Project.__init__` |
| `src/serena/resources/serena_config.template.yml` | Add config option with documentation |
| `src/serena/resources/project.template.yml` | Add config option with documentation |

## Test plan

- [ ] Verify `activate_project` output excludes files in configured directories
- [ ] Verify `read_memory("_archive/...")" still works for excluded directories
- [ ] Verify backward compatibility with existing config files (missing key defaults to `[]`)
- [ ] Verify global + project config merging works correctly


opcode81:
If you do not actually want to use these memories anymore, why do you not move them out of the memories folder or delete them entirely?

YouJianFengXue:
Good question! The archived memories are still needed — they just shouldn't clutter the listing output.

**Why not delete them?**
They contain historical design rationale, bug investigations, and session context that are occasionally referenced. Deleting them would lose valuable project history.

**Why not move them out of `memories/`?**
We still need to access them via `read_memory("_archive/2026-03/some-topic")` for on-demand recall. Moving them outside the memories directory would break this path resolution, since `get_memory_file_path` resolves relative to the project's memory dir.

**The core use case:**
Over time, a project accumulates hundreds of memory files. We compress and archive old ones into `_archive/YYYY-MM/` subdirectories to keep the active set small, but they must remain readable. The problem is that `_list_memories` uses `rglob("*.md")` which recursively scans everything — in our case, 1,700+ archived files that balloon `activate_project` output to ~26k tokens.

This is analogous to how `ignored_paths` works for code navigation — the files still exist and are accessible, but they don't appear in directory listings or symbol searches. Similarly, `excluded_memory_listing_dirs` hides directories from listing while preserving `read_memory` access.

An alternative approach could be to support glob/regex patterns (like `read_only_memory_patterns` does) instead of just directory names, but directory-level exclusion seemed like the simplest solution that covers the common case. Happy to adjust the approach if you have a different preference!

MischaPanch:
>  support glob/regex patterns

I prefer this, it's more flexible and still easy to use. The directory exclusion is easy to cover with this. An option `ignored_memories: [regex_list]` would make sense to me.

@YouJianFengXue I disagree that `read_memory` should still work on ignored memories, this is counterintuitive IMO. If you want the agent to read an ignored memory, the explicit instruction could be "read the file in `.serena/memories/archive/...`" instead of using the memory tool for that. WDYT?

@opcode81

YouJianFengXue:
Thanks for the feedback! I've reworked the PR based on your suggestions:

**Key changes (v1 → v2):**

1. **Renamed to `ignored_memories: [regex_list]`** — regex patterns instead of directory names, consistent with `read_only_memory_patterns`
2. **`read_memory` no longer works on ignored memories** — accessing an ignored memory via `read_memory`, `write_memory`, `delete_memory`, `move_memory`, or `edit_memory` now raises a `ValueError` with a message directing the user to `read_file` on the raw path instead
3. **Same merge pattern** — global + project configs merged additively, missing key defaults to `[]`

**Example config:**
```yaml
ignored_memories: ["_archive/.*", "_episodes/.*"]
```

**Behavior:**
- `list_memories` / `activate_project`: ignored memories are excluded from output
- `read_memory("_archive/2026-03/some-topic")`: raises error, suggests using `read_file`
- `read_file(".serena/memories/_archive/2026-03/some-topic.md")`: still works (raw file access)

The implementation mirrors the existing `read_only_memory_patterns` pattern (`fullmatch` on compiled regex, check in all access methods).

src/serena/project.py
```diff
@@ -35,24 +35,46 @@ class MemoriesManager:
     GLOBAL_TOPIC = "global"
     _global_memory_dir = SerenaPaths().global_memories_path
 
-    def __init__(self, serena_data_folder: str | Path | None, read_only_memory_patterns: Sequence[str] = ()):
+    def __init__(
+        self,
+        serena_data_folder: str | Path | None,
+        read_only_memory_patterns: Sequence[str] = (),
+        ignored_memories: Sequence[str] = (),
```
MischaPanch: pls rename to ignored_memory_patterns

MischaPanch:
Thanks @YouJianFengXue . Could you pls extend changelog and docs?

src/serena/project.py
MischaPanch: Here and everywhere else, including in the templates . If something is a pattern, it should be highlighted as such in the key name

YouJianFengXue:
Done! Updated per your feedback:

1. **Renamed** `ignored_memories` → `ignored_memory_patterns` (consistent with `read_only_memory_patterns`)
2. **Added changelog** entry under "Memories" section in `CHANGELOG.md`
3. **Added docs** section "Ignoring Memories" in `docs/02-usage/045_memories.md` with usage example and behavior description

MischaPanch:
Note for us - we should extend the changelog with other recent changes to the memory system, just linking to docs should be enough

PR #1146, opened 2026-03-10
docs: add HTTP mode guidance to Project Workflow section

### LINKED ISSUE #1118: Improve documentation of HTTP mode
Right now, we don't discuss the ramifications at all, when to use, etc.

Specifically, we need to talk about this in the "Project Workflow" section, where we already talk about how to handle multiple projects, etc.

zerone0x:
Closes #1118

Adds a new **Running a Persistent Server in HTTP Mode** section to the Project Workflow documentation, addressing the gap noted in #1118.

The section explains when HTTP mode is the better choice over the default stdio mode:
- multiple simultaneous clients sharing one server instance and its state
- persistent language server state between conversations (avoids re-initialization delays)
- Docker/remote deployments where the server runs as a background process
- explicit control over server lifecycle

Also includes a security note about network port exposure, with a link to the existing Security Considerations page.

### DIFF
--- a/docs/02-usage/040_workflow.md
+++ b/docs/02-usage/040_workflow.md
@@ -197,3 +197,40 @@ Depending on the language backend being used, the management of resources for th
   where `<serena>` is your way of running Serena. For example, when using `uvx`, run
 
       uvx --from git+https://github.com/oraios/serena serena start-project-server
+
+(http-mode-workflow)=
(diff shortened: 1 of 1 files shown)


opcode81:
Thanks for this proposal, but the documentation in the workflow section needed to be more focused on workflow aspects (and less on technical aspects).

PR #1152, opened 2026-03-10
Add Luau language support

StayBlue:
Following https://github.com/oraios/serena/pull/1049#discussion_r2827885299, I created a new PR that moves the previous download implementation to the `DependencyProvider` path already used by existing language servers.

Luau is the language developed and used by Roblox, but it has seen some noteworthy adoption as a Lua replacement in some other areas. It has noteworthy features such as a native type system, sandboxing, and better performance.

Supersedes #1049
Closes #1049

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -544,6 +544,10 @@ jobs:
             echo "ERROR: FPCDIR does not exist: $FPCDIR"
             exit 1
           fi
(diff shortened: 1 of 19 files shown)


StayBlue:
@MischaPanch Let me know if you'd like me to change anything.

MischaPanch:
Thank you for the contribution, merging

PR #1154, opened 2026-03-10
fix: allow project creation when no LSP-supported languages are detected

### LINKED ISSUE #1053: Improve project creation for JetBrains users
The LSP languages are meaningless, and creating a project for which there is no supported LS should not be a problem.

When in JetBrains mode:
* Skip language detection
* Set languages to empty list (allow this)

zerone0x:
## Summary

Fixes #1053

Previously, `ProjectConfig.autogenerate()` raised a `ValueError` when no recognized source files were found (e.g. an empty directory, a JetBrains project without recognised extension types, or a project whose files are all gitignored). This blocked users from creating Serena projects in perfectly valid scenarios.

The JetBrains backend doesn't use language servers at all, so language detection is meaningless there — but the error still fired. The same applied to any user wanting to add a project before adding source files, or whose project uses only file types not matched by serena's language detector.

## Changes

- Replace the `ValueError` in `ProjectConfig.autogenerate()` with a `log.warning()` and an empty languages list when no source files are detected
- Move the language-selection / interactive-prompt logic into the `else` branch so it's only executed when languages were actually found
- Update `test_autogenerate_empty_directory` to reflect the new behaviour (expects empty languages, not an exception)
- Update `test_autogenerate_with_gitignored_files_only` similarly
- Add `test_autogenerate_empty_directory_logs_warning` to assert the warning is emitted

All 51 config tests pass locally.

### DIFF
--- a/src/serena/config/serena_config.py
+++ b/src/serena/config/serena_config.py
@@ -292,40 +292,40 @@ class ProjectConfig(SharedConfig):
                 language_composition = determine_programming_language_composition(str(project_root))
                 log.info("Language composition: %s", language_composition)
                 if len(language_composition) == 0:
-                    language_values = ", ".join([lang.value for lang in Language])
-                    raise ValueError(
-                        f"No source files found in {project_root}\n\n"
-                        f"To use Serena with this project, you need to either\n"
-                        f"  1. specify a programming language by adding parameters --language <language>\n"
-                        f"     when creating the project via the Serena CLI command OR\n"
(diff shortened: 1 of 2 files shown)


opcode81:
Thanks, @zerone0x, this the correct first step, so I will just merge this.
@MischaPanch we can think about refining the behaviour later (e.g. issue a warning when calling `find_symbol` with 0 active language servers instead of returning an empty list of results).

PR #1157, opened 2026-03-11
fix: add dedicated max_memory_length config for write_memory tool

### LINKED ISSUE #527: Remove length restrictions of memory tools
@bdruth thanks for the report! It was a bug, this tool was abusing the max_answer_chars concept. I pushed a hotfix in d0fa9bd53776ed, but we should change it to use an independent number, like max_memory_length.

_Originally posted by @MischaPanch in https://github.com/oraios/serena/issues/523#issuecomment-3217445220_

zerone0x:
fixes #527

write_memory was using `default_max_tool_answer_chars` to cap memory content length. that's the config for tool response sizes (what gets sent to the LLM), not for what gets written to disk — two different things.

added `max_memory_length: 50_000` to `SerenaConfig` as an independent limit. also added it to `serena_config.template.yml` so users can see and override it.

changes:
- `src/serena/config/serena_config.py`: new `max_memory_length` field (default 50k chars)
- `src/serena/resources/serena_config.template.yml`: documented the new option
- `src/serena/tools/memory_tools.py`: use `max_memory_length` instead of `default_max_tool_answer_chars`

### DIFF
--- a/src/serena/config/serena_config.py
+++ b/src/serena/config/serena_config.py
@@ -604,6 +604,12 @@ class SerenaConfig(SharedConfig):
     Even though the value of the max_answer_chars can be changed when calling the tool, it may make sense to adjust this default 
     through the global configuration.
     """
+    max_memory_length: int = 50_000
+    """Maximum number of characters that can be written to a memory file via the write_memory tool.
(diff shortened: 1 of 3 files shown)


opcode81:
As discussed in the ticket, I am closing this. Commit added which removes the length restriction in `write_memory`

PR #1164, opened 2026-03-14
Fix HLSL language server leaving orphaned processes on shutdown

TheSmallKiwi:
## Summary
- The base `_shutdown()` in `SolidLanguageServer` calls `process.terminate()` directly on the subprocess, which on Windows with `shell=True` only kills the `cmd.exe` wrapper, leaving the actual `shader-language-server` binary running as an orphan
- Overrides `stop()` in `HlslLanguageServer` to use `psutil` to terminate the full process tree before calling the standard shutdown, following the same pattern used by `VueLanguageServer`

## Test plan
- [x] All 10 existing HLSL tests pass
- [x] Verified `shader-language-server` process appears in tasklist after start
- [x] Verified no orphaned `shader-language-server` processes remain after shutdown

### DIFF
--- a/src/solidlsp/language_servers/hlsl_language_server.py
+++ b/src/solidlsp/language_servers/hlsl_language_server.py
@@ -9,6 +9,7 @@ import pathlib
 import shutil
 from typing import Any, cast
 
+import psutil
 from overrides import override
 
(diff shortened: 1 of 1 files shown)


opcode81:
Thanks for the improvement.

Issue #876, opened 2025-12-24
Indexing time out error on a Swift project


murad1981:
When I index my iOS Swift project using the command: `uvx --from git+https://github.com/oraios/serena serena project index` it indexes ~90% of the files successfully, the log is as follows:
```
xxxx/Common/AppNavigation/AppCoordinator.swift
Request timed out (timeout=10.0)
xxxx/InputFields/DirectionalTextField.swift
Request timed out (timeout=10.0)

with many other files has the same issue.
```

 I did the following:
- Creating the project
- Activating it
- Indexing it (90% success, 10% failed)
- on boarding it

Any hint on this would be appreciated.

opcode81:
We would need the Serena logs to analyse this.
So if the issue persists, please provide some logs.

Note, however, that the Serena developers are not mac users, so we cannot directly test any of this. Swift support is a community contribution.

opcode81:
No information provided.

PR #1166, opened 2026-03-14
Fix Swift find_referencing_symbols: add --scratch-path and resolve via xcrun

Asher-:
## Summary

- **Pass --scratch-path to sourcekit-lsp** — without it, there is no index store for background indexing, so textDocument/references always returns empty results. This is the root cause of #876.
- **Resolve sourcekit-lsp via xcrun with DEVELOPER_DIR** — on macOS, bare sourcekit-lsp can resolve to the Command Line Tools version (limited capabilities). Using xcrun --find with DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer gives the full Xcode version.
- **Increase local indexing delay from 5s to 10s** — real-world Swift projects need more time for background indexing before the first references request.
- **Add retry logic for local runs** — previously only CI retried on empty references; now local also retries once after a 3s delay.

## Root Cause Analysis

Two issues combined to make find_referencing_symbols return empty for Swift:

1. ProcessLaunchInfo launched sourcekit-lsp with no arguments. SourceKit-LSP needs --scratch-path to have a location for its background index store. Without it, textDocument/references returns empty because there is no index data for cross-file symbol resolution.

2. On macOS with both Xcode and Command Line Tools installed, bare sourcekit-lsp resolves to /Library/Developer/CommandLineTools/usr/bin/sourcekit-lsp which has limited indexing capabilities compared to the Xcode version.

## Test plan

- [x] Verified find_referencing_symbols returns cross-file references for Swift structs and methods
- [x] Verified find_symbol and get_symbols_overview continue to work
- [x] Verified sourcekit-lsp process launches with correct path and --scratch-path argument
- [x] Verified fallback to bare sourcekit-lsp when xcrun is unavailable (Linux compatibility)
- [ ] Run existing Swift test suite (test/solidlsp/swift/test_swift_basic.py)

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/solidlsp/language_servers/sourcekit_lsp.py
+++ b/src/solidlsp/language_servers/sourcekit_lsp.py
@@ -31,6 +31,36 @@ class SourceKitLSP(SolidLanguageServer):
         return super().is_ignored_dirname(dirname) or dirname in [".build", ".swiftpm", "node_modules", "dist", "build"]
 
     @staticmethod
+    def _resolve_sourcekit_lsp_path() -> str:
+        """Resolve the path to sourcekit-lsp, preferring Xcode's version over Command Line Tools.
+
+        On macOS, bare `sourcekit-lsp` may resolve to the Command Line Tools version which
+        has limited capabilities. Using `xcrun --find sourcekit-lsp` resolves to Xcode's
+        full-featured version when Xcode is installed.
+        """
+        # Try xcrun with DEVELOPER_DIR pointing to Xcode (not Command Line Tools)
+        xcode_path = "/Applications/Xcode.app/Contents/Developer"
+        for env_override in [{"DEVELOPER_DIR": xcode_path}, {}]:
+            try:
+                run_env = {**os.environ, **env_override} if env_override else None
+                result = subprocess.run(
+                    ["xcrun", "--find", "sourcekit-lsp"],
+                    capture_output=True, text=True, check=False, timeout=10,
+                    env=run_env,
+                )
+                if result.returncode == 0:
+                    resolved = result.stdout.strip()
+                    if resolved and os.path.isfile(resolved):
+                        log.info(f"Resolved sourcekit-lsp via xcrun: {resolved}")
+                        return resolved
+            except (FileNotFoundError, subprocess.TimeoutExpired):
+                pass
+
+        # Fall back to bare command (Linux, or macOS without Xcode)
+        log.info("xcrun not available or failed, falling back to bare sourcekit-lsp")
+        return "sourcekit-lsp"
+
+    @staticmethod
     def _get_sourcekit_lsp_version() -> str:
         """Get the installed sourcekit-lsp version or raise error if sourcekit was not found."""
         try:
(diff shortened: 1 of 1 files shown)


src/solidlsp/language_servers/sourcekit_lsp.py
```diff
@@ -344,24 +384,26 @@ def request_references(self, relative_file_path: str, line: int, column: int) ->
...
+        # Get references with retry logic — indexing may not be complete on first request
         references = super().request_references(relative_file_path, line, column)
 
-        # In CI, if no references found, retry once after additional delay
-        if os.getenv("CI") and not references:
-            log.info("No references found in CI - retrying after additional 5s delay")
-            time.sleep(5)
+        # If no references found, retry once after additional delay (indexing may still be in progress)
+        if not references:
+            retry_delay = 5 if os.getenv("CI") else 3
```
MischaPanch, line 404: This waits on every request, not just the first one. It shouldn't be needed, right? Empty references may be the correct result and always waiting for a long time there is not nice.

We introduced `_get_wait_time_for_cross_file_referencing` in the base class that is meant to be overridden in the child classes, seems like the implementation of sourcekit_lsp was never adapted to use it. This would be the preferred mechanism, there it's ensured that we wait only once.

src/solidlsp/language_servers/sourcekit_lsp.py
```diff
@@ -30,6 +30,36 @@ def is_ignored_dirname(self, dirname: str) -> bool:
...
+    @staticmethod
+    def _resolve_sourcekit_lsp_path() -> str:
+        """Resolve the path to sourcekit-lsp, preferring Xcode's version over Command Line Tools.
+
+        On macOS, bare `sourcekit-lsp` may resolve to the Command Line Tools version which
+        has limited capabilities. Using `xcrun --find sourcekit-lsp` resolves to Xcode's
+        full-featured version when Xcode is installed.
+        """
+        # Try xcrun with DEVELOPER_DIR pointing to Xcode (not Command Line Tools)
+        xcode_path = "/Applications/Xcode.app/Contents/Developer"
```
MischaPanch, line 42: is this path always the same on each mac? Otherwise it could be passed with `ls_specific_settings`

src/solidlsp/language_servers/sourcekit_lsp.py
```diff
@@ -49,8 +79,18 @@ def __init__(self, config: LanguageServerConfig, repository_root_path: str, soli
         sourcekit_version = self._get_sourcekit_lsp_version()
         log.info(f"Starting sourcekit lsp with version: {sourcekit_version}")
 
+        # Resolve sourcekit-lsp path — prefer Xcode's version over Command Line Tools
+        sourcekit_path = self._resolve_sourcekit_lsp_path()
+
+        # sourcekit-lsp needs --scratch-path for background indexing and cross-file references.
+        # Without it, textDocument/references returns empty because there's no index store.
+        scratch_path = os.path.join(repository_root_path, ".build", "sourcekit-lsp")
```
MischaPanch, line 87: our tests include references search without this. I'm perplexed as to why

MischaPanch (commented):
Thanks for the contribution! Just a few questions (see comments)

MischaPanch:
@Asher- do you want to finalize this?

TyceHerrman:
Hey @opcode81 pinging you since I see you were the last person to touch this. I'm willing to work on this and resolve ci checks failing so this can be merged. just ran into this issue locally myself.

MischaPanch:
Hi @TyceHerrman . Github shows every force-push on main here, @opcode81 never worked on this (and is travelling at the moment).

Feel free to pick it up and thank you, would be nice to fix this long-standing issue!

PR #1167, opened 2026-03-14
fix: add meta property to flake.nix

MikhaD:
## Summary
Add the meta property to the serena package in the `flake.nix`.

## Rationale
I wanted this specifically to get the `lib.getExe` nix function work.

## Testing
I tested this by changing my personal system's nix flake to point to my fork, and confirmed that serena continues to work as expected, and the `lib.getExe` function now works.


MischaPanch:
Thanks! Merging

PR #1169, opened 2026-03-14
Force-update all comments in .yml files when resaving

opcode81:


src/serena/util/yaml.py
```diff
@@ -203,13 +203,14 @@ def yaml_comment_entry_is_empty(comment_entry: Any) -> bool:
 
 
 def transfer_missing_yaml_comments_by_index(
-    source: CommentedMap, target: CommentedMap, indices: list[int], forced_update_keys: Sequence[str] = ()
+    source: CommentedMap, target: CommentedMap, indices: list[int], forced_update_keys: Sequence[str] = (), force_update_all: bool = False
```
MischaPanch, line 206: I like force updating by default, but a few powerusers may want to keep their comments. We could accommodate both by adding an "advanced" global config option `allow_config_comment_overwrites`. Wdyt? Or not worth the complexity?

PR #1172, opened 2026-03-15
feat: add MiniMax as an OpenAI-compatible model provider

octo-patch:
## Summary

Add [MiniMax](https://www.minimax.io/) as a supported model provider via Agno's `OpenAILike` class. MiniMax offers models like MiniMax-M2.5 and MiniMax-M2.5-highspeed with 204K context windows through an OpenAI-compatible API at `https://api.minimax.io/v1`.

## Changes

- **`scripts/agno_agent.py`**: Add `OpenAILike` import and commented-out MiniMax model example alongside existing Claude and Gemini options
- **`.env.example`**: Add `MINIMAX_API_KEY` placeholder
- **`docs/03-special-guides/custom_agent.md`**: Document how to use MiniMax via `OpenAILike` with a code snippet
- **`README.md`**: Mention MiniMax as an example of OpenAI-compatible providers in the LLM Integration section

## Why MiniMax?

- OpenAI-compatible API — works out of the box with Agno's `OpenAILike`
- 204K context window — ideal for large codebase analysis with Serena
- Available models: `MiniMax-M2.5` and `MiniMax-M2.5-highspeed`

## Test Plan

- [x] Python syntax verified (`ast.parse`)
- [x] `ruff check` passes
- [x] `black --check` passes
- [x] Verified `OpenAILike` import works with installed Agno version

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,2 +1,3 @@
 GOOGLE_API_KEY=<your_google_api_key>
 ANTHROPIC_API_KEY=<your_key>
+MINIMAX_API_KEY=<your_minimax_api_key>
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ In general, Serena can be integrated with an LLM in several ways:
 * by using [mcpo to connect it to ChatGPT](docs/03-special-guides/serena_on_chatgpt.md) or other clients that don't support MCP but do support tool calling via OpenAPI.
 * by incorporating Serena's tools into an agent framework of your choice, as illustrated [here](docs/03-special-guides/custom_agent.md).
   Serena's tool implementation is decoupled from the framework-specific code and can thus easily be adapted to any agent framework.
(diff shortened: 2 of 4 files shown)


MischaPanch:
Our script is only an example, we don't document all of Agno's models or functionalities. I'd say we don't need to enhance our docs here, a user of Agno should read Agno's docs on model support

octo-patch:
Thanks for the feedback, @MischaPanch! That makes sense — if the script is intended as a minimal example rather than comprehensive documentation, I understand not wanting to expand it.

Would you prefer to close this PR, or is there another area of the project where MiniMax provider support would be a better fit? Happy to adjust the approach either way.

MischaPanch:
Let's close this, I think our docs sufficiently cover this :).

Thanks nevertheless for the PR

PR #1173, opened 2026-03-15
fix: detect rust-analyzer from PATH without rustup

### LINKED ISSUE #800: rust-analyzer fails if rustup is not installed even when rust-analyzer exists in PATH
## Suspected root cause 

In the Serena version I'm using, `RustAnalyzer._ensure_rust_analyzer_installed()` only relies on `RustAnalyzer._get_rust_analyzer_path()`.
In that version, `_get_rust_analyzer_path()` appears to only call `rustup which rust-analyzer` and does not fall back to the system `PATH`.

As a result, even if `rust-analyzer` is installed independently and is present on `PATH`, Serena still reports:

> Neither rust-analyzer nor rustup is installed.
> Please install Rust via https://rustup.rs/ or install rust-analyzer separately.

and there is no way to satisfy the “install rust-analyzer separately” part with just a standalone `rust-analyzer` on `PATH`.

---

## Environment

* OS: NixOS
* `rust-analyzer`:
  * Path: `/home/lum1narie/.nix-profile/bin/rust-analyzer`
    * installed via home-manager
  * Version:
    ```bash
    $ which rust-analyzer
    /home/lum1narie/.nix-profile/bin/rust-analyzer
    $ rust-analyzer --version
    rust-analyzer 2025-05-12
    ```
* `rustup`: not installed (`shutil.which("rustup")` returns `None`)
* Serena invocation (when reproducing):
  ```bash
  env PATH="$HOME/.nix-profile/bin:$PATH" \
    uvx --from git+https://github.com/oraios/serena \
    serena start-mcp-server \
      --project /home/lum1narie/Development/rust_scad/scadman
  ```
  * tried `env PATH="$HOME/.nix-profile/bin:$PATH"` to ensure `~/.nix-profile/` is in the `$PATH` for serena
* Serena source commit inspected:
  `855818da6357ab37fe1d413f5451f1b7027a99e6`, file
  `src/solidlsp/language_servers/rust_analyzer.py`
  (the error originates around L82 in that file).

## Actual behaviour

Even though `rust-analyzer` is installed and visible on `PATH` via Nix,
Serena fails to start the Rust language server with:

```text
INFO  ... Language server startup (language=rust) starting ...
INFO  ... Creating language server instance for /home/lum1narie/Development/rust_scad/scadman, language=rust.
ERROR ... Language server startup (language=rust) failed after 0.028 seconds
ERROR ... Error starting language server for language rust: Neither rust-analyzer nor rustup is installed.
Please install Rust via https://rustup.rs/ or install rust-analyzer separately.
  File ".../serena/ls_manager.py", line 93, in start_language_server
    language_server = factory.create_language_server(language)
  ...
  File ".../solidlsp/language_servers/rust_analyzer.py", line 103, in __init__
    rustanalyzer_executable_path = self._ensure_rust_analyzer_installed()
  File ".../solidlsp/language_servers/rust_analyzer.py", line 82, in _ensure_rust_analyzer_installed
    raise RuntimeError(
RuntimeError: Neither rust-analyzer nor rustup is installed.
Please install Rust via https://rustup.rs/ or install rust-analyzer separately.
```

The same error also occurs when using the Nix-packaged `serena` binary (instead of `uvx`).

## Additional diagnostics

To confirm that `rust-analyzer` is indeed visible from within the same uv environment, I ran a small test script.

`check_ra.py`:

```python
import os, shutil

print("PATH =", os.environ.get("PATH", ""))
print("rust-analyzer =", shutil.which("rust-analyzer"))
print("rustup =", shutil.which("rustup"))
```

Execution:

```bash
$ env PATH="$HOME/.nix-profile/bin:$PATH" \
    uvx --from git+https://github.com/oraios/serena \
    python check_ra.py

PATH = /home/lum1narie/.cache/uv/archive-v0/wRXg1E3NpepOlKRM2fjxe/bin:/home/lum1narie/.nix-profile/bin:...
rust-analyzer = /home/lum1narie/.nix-profile/bin/rust-analyzer
rustup = None
```

`check_ra2.py`:

```python
import os
from solidlsp.language_servers.rust_analyzer import RustAnalyzer

print("PATH =", os.environ.get("PATH", ""))
print("RustAnalyzer._get_rust_analyzer_path() =", RustAnalyzer._get_rust_analyzer_path())
```

Execution:

```bash
$ env PATH="$HOME/.nix-profile/bin:$PATH" \
    uvx --from git+https://github.com/oraios/serena \
    python check_ra2.py

PATH = /home/lum1narie/.cache/uv/archive-v0/wRXg1E3NpepOlKRM2fjxe/bin:/home/lum1narie/.nix-profile/bin:...
RustAnalyzer._get_rust_analyzer_path() = None
```

So:
* Inside the same uv-created Python environment, `shutil.which("rust-analyzer")` correctly finds the Nix-provided binary
* However, the Rust language server startup logic still throws the “Neither rust-analyzer nor rustup is installed” error

## Code-level analysis (suspected issue)

In the linked `rust_analyzer.py` (commit `855818da...`),
the RuntimeError is raised inside `RustAnalyzer._ensure_rust_analyzer_installed()` around L82.

https://github.com/oraios/serena/blob/855818da6357ab37fe1d413f5451f1b7027a99e6/src/solidlsp/language_servers/rust_analyzer.py#L82-L85

In the version I’m running, `_get_rust_analyzer_path()` appears to only call `rustup which rust-analyzer` and does not look at the system `PATH` at all (no `shutil.which("rust-analyzer")` fallback), despite the error text suggesting that a standalone `rust-analyzer` installation would be sufficient.

This leads to the following behaviour:

* On a system without `rustup`, `_get_rust_analyzer_path()` always returns `None`
* `_ensure_rust_analyzer_installed()` immediately raises the RuntimeError
* Even if `rust-analyzer` is installed separately and is available on `PATH`, it is never considered

This exactly matches my setup:

* `rust-analyzer` is installed via Nix and present on `PATH`
* `rustup` is not installed
* Yet Serena always reports “Neither rust-analyzer nor rustup is installed.”

## Expected behaviour

I would expect either:

1. Implementation to match the message:
   * `_get_rust_analyzer_path()` should check both `rustup` and the system `PATH`, e.g. by using `shutil.which("rust-analyzer")` as a fallback.
   * If `rust-analyzer` is found on `PATH`, use that without requiring `rustup`.
2. Or the error message to match the implementation:
   * If the intended behaviour is to support only `rustup`-managed toolchains, the message should not suggest that installing `rust-analyzer` separately is a viable fix when it is not actually checked.

VoidChecksum:
## Summary

Closes #800

- **Moves the `shutil.which("rust-analyzer")` PATH check before hardcoded common location fallbacks** in `_ensure_rust_analyzer_installed()`. This ensures standalone installs (Nix, distro packages, manual installs) are found promptly.
- **Adds `_is_rust_analyzer_functional()` validation** that runs `rust-analyzer --version` on PATH-found binaries. This catches the edge case where `rust-analyzer` in PATH is a rustup proxy symlink that fails when the component isn't installed (the concern from #800's discussion with @alichtl).
- **Detection priority is now**: rustup → rustup auto-install → PATH (with validation) → common locations.

### Root cause

On NixOS (and similar setups), `rust-analyzer` is installed standalone via the system package manager and available in `$PATH`, but `rustup` is not installed. The previous detection order checked common hardcoded paths before `$PATH`, meaning Nix paths like `~/.nix-profile/bin/rust-analyzer` were only found as a last resort — or not at all on older versions that lacked the PATH fallback entirely.

## Test plan

- [x] New test: `test_detect_from_nix_path_without_rustup` — verifies Nix-style PATH detection works
- [x] New test: `test_skip_broken_rustup_proxy_in_path` — verifies broken rustup proxy symlinks are skipped
- [x] Updated test: `test_detect_from_path_when_functional` — updated for new priority order
- [x] All 17 existing rust-analyzer detection tests pass
- [x] `uv run poe type-check` — clean (mypy passes)

### DIFF
--- a/test/solidlsp/rust/test_rust_analyzer_detection.py
+++ b/test/solidlsp/rust/test_rust_analyzer_detection.py
@@ -102,17 +102,26 @@ class TestRustAnalyzerDetection:
         """
         from solidlsp.language_servers.rust_analyzer import RustAnalyzer
 
-        cargo_path = os.path.expanduser("~/.cargo/bin/rust-analyzer")
+        # Use platform-appropriate paths and binary names
+        home = pathlib.Path.home()
+        if IS_WINDOWS:
+            binary_name = "rust-analyzer.exe"
+            broken_proxy_path = str(home / "AppData" / "Local" / "Microsoft" / "WindowsApps" / binary_name)
+            cargo_path = str(home / ".cargo" / "bin" / binary_name)
+        else:
+            binary_name = "rust-analyzer"
+            broken_proxy_path = "/usr/bin/rust-analyzer"
+            cargo_path = os.path.expanduser("~/.cargo/bin/rust-analyzer")
 
         def mock_isfile(path):
-            return path in ["/usr/bin/rust-analyzer", cargo_path]
+            return path in [broken_proxy_path, cargo_path]
 
         def mock_access(path, mode):
-            return path in ["/usr/bin/rust-analyzer", cargo_path]
+            return path in [broken_proxy_path, cargo_path]
 
         with patch.object(RustAnalyzer.DependencyProvider, "_get_rust_analyzer_via_rustup", return_value=None):
             with patch.object(RustAnalyzer.DependencyProvider, "_get_rustup_version", return_value=None):
-                with patch("shutil.which", return_value="/usr/bin/rust-analyzer"):
+                with patch("shutil.which", return_value=broken_proxy_path):
                     with patch("os.path.isfile", side_effect=mock_isfile):
                         with patch("os.access", side_effect=mock_access):
                             # The PATH binary is a broken rustup proxy


MischaPanch:
Rust tests now fail on windows in CI, pls have a look.

Previously, we had the comment that using rustup from PATH `(can pick up incorrect aliases, hence checked last)`. Was this assessment incorrect?

NetVar1337:
@MischaPanch Thanks for the review! I've pushed a fix for the Windows test failure.

**Root cause:** The `test_skip_broken_rustup_proxy_in_path` test used hardcoded Unix paths (`/usr/bin/rust-analyzer`) and `os.path.expanduser("~/.cargo/bin/rust-analyzer")` which doesn't include the `.exe` extension on Windows. The production code builds Windows common paths with `rust-analyzer.exe` via `pathlib`, so the mock paths never matched any common location, causing a `RuntimeError` instead of falling through to the cargo path.

**Fix:** The test now uses platform-appropriate paths — on Windows it uses `.exe` extension and `pathlib.Path.home()` to match how the production code constructs its search paths.

**Regarding your question about PATH alias concerns:** The original comment ("can pick up incorrect aliases, hence checked last") was partially correct — PATH *can* pick up broken rustup proxy symlinks. However, the new code addresses this properly by calling `_is_rust_analyzer_functional()` which runs `--version` to verify the binary actually works before accepting it. This means:

- A broken rustup proxy in PATH → detected and skipped (fails `--version` check)
- A legitimate standalone install (Nix, distro package, etc.) → detected and used (passes `--version` check)

So the safety concern is preserved through functional validation rather than positional ordering. Checking PATH earlier (after rustup, before common hardcoded paths) is actually better because it respects the user's environment configuration while still filtering out broken proxies.

Note: The 5 F# test failures and 0-1 Ansible failures on Windows are pre-existing on `main` — not introduced by this PR.

MischaPanch:
Thanks for the contribution and the explanation, merging

PR #1175, opened 2026-03-16
feat(config): support local project config overrides via project.local.yml

arikon:
## Summary
- Add support for `.serena/project.local.yml` that is applied on top of the main `project.yml`, allowing user-specific local configuration overrides
- Add `project.local.yml` to `.serena/.gitignore` so local configs are not committed

### DIFF
--- a/.serena/.gitignore
+++ b/.serena/.gitignore
@@ -1,1 +1,2 @@
 /cache
+/project.local.yml
--- a/src/serena/config/serena_config.py
+++ b/src/serena/config/serena_config.py
@@ -475,6 +475,14 @@ class ProjectConfig(SharedConfig):
         if "project_name" not in yaml_data:
             yaml_data["project_name"] = project_folder_name
 
+        # apply local overrides from project.local.yml if present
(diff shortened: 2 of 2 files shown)


arikon:
@MischaPanch Could you have a look, please?

MischaPanch:
I though we already had that functionality. Pinging @opcode81

opcode81:
@MischaPanch, as you might remember, when I implemented this and it was already fully complete (and working), I lost the entire source code and tried to recover it using the IDE's local history, but this was messy. Apparently, I failed to recover the changes in one of the most critical functions, `ProjectConfig.load`. This is a glaring omission, of course.

Anyway, this PR tries to fix this, but it contains some critical bugs, which I will fix.

opcode81:
I ultimately committed my own variant of the fix directly to main; see commit b659f91.

Issues with this PR:
* Major: Re-saving of an incomplete configuration would save project.yml with local overrides applied.
* Subtle (future): Loading of the local config did not use the class method which accounts for backward compatibility, which would likely have caused a problem down the line.
* Minor: Inappropriate logging
* Minor: Misleading commit message (suggesting feature addition instead of fix)

PR #1183, opened 2026-03-17
Handle OSError when validating registered project paths on startup

diegobonato:
## Summary

- When a registered project path points to a stale or unreachable filesystem (e.g. a disconnected SSHFS mount), `path.exists()` raises `OSError: [Errno 5] Input/output error` instead of returning `False`
- This uncaught exception crashes the entire MCP server on startup, making all Serena tools unavailable
- Wraps `path.exists()` in a `try/except OSError` so inaccessible paths are gracefully skipped with a warning, matching the existing behavior for non-existent paths

## Reproduction

1. Register a project path that points to a stale SSHFS mount (or any filesystem that returns I/O errors)
2. Start the MCP server — crashes with:
   ```
   OSError: [Errno 5] Input/output error: '/path/to/stale/mount'
   ```
   at `serena_config.py` line 829 (`path.exists()`)

## Fix

```python
# Before: crashes on stale mounts
if not path.exists() or ...

# After: gracefully skips inaccessible paths  
try:
    path_exists = path.exists()
except OSError as e:
    log.warning(f"Project path {path} is not accessible ({e}), skipping.")
    continue
if not path_exists or ...
```

## Test plan

- [ ] Verify MCP server starts normally with all valid project paths
- [ ] Verify stale/unreachable mount paths are skipped with a warning instead of crashing
- [ ] Verify existing behavior for non-existent paths is unchanged

### DIFF
--- a/src/serena/config/serena_config.py
+++ b/src/serena/config/serena_config.py
@@ -826,7 +826,12 @@ class SerenaConfig(SharedConfig):
         instance.projects = []
         for path in loaded_commented_yaml["projects"]:
             path = Path(path).resolve()
-            if not path.exists() or (path.is_dir() and not os.path.isfile(instance.get_project_yml_location(str(path)))):
+            try:
+                path_exists = path.exists()
+            except OSError as e:
+                log.warning(f"Project path {path} is not accessible ({e}), skipping.")
+                continue
+            if not path_exists or (path.is_dir() and not os.path.isfile(instance.get_project_yml_location(str(path)))):
                 log.warning(f"Project path {path} does not exist or no associated project configuration file found, skipping.")
                 continue
             if path.is_file():


diegobonato:
The 6 failing tests are pre-existing failures in the Windows CI runner, unrelated to this change:

- **1x Ansible** (`test_hover_on_module_contains_documentation`) — Ansible language server hover returning `None`
- **5x F#** (`test_find_symbol`, `test_get_document_symbols_program`, `test_get_document_symbols_calculator`, `test_nested_module_symbols`, `test_completion`) — F# language server returning empty symbols / missing `System.Array` assembly reference

All failures are in language server integration tests for Ansible and F#. This PR only touches project path validation in `serena_config.py` and does not affect any language server functionality. The core suite (871 passed) is unaffected.

MischaPanch:
This looks reasonable, but I want to clarify the desired/current behavior.

If I recall correctly, we clean up stale paths from the config yaml, right @opcode81 ?
If a path suddenly becomes unreachable, I guess we should also clean it up.

@diegobonato how did you run into this? Did registration of a project path that points to a stale SSHFS not lead to a mistake? If that's the case, this is not good behavior IMO. We should not allow registration of unavailable paths.

opcode81:
> If I recall correctly, we clean up stale paths from the config yaml, right @opcode81 ? If a path suddenly becomes unreachable, I guess we should also clean it up.

The code in this PR achieves that it is removed, since `continue` is used, preventing the project from being added to the list.
 
> @diegobonato how did you run into this? Did registration of a project path that points to a stale SSHFS not lead to a mistake? If that's the case, this is not good behavior IMO. We should not allow registration of unavailable paths.

Since this is a mount, the path was available at some point, then it was no longer, yet `exists()` doesn't return `False` and instead raises an exception. I would regard this as a bug in `exists` to be honest. It makes no sense for this to ever raise an exception.

MischaPanch:
Ok, then we can merge this, right @opcode81 ? I see no need for adjustment

opcode81:
Yes, merging it.

MischaPanch:
Thanks for the contribution @diegobonato !

PR #1188, opened 2026-03-19
Fix: Handle Non-Standard SymbolKind Values from Language Servers

habaneraa:
## Bug Description

When a Language Server returns symbol information with kind values outside the standard LSP specification (1-26), the application would either crash with a ValueError or fail to properly handle these symbols. This issue occurs because some language servers, such as ccls (https://github.com/MaskRay/ccls/blob/d31cc9f07668a91c892d5f13367b9a1e773fbe2b/src/lsp.hh#L114), may return protocol-extended or custom symbol kind values that exceed the standard enum range.

## Call Chain

The bug manifests through the following call chain when processing document symbols:

1. Language Server Response: Returns JSON with "kind": 255 (non-standard value)
2. ls_process.py: Raw JSON parsed via `json.loads(body)` → `{'kind': 255, ...}` . This is received by ls.py: `SolidLanguageServer.request_document_symbols` (e.g. GetSymbolOverview tool will use this)
3. ls.py:1276: `convert_to_unified_symbol()` casts dict to `UnifiedSymbolInformation` without converting kind field
   - Before fix: kind remains as raw int(255) 
   - After fix: kind converted via `SymbolKind.from_int(255)` → `SymbolKind.Unknown`
4. symbol.py:225: symbol_kind property accessed, expecting SymbolKind enum
   - Before fix: `SymbolKind(self.symbol_kind).name` raises ValueError
   - After fix: Receives SymbolKind.Unknown, returns safely

## Fix Summary

1. Added `Unknown = 0` to SymbolKind enum (ls_types.py). Provides a fallback value for non-standard symbol kinds
2. Updated `from_int()` method to return Unknown instead of raising ValueError
3. Added explicit conversion in ls.py:convert_to_unified_symbol()

### DIFF
--- a/src/solidlsp/lsp_protocol_handler/lsp_types.py
+++ b/src/solidlsp/lsp_protocol_handler/lsp_types.py
@@ -10,7 +10,7 @@ This file provides the Python types corresponding to the Typescript types define
 """
 
 from enum import Enum, IntEnum, IntFlag
-from typing import Literal, NotRequired, Union, Any
+from typing import Any, Literal, NotRequired, Union
 
 from typing_extensions import TypedDict
 


src/solidlsp/ls_types.py
```diff
@@ -152,6 +152,7 @@ class SymbolKind(IntEnum):
 
     # TODO: This is a duplicate of SymbolKind in lsp_types.
 
+    Unknown = 0
```
MischaPanch: This is hitting an old technical debt, see todo above. Maybe we should remove the duplication now. The other copy of SymbolKind even already has a from_int method, though it's never used.

src/solidlsp/ls_types.py
```diff
@@ -179,6 +180,18 @@ class SymbolKind(IntEnum):
     Operator = 25
     TypeParameter = 26
 
+    @classmethod
```
MischaPanch: There are many calls to `SymbolKind()` that have not been replaced by `from_int`.

We should remove this and use the inbuilt `_missing_` mechanism as described [here](https://stackoverflow.com/a/55259756), that would be a safe way to address the issue. Should be tested if this still works in python 3.11

src/solidlsp/ls_types.py
MischaPanch: @opcode81 FYI

src/solidlsp/ls_types.py
habaneraa: `_missing_` is cool. This is better than `from_int`. We still need something like "Unknown" for avoiding ValueError.

MischaPanch: yes, Unknown will be the default value

src/solidlsp/ls_types.py
opcode81: Yes, we should remove the duplication now and only keep the one in `lsp_types` as this is where it belongs.
It should be easy:
* Remove class from `ls_types`
* Import class from `lsp_types` in `ls_types` and use sensai `mark_used` to ensure that it remains there.

In this way, we don't need to change anything anywhere else; imports remain valid.

MischaPanch:
@habaneraa I am taking this over to address the old todo. Thanks for the contribution!

src/solidlsp/ls_types.py
opcode81: @MischaPanch inspecting the code, I found that `lsp_types` is auto-generated and therefore should not be edited. So if we want to maintain the ability to auto-generate `lsp_types`, then we actually shouldn't edit it and instead consistently use the (edited) copy in `ls_types` going forward.

Not an ideal situation...

opcode81: Previously, it wasn't edited, so the TODO made sense.

opcode81: Maybe the best option is to ditch "blind" auto-generation and use it only selectively, if at all, going forward (i.e. auto-generate, pick new stuff manually), adapting the text at the top of the file accordingly:
We should know that contents of this were originally auto-generated (and how) but that they were edited afterwards.
wdyt?

MischaPanch: Done

PR #1192, opened 2026-03-20
feat: Support project-level ls_specific_settings

code4lala:
## Summary

  This PR adds support for configuring language server-specific settings (such as clangd path) at the project level through `.serena/project.yml`, previously this configuration was only possible globally via
  `~/.serena/serena_config.yml`.

 ## Background

  Previously, users could only specify custom clangd paths using ls_specific_settings.cpp.ls_path in the global `~/.serena/serena_config.yml` file. This limitation meant that different projects requiring different clangd
  versions or paths could not be configured independently, forcing users to either:
   - Use the same clangd for all projects, or
   - Manually modify the global configuration when switching between projects

##  Changes

  This PR introduces the following changes:

   1. `src/serena/config/serena_config.py`:
      - Added ls_specific_settings field to the ProjectConfig dataclass
      - Updated _from_dict() method to properly load ls_specific_settings from YAML configuration

   2. `src/serena/project.py`:
      - Modified LanguageServerFactory creation logic to merge ls_specific_settings from both global (SerenaConfig) and project (ProjectConfig) configurations
      - Project-level settings take precedence over global settings, allowing per-project overrides

##  Configuration Example

  Users can now configure clangd path in their project's `.serena/project.yml`:

```yml
ls_specific_settings:
  cpp:
    ls_path: /path/to/project-specific/clangd
```

  The configuration hierarchy is:
   - Global `~/.serena/serena_config.yml` (base configuration)
   - Project `.serena/project.yml` (overrides global settings)

##  Testing

  The implementation has been verified by:
   1. Removing `ls_specific_settings.cpp.ls_path` from global `~/.serena/serena_config.yml`
   2. Adding the clangd path configuration to a project's `.serena/project.yml`
   3. Running `serena project index --log-level DEBUG` and confirming:
      - The custom clangd path is correctly loaded from the project configuration
      - Clangd starts successfully using the specified path without attempting to download the default version

 ## Impact

  This change improves the developer experience by:
   - Enabling project-specific language server configurations
   - Allowing different projects to use different clangd versions or paths
   - Reducing the need to modify global configuration when working with multiple projects
   - Making project configurations more self-contained and portable

### DIFF
--- a/src/serena/project.py
+++ b/src/serena/project.py
@@ -624,15 +624,14 @@ class Project(ToStringMixin):
 
             log.info(f"Creating language server manager for {self.project_root}")
             self._language_server_manager_init_error = None
-            # Merge ls_specific_settings from serena_config and project_config, with project_config taking precedence
-            merged_ls_specific_settings = {**self.serena_config.ls_specific_settings, **self.project_config.ls_specific_settings}
+            ls_specific_settings = {**self.serena_config.ls_specific_settings, **self.project_config.ls_specific_settings}
             factory = LanguageServerFactory(
                 project_root=self.project_root,
                 project_data_path=self._serena_data_folder,
                 encoding=self.project_config.encoding,
                 ignored_patterns=self._ignored_patterns,
                 ls_timeout=ls_timeout,
-                ls_specific_settings=merged_ls_specific_settings,
+                ls_specific_settings=ls_specific_settings,
                 trace_lsp_communication=self.serena_config.trace_lsp_communication,
             )
             self.language_server_manager = LanguageServerManager.from_languages(self.project_config.languages, factory)
--- a/docs/02-usage/050_configuration.md
+++ b/docs/02-usage/050_configuration.md
@@ -213,8 +213,10 @@ This ensures backward compatibility: existing projects that already have a `.ser
 Most users will not need to adjust these settings.
 :::
 
-Under the key `ls_specific_settings` in `serena_config.yml`, you can you pass per-language, 
-language server-specific configuration.
+Under the key `ls_specific_settings` in `serena_config.yml`, you can you pass global per-language, 
+language server-specific configuration. You can use the same key in the project configuration files (`project.yml`
+and `project.local.yml` ) to override or extend the global settings for a specific project.
+The settings are merged on top-level, meaning that project-level settings for a language will replace global settings for the same language.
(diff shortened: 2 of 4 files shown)


code4lala:
@opcode81 This PR is ready for review! Thanks in advance for taking a look. 🙏 Please let me know if any updates are needed.

MischaPanch:
Thanks @code4lala . Some minor things are off, I pushed a fix.

@opcode81 could you have a quick look? From my side it's good to merge.

I remember having a PR with a similar change somewhere but I couldn't find it now

src/serena/project.py
```diff
@@ -624,13 +624,17 @@ def create_language_server_manager(self) -> LanguageServerManager:
 
             log.info(f"Creating language server manager for {self.project_root}")
             self._language_server_manager_init_error = None
+            # Merge ls_specific_settings from serena_config and project_config, with project_config taking precedence
+            project_ls_specific_settings = {k: v for k, v in self.project_config.ls_specific_settings.items() if v}
```
opcode81: This is problematic, because it effectively disallows unsetting the global setting, i.e. replacing the global settings with empty settings as in `python: {}`, which, however, is a valid use case.

MischaPanch: I thought about this, but then we introduce a difference between

`python: None`, `python:` and `python: {}` in yaml. We have talked about it in a different setting, where I had proposed to add a special identifier like `use_global` for this exact purpose.

While replacing the global settings with empty settings is valid, it think it is rare and it can always be overcome in some way, for example by `python: {"ls_path": null}`. So I propose to just bite that pill instead of introducing a difference between None and empty. I am not sure how reliable this difference is mirrored in yaml

opcode81: > I thought about this, but then we introduce a difference between
> python: None, python: and python: {} in yaml. 

No, we don't. The universal semantics are: If you set *anything* for a language, it overrides the global setting.
This includes the `None` setting, which you can write as either `python: null` or `python: `.
If you set these, it should override the global setting, but both of these are probably invalid, because a dictionary is required as far as I am aware - similar to how setting `python: foo` is also invalid.

Fundamtentally, it makes no sense for `python: {}` (and invalid null settings) to have no effect.

MischaPanch: Done

PR #1196, opened 2026-03-21
Fix off-by-one in _get_range_from_file_content

Asher-:
## Problem

`_get_range_from_file_content` set `end_line = len(lines)` which points one past the last valid line (LSP lines are 0-indexed). Used in fallback file-symbol creation in `request_referencing_symbols`.

## Fix

Changed to `end_line = len(lines) - 1`.

## Verification

- Format/type-check clean
- Re-audited — zero regressions


Asher-:
Closing this — on closer review, the LSP spec clamps out-of-bounds positions, so both values produce the same result at runtime. Not a real bug. Sorry for the noise.

PR #1197, opened 2026-03-21
Fix copy-paste bugs in language server init params

Asher-:
## Problem

Three copy-paste errors in language server initialization, each verifiable
by comparing against other LS implementations in the same repo:

1. **clangd + ccls** — workspace folder name is the literal string `"$name"`
   instead of the actual project name. Every other LS in the repo uses
   `os.path.basename(repository_absolute_path)`. This is not a template
   variable — Python sends the five characters $, n, a, m, e to
   the language server.

   To verify: open any other LS file (e.g. python, go, ruby) and check the
   workspace folder `name` field.

2. **Erlang** — `rootUri` uses `f"file://{path}"` instead of
   `pathlib.Path(...).as_uri()`. The f-string form doesn't percent-encode
   special characters and uses wrong slashes on Windows. Every other LS uses
   `pathlib.Path.as_uri()`.

3. **F#** — `symbolKind.valueSet` uses `range(1, 26)` which produces values
   1-25. The LSP spec defines SymbolKind values 1-26 (TypeParameter is 26).
   Every other LS in the repo uses `range(1, 27)`.

## Fix

- clangd/ccls: `os.path.basename(repository_absolute_path)`
- Erlang: `pathlib.Path(self.repository_root_path).as_uri()`
- F#: `range(1, 27)`

### DIFF
--- a/src/solidlsp/language_servers/ccls_language_server.py
+++ b/src/solidlsp/language_servers/ccls_language_server.py
@@ -118,7 +118,7 @@ class CCLS(SolidLanguageServer):
             "workspaceFolders": [
                 {
                     "uri": root_uri,
-                    "name": "$name",
+                    "name": os.path.basename(repository_absolute_path),
                 }
             ],
             # ccls supports initializationOptions but none are required for basic functionality
--- a/src/solidlsp/language_servers/clangd_language_server.py
+++ b/src/solidlsp/language_servers/clangd_language_server.py
@@ -228,7 +228,7 @@ class ClangdLanguageServer(SolidLanguageServer):
             "workspaceFolders": [
                 {
                     "uri": root_uri,
-                    "name": "$name",
+                    "name": os.path.basename(repository_absolute_path),
                 }
             ],
         }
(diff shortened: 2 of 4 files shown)


Asher-:
Rebased cleanly against main — removed unrelated sourcekit_lsp changes that were in the original submission.

src/solidlsp/language_servers/ccls_language_server.py
```diff
@@ -118,7 +118,7 @@ def _get_initialize_params(repository_absolute_path: str) -> InitializeParams:
             "workspaceFolders": [
                 {
                     "uri": root_uri,
-                    "name": "$name",
+                    "name": os.path.basename(repository_absolute_path),
```
MischaPanch, line 121: huh, I wonder how this made it past the tests. We do run ccls and clangd tests in CI (and I ran them manually). Thanks!

MischaPanch: I guess it's not such an important thing to pass

PR #1198, opened 2026-03-21
Use shlex.quote for shell argument escaping on POSIX

Asher-:
## Problem

`quote_arg()` wraps arguments containing spaces in double quotes. But
double quotes do not escape shell metacharacters — \$, backtick, ;,
|, &, and others are still interpreted by the shell.

This function is used to quote arguments in language server startup
commands. If any configured path or argument contains shell metacharacters,
they would be interpreted rather than passed literally.

The realistic risk is low (paths come from config files, not user input),
but the fix is trivial and uses the stdlib-standard approach.

## Fix

On POSIX, use `shlex.quote()` which single-quotes arguments, preventing
all shell interpretation. Retain the original Windows behavior since
`cmd.exe` does not interpret single-quoted strings.

### DIFF
--- a/src/solidlsp/util/subprocess_util.py
+++ b/src/solidlsp/util/subprocess_util.py
@@ -1,4 +1,5 @@
 import platform
+import shlex
 import subprocess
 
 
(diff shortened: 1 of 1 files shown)


Asher-:
Rebased cleanly against main — removed unrelated sourcekit_lsp changes that were in the original submission.

MischaPanch:
Thanks, nice catch!

PR #1200, opened 2026-03-21
Remove dead code and fix spurious f in log message

Asher-:
Dead code removed (all verified zero refs in src/ and test/):
- SolidLanguageServer.request_text_document_diagnostics
- SolidLanguageServer.request_signature_help
- SolidLanguageServer.request_container_of_symbol
- LanguageServerCodeEditor._get_code_file_content
- LanguageServerSymbolRetriever.find_by_location
- SERENA_ICON_DIR constant

Cosmetic: fix f-string producing f/Users/... instead of /Users/...

Net: +1/-110 lines. Format/type-check clean.

### DIFF
--- a/src/solidlsp/ls.py
+++ b/src/solidlsp/ls.py
@@ -1011,45 +1011,6 @@ class SolidLanguageServer(ABC):
 
         return ret
 
-    def request_text_document_diagnostics(self, relative_file_path: str) -> list[ls_types.Diagnostic]:
-        """
-        Raise a [textDocument/diagnostic](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_diagnostic) request to the Language Server
-        to find diagnostics for the given file. Wait for the response and return the result.
-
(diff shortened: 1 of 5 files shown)


src/solidlsp/ls.py
```diff
@@ -1906,28 +1836,6 @@ def is_position_in_range(line: int, range_d: ls_types.Range) -> bool:
         else:
             return None
 
-    def request_container_of_symbol(
```
MischaPanch, line 1909: Solidlsp is a library, there's no need to remove unused methods

Asher-:
Closing this — you are right that solidlsp is a library and removing public API methods is wrong regardless of internal call count. Apologies for the noise.

The f-string cosmetic fix is the only valid change here and I will resubmit it separately once I have cleaned up the other PRs.

I am also aware that all the PRs in this batch have unrelated sourcekit_lsp changes leaked in from a shared base branch. I will rebase each one cleanly against upstream main so they are properly isolated and reviewable. Will ping you when that is done.

Issue #1178, opened 2026-03-17
Disambiguate language detection beyond file extensions


MischaPanch:
The issue appeared during the implementation of ansible support. Ansible is using yaml files, so its source file matcher has 100% overlap with the yaml language server. An ansible project is currently detected as yaml.

For now it's not a problem, ansible support is experimental anyway (the LS lacks many important features). Eventually, we might need a possibility to detect the most suitable language not just from the file extension but also from the content structure.

MischaPanch:
@arikon FYI

jcogilvie:
Working on a helm-ls adapter (`mrjosh/helm-ls`) and hit the same overlap you flagged here:  Helm charts are `*.yaml`/`*.yml`/`*.tpl`, so the matcher would collide with `yaml` the same way `ansible` does. Marking Helm experimental and relying on explicit opt-in in `project.yml` with `Language.get_priority()` for tie-breaking solves it in the "user knows what they've configured" case.

But Helm has a convenient structural signal that Ansible lacks: every chart has a `Chart.yaml` at its root. So there's a narrower precursor to the full "content-structure sniffing" idea from this issue that would handle Helm cleanly today:

- New subclass `ChartAwareFilenameMatcher(FilenameMatcher)` in `src/solidlsp/ls_config.py` taking `anchor_filename` (e.g. `"Chart.yaml"`) and `max_walk_depth`.
- `is_relevant_filename(abs_path)` checks the extension pattern, then walks ancestors looking for the anchor file. Returns `False` when given a bare basename (degrades safely for the `determine_programming_language_composition` callsite at `src/serena/util/inspection.py:54`, which already skips experimental languages).
- `lru_cache` on `(dir, anchor)` so traversal over a 10k-file repo costs one stat per unique ancestor dir.

That's ~60 LoC plus unit tests. It doesn't cover Ansible (no equivalent anchor) or the general content-structure case, so this issue stays open as the broader design — my suggestion would just be a building block other DSL-over-YAML LSs could reuse (k8s manifests, compose files, GHA workflows under `.github/workflows/`, etc.).

Two questions before I write it:

1. Is this a direction you're happy to see land as a scoped precursor to this issue, or would you rather wait for the content-structure design?
2. If yes: any preference on where the class should live (`ls_config.py` alongside `FilenameMatcher` vs. a new `ls_matchers.py`)?

Happy to ship Helm with a plain `*.yaml/*.yml/*.tpl` matcher in the meantime if you'd rather this wait.

PR #1205, opened 2026-03-22
fix: allow adding experimental languages from dashboard

### LINKED ISSUE #1177: Bug: Not all language servers can be added via the dashboard
For instance, markdown cannot be added; none of the experimental ones can be added.

zerone0x:
## Summary
- expose experimental language servers in the dashboard add-language list
- add a focused test covering the new selection logic

Fixes #1177

## Testing
- uv run pytest test/serena/test_dashboard.py

### DIFF
--- a/src/serena/dashboard.py
+++ b/src/serena/dashboard.py
@@ -543,7 +543,7 @@ class SerenaDashboardAPI:
         from solidlsp.ls_config import Language
 
         def run() -> ResponseAvailableLanguages:
-            all_languages = [lang.value for lang in Language.iter_all(include_experimental=False)]
+            all_languages = [lang.value for lang in Language.iter_all(include_experimental=True)]
 
             # Filter out already added languages for the active project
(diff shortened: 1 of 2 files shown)


MischaPanch:
Thanks for the PR @zerone0x .

We need to think how we want to handle that. Adding experimental languages means the possibility of collisions on resolving an LS for a file. What happens if a user adds two python LS simultaneously? The restriction in the dashboard had been there for a reason.

In principle, we should prohibit such collisions. We could add a mechanism telling a user to "Pls remove language `<colliding>` before adding `<new>`", but for that we need to establish what collides with what. Also related to #1178 

@opcode81 FYI

opcode81:
We don't necessarily need to prevent collisions. In the worst case, a user will get duplicate search results - and it'll be clear why. In most cases, the first LS that handles a file "wins". 
IDEs do not prevent this either. You can install multiple name providers for the same language - and you'll get multiple results for everything.

PR #1219, opened 2026-03-25
Update Eclipse JDTLS to use 1.53.0

stewardc:
## Summary by CodeRabbit

* **Chores**
  * Updated bundled Java language server runtime environment and development tools to current compatible versions, including the Java Runtime Environment and supporting libraries across all supported platforms.

### DIFF
--- a/src/solidlsp/language_servers/eclipse_jdtls.py
+++ b/src/solidlsp/language_servers/eclipse_jdtls.py
@@ -3,12 +3,12 @@ Provides Java specific instantiation of the LanguageServer class. Contains vario
 """
 
 import dataclasses
+import hashlib
 import logging
 import os
 import pathlib
 import shutil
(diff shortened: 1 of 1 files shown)


MischaPanch:
Thanks @stewardc . We are tightening security around downloads in an ongoing PR, including checksum verifications. IT will collide with these changes, so I'd ping you again after the tightening is merged, and then we can bump JDTLS in the tightened version. Would that be ok?

stewardc:
> Thanks @stewardc . We are tightening security around downloads in an ongoing PR, including checksum verifications. IT will collide with these changes, so I'd ping you again after the tightening is merged, and then we can bump JDTLS in the tightened version. Would that be ok?

Sounds good - I'll keep this one in draft until you're ready. Thank you!

PR #1234, opened 2026-03-27
expand config project path variables in CLI and config loading

roosteer:
Allow project paths and Serena folder templates to resolve shell/environment variables consistently, so configured paths work across custom setups. Update CLI path resolution and related tests to validate variable expansion and configured .serena locations.

## Summary by CodeRabbit

* **New Features**
  * Project paths now expand template/env variables and ~; project creation can target new (non-existent) directories.
  * Dashboard now includes experimental languages when listing available options.
  * YAML comment transfer can be forced to update all keys when syncing config files.

* **Bug Fixes**
  * Improved detection/handling of unresolved environment-variable patterns to avoid misconfigured locations.
  * CLI project lookup retries using resolved paths while preserving the original input in messages.

* **Documentation**
  * Updated inline project and dashboard configuration comments and guidance.

* **Tests**
  * Added/updated tests for template/env expansion, dashboard language behavior, and config-derived paths.

### DIFF
--- a/src/serena/config/serena_config.py
+++ b/src/serena/config/serena_config.py
@@ -900,9 +900,10 @@ class SerenaConfig(SharedConfig):
                 continue
             path_was_config_file = path.is_file()
             if path_was_config_file:
-                path = cls._migrate_out_of_project_config_file(path)
-                if path is None:
+                migrated_project_root = cls._migrate_out_of_project_config_file(path)
+                if migrated_project_root is None:
                     continue
+                path = migrated_project_root
                 num_migrations += 1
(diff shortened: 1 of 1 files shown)


MischaPanch:
@roosteer thanks for wanting to contribute to Serena. This touches core features like the config that need to be done very well at the first implementation as they are hard to adjust after being released. Generally, all features of this kind have to be discussed in an issue first before a design is planned and a PR is created.

I am therefore closing this PR without review. It's not a comment on the quality of the code (I only had a very brief look) or the necessity of the feature. But we need to understand the actual problem you have and are trying to solve first before a solution is proposed.

Pls open an issue instead and clearly describe which features you are missing and why (especially the why).

PR #1237, opened 2026-03-28
fix: replace file-count-based cache saving with time-based during indexing

AndreyPolovodov:
## Problem

Saving cache every 10 files causes extremely slow indexing on large projects because `save_all_caches()` is an expensive operation. On a large project it gets triggered thousands of times during indexing, making the whole process take hours instead of minutes.

## Fix

Switch from a file-count-based trigger (every 10 files) to a time-based trigger (every 30 seconds). This keeps cache saves infrequent enough to not hurt performance, while still preserving progress regularly in case indexing is interrupted.

## Summary by CodeRabbit

* **Chores**
  * Reduced frequency of cache persistence during project indexing: saves are now time-throttled (at most once every 30 seconds) while maintaining a final flush at the end of indexing.

### DIFF
--- a/src/serena/cli.py
+++ b/src/serena/cli.py
@@ -706,7 +706,7 @@ class ProjectCommands(AutoRegisteringGroup):
             collected_exceptions: list[Exception] = []
             files_failed = []
             language_file_counts: dict[Language, int] = collections.defaultdict(lambda: 0)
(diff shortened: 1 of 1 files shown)


MischaPanch:
Sure, makes sense, thanks! Merging

Issue #1232, opened 2026-03-27
Follow up after pinning all LS runtime deps - ensure configured version is honored


MischaPanch:
Currently for many LS we ignore the version if the binary can be found (we only download if we don't find).

We should consider putting the version into the path where we store the binaries. Also consider migration of some sort.

Idea for migration:
For each LS we store the initially used version and we add a postfix to the path only for versions differing for that. Then no migration is necessary (assuming the users didn't override the version in the past - for a few LS it was possible)

PR #1238, opened 2026-03-28
Update jdtls version from `1.42.0-561` to `1.53.0-873`

stewardc:
## Summary

This PR updates the bundled Eclipse JDTLS runtime assets and fixes a reinstallation issue when Serena extracts over an existing read-only file.

## What changed

- bumped the default `vscode_java_version` from `1.42.0-561` to `1.53.0-873`
- updated the pinned SHA256 values for each platform-specific `vscode-java` bundle
- updated bundled JRE paths to `21.0.10`
- updated the Eclipse Equinox launcher reference to `1.7.100.v20251111-0406`
- centralized the default runtime dependency versions into shared constants to avoid repeating version strings across platforms
- kept version overrides configurable via settings, and added support for overriding the launcher version separately
- made ZIP extraction set an existing target file writable before overwriting it, which avoids failures when re-extracting over a previous read-only install

## Why

The JDTLS runtime references were still pinned to an older `vscode-java` release. The old version fails when a project contains many layers of mutually recursive generics (run Serena against https://github.com/stewardc/mutuallyRecursiveGenericsDemo to see the error). Updating the version resolves this issue.

The extraction fix addresses a practical upgrade path issue: if Serena reuses an existing install directory and encounters a read-only file, extraction can fail instead of refreshing the bundle.

## Testing

- added a unit test covering ZIP extraction over an existing read-only file

### DIFF
--- a/src/solidlsp/language_servers/eclipse_jdtls.py
+++ b/src/solidlsp/language_servers/eclipse_jdtls.py
@@ -3,6 +3,7 @@ Provides Java specific instantiation of the LanguageServer class. Contains vario
 """
 
 import dataclasses
+import glob
 import hashlib
 import logging
 import os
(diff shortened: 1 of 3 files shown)


stewardc:
New PR with the SHA checks in place. 

One thing to note, the workspace config.ini seems to pin the old JDTLS version. I can try to force an update, or leave it as-is (maybe add something to the project readme?) Let me know how you'd like to handle this. 

Thanks!

MischaPanch:
Thanks! I think we need to address the follow-up issue  (https://github.com/oraios/serena/issues/1232) first, we shouldn't just overwrite the existing version all the time

stewardc:
> Thanks! I think we need to address the follow-up issue before #1232, we shouldn't just overwrite the existing version all the time

Makes sense - I agree, it felt wrong to just blindly overwrite. If it's not urgent, I can take a look at that issue when I've got some spare time over the next couple of weeks.

MischaPanch:
It's not urgent, but @opcode81 and I need to converge on the most suitable approach before the issue is addressed. Feel free to comment in the issue to participate in the discussion but I'd ask to refer from making a PR until the solution is sufficiently sketched out in the issue

MischaPanch:
JDTLS was now updated

PR #1240, opened 2026-03-29
Add basedpyright language server support

arikon:
## Summary
- Add basedpyright as an alternative Python language server (experimental)
- basedpyright is a fork of pyright with additional features and stricter type checking
- Follows the same integration pattern as the existing pyright server

## Changes
- New `BasedPyrightServer` class in `src/solidlsp/language_servers/basedpyright_server.py`
- `PYTHON_BASEDPYRIGHT` enum value in `Language` (marked as experimental)
- `basedpyright==1.38.4` added as a bundled dependency
- Test repo alias and pytest marker added

## Usage
In `.serena/project.yml`:
```yaml
languages:
- python_basedpyright
```

## Test plan
- [x] Existing Python tests pass (`pytest test -m python` — 53 passed)
- [x] ruff lint and format pass
- [x] mypy type check passes (no new errors)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,6 +16,7 @@ classifiers = [
 dependencies = [
   "requests==2.33.0",
   "pyright==1.1.403",
+  "basedpyright==1.38.4",
   "fortls==3.2.2",
   "overrides==7.7.0",
   "python-dotenv==1.2.1",
@@ -293,6 +294,7 @@ addopts = "--snapshot-patch-pycharm-diff"
 markers = [
   "clojure: language server running for Clojure",
   "python: language server running for Python",
+  "python_basedpyright: language server running for Python (basedpyright)",
   "go: language server running for Go",
   "java: language server running for Java",
   "kotlin: language server running for kotlin",
--- a/src/solidlsp/language_servers/basedpyright_server.py
+++ b/src/solidlsp/language_servers/basedpyright_server.py
@@ -1,0 +1,163 @@
+"""
+Python language server integration using basedpyright (a fork of pyright with additional features).
+
+You can pass the following entries in ``ls_specific_settings["python_basedpyright"]``:
+    - ls_path: Override the executable used to start basedpyright-langserver.
+"""
+
+import logging
+import os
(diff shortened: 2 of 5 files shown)


MischaPanch:
Hi. We already support multiple python LS, including ty. What's the purpose of also supporting basedpyright?

arikon:
@MischaPanch It is better to use with agent the same LSP Server as in IDE — shared configs and so on

MischaPanch:
Ok, that makes sense. Pls adjust the PR to run basedpyright with uvx like I have recently done for `ty`, so it can be installed on the fly. Serena's users are essentially all `uv` users, and this way we can prevent having `basedpyright` as dependency. It's a fairly large package and offers no benefit to most users

opcode81:
> It is better to use with agent the same LSP Server as in IDE — shared configs and so on

Why though? Which problem is being solved? Why does the configuration even matter, and what, effectively, is the difference when using basedpyright instead of pyright as the Python language server with Serena?

TyceHerrman:
@opcode81 it would help ensure agents are type checking the same way you are if you're using basedpyright. 

Also why basedpyright vs pyright - it has new diagnostic rules pyright doesn't and reimplements some pylance features. 

And why basedpyright vs other lsps - slower than others, but best conformance as of today (though other lsps are closing the gap: https://htmlpreview.github.io/?https://github.com/python/typing/blob/main/conformance/results/results.html). 

As one data point, Zed editor has made basedpyright their default lsp and disabled pyright out of the box.

MischaPanch:
Closing since there are too many conflicts. With the extension provided in 1122166b7cfa8, a way to add this could be to add a configurable option for the python languageserver package and bundle all LS (pyright, ty, basedpyright and pyrefly) in a single implementation. I suspect the initialization options for all of them are the same, they all can be installed with uvx. Then one would select this over `ls_specific_settings` rather than through the choice of `language`

Since we recently added support for diagnostics, it even makes some sense that users would want that (previously it made no sense). If anyone wants to pick this up, you could open a PR going in that direction

Issue #1242, opened 2026-03-29
find_symbol fails to match Go methods by bare name (receiver prefix mismatch)


hustLer2k:
`find_symbol("splice")` returns `[]` even though the method exists. The same search with `substring_matching=True` works, and using the full gopls-qualified name `(*Char).splice` also works.

## Root cause

gopls returns Go methods with receiver-qualified names in `textDocument/documentSymbol`:

```
(*Char).splice    ← symbol name from gopls
```

`NamePathMatcher` (in `symbol.py`) does exact string comparison by default:

```python
# symbol.py line 162
if self.name != name_path_component.name:
    return False
```

So `"splice" != "(*Char).splice"` → no match.

## Reproduction

```
find_symbol("splice")                          → []
find_symbol("splice", substring_matching=True)  → found
find_symbol("(*Char).splice")               → found
```

This affects all Go methods. Structs and package-level functions work fine with bare names because gopls doesn't prefix those.

## Suggested fix

For Go, strip the receiver prefix (e.g. `(*Foo).` or `Foo.`) from the symbol name before matching, or treat the part after the last `.` as the method's simple name for matching purposes. This would make `find_symbol("monitorTimeoutBet")` work as users expect.

Alternatively, a per-language or per-project option to default `substring_matching` to `True` would work as a less precise but simpler solution.

## Environment

- Serena v0.1.4
- gopls v0.21.1
- Go 1.26.1

opcode81:
> Serena v0.1.4

Note that this is not a useful version spec, because you should not run Serena based on a released version but from source (e.g. using `uvx`). I'm assuming that you have tested this using the latest version.

Suggested way of proceeding:
1. Add tests which make the issue clear.
2. Implement the fix in class `Gopls` by overriding either `request_document_symbols` or `_request_document_symbols` to post-process the symbols (see docstrings of these methods in base class `SolidLanguageServer` for guidance on how to override/choose correctly).
3. Confirm that tests pass after fixing this.

MischaPanch:
I ran into this in the past, several more advanced tests would fail due to this behavior of gopls (they were never merged). Gopls is also not the only LS to behave this way, so I propose to address this on the level of SolidLanguageServer, where a name-processing method should be added (that does nothing) and that is then used by the gopls implementation.

MischaPanch:
A thorough solution of this issue would entail adding similar tests not only for go but for all language servers, making sure we receive the symbol names in the expected format (bare, no qualifiers)

[...]

PR #1243, opened 2026-03-29
Improve LS symbol name normalisation handling

opcode81:
*  Support name normalisation in `request_document_symbols` via new method

    * Introduce method `_normalize_symbol_name` to support name transformation
    * Apply in `CSharpLanguageServer`, fixing old override of `request_document_symbols`
      not accounting for overload handling (and invalidate high-level cache accordingly)
    * Apply in `ALLanguageServer`

*   Improve symbol typing

    * Clearly differentiate between raw and high-level symbol types
    * Remove the catch-all type `GenericDocumentSymbol`
    * Introduce `RawDocumentSymbol` alias for the raw LS types

### DIFF
--- a/src/solidlsp/language_servers/al_language_server.py
+++ b/src/solidlsp/language_servers/al_language_server.py
@@ -18,7 +18,7 @@ from overrides import override
 
 from solidlsp import ls_types
 from solidlsp.language_servers.common import quote_windows_path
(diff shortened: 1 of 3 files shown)


MischaPanch:
Should we add the tests and fix for go in this PR?

opcode81:
> Should we add the tests and fix for go in this PR?

I will merge this as soon as the tests pass. #1242 can then be addressed separately.

PR #1250, opened 2026-03-30
fix: swapped error message args, dead code after check=True, missing dotnet v9

Bortlesboat:
Three small bug fixes found through code inspection:

**1. `ls_process.py:324` — swapped values in error message**
The message reported "read {expected} of {actual}" instead of "read {actual} of {expected}". `num_bytes` is the target to read and `len(data)` is what was actually read before termination.

**2. `common.py:120` — dead code after `check=True`**
`subprocess.run(..., check=True)` raises `CalledProcessError` on non-zero exit, making the `returncode != 0` branch unreachable. Removed `check=True` so the warning logs execute before raising the exception explicitly. Behavior is preserved — callers still see `CalledProcessError`.

**3. `ls_utils.py:552` — missing dotnet v9 in error message**
The version detection code (lines 538-548) checks for versions 9, 8, 7, 6, and 4, but the error message only listed "4, 6, 7, 8".

### DIFF
--- a/src/solidlsp/language_servers/common.py
+++ b/src/solidlsp/language_servers/common.py
@@ -117,6 +117,7 @@ class RuntimeDependencyCollection:
         completed_process = subprocess.run(
             command,
             shell=True,
+            check=False,
             cwd=cwd,
             stdout=subprocess.PIPE,
             stderr=subprocess.STDOUT,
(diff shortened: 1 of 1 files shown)


opcode81:
Please apply the formatter and amend the changes to your commit.

opcode81:
Thanks for these improvements, @Bortlesboat!

Issue #1255, opened 2026-03-31
Nix flake should use uv


MischaPanch:
The core team doesn't have nix experience.

We recently ran into issues and noticed that our nix flake is not using uv to install dependencies. This is a problem and makes the nix flake unmaintainable for us as it is using a different mechanism from what we are used to and from what is tested in CI.

It would be great to get help from the community to rectify this. It should be completely sure that serena as installed via nix flake has the same environment as via uv

MischaPanch:
@jcdickinson @Eastwdd @MikhaD FYI

MischaPanch:
This might help
https://open.substack.com/pub/pierretdev/p/how-to-create-a-python-uv-project?utm_campaign=post&utm_medium=web

MikhaD:
I will take a look in a bit.

MischaPanch:
Thanks @MikhaD !

jakeefr:
The Nix flake already uses uv2nix on the current main branch — this appears to be resolved. Can this issue be closed?

(I noticed while contributing to #1334 — Gleam language support.)

opcode81:
@jakeefr does the use of uv2nix imply that all Python dependencies will be resolved by uv, i.e. will it be guaranteed that the Python environment uses the right dependencies?

jakeefr:
@opcode81 Yes, uv2nix uses a two-stage model where uv owns the resolution and Nix owns the build.

1. uv lock produces uv.lock with the full dependency graph and cryptographic hashes
2. uv2nix reads uv.lock and translates it into a Nix overlay, it does not re-resolve dependencies

So Python dependency versions are determined entirely by uv's resolver. uv.lock is the single source of truth. Nix just builds what the lock file specifies.

One caveat: uv.lock doesn't capture build-system metadata (PEP 517 deps like setuptools, hatchling, etc.). Packages built from sdist may need manual Nix overrides for build inputs, but runtime dependencies still match the lock file exactly.

Short answer: if uv lock resolves correctly, the Nix environment is guaranteed to have the same Python dependency versions.

opcode81:
@MischaPanch, what was the discrepancy that prompted the creation of this ticket? Do you remember?

opcode81:
@jakeefr thanks for your input. I talked to @MischaPanch and we can consider this resolved.

PR #1258, opened 2026-03-31
Follow uv2nix best practices in nix flake

MikhaD:
Start integrating uv2nix best practices from https://pierretdev.substack.com/p/how-to-create-a-python-uv-project in order to address #1255 . I still need to test and do some more reading. I am not very familiar with uv.


MischaPanch:
No activity here and in fact maybe there was nothing to do here in the first place, closing

PR #1262, opened 2026-04-01
Implement macOS dock handling and window state management in SerenaDa…

### LINKED ISSUE #1221: macOS support for pywebview-based Dashboard
We need a Mac user to test/improve the pywebview-based dashboard implementation in PR #1117.

To test it:
* In your `serena_config.yml`, make sure the dashboard is enabled
* Run the MCP server

Expected behaviour:
1. With `web_dashboard_open_on_launch` enabled, the new dashboard window should be shown upon startup.
2. With the flag disabled, the window should not be shown, but an icon should appear in the system tray. Clicking on the icon should open the window.
3. Dock/tray behaviour: While the window is open, it should be shown in the taskbar/dock. Closing the window should make it disappear from the taskbar/dock (but it can be reopened from the tray).

Most likely, code changes are necessary to get this to fully work as described.

Please issue a pull request to the branch of PR #1117 with the necessary changes and perhaps make some screenshots so we can see how this looks on macOS.
We appreciate your help!

arnaujc91:
Resolves #1221 

The core issue was macOS app lifecycle, not just window show/hide.

Before, the code in [dashboard.py](https://github.com/oraios/serena/blob/main/src/serena/dashboard.py) hid the window and also hid the app from the Dock by switching the app’s activation policy in `_set_dock_icon_visible(False)`. That meant reopening depended on manually restoring the Dock presence later and on macOS sending the right reopen events. The fragile part was that `pywebview` installs its own `NSApplication` delegate during `webview.start()`, so any custom dock-click handling attached too early could be overwritten. In practice, the app could end up hidden with no reliable dock-driven path to bring the window back.

Now it works because the patch changed the model:

- It stops toggling Dock visibility entirely and just hides/shows the window via `_hide_window()` and `_show_window()`.
- It waits until after `webview.start()` to install the macOS handler in `_setup_macos_dock_handler()`, so it wraps pywebview’s final delegate instead of losing to it.
- The new delegate explicitly handles `applicationShouldHandleReopen_hasVisibleWindows_`, which is the Dock-click reopen path macOS uses when the app has no visible windows.
- It also adds `applicationDidBecomeActive_` as a fallback for activation paths that do not fire the reopen callback.
- `forwardingTargetForSelector_` passes everything else through to pywebview’s original delegate, so normal app behavior is preserved.
- `_suppress_next_activate_reopen` prevents the startup-minimized case from immediately reopening itself on the first app activation.

So the short version is: before, the app relied on hiding/showing the entire Dock presence and on delegate hooks that could get replaced; now it keeps the app alive in the normal macOS app model and installs the reopen logic at the one point where pywebview can no longer overwrite it.

One caveat from the review still stands: the new `confirm_close=self.tray` line introduces a separate tray-quit regression, but that’s independent from why the Dock reopen behavior now works.


test/serena/test_dashboard.py
```diff
@@ -53,3 +55,106 @@ def test_available_languages_exclude_project_languages():
     assert Language.MARKDOWN.value not in available
     # ensure experimental languages remain available for selection
     assert Language.ANSIBLE.value in available
+
+
+def test_dashboard_viewer_initializes_macos_delegate_slot():
```
MischaPanch: let's remove all these new tests. Testing private attr properties is useless, and I don't have confidence that the monkeypatching part is testing anything useful either

arnaujc91: Done

arnaujc91:
Small thing, right now when you hover on the serena icon on the dock it says `python3.12` instead of `serena`. Maybe we could fix that on the PR.

MischaPanch:
> Small thing, right now when you hover on the serena icon on the dock it says python3.12 instead of serena. Maybe we could fix that on the PR.

Yeah, would be nice, would you be up for addressing it?

arnaujc91:
> > Small thing, right now when you hover on the serena icon on the dock it says python3.12 instead of serena. Maybe we could fix that on the PR.
> 
> Yeah, would be nice, would you be up for addressing it?

As agreed there is no easy fix, so we leave it as is

opcode81:
This is a bug in pystray. This line already sets the title, and it works on Windows:

```python
self._tray_icon = pystray.Icon("dashboard_viewer", icon_img, self.title, menu, **kwargs)
```

opcode81:
I have tested this branch on Windows (since the changes affect not only macOS), and everything still works - so that's good!

opcode81:
I think we should merge this asap, as users are already annoyed by this. #1263
Any reason not to merge?

PR #1265, opened 2026-04-01
Fix: dock icon disappears at closing serena window and reappears after clicking open in the tray

arnaujc91:
This PR updates the macOS dashboard viewer behavior to match the requested tray/Dock workflow.

What changed:

- When web_dashboard_open_on_launch is enabled, the dashboard window opens automatically on startup.
- When it is disabled, Serena starts without showing the window, but remains available from the tray/status-bar icon.
- While the dashboard window is open, Serena appears in the Dock.
- Closing the window hides it and removes Serena from the Dock instead of quitting the app.
- The window can be reopened from the tray Open action, and when reopened it appears in the Dock again.

Implementation details:

- Centralized macOS show/hide behavior in the dashboard viewer.
- Tied Dock visibility to the actual window visibility by switching the macOS activation policy between regular and accessory mode.
- Ensured macOS window restore/show logic runs on the AppKit thread so tray reopen works reliably.
- Reapplied the app icon when restoring the Dock entry after reopening.
- This brings the dashboard viewer behavior in line with the owners’ expected macOS behavior for startup, tray access, and Dock visibility.


arnaujc91:
Regarding the multiple icons appearing for multiple processes in MacOS. Findings are:

### Can we get “at most one tray icon” across multiple Serena processes?
**Not with a small tweak.**

Why:

- macOS does not merge two separate NSStatusItem/pystray items from two separate processes into one.
- Right now, every Serena process starts its own viewer process, and every viewer process creates its own tray icon.
- So if you want only one tray icon, you need some form of cross-process coordination.


### Realistic options:

---

#### Enforce a single tray owner. Simplest option.

Behavior:

- first Serena process creates the tray/viewer
- later Serena processes detect it and do not create another tray icon

How:

- file lock, socket lock, or PID lock
- secondary processes skip _start_dashboard_viewer

Tradeoff:

- only one process would own the tray UI
- not obvious how users would open the dashboard for the “other” Serena instances

---

#### Build a shared tray manager. Most correct, but much bigger change.

Behavior:

- one long-lived tray helper process
- all Serena instances register themselves with it

tray menu could show:

- Open dashboard for server A
- Open dashboard for server B

Tradeoff:

- needs IPC and lifecycle management
- definitely not a small patch

---

#### Enforce a single Serena instance. Most restrictive.

Behavior:

- starting a second Serena process either fails or reuses the first one

Tradeoff:

- changes product behavior significantly
- probably not what you want unless the project explicitly wants singleton Serena

---

My recommendation:

If you really want at most one tray icon, the cleanest path is:

Do not try to “merge” icons. Instead, make only one process allowed to create the tray/viewer.

That is feasible, but it’s a product decision, because then we must define:

- what should happen for the second server process?
- should it have no tray at all?
- should it reuse the first viewer?
- if so, how does the viewer know which dashboard URL to open?

So my conclusion is:

I do not think we can get one tray icon for several simultaneous Serena processes with an easy local patch.
We can do it with a singleton tray/viewer design, but that is a real architectural change, not a quick fix.

PR #1266, opened 2026-04-01
Add centralised Serena dashboard tray manager

opcode81:
Add class `SerenaDashboardTrayManager`:
A single tray app to manage all Serena instances (detached process which is informed about running instances via network requests)

Tray icon functionality was consequently removed from `SerenaDashboardViewer`, which is now a plain viewer.

### DIFF
--- a/src/serena/agent.py
+++ b/src/serena/agent.py
@@ -2,17 +2,16 @@
 The Serena Model Context Protocol (MCP) Server
 """
 
-import multiprocessing
 import os
 import platform
 import subprocess
 import sys
 from collections.abc import Callable, Iterator, Sequence
 from contextlib import contextmanager
+from datetime import datetime
 from logging import Logger
 from typing import TYPE_CHECKING, Optional, TypeVar
 
-import webview
 from sensai.util import logging
 from sensai.util.logging import LogTime
 from sensai.util.string import dict_string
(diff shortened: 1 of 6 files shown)


opcode81:
Current state/points worth mentioning:
1. This currently delays agent startup by about two seconds on Windows. This is due to port probing being notoriously slow.
2. Opening a dashboard window via the tray icon is more sluggish than before, because we need to dynamically spawn a new process on demand. On macOS, this is even due to a technical limitation, because (as you probably know) UIs must live on the process' main thread, so we cannot prepare the dashboard windows in the tray process.
3. Tray app termination is still flawed; doesn't always close reliably (even though context menu shows no instances).

MischaPanch:
Issues:

4. Left click on the tray icon does nothing
5. Right click shows the list of sessions but doesn't mention the active project. If a project is active, it would be beneficial to display in the list instead of the generic name. Otherwise, we could display "No active project"
6. The text in the list is pixelated

image

opcode81:
(4): Spec was: Left click should (a) open the instance if there is only one or (b) show the menu if there is more than one - but apparently, (b) could not be satisfied (yet).
(5): It does. You have no project activated in both instances. This works for me.
(6): There is nothing we can do about this.

MischaPanch:
7. There is a  delay of almost one second before the click on the tray icon opens the dashboard. This makes it easy to want to click multiple times. As a result, multiple dashboard instances are shown (for the same session).

We should either:
1. never show more than one dashboard per session at the same time (not sure if technically possible)
2. disable the special behavior for the single-session case

2 is an improvement in UX since after right-click, select, left-click the selection menu disappears, and the dashboard will open before the user will reopen the selection menu. With the special handling of the single-session case, it's just too easy to click on the tray icon multiple times in a row

MischaPanch:
> (5): It does. You have no project activated in both instances. This works for me.

Confirmed, sorry for the noise. Though I'd prefer a different string than "Serena Agent" shown in the no project case

opcode81:
(7) duplicates (2). I have already explained why this is not really easy to avoid. Though we could handle it differently (much faster) on Windows ...

opcode81:
Superseded by #1380

PR #1294, opened 2026-04-07
Add Crystal language support

jochenseeber:
* Adds [Crystalline](https://github.com/elbywan/crystalline), the [Crystal](https://crystal-lang.org) LSP server

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 Status of the `main` branch. Changes prior to the next official version change will appear here.
 
 * General:
-  - Added `serena init` command  
+  - Added `serena init` command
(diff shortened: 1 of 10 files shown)


MischaPanch:
Hi, thanks for the contribution. I will merge this, the very limited functionality of the crystal LS, the small community and the fact that we don't install or test this in CI means that the support in Serena is entirely untested and in a pre-alpha state (so is the crystalline LS itself, from what I gathered).

I think this is sufficiently explained in the note you added though and in the readme of the crystalline project itself. So I am ok with merging this untested state, trusting that you have actually tested it yourself :).

If not, pls do make sure that the tests run through. Finding definition working only once per LS session is really a bit crazy ^^

PR #1299, opened 2026-04-08
JetBrainsFindReferencingSymbolsTool: Add computation of context lines

opcode81:
... based on line number now returned by plugin


MischaPanch (approved):
Nice, even backwards compatible!

You can consider using the progressive shortening from the LS tool, but it's really optional

MischaPanch:
@opcode81 SymbolDTOKey should be updated (see failing test)

PR #1300, opened 2026-04-08
Feature/haxe language support

Neverbirth:
## Summary

  - Add Haxe language support via the [vshaxe/haxe-language-server](https://github.com/vshaxe/haxe-language-server)
  - Supports `.hx` files with go-to-definition, find references, document symbols, hover, and completion
  - Requires Node.js and the Haxe compiler (3.4.0+), only production tested with latest stable 4.3.x version though.

  ## Language server resolution

  The `server.js` binary is resolved in this order:

  1. **User-provided `ls_path`** in `ls_specific_settings`
  2. **Previously built copy** in Serena's resources directory
  3. **VS Code extension** — discovers `nadako.vshaxe-*` or `vshaxe.haxe-*` from `~/.vscode/extensions/`
  4. **Build from source** — clones the repo and builds via `npm ci` + `npx lix run vshaxe-build`

  ## Configurable settings (`ls_specific_settings["haxe"]`)

  | Setting | Purpose | Default |
  |---|---|---|
  | `ls_path` | Path to a pre-built `server.js` | auto-resolved |
  | `build_file` | `.hxml` build file for display arguments | auto-detected from project root |
  | `haxe_executable` | Path to the Haxe compiler | `haxe` from PATH |
  | `working_directory` | Subdirectory for Haxe CWD ("monorepo" support, unfortunately the Haxe LS doesn't support workspaces, in my local version I have some extra haxe language defined to support multiple projects...) | project root |

  ## Implementation details

  - **AI coded**: I hope everything is in acceptable quality and state.
  - **Dynamic capability registration**: The Haxe LS registers features (references, definition, completion) dynamically
   via `client/registerCapability` after initialization. Startup waits for this.
  - **No `didOpen` notifications**: Sending `didOpen` triggers a recompilation that causes subsequent requests to return
   empty results. File opening is suppressed; the LS uses its own file system access.
  - **Compiler request queuing**: References requests sent during the initial compilation are queued by the LS and
  answered correctly once compilation finishes. Requests sent after compilation return empty results. The server
  deliberately does not block on compilation.
  - **External classpaths**: Parses `-cp` entries from `.hxml` files (including recursive includes) to scan source
  directories outside the repo root, enabling cross-project symbol discovery.
  - **`workspace/didChangeConfiguration`**: Sent after `initialized` to trigger the display server (compiler) startup,
  without which no features work.


Neverbirth:
Hm, I'm surprised about the failing tests, in theory they were run and passing before creating the PR. I will fix them.

Neverbirth:
I will keep looking into it tomorrow

Neverbirth:
I hope the test approach taken is ok for you, but let me know.

MischaPanch:
@Neverbirth thanks for the PR. The tests have to run through in CI for us to merge this, they should not be skipped. We have node in the CI runners. What was the problem there?

The skip condition should only be applied when not in ci (we have an `is_ci` variable in conftest that you can use for that).

I haven't reviewed yet and will do so once tests pass in CI

Neverbirth:
No problem at all. I read the reply to this PR https://github.com/oraios/serena/pull/1294 and thought it could be ok. It fails to run because the tests need to download Haxe itself and the LS support for it. I will then look into it.

MischaPanch:
@Neverbirth my comment on #1294 was because the Crystal LS that was committed there was non-functioning to such a large degree that testing it would just not make sense. I don't really see why one would benefit from having such limited Crystal support in Serena, but since the PR's author wanted to have it, it didn't hurt to merge it either.

I hope the haxe LS actually works and is not in such a sorry state as the Crystal LS. If it does work, it should be tested. Pls check if it can be installed on the fly, like we do with other language servers

Neverbirth:
Yeah, it does work, it’s been around for some time, tests pass locally on my machine and I have been testing the Serena support in two medium-big sized projects, although I didn’t have the opportunity to test on long running and complex tasks. 

But I will look into having the tests passing on the ci.

MischaPanch:
Great, happy to hear that, then it should be possible to make it work in CI as well :)

Neverbirth:
Finally green. It was a bit painful heh.

src/solidlsp/language_servers/haxe_language_server.py
```diff
@@ -0,0 +1,690 @@
...
+            "export",
+            "bin",
+            "dump",
+            "node_modules",
+        ]
+
+    class DependencyProvider(LanguageServerDependencyProviderSinglePath):
+        # URL for the vshaxe VS Code extension package (contains pre-built server.js)
+        _VSHAXE_VSIX_URL = (
+            "https://marketplace.visualstudio.com/_apis/public/gallery/publishers/nadako/vsextensions/vshaxe/latest/vspackage"
```
MischaPanch, line 107: pls use a fixed default version and allow overriding the version via ls_specific_settings, like in other LS implementations

src/solidlsp/language_servers/haxe_language_server.py
```diff
@@ -0,0 +1,690 @@
...
+            return None
+
+        @classmethod
+        def _download_from_marketplace(cls, haxe_ls_dir: str, server_js_path: str) -> str:
+            """Download pre-built server.js from the VS Code marketplace.
+
+            The vshaxe extension (.vsix) is a gzip-compressed zip archive containing
+            the pre-built language server at extension/server/bin/server.js.
+            """
+            import gzip
```
MischaPanch, line 163: use global imports

.github/workflows/pytest.yml
```diff
@@ -44,6 +44,10 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: '20.x'
+      - name: Set up Haxe
```
MischaPanch, line 47: should not be necessary, right? The LS is downloaded on the fly

src/solidlsp/language_servers/haxe_language_server.py
```diff
@@ -0,0 +1,690 @@
...
+            is_subdir = any(pathlib.Path(path).is_relative_to(other) for other in external if other != path)
+            if not is_subdir:
+                filtered.append(path)
+
+        if filtered:
+            log.info("External Haxe classpaths detected: %s", filtered)
+        return filtered
+
+    @override
+    def request_full_symbol_tree(self, within_relative_path: str | None = None) -> list["ls_types.UnifiedSymbolInformation"]:
```
MischaPanch, line 310: this violates some of Serena's paradigms and @opcode81 and I need to have a look at that before this is merged. It seems very weird to me that one needs to analyze code outside the repository.

We have many high-priority topics now and a proper review of this will have to wait a bit

MischaPanch (changes requested):
>It was a bit painful heh.

It usually is ^^

Thanks for the PR! I did a quick initial review, the rest will have to wait for the reasons explained in the comment.

src/solidlsp/language_servers/haxe_language_server.py
Neverbirth: I did that because my biggest project has a structure a la:
/projectA/src/
/projectB/src/
/projectC/src/
/sharedLibA/src/
/sharedLibB/src/

And the files in the sharedLibs were not being indexed by Serena while working on the different projects, and if I needed to do some work on them the request was falling back to the AI model fully.

PR #1302, opened 2026-04-09
fix: avoid ignoring root path in gitignore checks

### LINKED ISSUE #1054: find_symbol with empty relative_path returns [] — root dir treated as hidden
## Bug

`find_symbol` (which calls `request_full_symbol_tree`) returns empty results when `relative_path` is not provided (or is empty string).

## Root Cause

When `relative_path=""`, it resolves to `"."`. The function `is_ignored_dirname()` in solidlsp returns `True` for any dirname starting with `.` — including the literal `"."` (current directory). This causes the entire root directory to be skipped as "hidden".

## Reproduction

```python
# In a Swift project with symbols in Sources/
find_symbol(name_path_pattern="MyClass")  # returns []
find_symbol(name_path_pattern="MyClass", relative_path="Sources")  # returns correct results
```

## Expected Behavior

`find_symbol` without `relative_path` should search the entire project (respecting `.gitignore` and `ignored_paths` from `project.yml`).

## Workaround

Always pass `relative_path="Sources"` (or a more specific path) to scope the search.

## Environment

- Serena: latest from `git+https://github.com/oraios/serena` via uvx
- Language: Swift (sourcekit-lsp)
- Swift version: 6.2.4
- Context: claude-code

zerone0x:
Fixes #1054

Prevent pathspec matching from treating the project root (".") as an ignored path, which can cause find_symbol to skip the entire tree when .gitignore contains '/.*/'.

Tests: not run (missing local deps: sensai).

### DIFF
--- a/src/serena/util/file_system.py
+++ b/src/serena/util/file_system.py
@@ -341,6 +341,9 @@ def match_path(relative_path: str, path_spec: PathSpec, root_path: str = "") -> 
     :param root_path: the root path from which the relative path is derived
     :return:
     """
+    if str(relative_path) in {"", "."}:
+        return False
+
     normalized_path = str(relative_path).replace(os.path.sep, "/")
 
     # We can have patterns like /src/..., which would only match corresponding paths from the repo root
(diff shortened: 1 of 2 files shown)


MischaPanch:
Thanks, it's a very niche case, but it can't hurt to include this special handling. Merging

> Tests: not run (missing local deps: sensai).

What was the problem with that? Did you not create the env with `uv sync`?

PR #1304, opened 2026-04-09
feat: add OpenTofu language support

Sensoramo:
## Summary

  Adds first-class `opentofu` language support to Serena by reusing the existing
  Terraform language server integration.

  ## Changes

  - add `Language.OPENTOFU`
  - map OpenTofu to the existing `TerraformLS`
  - allow CLI detection via `tofu`, `terraform`, `TOFU_CLI_PATH`, and
  `TERRAFORM_CLI_PATH`
  - add OpenTofu language tests
  - exclude `.tfvars` and `.tfstate` from symbol indexing to avoid `terraform-ls`
  documentSymbol failures
  - update docs, changelog, pytest markers, and project template

  ## Verification

  - `uv run --extra dev python -m pytest test/solidlsp/terraform/
  test_terraform_basic.py test/solidlsp/opentofu/test_opentofu_basic.py -q`
  - `uv run --extra dev poe lint`
  - `uv run --extra dev poe type-check`

  ## Notes

  OpenTofu support is implemented as a low-risk extension of the existing Terraform
  integration rather than a separate language server implementation.


test/solidlsp/opentofu/test_opentofu_basic.py
```diff
@@ -0,0 +1,40 @@
...
+
+from solidlsp.ls import SolidLanguageServer
+from solidlsp.ls_config import Language
+
+
+@pytest.mark.opentofu
+class TestLanguageServerBasics:
+    """Test basic functionality of the OpenTofu language server configuration."""
+
+    @pytest.mark.parametrize("language_server", [Language.OPENTOFU], indirect=True)
```
MischaPanch, line 11: I think these tests can be safely deleted, right? The language server is the same for opentofu and terraform, so you are just repeating the terraform tests here.

A better approach would have been to add some opentofu specific things to existing test resources and tests, if this exists

PR #1305, opened 2026-04-09
Fix HLSL language server crash on CRLF files pulled in via #include

TheSmallKiwi:
## Summary

Fixes a 100%-reproducible crash in the HLSL language server on Windows when a shader file is first pulled into `shader-language-server`'s cache via an `#include` and then later opened directly. The LS process dies mid-request with `LanguageServerTerminatedException`, which in turn fails tools like `find_symbol` that traverse the workspace.

## Root cause

Upstream bug in [antaalt/shader-sense](https://github.com/antaalt/shader-sense) v1.3.0 at `shader-language-server/src/server/server_file_cache.rs` in `watch_main_file`:

```rust
cached_file.is_main_file = true;
debug_assert!(cached_file.shader_module.content == *text,
              "Server deps content different from client provided one.");
cached_file.shader_module.content = text.into();
```

When a `didOpen` arrives for a file that's already cached as an include-dependency, the server assigns the client-provided text to the module's `content` **without re-parsing the tree-sitter tree**. The `debug_assert!` encodes the maintainer's assumption that the two reads are byte-identical. In release builds it's silently violated.

The violation occurs when:

- The dependency-path read goes through `read_string_lossy` → `std::fs::read_to_string`, which preserves CRLF verbatim.
- The main-file-path text comes from Serena's `FileUtils.read_file`, which opens in Python text mode and performs universal-newlines translation (CRLF → LF), shrinking the content by 1 byte per line.

After assignment, `shader_module.content` is the shorter LF buffer but `shader_module.tree` still references byte offsets into the longer CRLF buffer. The next `query_file_symbols` walks the stale tree and calls `get_name` at `shader-sense/src/symbols/symbol_parser.rs:25`, which does `&shader_content[range.start_byte..range.end_byte]` → out-of-bounds panic → process exit.

### Arithmetic confirmation

In the failure I observed with `TerrainSDF.hlsl` (543 lines, CRLF):

| | |
|---|---|
| On-disk size (CRLF) | 19,390 bytes |
| After CRLF → LF normalization | 19,390 − 543 = **18,847** bytes |
| Panic index reported by the server | **18,849** (~2 past the new EOF) |

## Fix

Override `open_file` in `HlslLanguageServer` to preload the `LSPFileBuffer` with a CRLF-preserving read (`open(path, encoding=..., newline="")`) before `ensure_open_in_ls` sends the `didOpen` notification. The bytes Serena sends now match what the server reads from disk, so `watch_main_file`'s content assignment becomes a no-op and the stale-tree panic is avoided entirely.

The override mirrors `SolidLanguageServer.open_file` verbatim except that in the buffer-creation branch, construction is deferred (`open_in_ls=False`) so contents can be preloaded before the LSP notification fires.

**Scope:** the change is deliberately confined to `src/solidlsp/language_servers/hlsl_language_server.py`. No core files are modified, and no other language server is affected.

### Why not fix it upstream?

The root-cause fix belongs in shader-sense (`watch_main_file` should call `update_module` rather than assigning `.content` raw, or should bounds-check in `get_name` to degrade gracefully). I'll file that separately. This PR is the client-side workaround that unblocks Windows users today.

## Reproduction

Minimal 2-file repro:

1. `textDocument/didOpen` for a `.shader` file that `#include`s a `.hlsl` with CRLF line endings → returns OK.
2. `textDocument/didOpen` for that `.hlsl` directly → LS panics with `byte index N is out of bounds` in `symbol_parser.rs:25`.

## Verification

Tested against a Unity HDRP workspace with 28 shader files:

**Before:** crash on file 25 of the alphabetical traversal (`TerrainSDF.hlsl`, pulled in earlier by `TerrainRayMarch.shader`). Error: `LanguageServerTerminatedException: Language server stdout read process terminated unexpectedly`.

**After:** all 28 files return symbols cleanly. `TerrainRayMarch.shader` → 138 symbols, `TerrainSDF.hlsl` → 174 symbols, full sweep passes.

## Test plan

- [x] Minimal 2-file repro (parent `.shader` → child `.hlsl`) passes
- [x] Full 28-file shader sweep passes in the workspace where the crash was reproducible
- [x] `uv run poe type-check` clean (mypy: no issues in 119 source files)

### DIFF
--- a/src/solidlsp/language_servers/hlsl_language_server.py
+++ b/src/solidlsp/language_servers/hlsl_language_server.py
@@ -7,13 +7,22 @@ import logging
 import os
 import pathlib
 import shutil
+from collections.abc import Iterator
+from contextlib import contextmanager
+from pathlib import Path
 from typing import Any, cast
 
 import psutil
 from overrides import override
 
-from solidlsp.ls import LanguageServerDependencyProvider, LanguageServerDependencyProviderSinglePath, SolidLanguageServer
+from solidlsp.ls import (
+    LanguageServerDependencyProvider,
+    LanguageServerDependencyProviderSinglePath,
+    LSPFileBuffer,
+    SolidLanguageServer,
+)
 from solidlsp.ls_config import LanguageServerConfig
+from solidlsp.ls_exceptions import SolidLSPException
 from solidlsp.lsp_protocol_handler.lsp_types import InitializeParams
 from solidlsp.settings import SolidLSPSettings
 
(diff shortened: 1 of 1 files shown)


MischaPanch:
Thanks for the PR. I stared at the long explanations for a long time, it somewhat makes sense to me, though it is phrased in a very complicated way. The solution here is an unelegant hack but it's contained and is documented as such.

We don't have the capacity for verifying whether this is really necessary, which would require repeating the complex root cause analysis, so I'm going to trust that @TheSmallKiwi correctly identified the issue and am merging this. I did verify that the code doesn't introduce anything dangerous or obviously wrong :)

TheSmallKiwi:
> Thanks for the PR. I stared at the long explanations for a long time, it somewhat makes sense to me, though it is phrased in a very complicated way. The solution here is an unelegant hack but it's contained and is documented as such.
> 
> We don't have the capacity for verifying whether this is really necessary, which would require repeating the complex root cause analysis, so I'm going to trust that @TheSmallKiwi correctly identified the issue and am merging this. I did verify that the code doesn't introduce anything dangerous or obviously wrong :)

LOL it took us a while to track down what was going on, I should have edited the explanation down. Hopefully a temporary fix, I submitted a PR to shader-sense that fixes the root cause.

MischaPanch:
I imagine it was very hard to find the cause! Pls don't read the "inelegant hack" comment as a critique of your contribution, when a language server misbehaves in such a way, hacks are the only resort. I just wrote down the comment as a reference for future is if we have to wonder why this was merged :)

antaalt:
Hey, FYI, he last version of shader-sense 1.3.1 is released and should fix the root cause of the issue thanks to @TheSmallKiwi's PR !

PR #1318, opened 2026-04-11
Allow disabling usage reporting via SERENA_USAGE_REPORTING env var

jochenseeber:
## Summary

- Add support for `SERENA_USAGE_REPORTING=false` environment variable to skip usage reporting in `_send_usage_info`, alongside the existing CI detection


opcode81:
Thanks. The addition of the variable is a good way to handle this. I added the necessary documentation.

PR #1319, opened 2026-04-11
HLSL: bump shader-language-server to v1.3.1, auto-resolve latest on install

TheSmallKiwi:
## Summary

- **Removes the client-side CRLF workaround** (`open_file` override + `_preload_crlf_content`) added in #1305. The upstream root-cause fix (antaalt/shader-sense#44) was merged and released as v1.3.1, so the workaround is no longer needed.
- **Bumps `_DEFAULT_VERSION` to `1.3.1`** and updates SHA-256 checksums for all three pre-built archives (`win-x64`, `linux-x64`, `win-arm64`).
- **Auto-resolves the latest release on first install**: adds `_resolve_version()` which queries the GitHub releases API (`/releases/latest`) when no version is pinned by the user. Falls back to the hardcoded `_DEFAULT_VERSION` (with SHA-256 verification) when the network is unavailable. User-pinned `version:` settings pass through unchanged.

## Behaviour

| Scenario | Version used | SHA-256 checked |
|---|---|---|
| User sets `version: "1.2.2"` | `1.2.2` | No |
| No config, GitHub reachable | latest from API | No (unknown hash) |
| No config, GitHub unreachable | `1.3.1` (fallback) | Yes |

## Test plan

- [x] All 14 HLSL tests pass (`uv run poe test -m hlsl`), including `test_include_file_document_symbols_directly` which exercises the original CRLF crash scenario
- [x] `uv run poe type-check` — one pre-existing error in `agent.py` unrelated to this change

### DIFF
--- a/docs/02-usage/050_configuration.md
+++ b/docs/02-usage/050_configuration.md
@@ -472,7 +472,7 @@ Supported settings:
 | Setting | Default | Description |
 |---|---|---|
 | `ls_path` | managed install or build | Override the `shader-language-server` executable path. |
-| `version` | `1.3.0` | Override the bundled version Serena downloads, or builds from source on macOS, when `ls_path` is not set. |
+| `version` | `1.3.1` | Override the bundled version Serena downloads, or builds from source on macOS, when `ls_path` is not set. |
 
 
 #### Haxe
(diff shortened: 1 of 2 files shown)


MischaPanch:
Thanks, that was a very quick upstream fix! I don't think the default behavior should be pulling latest though, can we instead pull the release where the issue was fixed by default?

TheSmallKiwi:
Good call — I've reverted the auto-latest logic. The PR now simply bumps to v1.3.1 (the release that contains the upstream fix) with updated SHA-256s and removes the client-side workaround. No dynamic version resolution.

PR #1326, opened 2026-04-13
Followup after 34e1fae7 - added more ignored directories

MischaPanch:


MischaPanch:
tests run through but doc-build fails for some reason, can't recreate locally. I think it's just some weird CI issue and it will run through on main, merging

PR #1331, opened 2026-04-13
Add cursor-based code navigation via LSP graph edges

Asher-:
## Motivation

Serena's existing symbol tools (`find_symbol`, `find_referencing_symbols`, `get_symbol_body`) support point lookups — you ask about a specific symbol and get an answer. But an agent exploring an unfamiliar codebase needs something more: a **map**. It needs to land on a symbol, see what's around it, pick a direction, move there, and keep going — building up a mental model of the code's structure as it navigates, rather than teleporting between disconnected lookups.

Today this requires the agent to repeatedly invoke `find_symbol` + `find_referencing_symbols` in separate tool calls, manually tracking what it's already visited and which direction to go next. The cursor changes this by making the code graph *navigable*: the agent places a cursor on a symbol and sees its neighborhood — what it contains, what references it, what it calls, what inherits from it. Each `cursor_move` shifts the viewport to an adjacent symbol, and the trail records everywhere the cursor has been. The agent uses Serena's code graph the way you'd use a map: orient, explore a direction, backtrack, try another path.

The design uses **pure LSP queries** (no persistent index, no external dependencies) so it works for all Serena users out of the box. Every neighborhood resolution is a live LSP request, so the cursor always reflects current code state.

## Summary

- **CursorManager engine** (`src/serena/cursor.py`): Stateful cursors with trail tracking, 7 configurable edge types (contains, references, referenced-by, calls, called-by, inherits, inherited-by), multi-cursor support
- **6 MCP tools** (`src/serena/tools/cursor_tools.py`): `cursor_start`, `cursor_move`, `cursor_look`, `cursor_configure`, `cursor_history`, `cursor_close` — **default-enabled** (users can disable via `excluded_tools`)
- **4 new SolidLanguageServer methods** (`src/solidlsp/ls.py`): `request_call_hierarchy_incoming`, `request_call_hierarchy_outgoing`, `request_type_hierarchy_supertypes`, `request_type_hierarchy_subtypes`
- **sourcekit-lsp fixes**: Resolve binary via `xcrun` with `DEVELOPER_DIR`, add `--scratch-path` for `find_referencing_symbols`
- **Documentation**: `docs/02-usage/042_cursor_navigation.md` — concepts, all 6 tools with example output, usage patterns

## Audit-driven hardening

A 6-lens code audit (error propagation, exception safety, state machine, type system, test architecture, resource lifecycle) was run against the implementation. Key fixes applied:

- Narrowed `except Exception` to `except SolidLSPException` in `resolve_neighbors` so programming errors (KeyError, AttributeError from malformed LSP responses) surface instead of being silently swallowed
- Added summary warning log when edge types fail, so partial failures are visible at default log level (previously silent at DEBUG)
- Added duplicate cursor ID guard in `start_cursor` — raises `ValueError` instead of silently overwriting
- Fixed test isolation: autouse cleanup fixture closes cursors between tests to prevent order-dependent state accumulation

## Test plan

- [x] 39 unit tests — CursorManager with mocked LSP (`test/serena/test_cursor.py`)
- [x] 14 integration tests — CursorManager with live Python LSP (`test/serena/test_cursor_integration.py`)
- [x] 32 SerenaAgent integration tests — full MCP tool chain via `SerenaAgent.get_tool()` (`test/serena/test_cursor_navigation.py`)
- [x] Existing 87 Python tests + 21 snapshot tests unaffected
- [x] Ruff lint and format checks pass
- [ ] CI pipeline passes

### DIFF
--- a/docs/02-usage/042_cursor_navigation.md
+++ b/docs/02-usage/042_cursor_navigation.md
@@ -36,8 +36,7 @@ gracefully skip unavailable edges.
 
 ## Tools
 
-Cursor navigation is available through six MCP tools. These tools are **optional** and
-must be enabled in your project configuration.
+Cursor navigation is available through six MCP tools, enabled by default.
 
 ### cursor_start
 Start a new cursor at a symbol. Returns the symbol's neighborhood showing all reachable
@@ -134,7 +133,6 @@ You can have multiple cursors active simultaneously. Each cursor maintains indep
 state (position, trail, edge type configuration). This is useful for comparing different
 parts of the code or maintaining context while exploring a call chain.
 
-## Enabling Cursor Tools
-Cursor tools are optional and not enabled by default. To enable them, ensure your project
-configuration does not exclude them. The tools are marked with `ToolMarkerOptional`, so they
-will appear in the tool list only when explicitly included.
+## Disabling Cursor Tools
+Cursor tools are enabled by default. To disable them, add the cursor tool names to the
+`excluded_tools` list in your project configuration.
--- a/docs/02-usage/042_cursor_navigation.md
+++ b/docs/02-usage/042_cursor_navigation.md
@@ -1,16 +1,24 @@
-# Cursor-Based Code Navigation
+# Cursor-Based Code Navigation and Editing
 
-Serena provides a cursor-based navigation system for incrementally exploring code structure
-through LSP graph edges. Instead of jumping directly to a symbol, you place a **cursor** on
-a symbol and then **move** it along relationships like containment, references, calls, and
-type hierarchy.
+Serena provides a cursor-based interface for incrementally exploring and editing code
+structure through LSP graph edges. Instead of jumping directly to a symbol, you place
+a **cursor** on a symbol and then **move** it along relationships like containment,
+references, calls, and type hierarchy. Edits (replace body, insert before/after,
+rename) are issued at the cursor's current position.
+
+The cursor is the full MCP-exposed interface for LSP-addressable (symbol-level)
+activity: search, navigation, overview, read, and edit. The older path-and-name-path
+symbol tools (`find_symbol`, `get_symbols_overview`, `find_referencing_symbols`,
+`replace_symbol_body`, `insert_before_symbol`, `insert_after_symbol`, `rename_symbol`)
+are still available as Python classes for direct CLI / API use, but are no longer
+surfaced as default MCP tools — the cursor primitives cover all of their uses.
 
 ## Concepts
 
 ### Cursor
-A cursor is a named position in the code graph. You start it at a symbol, then move it to
-neighboring symbols. Each cursor maintains a **trail** of every position it has visited,
-giving you a breadcrumb path through the code.
+A cursor is a named position in the code graph. You start it at a symbol, then move
+it to neighboring symbols. Each cursor maintains a **trail** of every position it has
+visited, giving you a breadcrumb path through the code.
 
 ### Edge Types
 Cursors navigate along seven types of edges:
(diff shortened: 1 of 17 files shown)


MischaPanch:
Hi @Asher- . Would you be up for a call to discuss the problem you are trying to solve here? There are many possible approaches to simplify the "map a repository" task, the one you propose is only one of them.

Asher-:
Absolutely— I sent you an email to coordinate a chat. 

You're certainly right about there being many ways to map a repository; but the point here is not to create a map, rather to treat the existing information _like a map_. Claude can only see what makes it into Claude's context window. When MCPs operate through a query interface where all activity happens between query (send) and response (receive), everything in-between is hidden from Claude. 

The goal here is to let Claude be in control of the navigation through the information according to the terms in which the information is used. It gives Claude a geographical location _in_ the LSP and lets it relate to the LSP content like moving through a labyrinth. Any given location is a node in a graph and permits Claude to 1. investigate what is at that node and where it connects along various axes of determination 2. to move along any of these axes to follow where it leads 3. to maintain multiple locations in the code space and navigate their relations as they interact. 

Of course all of this can be achieved from a top-down summary approach, but that relies on presenting "everything" at once and then having Claude sort through what matters and what to do with it as well as keeping track of it in the context. The cursor makes the context tracking a part of the history of the cursor that can also be investigated and recalled and that naturally includes the direction of inquiry because it is structured according to taking steps through the inquiry.

It also functions to narrow down the reporting criterion, allowing any particular report to 1. report less 2. format more coherently and with less structure 3. imply options for next steps automatically 4. utilize far fewer cumulative tokens.

opcode81:
@Asher- @MischaPanch please invite me to the call as well.

Asher-:
The red X on this run (`24411939757`) is CI infrastructure flakiness, not a code regression. Diagnosis:

**Ubuntu (2 FAILED tests, both haxe)**
```
FAILED test/serena/test_serena_agent.py::TestSerenaAgent::test_find_symbol_stable[haxe-Main-Class-Main.hx]
FAILED test/serena/test_serena_agent.py::TestSerenaAgent::test_find_symbol_references_stable[haxe-addNumbers-src/utils/Helper.hx-src/Main.hx]
```
Root cause: `urllib.error.HTTPError: HTTP Error 403: Forbidden` while downloading the Haxe Language Server from Open VSX in the test-setup step. External mirror rejected the request.

**Windows (setup failure, 0 tests run)**
```
100   108    0   108    0     0   1090      0 [fpc-source.zip]
End-of-central-directory signature not found. ... unzip: cannot find zipfile directory
##[error]Process completed with exit code 9.
```
Root cause: `fpc-source.zip` download returned 108 bytes of HTML (likely an error page) instead of the Free Pascal compiler source archive. Job exited before pytest ever ran.

**macOS:** PASS.

**Comparison:** The previous Tests run on this branch (`24380392955` at `5d6f67ee`, 12 hours earlier) was fully green — both haxe tests PASSED on ubuntu and the Windows fpc setup succeeded. The infra code hasn't changed between `5d6f67ee` and `9db8c3ab`; only cursor/symbol-tool changes landed.

**Local verification:** All 101 cursor tests (`test_cursor.py`, `test_cursor_integration.py`, `test_cursor_navigation.py`) pass locally on `9db8c3ab`.

No FAILED test in the CI output touches code this PR modifies. A rerun of the failed jobs should clear it — I don't have permission to trigger that on this repo. @opcode81 / @MischaPanch, could one of you rerun, or let me know if you'd prefer I push a no-op trigger commit instead?

MischaPanch:
We discussed this online and concluded that @Asher- would experiment with this and prepare a demonstration that this approach can be beneficial. Since this hasn't happened in multiple month, closing this PR.

@Asher- let me know if you want to discuss this

PR #1339, opened 2026-04-14
feature: add Dockerfile.maximal to optimize CI (#1252)

### LINKED ISSUE #1252: Accelerate CI, extend docker build
Most time in the CI is spent installing dependencies. We already use caching there, but likely not perfectly.

We also only build a minimal docker image that doesn't include all dependencies.

It would be useful to also build a maximal docker image that includes all preparatory steps from the CI workflow. In this image, all tests for all languages would run out of the box without further installations. We could then use this image for CI tests, though since we also test on windows and macOS it's likely not enough and we'd need to use some other strategy to not have to install so much from scratch all the time.

Help from the community on this would be appreciated, ideally by someone with experience on how to optimize GitHub CI workflows.

Note that making the maximal docker image is not completely trivial, in the current CI workflow we use a lot of setup actions from the GH action marketplace which would need to be translated into appropriate docker commands.

tirthpatel90:
This PR introduces Dockerfile.maximal to resolve #1252. It successfully ports the heaviest dependencies (R, Julia, Rust, Go, C++, Node, Ruby, etc.) into a native Docker image, bypassing OS security restrictions and memory limits.
Note: For stability and to avoid GitHub Actions OOM limits, extremely niche toolchains (like Swift, Haskell, Lean4) are deferred to a Phase 2 update. This Phase 1 image immediately resolves the primary CI compilation bottlenecks.


MischaPanch:
Hey @tirthpatel90 , thanks for the PR. This just adds a new dockerfile, I don't see how this is accelerating CI. The dockerfile won't be built into an image or used anywhere in CI, or am I missing something?

tirthpatel90:
Hi @MischaPanch, you are absolutely right! 

My main focus initially was to successfully build this massive `Dockerfile.maximal` without hitting GitHub's 6-hour timeout or OOM limits. (Note: I intentionally included the heaviest bottlenecks like R, Julia, Rust, Node, Go, C++, etc., but deferred a few remaining ones like Swift, Haskell, and OCaml to ensure the build remains stable within runner limits for now).

Now that the core image builds perfectly, I'm ready to wire it up to the CI! 

To actually accelerate the CI, what is your preferred strategy? Should we add a workflow to publish this image to the GitHub Container Registry (`ghcr.io`), and then modify the main test workflows to pull and run inside it? 

Let me know how you'd like to handle the image hosting, and I'll push the necessary CI workflow changes to this PR right away!

MischaPanch:
Yes, let's build it, push it to the GH container registry, adjust the pytest workflow to use it and see how much it accelerates. Would it be possible for you to test this out in your fork and link the actions here? I suppose you can't write to the right place in the container registry from the actions triggered in a PR, right?

MischaPanch:
I would prefer to review the dockerfile after things are running and acceleration is visible :). That's why I'm asking

MischaPanch:
Btw, setting up OCaml eats up a lot of time in CI, if it's possible to include it here, that would be great. If not, we might just disable it at some point

tirthpatel90:
Sounds like a plan, @MischaPanch! 

I'll add OCaml to the maximal image, set up a workflow on my fork to build and publish it to my personal GHCR, and then adjust the `pytest.yml` to run inside that container. 

I'll ping you here with the action run links showing the acceleration metrics once it's running smoothly on my end. Working on it now!

tirthpatel90:
Hi @MischaPanch,

I have successfully published the maximal image to GHCR and wired it up to a test workflow on my fork. 

The good news: The environment works perfectly and immediately breezes through ~65% of the test suite (including C++, Go, Java, Rust, etc.) without any setup overhead! 

However, the test execution hangs/slows down significantly after the 65% mark. Based on the logs, I suspect two reasons for this:
1. Missing Toolchains: The test suite might be automatically downloading and building the deferred niche toolchains (Haskell,    Swift, Lean 4, etc.) on the fly because they aren't baked into this Phase 1 image.
2. Julia JIT Precompilation: Since we disabled Julia package precompilation in the Docker build to prevent OOM crashes, it  might be running that heavy compilation during the test itself.

Does the test suite automatically attempt to install missing language servers during execution? If so, is there an environment variable or a `pytest` flag I should pass to strictly skip tests for toolchains not present in the image? This would allow us to accurately benchmark the speed of the pre-installed languages!

tirthpatel90:
Closing this PR in favor of the parallelized CI architecture discussed in #1362. 

The monolithic maximal image successfully proved that we can drastically cut down test times by pre-baking dependencies. However, moving forward, we will pivot to segmenting the tests into dynamic matrix batches and running them in parallel using a leaner base image. This will be more scalable and maintainable. 

Thanks for the feedback, everyone! I'll be opening a new PR for the parallel matrix workflows soon.

PR #1349, opened 2026-04-14
Fix duplicated YAML comments

opcode81:
#1285


MischaPanch (commented):
Nice! I suspect you tested this, good to go from my side.

Although it might be almost worth to start adding unit tests for this

MischaPanch (approved):

PR #1352, opened 2026-04-15
feat: add Gleam language support

### LINKED ISSUE #1334: Add support for Gleam
**Additional context**
[Gleam](https://github.com/gleam-lang/gleam) is an increasingly popular language and is known for putting a lot of effort into a [really good LSP](https://gleam.run/language-server/) (which already has a lot of very useful code actions). 

(As an interesting example, given a Gleam type, the LSP can actually generate code for a json encoder and decoder for that type.  As a test, I asked a cheap LLM to generate encoders/decoders for my types and it got completely stuck, even with good documentation available. It did much better when it already had the encoder/decoder structure that I got from the LSP and only needed to make some adjustments according to the JSON schema. 

Personally I only use coding agents in a very constrained way; when I know they can converge towards the solution I actually want. Giving them access to these kinds of code actions would really help with that.)

jakeefr:
Fixes #1334. Added Gleam language support following the existing pattern for other language integrations.

- `GleamLanguageServer` class using the `DependencyProvider` pattern, finds `gleam` on PATH and launches via `gleam lsp`
- `Language.GLEAM` enum variant with `*.gleam` file matcher
- Updated README and programming languages docs
- Added pytest marker for gleam tests

Gleam's LSP is built into the compiler so no additional tooling is needed beyond the Gleam installation itself.

I maintain [PRISM](https://github.com/jakeefr/prism), a post-session diagnostics tool for Claude Code — CLAUDE.md adherence analysis and session health scoring. Serena's semantic retrieval and PRISM's session diagnostics are complementary tools for Claude Code power users.

### DIFF
--- a/src/solidlsp/language_servers/gleam_language_server.py
+++ b/src/solidlsp/language_servers/gleam_language_server.py
@@ -1,0 +1,145 @@
+"""
+Provides Gleam specific instantiation of the LanguageServer class.
+Uses the language server built into the Gleam compiler (gleam lsp).
+"""
+
+import logging
+import os
+import pathlib
+import shutil
+import threading
+
+from overrides import override
+
+from solidlsp.ls import LanguageServerDependencyProvider, LanguageServerDependencyProviderSinglePath, SolidLanguageServer
+from solidlsp.ls_config import LanguageServerConfig
+from solidlsp.lsp_protocol_handler.lsp_types import InitializeParams
+from solidlsp.settings import SolidLSPSettings
+
+log = logging.getLogger(__name__)
+
+
+class GleamLanguageServer(SolidLanguageServer):
+    """
+    Provides Gleam specific instantiation of the LanguageServer class.
+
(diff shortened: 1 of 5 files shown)


src/solidlsp/language_servers/gleam_language_server.py
```diff
@@ -0,0 +1,145 @@
...
+            None,
+            "gleam",
+            solidlsp_settings,
+        )
+        self.server_ready = threading.Event()
+
+    def _create_dependency_provider(self) -> LanguageServerDependencyProvider:
+        return self.DependencyProvider(self._custom_settings, self._ls_resources_dir)
+
+    class DependencyProvider(LanguageServerDependencyProviderSinglePath):
```
MischaPanch, line 44: You don't need a dep. provider when there's nothing to download

MischaPanch (changes requested):
Thanks for the contribution. Tests are missing completely, pls have a look at our instructions for adding language support. I will review once tests pass in CI

MischaPanch:
No reaction, closing

PR #1353, opened 2026-04-15
Implement security recommendations

MischaPanch:


.github/workflows/publish.yml
```diff
@@ -52,18 +27,11 @@ jobs:
       - name: Build package
         run: uv build
 
-      - name: Upload artifacts to GitHub Release
-        if: env.PUBLISH_TO_PYPI == 'true'
-        uses: softprops/action-gh-release@v2
+      - name: Create GitHub Release
```
MischaPanch, line 30: @opcode81 fyi - this should be enough to publish a release, the build job was not necessary. We will verify with the next release

Issue #1348, opened 2026-04-14
rename_symbol applies fewer changes on repeated rename of the same symbol (vue)


ek-133:
## Description                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                
*First of all, thanks for the quick fix on the previous issue — really appreciated! It would be great if this one could get the same treatment and land in a new release soon.*
                                                                                              
 When renaming a symbol back to its original name, rename_symbol only updates the definition file and misses usages in other files that were correctly updated in the first rename.

## Steps to reproduce                                                                                                          

  1. Have an enum FeatureFlags defined in src/useFeatureFlags.ts and imported/used in src/App.vue                                                                                                                                               
  2. Call rename_symbol to rename FeatureFlags → TestFeatureFlags
    - Result: 2 changes applied (.ts + .vue) ✅                                                                                                                                                                                                                                 
  3. Call rename_symbol to rename TestFeatureFlags → FeatureFlags (reverse)                                                   
    - Result: 1 change applied (.ts only) ❌ — .vue file retains TestFeatureFlags                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                
## Expected behavior
                                                                                                                                                                                                                                                                                
  Both calls should apply changes in all files where the symbol is referenced (2 changes each).                               

## Actual behavior

  The second (reverse) rename reports 1 change and leaves App.vue with the stale name TestFeatureFlags.                                                                                                                                                                    
   
## Environment                                                                                                                                                                                                                                                                   
                                                                                                                              
  - Project: Vue 3 + TypeScript
  - Files involved:
    - src/useFeatureFlags.ts (definition)
    - src/App.vue (usage)                                                                                                                                                                                                                                   
   
## Hypothesis                                                                                                                                                                                                                                                                    
                                                                                                                              
  Possibly a caching/indexing issue — after the first rename, the language server index may not have re-indexed usages in .vue files before the second rename was executed, so the reverse rename only found the symbol in the .ts definition file.

---

- [x] read the relevant parts of the documentation and verified that the issue cannot be solved by adjusting configuration
- [x] understood that the Serena Dashboard can be disabled through the config
- [x] understood that, by default, a client session will start a separate instance of a Serena server. 
- [x] understood that, for multi-agent setups, the Streamable HTTP/SSE mode should be used.
- [x] understood that non-project files are ignored using either .gitignore or the corresponding setting in `.serena/project.yml`
- [x] looked for similar issues and discussions, including closed ones
[...]

PR #1355, opened 2026-04-15
Issue #1348 Vue rename_symbol fix

ek-133:


MischaPanch:
Thanks! The managing of the state of the two LS is getting increasingly complicated :(

MischaPanch:
Could you add an entry to the changelog before we merge this pls?

PR #1356, opened 2026-04-16
Fix Lean4 stale cache: skip caching empty document symbol responses

### LINKED ISSUE #1217: Lean4 document symbol cache ignores build state and serves stale empty results
Text below is written by codex, so there may be inaccuracies. My summary - the lean LS doesn't work when lean build was not called before the LS was started, and moreover calling before the build can even poison our cache, so we need to invalidate the cache somehow.

This never appears as a problem in CI since there the build happens in the lean-setup GH action.

## Summary

  The Lean4 integration caches documentSymbol responses without considering whether the Lean project has been built. If
  Serena queries a Lean file before lake build has produced the .olean artifacts, the language server can return empty
  or incomplete symbol data. That empty result is then cached and kept even after the project is built, because the
  cache key only depends on file path and file contents.

##  Observed behavior

  In test/resources/repos/lean4/test_repo, querying Main.lean before building the repo produced empty document symbols
  and missing cross-file navigation. After lake build, cross-file references/definitions started working, but Main.lean
  symbols still stayed empty because Serena reused the stale cached result from .serena/cache/lean4/
  raw_document_symbols.pkl.

##  Why CI passes

  CI builds the Lean test project first in /C:/Users/misch/Projects/oraios/serena/.github/workflows/pytest.yml:550, so
  the language server sees a ready project and never writes the bad cache entry.

##  Root cause

  The raw document symbol cache in /C:/Users/misch/Projects/oraios/serena/src/solidlsp/ls.py:1481 is keyed by relative
  path plus file content hash, but not by Lean build state. For Lean4, build artifacts from lake build materially affect
  symbol resolution, so cache validity depends on more than source contents.

##  Needed fix

  Invalidate Lean4 document symbol caches when Lean build state changes, and include Lean build state in cache validity.
  Concretely:

  - take lake build output state into account for Lean4 cache invalidation
  - avoid reusing cached empty symbol results created before the project was built
  - ideally include relevant build artifacts or a Lean-project-specific build fingerprint in the Lean4 cache version/key

##  Acceptance criteria

  - querying Lean files before build does not poison the cache permanently
  - after lake build, a fresh request for Main.lean returns correct symbols without manual cache deletion
  - local behavior matches CI for the Lean4 test project

Koushik-Salammagari:
Fixes #1217

## Problem

When Serena queries a Lean file **before** `lake build` has produced the `.olean` artifacts, the language server returns an empty symbol list. Previously this empty result was written to the persistent `_raw_document_symbols_cache`. On subsequent requests — even after a successful build — the cached empty result was returned instead of querying the now-ready language server, permanently hiding all symbols until the user manually deleted the cache.

## Fix

In `SolidLanguageServer._request_document_symbols` (ls.py), only write to the cache when the response is non-empty:

```python
if response:
    self._raw_document_symbols_cache[cache_key] = (fd.content_hash, response)
    self._raw_document_symbols_cache_is_modified = True
```

An empty/`None` response is still returned to the caller so the immediate request behaves correctly; it is simply not persisted.

## Test

Added `test_empty_symbol_response_not_cached` to `test/solidlsp/lean4/test_lean4_basic.py`. It patches `server.send.document_symbol` to return `[]` (simulating a pre-build response) and asserts that `_raw_document_symbols_cache` does not contain an entry for that file afterwards.

## Notes

- The fix is in the base class (`SolidLanguageServer`), so it guards all language servers against accidentally caching empty results, not only Lean4.
- CI already passes for Lean4 because the build runs before tests; this fix only affects the local/pre-build scenario described in the issue.


MischaPanch (changes requested):
Thanks! Pls run poe format so CI can run through and add an entry to the changelog, then this can be merged

Koushik-Salammagari:
@MischaPanch — both points from your review are addressed: ran `poe format` (fixed one lint issue) and added a changelog entry under `# Unreleased (main)`. Should be ready to merge now!

PR #1359, opened 2026-04-16
Fix clangd cache invalidation and capability checks

### LINKED ISSUE #1358: C++ clangd symbol cache can preserve malformed document symbols after LS/config recovery
## Summary

In C++ projects, Serena can keep returning malformed or empty results from `get_symbols_overview` / `find_symbol` even after the underlying `clangd` setup is fixed.

The root cause appears to be stale cached C++ document symbols under `.serena/cache/cpp/`, which are reused based only on file-content hash. If bad symbol data is cached while `clangd` is misconfigured or degraded, Serena continues serving those bad symbols even after the language server environment is repaired.

## What I observed

I hit this while debugging a C++ project that initially had multiple clangd-related environment issues. After fixing those issues:

- `clangd` itself could parse the project correctly
- `compile_commands.json` was valid
- generic LSP symbol queries returned correct C++ symbols
- but Serena's `get_symbols_overview` / `find_symbol` still returned stale bad results

Examples from the affected repo:

- `main.cpp` was cached as only `SIGINT`
- `src/utils/logger.cpp` was cached as having no symbols
- `get_symbols_overview` kept returning those incorrect results
- clearing `.serena/cache/cpp/` and restarting Serena fixed the issue immediately

## Why this looks like a Serena cache invalidation issue

From tracing the local installed code:

- `request_overview -> request_document_overview -> request_document_symbols`
- `request_document_symbols` prefers cached `DocumentSymbols` if the file content hash matches
- `ClangdLanguageServer` does not appear to contribute a document-symbol cache fingerprint that reflects LS/runtime/config changes
- so cached bad symbol trees remain "valid" as long as file contents are unchanged

This means cache entries created during a broken/degraded clangd state can survive later LS recovery.

## Expected behavior

Serena should invalidate or bypass cached C++ document symbols when clangd-relevant state changes, for example:

- clangd executable/version changes
- clangd launch args change
- compile commands directory / relevant LS config changes
- other LS-specific conditions that affect document symbols change

## Actual behavior

Serena continues returning malformed cached C++ symbols until `.serena/cache/cpp/` is manually deleted.

## Minimal workaround

Delete:

- `.serena/cache/cpp/document_symbols.pkl`
- `.serena/cache/cpp/raw_document_symbols.pkl`

Then restart Serena.

## Suggested fix direction

Add a clangd-specific document-symbol cache fingerprint so C++ symbol caches are invalidated when the effective clangd/config/runtime state changes.

I also prepared a local fix for this, together with a small clangd capability-compatibility adjustment and a C++ cache invalidation regression test.

mesa3:
## Summary

- relax clangd initialize capability checks to tolerate valid response shape differences
- add a clangd-specific document-symbol cache fingerprint so stale C++ symbol caches are invalidated when backend/config state changes
- add a regression test covering C++ document-symbol cache invalidation

## Problem

For C++ projects, Serena's symbol tools could keep returning malformed cached document symbols even after clangd / compile commands / backend state had been repaired. The cache key was effectively stable as long as file contents did not change, so previously bad symbol trees could survive later LS recovery.

In addition, clangd capability checks were too strict and assumed a single exact initialize-response shape.

## What this changes

- accepts both integer and object-style `textDocumentSync` responses
- validates `completionProvider` more conservatively by checking required trigger characters instead of exact equality
- fingerprints clangd document-symbol caches using:
  - `clangd_version`
  - `ls_path`
  - `compile_commands_dir`
  - `compile_commands.json` content hash
- adds a C++ regression test similar to the existing Go cache invalidation coverage

## Validation

- `uv run --extra dev -m pytest test/solidlsp/cpp/test_cpp_basic.py -q`
- result: `9 passed, 2 xfailed`

Closes #1358

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ Status of the `main` branch. Changes prior to the next official version change w
 * Dependencies:
   - `pywebview`: Switch back to official release (new version 6.2) #1253
 
+* Language Servers:
+  - Fix clangd capability checks to tolerate valid initialize response shape differences and invalidate cached C++ document symbols when clangd/compile commands context changes #1359
+
 # v1.1.2 (2026-04-14)
 
 * General:
--- a/src/solidlsp/language_servers/clangd_language_server.py
+++ b/src/solidlsp/language_servers/clangd_language_server.py
@@ -1,8 +1,10 @@
+import hashlib
 import json
 import logging
 import os
 import pathlib
 import threading
+from collections.abc import Hashable
 from typing import Any, cast
 
 from overrides import override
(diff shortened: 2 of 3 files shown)


MischaPanch:
Thanks! Pls add an entry to the changelog, then this is good to go

mesa3:
I’ve added the changelog entry under the unreleased. Also apologies for the earlier force-push XD.  it was only to correct the automatically generated agent commit metadata, not to change the actual implementation.

MischaPanch:
Np, on the contrary the force push is appreciated, we want to keep the git history clean

Issue #1147, opened 2026-03-10
Codex on Linux: dashboard starts but browser auto-open can fail with "no usable display" in graphical sessions


rldyourmnd:
Hi, I ran into this while setting up Serena with Codex on a normal graphical Linux session.

The dashboard itself starts fine, but Serena decides there is no usable display and skips opening the browser:

```text
INFO  serena.dashboard:run_in_thread:642 - Starting dashboard (listen_address=127.0.0.1, port=24283)
INFO  serena.agent:__init__:361 - Serena web dashboard started at http://127.0.0.1:24283/dashboard/index.html
WARNING serena.agent:open_dashboard:471 - Not opening the Serena web dashboard because no usable display was detected.
```

This was not a headless shell or a container. It happened in an active GNOME / Wayland desktop session.

## Environment

- OS: Linux 6.17.0-14-generic x86_64 GNU/Linux
- Desktop session: GNOME / Wayland
- Codex CLI: 0.112.0
- Python: 3.14.3
- uvx: 0.10.9
- Serena: `0.1.4-4048c3d0-dirty` (from the startup log)

## Config

Codex MCP entry:

```toml
[mcp_servers.serena]
command = "uvx"
args = ["--from", "git+https://github.com/oraios/serena", "serena", "start-mcp-server", "--context", "codex"]
```

Global Serena config:

```yaml
web_dashboard: true
web_dashboard_open_on_launch: true
```

## How to reproduce

1. Start from a regular graphical Linux session.
2. Configure Serena as a stdio MCP server under Codex.
3. Launch Codex and let it start Serena.
4. Check the Serena startup logs.

## What I expected

Either the browser opens normally, or Serena can tell me more explicitly that the MCP client started it with an incomplete GUI/session environment.

## What fixed it locally

I wrapped the Serena startup command and restored the usual desktop session variables when they were missing:

- `XDG_RUNTIME_DIR`
- `DBUS_SESSION_BUS_ADDRESS`
- `WAYLAND_DISPLAY`
- `DISPLAY`

After that, Serena started the dashboard and no longer logged `no usable display was detected`.

So this looks less like "there is no display" and more like "the MCP subprocess was started without enough session information to detect one".

## Suggestion

Would it make sense for Serena to be a bit more forgiving here on Linux before deciding that the session is headless?

For example, if these values are missing, Serena could try a lightweight recovery path before giving up:

- derive `XDG_RUNTIME_DIR` from `/run/user/$UID`
[...]

Issue #898, opened 2026-01-03
MCP error -32000: Connection closed when fetching tools in Claude Code CLI


fregataa:
## Description

Serena MCP server connects successfully to Claude Code CLI, but immediately fails when Claude Code attempts to fetch tools. The server shows "Successfully connected" status, but tool calls fail with "No such tool available" error.

## Environment

- **OS**: macOS Darwin 24.6.0 (arm64)
- **Claude Code Version**: 2.0.76
- **Serena Version**: 0.1.4 (commit: 5e7d5469cf356c7ef72503243b5d3152d75cdbfa)
- **Python**: via uvx
- **uvx Version**: 0.9.21

## MCP Server Configuration

```json
{
  "mcpServers": {
    "serena": {
      "type": "stdio",
      "command": "uvx",
      "args": [
        "--from",
        "git+https://github.com/oraios/serena",
        "serena",
        "start-mcp-server",
        "--context",
        "claude-code",
        "--project",
        "/Users/sh/TrueGul"
      ],
      "env": {}
    }
  }
}
```

## Steps to Reproduce

1. Configure Serena as MCP server in Claude Code (as shown above)
2. Start Claude Code in the project directory
3. Run `claude mcp list` - shows serena as "Connected"
4. Try to use any Serena tool (e.g., `list_dir`)
5. Error: "No such tool available: mcp__serena__list_dir"

### Claude Code Debug Log

```
2026-01-04T01:03:35.338Z [DEBUG] MCP server "serena": Starting connection with timeout of 30000ms
2026-01-04T01:03:40.155Z [DEBUG] MCP server "serena": Successfully connected to stdio server in 4820ms
2026-01-04T01:03:40.412Z [ERROR] MCP server "serena" Failed to fetch tools: MCP error -32000: Connection closed
2026-01-04T01:03:40.412Z [ERROR] MCP server "serena" Failed to fetch commands: MCP error -32000: Connection closed
2026-01-04T01:03:40.412Z [ERROR] MCP server "serena" Failed to fetch resources: MCP error -32000: Connection closed
2026-01-04T01:05:35.620Z [ERROR] Error: Error: Tool mcp__serena__list_dir not found
```

### Serena Server Log

Server starts successfully and loads 21 tools:

```
INFO  serena.cli:start_mcp_server:229 - Initializing Serena MCP server
INFO  serena.agent:__init__:239 - Starting Serena server (version=0.1.4-69d39645-dirty, ...)
[...]

PR #1364, opened 2026-04-16
fix(agent): swallow OSError in _open_dashboard_in_browser (#1363)

### LINKED ISSUE #1363: Fatal OSError [Errno 14] Bad address in _open_dashboard_in_browser on Linux under Claude Code stdio MCP
## Summary

On Linux, when `web_dashboard_open_on_launch: true` (the default) and serena is started as an stdio MCP server by Claude Code CLI, `_open_dashboard_in_browser` raises a fatal `OSError: [Errno 14] Bad address` from `subprocess.Popen`. The exception propagates out of `SerenaAgent.__init__`, which means `create_mcp_server` never completes. The process stays alive long enough for pyright-langserver to initialize, but the stdio handshake never responds, so Claude Code times out after 30s with `MCP server \"serena\" connection timed out after 30000ms`.

No existing issue mentions EFAULT / `Errno 14` / `Bad address` for this code path (searched `oraios/serena`).

## Environment

- serena **1.1.2** (installed via `uv tool install`)
- Python **3.13.12** (uv-managed, `/home/marcos/.local/share/uv/tools/serena-agent/bin/python`)
- Linux **6.17.0-14-generic** (Ubuntu/derivative, Wayland session, no system tray)
- Client: **Claude Code CLI** (stdio transport, `--context claude-code --project-from-cwd`)

## Symptom

Claude Code log:

```
MCP server \"serena\" connection timed out after 30000ms
```

MCP stderr log from `~/.cache/claude-cli-nodejs/<project>/mcp-logs-serena/<ts>.jsonl`:

```
ERROR  serena.agent:__init__:405 - Fatal exception while initializing Serena agent:
  File \".../serena/mcp.py\", line 324, in create_mcp_server
  File \".../serena/mcp.py\", line 264, in <lambda>
  File \".../serena/agent.py\", line 405, in __init__
    self._start_dashboard_viewer(minimized=not self.serena_config.web_dashboard_open_on_launch)
  File \".../serena/agent.py\", line 612, in _start_dashboard_viewer
    self._open_dashboard_in_browser(url)
  File \".../serena/agent.py\", line 633, in _open_dashboard_in_browser
    subprocess.Popen(
  File \".../subprocess.py\", line 1991, in _execute_child
OSError: [Errno 14] Bad address: '/home/marcos/.local/share/uv/tools/serena-agent/bin/python'
```

The `sys.executable` path is valid (symlink → `cpython-3.13.12`); the same binary runs fine outside the MCP stdio environment. The kernel returns EFAULT from `execve` only inside the stdio-hosted agent process — suspected interaction between the fd layout / address space inherited from the MCP parent and `subprocess._execute_child` on this kernel.

Preceding INFO line:

```
INFO  serena.agent:_start_dashboard_viewer:610 - Not starting Serena dashboard viewer because the current platform does not support it; using browser-based fallback
```

So the branch taken is the Linux \"no system tray\" fallback that calls `_open_dashboard_in_browser`.

## Root cause

`SerenaAgent._open_dashboard_in_browser` (agent.py L630–L639 in 1.1.2):

```python
@staticmethod
def _open_dashboard_in_browser(url: str) -> None:
    # Use a subprocess to avoid any output from webbrowser.open being written to stdout
    subprocess.Popen(
        [sys.executable, \"-c\", f\"import webbrowser; webbrowser.open({url!r})\"],
        stdin=subprocess.DEVNULL,
        stdout=subprocess.DEVNULL,
        stderr=subprocess.DEVNULL,
        start_new_session=False,
    )
```

No `try/except`. The only upstream guard is `system_has_usable_display()` in `_start_dashboard_viewer` (L598), which returns `True` here because `$DISPLAY`/`$WAYLAND_DISPLAY` is set — it just doesn't protect against `execve` failing.

Because this call sits inside `SerenaAgent.__init__`, any `OSError` aborts agent construction. `create_mcp_server` never returns → stdio never responds → client timeout. A best-effort \"open the browser\" side-effect shouldn't be able to brick the MCP handshake.

## Workaround

Set in `~/.serena/serena_config.yml`:

```yaml
web_dashboard_open_on_launch: false
```

Dashboard itself (`web_dashboard: true`) keeps working — reachable at `http://localhost:24282/dashboard/` manually. Only the auto-open is suppressed. This matches the workaround from #898 (macOS / different traceback, same general class).

## Suggested fix

Wrap the `Popen` in `try/except OSError` and log a warning. Opening the browser is best-effort and must never be fatal:

```python
@staticmethod
def _open_dashboard_in_browser(url: str) -> None:
    try:
        subprocess.Popen(
            [sys.executable, \"-c\", f\"import webbrowser; webbrowser.open({url!r})\"],
            stdin=subprocess.DEVNULL,
            stdout=subprocess.DEVNULL,
            stderr=subprocess.DEVNULL,
            start_new_session=False,
        )
    except OSError as e:
        log.warning(
            \"Failed to auto-open dashboard in browser (%s). \"
            \"Dashboard still accessible at %s\", e, url,
        )
```

Same treatment would apply to the `open_dashboard()` tool path (L627).

Optional hardening: also guard `_start_dashboard_viewer_process_function` / the `multiprocessing.Process` path — the tray/viewer lifecycle has been fragile in other reports (#1281 WebView2 leak, #1147 Codex).

## Related

- #898 — Claude Code CLI + serena stdio on macOS, handshake failure; workaround was disabling `web_dashboard` + `web_dashboard_open_on_launch`. Same root class (dashboard side-effect breaks MCP handshake), different traceback.
- #613 — request to disable dashboard by default for Claude Code users; closed without changing the default.
- #1147 — Codex CLI + Linux, browser auto-open falls through cleanly because \"no usable display\" is detected first. That path is fine; this bug is when `system_has_usable_display()` returns True but `execve` then fails.
- #1266 (open PR by @opcode81) — touches `_start_dashboard_viewer` but keeps `_open_dashboard_in_browser` unchanged; does **not** address this failure mode.

Happy to submit a PR for the try/except patch if it'd help.

marcosrvaz:
## Summary

Fixes #1363.

`SerenaAgent._open_dashboard_in_browser` currently calls `subprocess.Popen` with no error handling. Because it runs inside `SerenaAgent.__init__` (via `_start_dashboard_viewer` on Linux systems without a usable tray), any `OSError` from `Popen` aborts agent construction. When that happens under Claude Code's stdio MCP transport, `create_mcp_server` never returns and the client times out after 30s with `MCP server "serena" connection timed out after 30000ms`.

Observed failure on Linux 6.17 / Python 3.13 / serena 1.1.2:

```
OSError: [Errno 14] Bad address: '/…/serena-agent/bin/python'
```

(full traceback in #1363 — `execve` EFAULT inside the stdio-hosted agent, while the same binary runs fine outside that context).

Opening a browser is best-effort and must never be fatal. This PR wraps the `Popen` in `try/except OSError` and logs a warning; the dashboard itself remains reachable at its URL.

## Changes

- `src/serena/agent.py`: catch `OSError` in `_open_dashboard_in_browser`, log a warning pointing at the dashboard URL instead of propagating.
- `test/serena/test_dashboard.py`: regression test — monkeypatches `subprocess.Popen` to raise `OSError(14, "Bad address")` and asserts the call returns normally with a `WARNING` on the `serena.agent` logger.

## Test plan

- [x] `poe format` — clean
- [x] `poe type-check` — `Success: no issues found in 124 source files` (both src and test passes)
- [x] `pytest test/serena/test_dashboard.py -v` — 3 passed (2 pre-existing + new regression test)

## Notes

- Scope intentionally minimal: only the `Popen` call in `_open_dashboard_in_browser` is touched. The same method is also called from the user-triggered `open_dashboard()` tool path (agent.py:627), which benefits from the same protection.
- Existing `test_serena_agent.py` fixtures already set `web_dashboard=False`, so the buggy path had no CI coverage. This PR adds targeted coverage.
- Related but not a duplicate: #898 (macOS, same class — dashboard side-effect breaking MCP handshake), #1147 (Codex on Linux — different branch, handled via `system_has_usable_display` early-return), #1266 (open PR that refactors `_start_dashboard_viewer` but keeps `_open_dashboard_in_browser` unchanged).

cc @opcode81

### DIFF
--- a/src/serena/agent.py
+++ b/src/serena/agent.py
@@ -629,14 +629,24 @@ class SerenaAgent:
 
     @staticmethod
     def _open_dashboard_in_browser(url: str) -> None:
-        # Use a subprocess to avoid any output from webbrowser.open being written to stdout
-        subprocess.Popen(
-            [sys.executable, "-c", f"import webbrowser; webbrowser.open({url!r})"],
-            stdin=subprocess.DEVNULL,
-            stdout=subprocess.DEVNULL,
-            stderr=subprocess.DEVNULL,
-            start_new_session=False,
-        )
+        # Use a subprocess to avoid any output from webbrowser.open being written to stdout.
+        # Auto-opening the browser is best-effort: any failure here must be non-fatal, since
+        # this runs inside SerenaAgent.__init__ and a propagating OSError would abort agent
+        # construction and break the MCP handshake (see oraios/serena#1363).
+        try:
+            subprocess.Popen(
+                [sys.executable, "-c", f"import webbrowser; webbrowser.open({url!r})"],
+                stdin=subprocess.DEVNULL,
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
+                start_new_session=False,
+            )
+        except OSError as e:
+            log.warning(
+                "Failed to auto-open dashboard in browser (%s); dashboard remains accessible at %s",
+                e,
+                url,
+            )
 
     def get_exposed_tool_instances(self) -> list["Tool"]:
         """
--- a/test/serena/test_dashboard.py
+++ b/test/serena/test_dashboard.py
@@ -1,5 +1,9 @@
+import logging
 from types import SimpleNamespace
 
+import pytest
+
+from serena.agent import SerenaAgent
 from serena.dashboard import SerenaDashboardAPI
 from solidlsp.ls_config import Language
 
@@ -47,3 +51,19 @@ def test_available_languages_exclude_project_languages():
     assert Language.MARKDOWN.value not in available
     # ensure experimental languages remain available for selection
     assert Language.ANSIBLE.value in available
+
+
+def test_open_dashboard_in_browser_swallows_os_error(monkeypatch: pytest.MonkeyPatch, caplog: pytest.LogCaptureFixture) -> None:
+    """`_open_dashboard_in_browser` must never propagate OSError: it runs inside
+    SerenaAgent.__init__ and a fatal here breaks the MCP handshake (see #1363).
+    """
+
+    def _raise(*_args: object, **_kwargs: object) -> None:
+        raise OSError(14, "Bad address")
+
+    monkeypatch.setattr("serena.agent.subprocess.Popen", _raise)
+
+    with caplog.at_level(logging.WARNING, logger="serena.agent"):
+        SerenaAgent._open_dashboard_in_browser("http://localhost:24282/dashboard/")
+
+    assert any("Failed to auto-open dashboard" in record.message and record.levelno == logging.WARNING for record in caplog.records)


opcode81:
The code has moved, but I have now added the try-catch block in the new location.

PR #1365, opened 2026-04-17
Fix in request_overview: explicitly passed file path bypassed ignore check

### LINKED ISSUE #1361: find_symbol returns empty results for files in .claude/worktrees/
Environment

- Serena version: 1.1.2
- Backend: LSP / Claude Code
- OS: macOS
- Language: Vue 3 + TypeScript
- Project path: /Users/xxx/.claude/worktrees/feat/ (worktree inside .claude subdirectory)

Description

find_symbol returns [] for all symbols inside .vue Single File Components located under .claude/worktrees/, while get_symbols_overview successfully lists symbols from the same files.

Steps to reproduce:
1. Call find_symbol(name_path_pattern="handleBack", relative_path="frontend/src/views/ClothDetailView.vue")
2. Result: []
3. Call get_symbols_overview(relative_path="frontend/src/views/ClothDetailView.vue") on the same file
4. Result: Successfully returns all symbols including handleBack, enterEdit, loadDetail, etc.

// find_symbol → []
// get_symbols_overview → ["handleBack", "enterEdit", "loadDetail", ...]

get_symbols_overview works, confirming the file is indexed and the language server can access it. find_symbol silently returns empty — no error, no indication of why.

Expected behavior

find_symbol should return symbol information for .vue files, consistent with get_symbols_overview.

Related issues

- #853: find_symbol vs get_symbols_overview inconsistency (dot-directory files like .claude/lib/)
- #1187: In current version (0.1.4), both tools fail with ValueError on dot-prefixed directory paths
- #1333: rename_symbol skips .vue files (Vue + Serena LSP compatibility issue)
- Commit 1558a4e fixed Vue LSP readiness in request_references, but find_symbol still fails

Root cause (suspected)

Combination of two known bugs:
1. .claude/worktrees/ is a dot-prefixed path — find_symbol has systematic issues with such paths (#853 / #1187)
2. Vue SFC <script setup> symbol retrieval is inherently unstable in Serena (#1333)

get_symbols_overview uses request_document_symbols (works), while find_symbol uses request_full_symbol_tree which fails on dot-prefixed Vue files.

MischaPanch:
Closes: #1361

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Status of the `main` branch. Changes prior to the next official version change w
 * General:
   - Support `serena --version` CLI command for displaying the current version #1347
   - Fix: Check for ignored path ignored `.git` folder only at the top level, not in every subdirectory (`Project._is_ignored_relative_path`) #1350
+  - `GetSymbolsOverviewTool`: ignored paths were not respected in LSP variant (fix in `SolidLanguageServer`)
   - Fix: Duplicate comments in re-saved YAML configuration files #1285
 
 * JetBrains:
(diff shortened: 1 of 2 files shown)


MischaPanch:
@opcode81 merging this, it's uncontroversial IMO

PR #1371, opened 2026-04-18
Improve project-specific prompt provision

### LINKED ISSUE #1370: Prompt of newly activated mode is not passed to LLM when using dynamic project activation
Im using a custom mode (project.yml - default_modes:), it is correctly displayed in Dashboard under my project.

When are the modes injected?  I dont see the custom modes prompt during MCP initialize (which would be expected for a project based mode) or "activate project". Do i need to put a placeholder or include into .serena/project.yml initial_prompt?
 (im debugging with mcp inspector)

_Originally posted by @tomelgato in https://github.com/oraios/serena/discussions/1369_

### LINKED ISSUE #1372: Project activation message can be provided more than once
When using dynamic project activation and `initial_instructions` is called after project activation, then the message is sent twice (also in response to `initial_instructions`)

opcode81:
* Introduce new abstraction to handle prompt provision.
* Prompt provision status management is session-aware and thus works for HTTP mode as well
* Refactor ActiveModes

Fixes #1370 
Fixes #1372


MischaPanch (approved):
Overall looks good, a shame we are forced to use even more reflection but better like this than the alternative.

I suspect you tested it, and while I would personally feel better with some test coverage, it's good to go if you are confident that the feature works as intended

opcode81:
I have added tests that verify the expected behaviour.

PR #1380, opened 2026-04-19
Introduce DashboardManager with three interface modes for the dashboard

opcode81:
The manager supports three modes, which one can switch between:
* `browser` (legacy mode) - all platforms
* `app` (pywebview with tray as currently used on Windows) - win/macOS
* `tray_manager` (new; central tray icon for all Serena instances, click opens in browser) - win/macOS

The setting is configured in the global config, with the default (null) resulting in a platform-specific selection:
* Windows: `app` remains the default
* Linux: `browser` remains the default
* macOS: `tray_manager` is new default

Things to test:
* The `tray_manager` must be tested on macOS
* The `tray_manager` mode may also work on Linux, but it depends on the DE. We could test this (but we most definitely should not enable it by default on Linux)


opcode81:
@arnaujc91 if you have time, maybe you can test this on macOS:

* Try running several Serena instances from the branch of this PR (`dashboard-manager`)
* As soon as the first one is up, a tray icon should appear.
* The tray icon's menu should have one item for every running instance.
* Clicking on an item should open a browser window.
* The tray icon should disappear when all Serena instances have terminated.

opcode81:
@MischaPanch: We could just merge this if we make minor changes:
* Default on macOS remains `browser`
* Document `tray_manager` option as experimental/alpha

Since the entry in serena_config defaults to `null` meaning "use OS-specific default", we can later change the macOS default to `tray_manager` if it turns out that it works correctly.

BrunoKrugel:
How to disable to icon to appear on macOS menu bar?

Image

I have web_dashboard_open_on_launch: False to prevent the dashboard from being open, but it still shows on menu bar.

opcode81:
@BrunoKrugel `web_dashboard_open_on_launch` controls whether a window/browser tab with the dashboard is opened as soon as Serena starts. The second relevant configuration option is `web_dashboard_interface`, which controls the type of the interface.

v1.7 of Serena has changed the default dashboard interface on macOS to use the tray manager, which displays the icon for convenient access to dashboard instances. If you would prefer not to have the icon, you can set `web_dashboard_interface` to `browser`, which was the old default for macOS. In this mode, you will have to enter the URL manually if `web_dashboard_open_on_launch` is disabled.

opcode81:
Documentation is here: https://oraios.github.io/serena/02-usage/060_dashboard.html#the-serena-dashboard

Issue #1360, opened 2026-04-16
insert_after_symbol inserts inside multi-line constant/variable values (dict, list, etc.)


walkingpendulum:
## Problem

When `insert_after_symbol` is called on a top-level **constant or variable with a multi-line value** (e.g. a dict or list literal), the insertion lands **inside the value** — after the opening `{`/`[` — instead of after the closing `}`/`]`.

The root cause: `find_symbol` reports `body_location.start_line == body_location.end_line` for such constants (the LSP returns only the declaration line, not the full span of the value). `insert_after_symbol` then inserts after that single line, splitting the literal in half.

## Minimal reproduction

**Input file (`example.py`):**

```python
"""Minimal repro."""

SOME_CONFIG: dict[str, dict] = {
    "key_a": {
        "value": 1,
        "notes": "first entry",
    },
    "key_b": {
        "value": 2,
        "notes": "second entry",
    },
}


def some_function() -> str:
    return "hello"
```

**Step 1 — find_symbol reports a single-line body for a 10-line dict:**

```
find_symbol("SOME_CONFIG", relative_path="example.py", include_body=True)
→ body_location: {start_line: 2, end_line: 2}
  body: "SOME_CONFIG"
```

The dict spans lines 3–12 (1-indexed) but `end_line` equals `start_line`.

**Step 2 — insert_after_symbol inserts inside the dict:**

```
insert_after_symbol("SOME_CONFIG", "example.py", '\nNEW_CONSTANT = "inserted"\n')
→ OK
```

**Resulting file (broken):**

```python
"""Minimal repro."""

SOME_CONFIG: dict[str, dict] = {


NEW_CONSTANT = "inserted"
    "key_a": {
        "value": 1,
        "notes": "first entry",
    },
    "key_b": {
        "value": 2,
        "notes": "second entry",
    },
}


def some_function() -> str:
    return "hello"
```

`NEW_CONSTANT` is injected after the opening `{`, corrupting the dict literal.

## Expected behaviour

The inserted code should appear **after line 12** (the closing `}`), i.e.:

```python
SOME_CONFIG: dict[str, dict] = {
    ...
}

NEW_CONSTANT = "inserted"
```

## Environment

- Serena via Claude Code MCP
- Python project (pyright LSP)
- Confirmed with both `dict[str, dict]` and typed assignments (`_STATIC_TABLE_CATALOGUE: dict[str, dict] = { ... }`)

## Notes

[...]

PR #1399, opened 2026-04-22
fix(insert_after_symbol): handle multi-line constant/variable symbol locations

Will-hxw:
## Summary
- Detect when LSP reports Constant/Variable symbols with start_line == end_line and scan forward to find the actual closing quote for multi-line string literals

## Why
Issue #1360: LSP reports body start_line == end_line for multi-line constants (pointing to opening quote), causing insert_after_symbol to insert inside the constant instead of after it.

## Validation
- [x] Code review passed
- [x] Branch pushed to fork

### DIFF
--- a/src/serena/code_editor.py
+++ b/src/serena/code_editor.py
@@ -142,6 +142,33 @@ class CodeEditor(Generic[TSymbol], ABC):
 
         pos = symbol.get_body_end_position_or_raise()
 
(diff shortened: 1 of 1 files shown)


opcode81:
Sorry, @Will-hxw, but this is heuristic workaround, which we cannot merge.
@MischaPanch has already made some changes which should prevent this from being an issue.

PR #1403, opened 2026-04-23
Add JetBrainsDebugTool

opcode81:


src/serena/tools/jetbrains_tools.py
```diff
@@ -565,3 +565,94 @@ def apply(
...
+    Provides debugging functionality (run configs, breakpoints, stepping, inspection, and evaluation)
+    via a persistent debug REPL connected to the JetBrains IDE.
+    """
+
+    def apply(
+        self,
+        expression: str,
+        repl_key: str = "default",
+    ) -> str:
+        """
```
MischaPanch, line 581: We have to check that this works with Codex. It certainly won't work with chatgpt, but that's ok.

Another approach to the long tool description is to instead add a tool "get_instructions_on_debugging" with the description "If you were asked to debug, call this first before using the debugging tool".

In scenarios with non-deferred tool loading, the long description here unnecessarily bloats the context

opcode81: Now resolved by adding info tool with 1fd0fe2cb5b962060d98de4a3e759e0fbd2c4aec

Issue #1259, opened 2026-04-01
CPP language should include .m and .mm file extensions for Objective-C/Objective-C++ support


freelf:
## Problem

The `cpp` language's `FilenameMatcher` in `solidlsp/ls_config.py` only matches C/C++ extensions:

```python
case self.CPP | self.CPP_CCLS:
    return FilenameMatcher("*.cpp", "*.h", "*.hpp", "*.c", "*.hxx", "*.cc", "*.cxx")
```

This means `.m` (Objective-C) and `.mm` (Objective-C++) files are not associated with clangd, even though clangd fully supports these file types.

Currently `.m` is mapped to MATLAB, which makes Objective-C source files inaccessible through the `cpp` language server.

## Impact

For iOS/macOS projects, `get_symbols_overview`, `find_symbol`, and `find_referencing_symbols` all fail on `.m` and `.mm` files with:

```
ValueError - Cannot extract symbols from file xxx.m. Active languages: ['cpp']
```

While `.h` header files work fine (since they match `*.h`), the implementation files are completely excluded.

## Suggested Fix

Add `"*.m"` and `"*.mm"` to the CPP FilenameMatcher:

```python
case self.CPP | self.CPP_CCLS:
    return FilenameMatcher("*.cpp", "*.h", "*.hpp", "*.c", "*.hxx", "*.cc", "*.cxx", "*.m", "*.mm")
```

Note: `*.m` currently conflicts with MATLAB. A possible solution is to either:
1. Add `.m`/`.mm` to cpp and remove `.m` from MATLAB (breaking change for MATLAB users)
2. Allow configurable file extension overrides via `ls_specific_settings`
3. Add a dedicated `objc` language alias that maps to clangd with ObjC extensions

## Environment

- Serena version: 0.1.4
- Platform: macOS (ARM64)
- Use case: Cross-platform mobile SDK with C++, Objective-C++, and Kotlin code

MischaPanch:
Thx for reporting, @opcode81 and I will discuss how to best handle overlapping file extensions.

> Add a dedicated objc language alias that maps to clangd with ObjC extensions

As to this, could you elaborate what would be needed to enable ObjC extensions in Serena's clangd support? Does it also work with ccls?

freelf:
> Thx for reporting, [@opcode81](https://github.com/opcode81) and I will discuss how to best handle overlapping file extensions.
> 
> > Add a dedicated objc language alias that maps to clangd with ObjC extensions
> 
[...]

PR #1410, opened 2026-04-24
Add more cpp extensions

silvanshade:
This PR adds several missing extensions that `clangd` understands to `cpp`. This enables recognition of C++20 module interface files among other things.

Most of these extensions are taken directly from the `clang` driver frontend:

https://github.com/llvm/llvm-project/blob/08992737d23113d3b9a78974b87b1ba3e7ecb6cb/clang/lib/Driver/Types.cpp#L309

One exception is `*.ixx` which is the default extension Visual Studio uses for C++20 module interface files:

https://learn.microsoft.com/en-us/cpp/cpp/modules-cpp?view=msvc-170#single-partition-modules

Related: #1259

### DIFF
--- a/src/solidlsp/ls_config.py
+++ b/src/solidlsp/ls_config.py
@@ -232,7 +232,41 @@ class Language(str, Enum):
             case self.RUBY_SOLARGRAPH:
                 return FilenameMatcher("*.rb")
             case self.CPP | self.CPP_CCLS:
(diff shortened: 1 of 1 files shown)


silvanshade:
Closing this because I have a more comprehensive PR in the pipeline which addresses several other issues with the clangd integration.

PR #1413, opened 2026-04-25
Make claude code great again

MischaPanch:
I tested the proposed workaround and it seems to work. Might also work with less repetition or emphasis, but at this point, I'd rather leave it as proposed here.


MischaPanch:
@opcode81 We can add an entry to the changelog and a newsitem after the review is finished

opcode81:
Merged into dev-v1.2.0

Issue #1260, opened 2026-04-01
Improve monorepo support by separating memory and LSP context


arikon:
**Is your feature request related to a problem? Please describe.**  
Yes. Serena is hard to use in very large monorepos where many logically separate projects live in the same repository, but parts of the codebase are still connected through dependencies.

In this setup, I want LSPs to run from the monorepo root so navigation, references, and dependency resolution work across the whole repository. At the same time, I want `.serena/memory` to stay project-specific, so each logical project can have its own isolated memory/context.

Right now, `activate_project` switches both:
- the memory context, and
- the code/LSP root.

That makes it difficult to use Serena effectively in large monorepos, because switching to a project also narrows the LSP scope to that project root.

**Describe the solution you'd like**  
It would be great to decouple **memory context** from **code/LSP context**.

A few possible ways to support this:

- Add a configuration option that changes `activate_project` so it only switches memory, while keeping the current code/LSP root unchanged.
- Or add separate commands for switching only the memory context without affecting LSPs / code indexing scope.

This would make it possible to:
- run LSPs from the monorepo root,
- keep full cross-project code intelligence,
- and still have separate `.serena/memory` contexts for individual projects.

**Describe alternatives you've considered**  
The main alternative is using `activate_project` as-is, but that does not work well because it switches both memory and LSP context together.

Another option is to use a single shared memory for the whole monorepo, but in practice that becomes too noisy and mixes context from unrelated projects.

**Additional context**  
This is mainly relevant for very large enterprise monorepos containing many unrelated codebases in one repository.

In that kind of setup:
- project boundaries are logical, not repo-based,
- dependencies can span multiple projects,
- LSPs need repo-wide visibility,
- but memory should stay isolated per project.

A feature that allows switching memory independently from LSP/code context would make Serena much more usable in this scenario.

opcode81:
You can use the [query_project tool](https://oraios.github.io/serena/02-usage/040_workflow.html#reading-from-external-projects) to read memories from the subprojects, provided that the subprojects are also Serena projects already (which seems to be the case).
You would only need to instruct the LLM to specifically do this.

Would that be a good solution?

PS: Note that for reading memories, you do not need the project server/JetBrains IDE to be open for the other projects.

arikon:
@opcode81 I don't only want to read memories from different projects, but also to write them.
I want to open single monorepo workspace to work on many linked projects together, which have separate Serena memory banks, but LSP running from the root of monorepo.

[...]

PR #1418, opened 2026-04-26
Add cross-package TypeScript reference support via additional_workspace_folders

### LINKED ISSUE #750: Feature Request: search other libraries for symbols
Hi! thanks for your work on Serena, and thanks considering my request:

I have a package which contains clients (python) for
- many HTTP services that I need to talk to
- many SQL tables that I need to query

When I'm using Serena to develop that project, I can use natural language to describe what I want, and the LLM will write reasonably smart code which relies on the packaged clients to get the job done.  Hurray, nice job Serena.

I have more than one need for this ability, so I have made this a python package and paid special attention to ensuring the type info is made available.  This allows me to import it into other (smaller, simpler) projects and get tab completion re: the tables' columns, the http verbs, the http params, and the attributes on the response objects.  Hurray, nice job LSP.

But I can see that when I use Serena in those other projects, she can't see into the type information which I have prepared there.  I made a smaller example to test this claim:

```python
import pyfiglet
pyfiglet.print_figlet("hello world!", colors="MAGENTA")
```

`pyfiglet` supports a `font` kwarg on `print_figlet`.  It's easy for me to see because when I open my editor in the python virtual environment and place a comma before the `)`, a small tooltip appears:

```
(text: str, font: str = DEFAULT_FONT, colors: str = ":", **kwargs: Any) -> None
```

But if an agent instructs Serena to search for this information, Serena can't find it:

```
serena - search_for_pattern (MCP)(substring_pattern: "def print_figlet", restrict_search_to_code_files: false, context_lines_after: 10)
  ⎿  {
       "result": "{}"
     }
```

I'd like to be able to configure serena (perhaps via a config file or a cli param) such that she can search among my dependencies as if they were local code.  If this is a performance concern, then I'd like to specify only certain libraries for her to index.  

I usually launch the agent from inside of a python virtual environment, so I suspect that serena, as a subprocess thereof, will also have access to that environment, but just to be sure I've also configured the agent to explicitly place serena in such an environment:

```
$ uv run serena start-mcp-server --context ide-assistant --project $PWD
```

...still no luck.

The workflow I've settled on right now is to use serena in the project with the knowledge to write and tests an exemplar script which shows the various tables and API's that I need the tool to use, and then copy that script over into the project with the task and say "this script works, now incorporate it into the bigger picture here", but it's kind of clunky compared to the possibility that the tool-focuse agent might be able to just import the knowledge from the other project.

I suppose I could also just make it a monorepo but... yuck.

meirLixen:
Closes #750 (partially)
Related to #1260

## Summary

- Adds `additional_workspace_folders` setting in `ls_specific_settings["typescript"]` to enable cross-package `find_referencing_symbols` and `request_references` in monorepos
- Each listed folder is registered as an LSP workspace folder during initialization, and a representative `.ts` file is opened to trigger tsserver project loading
- Adds `_is_cross_workspace_path()` / `_resolve_file_uri()` helpers in the base `SolidLanguageServer` class to canonicalize URIs for paths with `..` segments, applied consistently across all LSP request methods

## Configuration

```yaml
# project.yml or project.local.yml
ls_specific_settings:
  typescript:
    additional_workspace_folders:
      - ../sibling-package
      - /absolute/path/to/other-package
```

Paths can be absolute or relative to the project root. Results are cached, deduplicated, and validated (non-existent paths are skipped with a warning, paths overlapping with the repo root are ignored).

## How it works

1. During TS language server initialization, additional folders are included in the LSP `workspaceFolders` array
2. After server startup, a representative `.ts` file from each folder is opened via `didOpen` to trigger tsserver project loading (tsserver only loads projects on-demand)
3. Files are registered in `open_file_buffers` with permanent ref_count to stay open for the server lifetime
4. Cross-workspace paths (`../other-package/src/file.ts`) are resolved to canonical URIs via `_resolve_file_uri()` to prevent mismatches between tsserver-returned URIs and Serena-computed URIs

## Test plan

- [x] `test_cross_package_find_references` — verifies `request_references` finds refs across packages
- [x] `test_cross_package_referencing_symbols` — verifies `request_referencing_symbols` resolves containing symbols across packages
- [x] `test_without_additional_workspace_no_cross_refs` — baseline confirming cross-refs don't appear without config
- [x] All 3 existing TypeScript tests pass (no regressions)
- [x] Edge cases handled: non-existent paths, duplicates, overlap with repo root, robust path component `..` detection


MischaPanch (commented):
Thanks for the PR. Is there a reason why we can't do that centrally in the base class, such that it works for all language servers?

meirLixen (commented):
Good question! The URI normalization (`_is_cross_workspace_path` / `_resolve_file_uri`) is already in the base `SolidLanguageServer` class and works for all language servers.

The TS-specific part is the `additional_workspace_folders` configuration and the file-opening logic in `_open_additional_workspace_files`. This is TS-specific because:

1. **LSP workspace folder initialization differs per server** — each language server subclass builds its own `_get_initialize_params()`, and not all language servers support multi-root workspaces the same way
2. **tsserver requires explicit file opening** to load projects — other language servers (e.g. Java/jdtls, Rust/rust-analyzer) may handle project discovery differently
3. **The representative file finder** (`_find_representative_ts_file`) looks for `.ts`/`.tsx` files near `tsconfig.json`

That said, moving the workspace folder resolution to the base class (with per-language override hooks for the file-opening strategy) would be a natural next step if other language servers need this. Happy to refactor if you'd prefer that approach.

MischaPanch:
Yes, pls refactor this to be future proof, even if the implementation will only work for typescript now. Pls change the logic such that the user can always provide the config in a uniform way, so we have a fixed API for users for providing additional workspaces. Then add methods in the base class where needed which raise if this option is not supported/implemented yet, and override the methods accordingly in typescript. Alternatively, you could task an agent with actually implementing and testing it for the most important LS.

You could investigate whether it makes sense to post-process the initialization params and include the workspaces in a default fashion directly in the base class, overriding the default behavior where needed. But I don't know the best way to handle this, so I'll leave proposing the most suitable design to you at this stage - so this is just a suggestion to consider. @opcode81 and might go back to proposing a different structure after your initial proposal, but we don't have time to dive into it ourselves now.

Since this is a significant extension, it should be thoroughly documented in our docs and in the changelog as well. Thanks for laying the groundwork for this!

MischaPanch:
@meirLixen just FYI - the JetBrains variant of Serena has first-class support for all external dependencies out of the box

meirLixen:
Refactored as requested in c1a72a7c:

- `additional_workspace_folders` is now a **top-level project.yml setting** (uniform API for all languages)
- Flows through `ProjectConfig` → `LanguageServerFactory` → `SolidLSPSettings` → `SolidLanguageServer`
- Base class provides: `_resolve_additional_workspace_folders()`, `_build_workspace_folders_param()`, `_activate_additional_workspaces()`
- Subclasses override `_find_representative_source_file()` (raises `NotImplementedError` by default) and optional indexing hooks (`_signal_expect_indexing`, `_wait_for_additional_workspace_indexing`)
- TypeScript overrides just these three methods
- Added entry to `CHANGELOG.md` and documented in `project.template.yml`

Adding support for another language server would only require overriding `_find_representative_source_file()` and calling `self._build_workspace_folders_param()` in `_get_initialize_params()`.

MischaPanch:
Thanks, overall looks good! I had a rather superficial look though. Will resolve conflicts and possibly make minor adjustments after #1438 is merged

MischaPanch:
@meirLixen I improved some things in the implementation and docs, squashed your commits and merged main into this branch resolving conflicts. I force-pushed over your commits (due to the squashing and rewording).

Once CI runs through, I'll merge this

PR #1422, opened 2026-04-26
Improve clangd log level classification

silvanshade:
This PR improves the log record classification for clangd by relying on the actual diagnostic codes that clangd produces rather than scanning for keywords.

Fixes a false positive where a log record could be classified as `ERROR` if it contained `-fno-exceptions` or similar.


MischaPanch:
Thanks, merging

PR #1423, opened 2026-04-26
Add more extensions for clangd and ccls

silvanshade:
This is a follow up to #1410.

This PR registers several more extensions for `clangd` and `ccls`.

Most of these extensions are taken directly from the `clang` driver frontend source:

https://github.com/llvm/llvm-project/blob/08992737d23113d3b9a78974b87b1ba3e7ecb6cb/clang/lib/Driver/Types.cpp#L309

The one exception is `*.ixx`. This is the default extension Visual Studio uses for C++20 module interface files:

https://learn.microsoft.com/en-us/cpp/cpp/modules-cpp?view=msvc-170#single-partition-modules

The clang source doesn't reference the extension but `clangd` does read the files without issue.

The additional extensions add support for the following:

| language |`clangd` | `ccls` |
|----------|---------|--------|
| Objective-C | :white_check_mark: | :white_check_mark: |
| C++20 module interface files | :white_check_mark: | :x: |
| CUDA | :white_check_mark: | :x: |
| HIP | :white_check_mark: | :x: |
| OpenCL | :white_check_mark: | :x: |

I opted not to associate C++20 module interface files with `ccls` since it sounds like that isn't actually supported: https://github.com/MaskRay/ccls/issues/798 If enabling it in the tests, it does seem to return symbols for the very simple case we have in the test repo though.

Same story for CUDA, HIP, and OpenCL. However, in those cases `ccls` also seems to be returning keywords (e.g, `__global`) as symbols, suggesting it doesn't properly recognize the AST and likely the results would not be reliable.

Related: #1259

### DIFF
--- a/src/solidlsp/ls_config.py
+++ b/src/solidlsp/ls_config.py
@@ -13,13 +13,21 @@ if TYPE_CHECKING:
 
 
 class FilenameMatcher:
-    def __init__(self, *patterns: str) -> None:
+    patterns: list[str]
+
+    def __init__(self, *patterns: str, insensitive: bool = False) -> None:
         """
         :param patterns: fnmatch-compatible patterns
         """
-        self.patterns = patterns
+        self.insensitive = insensitive
+        # Pre-lowercase patterns if we are in insensitive mode
+        if self.insensitive:
+            self.patterns = [p.lower() for p in patterns]
+        else:
+            self.patterns = list[str](patterns)
 
     def is_relevant_filename(self, fn: str) -> bool:
+        fn = fn.lower() if self.insensitive else fn
         for pattern in self.patterns:
             if fnmatch.fnmatch(fn, pattern):
                 return True
@@ -272,6 +280,7 @@ class Language(str, Enum):
                     # OpenCL
                     "*.cl",
                     "*.clcpp",
+                    insensitive=True,
                 )
             case self.CPP_CCLS:
                 # From llvm-project/clang/lib/Driver/Types.cpp types::lookupTypeForExtension:
@@ -296,6 +305,7 @@ class Language(str, Enum):
                     # Objective-C
                     "*.m",
                     "*.mm",
+                    insensitive=True,
                 )
             case self.KOTLIN:
                 return FilenameMatcher("*.kt", "*.kts")
@@ -352,9 +362,7 @@ class Language(str, Enum):
             case self.JULIA:
                 return FilenameMatcher("*.jl")
             case self.FORTRAN:
-                return FilenameMatcher(
-                    "*.f90", "*.F90", "*.f95", "*.F95", "*.f03", "*.F03", "*.f08", "*.F08", "*.f", "*.F", "*.for", "*.FOR", "*.fpp", "*.FPP"
-                )
+                return FilenameMatcher("*.f90", "*.f95", "*.f03", "*.f08", "*.f", "*.for", "*.fpp", insensitive=True)
             case self.HASKELL:
                 return FilenameMatcher("*.hs", "*.lhs")
             case self.HAXE:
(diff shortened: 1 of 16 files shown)


MischaPanch:
Thanks for the PR! ccls tests on windows fail and then hang forever, I will review once tests run through in CI

```
test/solidlsp/cpp/test_cpp_basic.py::TestCppLanguageServer::test_find_symbol[cpp_ccls] FAILED [ 51%]
```

silvanshade:
> Thanks for the PR! ccls tests on windows fail and then hang forever, I will review once tests run through in CI
> 
> ```
> test/solidlsp/cpp/test_cpp_basic.py::TestCppLanguageServer::test_find_symbol[cpp_ccls] FAILED [ 51%]
> ```

@MischaPanch 

The windows CI installs `ccls` version `0.20220729` which is quote old and corresponds to LLVM-15. For whatever reason, that version crashes when launched in the test directory.

I didn't modify `test_find_symbol` and the crash still happens if I revert the changes to the associated extensions. That suggests the crashing behavior is caused just by `ccls` initializing in the directory with those files.

I didn't debug further since I don't see much point in trying to track down the error in the old version.

Building a recent `ccls` binary locally is an option but that would add both a lot of complexity to the configuration and a lot of build time overhead to the Windows test. You'd need to check out the `llvm-project` source locally, build Clang+LLVM first, then build `ccls`. Doesn't seem worth all the hassle.

Simplest fix seems to be just to disable the `ccls` test on Windows, so that's what I opted for.

MischaPanch:
Thanks, will merge after #1438 is merged (since conflicts will need to be resolved)

silvanshade:
Going to close this since I don't use Serena anymore and don't plan to fix the merge conflicts. Feel free to revive in a local branch though.

MischaPanch:
Ok, thanks for the update. Could you tell us why you don't use Serena anymore? Such feedback helps is a lot in improving the project

silvanshade:
> Ok, thanks for the update. Could you tell us why you don't use Serena anymore? Such feedback helps is a lot in improving the project

@MischaPanch sure. First I will just say thanks for the project. I did find it useful and I think it's good that it exists.

Main reason I am not using it now is because I've switched to pi-coding-agent harness and have less need for MCP.

Beyond that, there were already some things that made me have doubts about continuing to use Serena though:

1. CI is always randomly broken. This makes it hard to contribute to the project.
2. CI takes way too long. ~60+ min per run also makes it hard to contribute. It should be possible with to get that down to 3-5 min tops with aggressive caching strategies and/or container usage. Could be implemented in less than a day.
3. Project seems more focused on supporting every language server rather than ensuring the major ones are supported as well as possible. Example: ccls vs. clangd. Current codebase actually only tests ccls in some cases. Seems odd because ccls is rarely used in practice, hasn't been actively maintained for a long time (meaning support will get worse over time as C++ evolves), and is hard to even find packaged releases for.
4. Still not sure why this wasn't merged earlier. The ccls failure on Windows I explained awhile back. Could have just merged it back then.

MischaPanch:
Thanks for the explanations, I understand your position. 

The PR was not merged earlier because we were working on correcting technical debt that I knew would have collisions with this. The merge was not postponed due to CI failures, but ofc if changes introduce new failures, it is up to the PR author to correct them. 

Local testing is much faster and can be restricted to the language one is interested in through pytest markers, it's not necessary to wait for CI. There is work from the community right now on improving CI. It was not a priority for the core team because many other things were important and CI worked, even though it is very slow.

Serena's main goal is not supporting all possible LS, but the improvements to the language support generally come from the community. We don't write C/C++, so we relied on external contributions to improve the support. Maybe somewhat surprisingly, there were very few - far fewer than for ruby, C#, Java or other languages. So C/C++ support was lagging behind. Ofc you closing the PR doesn't exactly help (not trying to push anything, just an observation).

I might find time to pick this up later. I don't think that waiting a week for a merge is preposterous though.

MischaPanch:
Now merged this as part of #1451

`pytest -m cpp` runs through locally in a few seconds

PR #1431, opened 2026-04-28
fix(hooks): detect grep/read shell commands in Codex exec_command

### LINKED ISSUE #1394: `serena-hooks remind --client=codex` misses grep/read workflows inside `exec_command`
# `serena-hooks remind --client=codex` misses grep/read workflows inside `exec_command`

`serena-hooks remind --client=codex` classifies pre-tool calls by `tool_name` / `toolName`. In Codex, shell work arrives as the generic tool name `exec_command`, with the actual command under `tool_input.cmd`.

That means common Codex exploration like this is invisible to the `remind` counter:

```json
{
  "session_id": "demo",
  "tool_name": "exec_command",
  "tool_input": { "cmd": "rg \"SomeSymbol\" packages/server" }
}
```

Same for file-read style shell commands:

```json
{
  "session_id": "demo",
  "tool_name": "exec_command",
  "tool_input": { "cmd": "sed -n '1,220p' packages/server/foo.ts" }
}
```

Current relevant matcher behavior:

- `is_grep_tool()`: for non-Claude clients, checks whether `"grep" in tool_name`
- `is_read_file_tool()`: for non-Claude clients, checks whether `"file" in tool_name` and one of `read/view/open/show` is also in `tool_name`

Since the tool name is only `exec_command`, neither path fires.

I am not sure Serena should parse arbitrary shell commands. This may be better solved by Codex exposing richer structured tool-call metadata to hooks, for example a shell command category or parsed command segments. But as-is, the Codex `remind` hook can run successfully before every tool call while missing the main grep/read behavior it is meant to detect.

Environment:

- Serena: `serena-agent 1.1.2`
- Client: Codex
- Hook: `serena-hooks remind --client=codex`

Written with assistance from Codex.

octo-patch:
Fixes #1394

## Problem

`serena-hooks remind --client=codex` never counted Codex grep/read workflows because the Codex client sends all shell work as `exec_command` (with the actual command under `tool_input.cmd`). The previous detection checked only the tool name string, so `is_grep_tool()` and `is_read_file_tool()` always returned `False` for Codex — leaving the remind hook entirely blind to Codex exploration patterns.

## Solution

- Added `_GREP_SHELL_COMMANDS` and `_READ_SHELL_COMMANDS` frozensets listing common grep-like (`grep`, `rg`, `ag`, `ack`, `fgrep`, `egrep`) and read-file-like (`cat`, `head`, `tail`, `sed`, `less`, `more`, `bat`) shell commands.
- Added `_get_exec_command_name()` helper that extracts the basename of the first token from `tool_input.cmd`, handling both bare names (`rg`) and path-prefixed invocations (`/usr/bin/grep`).
- Updated `is_grep_tool()` and `is_read_file_tool()` to use these when the client is `HookClient.CODEX` and the tool name is `exec_command`; all other client paths are unchanged.

## Testing

Added `TestCodexExecCommandDetection` class with 13 tests covering:
- Positive cases: `rg`, `grep`, path-prefixed `grep`, `ag` → `is_grep_tool()`
- Positive cases: `sed`, `cat`, `head`, `tail` → `is_read_file_tool()`
- Negative cases: `ls`, `echo` → neither
- Non-`exec_command` tools for Codex still fall through to substring matching
- End-to-end: repeated `exec_command` + `rg` / `sed` calls trip the deny after threshold

All 13 new tests pass. The 2 pre-existing test failures (`test_outputs_activation_message`, `test_activate_command`) were already present on `main` due to an unrelated activation message text change.

### DIFF
--- a/src/serena/hooks.py
+++ b/src/serena/hooks.py
@@ -243,18 +243,46 @@ class PreToolUseRemindAboutSerenaHook(PreToolUseHook):
     #: are caught alongside ``read_file``, while ``write_file``/``edit_file`` are not.
     _READ_FILE_VERB_SUBSTRINGS: tuple[str, ...] = ("read", "view", "open", "show")
 
+    #: Shell commands that perform grep-like search; used to classify Codex
+    #: ``exec_command`` calls whose ``cmd`` field starts with one of these.
+    _GREP_SHELL_COMMANDS: frozenset[str] = frozenset(("grep", "rg", "ag", "ack", "fgrep", "egrep"))
+
+    #: Shell commands that perform file-read operations; used to classify Codex
+    #: ``exec_command`` calls whose ``cmd`` field starts with one of these.
+    _READ_SHELL_COMMANDS: frozenset[str] = frozenset(("cat", "head", "tail", "sed", "less", "more", "bat"))
+
     def __init__(self, client: HookClient):
         super().__init__(client)
         self._tool_call_counter = self.ToolUseCounter.load(self)
 
+    def _get_exec_command_name(self) -> str:
+        """Extract the command name from an ``exec_command`` tool input.
+
+        Returns the basename of the first whitespace-separated token of the
+        ``cmd`` field, lowercased, so that both bare names (``rg``) and
+        path-prefixed invocations (``/usr/bin/grep``) are normalised the same
+        way.  Returns an empty string when the input is absent or empty.
+        """
+        if self._tool_input is None:
+            return ""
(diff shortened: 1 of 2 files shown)


MischaPanch:
Thanks for the PR. We don't need the new tests, the current logic is very simple. Pls remove them, format the code and add an entry to the changelog. Then this can be merged

MischaPanch:
Superseded by #1447 , I built off your commit

PR #1433, opened 2026-04-28
Features: diagnostics and LS extensions

MischaPanch:
Adds support for diagnostics (JB and LSP); find_implementation and find_definition (LSP)


src/serena/tools/symbol_tools.py
```diff
@@ -324,7 +392,386 @@ def make_summary() -> str:
...
+            for s, s_dict in zip(implementing_symbols, symbol_dicts, strict=True):
+                if symbol_info := info_by_symbol.get(s):
+                    s_dict["info"] = symbol_info
+                    s_dict.pop("name", None)  # name is included in the info
+
+        result = self._to_json(symbol_dicts)
+        return self._limit_length(result, max_answer_chars)
+
+
+class FindDefiningSymbolAtLocationTool(Tool, ToolMarkerSymbolicRead, ToolMarkerOptional):
```
opcode81: Is there a point to this tool if we already have FindDefiningSymbolTool?

src/serena/tools/symbol_tools.py
```diff
@@ -324,7 +392,386 @@ def make_summary() -> str:
...
+            result = self._to_json(None)
+            return self._limit_length(result, max_answer_chars)
+
+        symbol_dict = self._defining_symbol_to_result_dict(symbol_retriever, defining_symbol, include_body, include_info)
+        assert symbol_dict is not None
+        result = self._to_json(symbol_dict)
+        return self._limit_length(result, max_answer_chars)
+
+
+class FindDefiningSymbolTool(Tool, ToolMarkerSymbolicRead):
```
opcode81: The implementation of this tool is rather complicated. I think it should be based on the same helper as `JetBrainsFindDeclarationTool`, which already does all of the heavy lifting.
And we should probably rename it to `FindDeclarationTool` or `FindSymbolDeclarationTool`.
wdyt?

src/serena/tools/symbol_tools.py
```diff
@@ -4,22 +4,90 @@
...
+        line_start_offsets.append(match.end())
+
+    if line >= len(line_start_offsets):
+        raise ValueError(f"Line out of range: {line}")
+
+    line_start_offset = line_start_offsets[line]
+    line_end_offset = line_start_offsets[line + 1] - 1 if line + 1 < len(line_start_offsets) else len(text)
+    if column > line_end_offset - line_start_offset:
+        raise ValueError(f"Column out of range for line {line}: {column}")
+    return line_start_offset + column
```
opcode81: We should probably move these helpers to the new module `ls_diagnostics` and encapsulate them in an abstraction. Some of the functionality is probably already duplicated - and this is partly true for the implementation of the diagnostics tools as well. 

We need to consolidate this.

src/serena/tools/symbol_tools.py
MischaPanch: The idea was that the user might want to switch to a position-based detection, but I don't think it's worth it now. The tool can be deleted

src/serena/tools/symbol_tools.py
MischaPanch: Yes, the implementation predates the work you did for the Jetbrains variant on main. I agre to rename to FindDeclaration and to use the same helper as in the JB tool

src/serena/tools/symbol_tools.py
MischaPanch: Yes, moving there sounds good. You can probably task an agent with deduplication and refactoring, the new tests and the demo script should allow it to perform this without breaking functionality

opcode81: Duplication of diagnostics-based functionality is now fixed in cdf2481410ff1292c00ff03d13259126ae893727 by introducing new abstraction `GroupedDiagnostics`. All of the functions were duplicates.

src/serena/tools/symbol_tools.py
opcode81: I removed it

src/serena/tools/symbol_tools.py
opcode81: I consolidated this with f6396f8d7a68f0af0ccdb25356e81c64a876840b

PR #1437, opened 2026-04-29
Revamp mode configuration

opcode81:
See change log

### DIFF
--- a/src/serena/resources/serena_config.template.yml
+++ b/src/serena/resources/serena_config.template.yml
@@ -37,13 +37,15 @@ gui_log_window: False
 # Further information: https://oraios.github.io/serena/02-usage/060_dashboard.html
 web_dashboard: True
 
-# whether to open the Dashboard window/browser tab when Serena starts (provided that web_dashboard is enabled).
-# If set to false, you can still open the dashboard manually by clicking on the Serena icon in your system
-# tray on Windows and macOS. On Linux, there is no system tray support, so you can only open the dashboard by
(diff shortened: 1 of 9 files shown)


src/serena/resources/serena_config.template.yml
```diff
@@ -112,18 +112,22 @@ included_optional_tools: []
 # This cannot be combined with non-empty excluded_tools or included_optional_tools.
 fixed_tools: []
 
-# list of mode names to that are always to be included in the set of active modes
-# The full set of modes to be activated is base_modes + default_modes.
-# If this is undefined, no base modes are included.
-# The project configuration (project.yml) may override this setting.
+# list of mode names to that are always to be included in the set of active modes.
+# The full set of modes to be activated is base_modes + default_modes + added_modes,
+# where added_modes can be defined by projects/CLI parameters.
```
MischaPanch, line 120:
```suggestion
# where default_modes can be overwritten by projects/CLI parameters. and added_modes can be defined there
```

MischaPanch (approved):
Nice! I suggested a minor change in wording, feel free to merge without applying it if you disagree

src/serena/resources/serena_config.template.yml
opcode81: I don't think this needs to be repeated in the comment on `base_modes`.
`default_modes` appears directly below `base_modes` in the file, and the associated comment says it already:

> These modes can be overridden by the project configuration (project.yml) or through the CLI (--mode).

PR #1439, opened 2026-04-30
feat: Add Angular, HTML, and SCSS / Sass language servers

paul-basanets:
Four new language keys, all opt-in via `project.yml` (none auto-detected).

| key       | LS                                                              |
| --------- | --------------------------------------------------------------- |
| `angular` | `@angular/language-server` + `typescript-language-server` (with `@angular/language-service` plugin) + `vscode-html-language-server` |
| `html`    | `vscode-html-language-server` (`vscode-langservers-extracted`)  |
| `css`     | `vscode-css-language-server` (`vscode-langservers-extracted`)   |
| `scss`    | `some-sass-language-server` (`.scss` and `.sass`)               |

### Implementation

**Angular** runs three child processes:

* `ngserver` — `.html` definition / references / hover (templates know component types).
* Companion `typescript-language-server` with `@angular/language-service` loaded as a tsserver plugin — `.ts` document symbols, definition, hover, rename, references that span templates.
* Companion `vscode-html-language-server` — `.html` `documentSymbol` only, because `ngserver` returns `-32601` for that method on every `.html` file.

Routing per `(operation, extension)` lives in the module docstring of `angular_language_server.py`. The companions are spawned before `ngserver`'s `initialize`, and the start path tears them down on init failure so we don't leak Node processes.

### Key decisions

* **Some Sass for `scss`, not the generic CSS LS.** The CSS LS technically parses SCSS, but Some Sass is the only LSP with full `@use` / `@forward` workspace navigation (cross-file go-to-definition / find-references for variables, mixins, functions).
* **`.sass` opens with language id `"sass"`, not `"scss"`.** The indented Sass syntax is a different parser; sharing the id silently breaks parsing.
* **HTML LS as a tertiary Angular companion.** `ngserver` does not implement `textDocument/documentSymbol`. Pulling in the HTML LS gives users a structural element outline on plain HTML *and* Angular templates without cluttering `Language.ANGULAR.get_source_fn_matcher()` with non-HTML files.
* **Don't subsume SCSS under Angular.** Angular projects routinely use SCSS, but mixing both LSPs into one matcher would force users into a worse SCSS experience. SCSS stays a separate language key.
* **Companion deps overridden via instance state, not class attribute.** The previous Vue pattern (set class attr → `super().__init__` → reset) raced when two `LanguageServer.create` calls overlapped. Both Vue and Angular now pass the resolved TS-LS executable through an instance-level `_explicit_ts_ls_executable`, dispatched via a per-instance `_create_dependency_provider`. Vue refactor is in the same branch.
* **Monorepo-aware `@angular/core` probe.** Walking up from `repository_root_path` until a `node_modules/@angular/core` shows up — covers Nx / yarn / pnpm workspaces. Stops at filesystem root, mount-point change, or a `package.json` declaring `"workspaces"` (the workspace root is the canonical install location).
* **CSS and HTML LS install into separate directories** (`vscode-langservers-html/`, `vscode-langservers-css/`) even though both default to the same npm package. Otherwise a user override of `vscode_langservers_package` on one would invalidate the other's `.installed_version` and trigger re-installs in a loop.
* **Angular LS does not support `ls_path`.** The path-override hook assumes a single binary; ngserver is a multi-process orchestration where the layout matters. Versions of every component are still overridable via `ls_specific_settings.angular`.

### Notable details

* **Tri-server cleanup invariant.** If anything between `self.server.start()` and `projectLoadingFinish` fails, both companion processes plus the partially-started `ngserver` are torn down before the exception propagates. Regression test in `test_angular_error_cases.py::TestAngularStartupCleanup`.
* **`@angular/core` is required *in the project*, not just installed by Serena.** Without it, `ngserver` reports every file as "not in an Angular project" and template-aware features silently return empty. Logged loudly at startup, but not fatal.
* **Test fixture installs npm deps once per checkout.** The Angular tests need real `@angular/core`; the conftest runs `npm install` in the test repo on first use, serialised across pytest-xdist workers via `filelock`. Lock file gitignored.
* **CI**: Angular `npm install` runs as a job step before pytest so language-server cache + npm install can happen in parallel where possible.
* **Cross-file references on `.ts` symbols are routed through `ngserver`, not the TS companion.** On real Angular projects, `typescript-language-server` alone misses template usages and under-reports cross-file `.ts` references unless every file is pre-opened. `ngserver` aggregates both; this is the only request type that bypasses the TS companion.

### Caveats

* All four languages are marked experimental — must be explicitly listed in `project.yml`, never auto-detected. Priority is `0` (lowest), so adding them won't override TypeScript / HTML / CSS auto-routing for non-Angular projects.
* HTML cross-file definition / references aren't meaningful and aren't exposed.
* `@import` navigation in CSS is limited (use `scss` if you need cross-file).

---------

P.S. I know it's a bit to much 4 langs in one PR but i wrote this for my project and was using it all together. If needed i can split it into 3 PRs - CSS, SCSS / Sass and Angular + HTML.


src/solidlsp/language_servers/some_sass_language_server.py
```diff
@@ -0,0 +1,191 @@
...
+
+from solidlsp.language_servers.common import RuntimeDependency, RuntimeDependencyCollection, build_npm_install_command
+from solidlsp.ls import LanguageServerDependencyProvider, LanguageServerDependencyProviderSinglePath, SolidLanguageServer
+from solidlsp.ls_config import LanguageServerConfig
+from solidlsp.lsp_protocol_handler.lsp_types import InitializeParams
+from solidlsp.settings import SolidLSPSettings
+
+log = logging.getLogger(__name__)
+
+DEFAULT_PACKAGE_VERSION = "2.3.8"
```
MischaPanch, line 42: pls include all versions in the installation paths (across all new LS)

test/solidlsp/angular/conftest.py
```diff
@@ -0,0 +1,80 @@
...
+NODE_MODULES = REPO_ROOT / "node_modules"
+ANGULAR_CORE_MARKER = NODE_MODULES / "@angular" / "core" / "package.json"
+# Lock file lives inside the repo (covered by its .gitignore) so xdist workers
+# can serialise on the install. Putting it elsewhere risks placing it on a
+# different filesystem from the install dir, which makes flock semantics fuzzy.
+INSTALL_LOCK = REPO_ROOT / ".angular-install.lock"
+
+
+@pytest.fixture(scope="session", autouse=True)
+def _install_angular_test_repo_node_modules() -> None:
```
MischaPanch, line 42: this shouldn't be needed, the LS should be autoinstalled on first start, right?

test/solidlsp/angular/test_angular_basic.py
```diff
@@ -0,0 +1,318 @@
...
+        component = next((s for s in all_symbols if s.get("name") == "AppComponent"), None)
+        assert component is not None, "AppComponent class symbol not found"
+
+        children = component.get("children") or []
+        child_names = {c.get("name") for c in children}
+        for expected in ("greeting", "setName", "title", "userName", "items"):
+            assert expected in child_names, f"Expected '{expected}' as child of AppComponent, got: {child_names}"
+
+    @pytest.mark.parametrize("language_server", [Language.ANGULAR], indirect=True)
+    def test_pipe_class_in_symbol_tree(self, language_server: SolidLanguageServer) -> None:
```
MischaPanch, line 310: pls add tests for diagnostics for all new LS. We recently added this capability and corresponding tools, see existing tests for diagnostics on main

test/solidlsp/conftest.py
```diff
@@ -1,10 +1,36 @@
+from pathlib import Path
+
 from solidlsp import SolidLanguageServer
 from solidlsp.ls_config import Language
 from solidlsp.ls_types import SymbolKind, UnifiedSymbolInformation
 
 PYTHON_BACKEND_LANGUAGES = [Language.PYTHON, Language.PYTHON_TY]
 
 
+def find_in_file(language_server: SolidLanguageServer, relative_path: str, needle: str, occurrence: int = 0) -> tuple[int, int]:
```
MischaPanch: we already have a util `find_text_coordinates`, pls use that instead

docs/01-about/020_programming-languages.md
```diff
@@ -29,6 +29,19 @@ We currently provide direct, out-of-the-box support for the programming language
 Some languages require additional installations or setup steps, as noted.
 
 * **AL**
+* **Angular**  
+  (experimental; requires Node.js and npm; automatically installs `@angular/language-server`,
```
MischaPanch: this is too detailed for user-facing docs, pls shorten it

docs/01-about/020_programming-languages.md
```diff
@@ -47,6 +60,12 @@ Some languages require additional installations or setup steps, as noted.
 * **Crystal**  
   (requires [Crystalline](https://github.com/elbywan/crystalline) language server to be installed and available on PATH;
   note: Crystalline has limited go-to-definition support and does not support find-references)
+* **CSS**
+  (experimental; requires Node.js and npm; automatically installs `vscode-langservers-extracted`;
+  must be explicitly specified in the `languages` entry in the `project.yml`;
+  uses `vscode-css-language-server` from Microsoft;
+  reports rules/selectors as document symbols; cross-file `@import` navigation is limited;
+  for SCSS/Sass, use the `scss` language instead)
```
MischaPanch: pls shorten. Can't a user just always use sass? What's the downside to doing that?

MischaPanch (changes requested):
Thanks for the PR! While the changes are indeed large, we can merge this in once, you don't need to split it apart. The tests overall look good and we can add this functionality.

We recently merged significant extensions to Serena, pls have a look at them and incorporate the missing parts if needed. I made comments in most places to help you guide through them, but roughly you need to make sure:

1. For all downloaded dependencies the version is part of the path. SHA verification is done for the default version where possible/useful. Users can provide versions and if a version is found it's not re-downloaded. (I think you have most of the logic already except the version as part of the path)
2. We added support for diagnostics, pls add basic tests for this like already done for the other LS.
3. We added support for finding implementations. Many LS don't support this endpoint, those which do are explicitly marked. If any of the LS you added support finding implementations, mark them as a such and add a corresponding test.

I did a shallow review, focusing on the test and download logic. Those parts seem good, so after the comments are addressed and after a more thorough pass, this can be merged

MischaPanch:
Should be merged after #1447 which introduces breaking changes to FilenameMatcher

docs/01-about/020_programming-languages.md
paul-basanets: Shortened. Also it was a good catch, i've checked and some-sass-language-server can be used for .css also, so i dropped separate CSS server

test/solidlsp/angular/test_angular_basic.py
paul-basanets: done

test/solidlsp/angular/conftest.py
paul-basanets: 1. The language server itself (ngserver + tsserver companion + @angular/language-service + typescript) is fully auto-installed by RuntimeDependencyCollection inside AngularLanguageServer — into Serena's managed ls_resources_dir, exactly like every other LS.                                                                                                                                                                                                       
2. The fixture project's runtime deps (@angular/core) need to be resolvable from the workspace root. ngserver explicitly looks for @angular/core in the workspace's node_modules — without it, every file gets reported as  "not in an Angular project" and template features silently return empty results. For Angular project it's expected to have @angular/core installed but for tests we have to take care about it, so the conftest runs npm install once per checkout, cached across sessions and serialised across xdist workers via filelock.

src/solidlsp/language_servers/some_sass_language_server.py
paul-basanets: done

docs/01-about/020_programming-languages.md
paul-basanets: Done, hope it's better now.

test/solidlsp/conftest.py
paul-basanets: Updated

paul-basanets:
1. Downloaded deps versions are included to the corresponding pathes now.
2. and 3. Done - both diagnostics and find implementations are supported. Find implementations - only for Angular projects, not HTML or SASS/SCSS.

src/solidlsp/language_servers/angular_language_server.py
```diff
@@ -0,0 +1,721 @@
...
+
+log = logging.getLogger(__name__)
+
+# Version pinning convention (see eclipse_jdtls.py for the full spec):
+#   INITIAL_* — frozen forever; legacy unversioned install dir is reserved for it.
+#   DEFAULT_* — bumped on upgrades; goes into a versioned subdir.
+# Angular installs four interdependent npm packages into a single ``node_modules`` (npm
+# hoists them so ngserver's plugin resolution works); the install-dir name therefore
+# encodes all four versions, so overriding any single one routes to a fresh subdir.
+INITIAL_ANGULAR_LANGUAGE_SERVER_VERSION = "21.2.10"
```
MischaPanch: since this is a new LS, we don't need the initial versions (true for all new LS). These were just to correct for an earlier mistake

test/solidlsp/angular/test_angular_basic.py
```diff
@@ -0,0 +1,348 @@
...
+    @pytest.mark.parametrize("language_server", [Language.ANGULAR], indirect=True)
+    def test_find_implementations_of_interface_method(self, language_server: SolidLanguageServer) -> None:
+        """``Greeter.greet`` is implemented by ``GreetingService``; ``request_implementation``
+        invoked at the interface declaration must point at the service.
+
+        Routed through the Angular LS' companion typescript-language-server (with the
+        @angular/language-service plugin loaded as a tsserver plugin) — bare ngserver
+        does not own ``textDocument/implementation``.
+        """
+        if not language_has_verified_implementation_support(Language.ANGULAR):
```
MischaPanch: pls never use this pattern - either it works or it doesn't, no conditional skips

MischaPanch (changes requested):
Nice work! Just two minor changes needed, then this can be merged

PR #1440, opened 2026-04-30
Minor dashboard UI polish (typography, palette, modal cleanup)

paul-basanets:
- Switch UI font to Inter, keep JetBrains Mono only for code/logs/paths/ identifiers (loaded via Google Fonts; system fallbacks)
- Refine the light/dark palette with softer borders, clearer text hierarchy, and a more nuanced shadow/elevation system; introduce --bg-elevated, --border-strong, --shadow-elevated tokens; retain the orange accent
- Introduce a consistent spacing scale (--space-1..6) and apply it across sections, modals, and controls
- Extract shared modal CSS classes (.modal-info, .modal-hint, .modal-prompt, .modal-field, .modal-input, .modal-select, .modal-textarea, .modal-actions, .btn-secondary) and strip duplicated inline styles from all seven modals; add an accent-colored focus ring for inputs/textareas and a subtle backdrop blur
- Default-expand the "Registered Projects" sidebar so the right column is not visually empty on first load
- Replace the native <select> in Add Language with a filterable combobox: type to filter, keyboard navigation (Up/Down/Enter/Esc), substring highlight, click-outside to close; the typed value is validated against the available languages list before submission
- Update CHANGELOG accordingly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 Status of the `main` branch. Changes prior to the next official version change will appear here.
 
+* Dashboard:
+  - UI polish: switch UI font to Inter (with system fallbacks) and use JetBrains Mono only for code/logs/paths/identifiers; refine the light/dark palette with softer borders, clearer text hierarchy, and a more nuanced shadow/elevation system; introduce a consistent spacing scale; keep the orange accent.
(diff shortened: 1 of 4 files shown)


opcode81:
Thanks for the improvements; I made some tweaks, and we can now merge this.

PR #1405, opened 2026-04-24
Fix: 1397 Expert deadlock by sending didOpen(mix.exs) before readiness wait

scytherswings:
Fixes: #1397 
Expert's build pipeline only starts after receiving a textDocument/didOpen notification. Without it, Serena waited 300 s for $/progress signals that Expert never emitted, causing a deadlock on every cold start.

PR #1444, opened 2026-05-03
fix: search subdirectories for mix.exs to prevent Expert deadlock in monorepos

scytherswings:
Problem

  PR #1405 fixed the Expert deadlock by sending a textDocument/didOpen notification for mix.exs before waiting for readiness. However, it only checks {repository_root_path}/mix.exs. In monorepo layouts where the Elixir project lives in a subdirectory (e.g.
  myproject/server/mix.exs), the file isn't found, didOpen is silently skipped, and Expert never starts compiling — causing the same 300-second deadlock timeout.

  Fix

  Adds a _find_mix_exs() method to ElixirTools that:
  1. Checks {repository_root_path}/mix.exs first (preserves existing behavior)
  2. If not found, scans immediate subdirectories (one level deep) for mix.exs
  3. Returns the first match, or None if nothing found
  4. Logs at info level when found in a subdirectory

  The rest of _start_server() is unchanged — it already guards on mix_exs_uri is not None.

  Tests

  Added 4 new tests to TestElixirToolsStartup:
  - test_finds_mix_exs_in_subdirectory — monorepo layout, didOpen sent with subdir URI
  - test_prefers_root_mix_exs_over_subdirectory — root takes precedence
  - test_closes_subdirectory_mix_exs_after_server_ready — didClose works for subdir path
  - test_no_didopen_when_no_mix_exs_in_subdirectories — graceful no-op when absent everywhere

  All 9 tests pass (5 existing + 4 new).


MischaPanch (changes requested):
Thanks! We can merge this, but pls remove the new tests (not needed to test such simple logic) and address the typing error from the CI run (it should probably be `if mix_exs_path is not None`)

MischaPanch:
You removed not only the new but also some old tests ^^. But that's ok

PR #1445, opened 2026-05-03
hooks: extend auto-approve to Claude Code's `auto` permission mode

### LINKED ISSUE #1386: Extend `serena-hooks auto-approve` to Claude Code's `auto` permission mode

`serena-hooks auto-approve` (introduced in v1.1, tracked in #920) only emits an allow decision when Claude Code reports `permission_mode == "acceptEdits"`. It is silent in every other documented mode.

Claude Code's [hooks docs](https://code.claude.com/docs/en/hooks) list six values for `permission_mode`:

> `permission_mode`: Current permission mode: `"default"`, `"plan"`, `"acceptEdits"`, `"auto"`, `"dontAsk"`, or `"bypassPermissions"`.

The most user-visible gap is `auto` mode. [Claude Code docs](https://code.claude.com/docs/en/permissions) describe it as:

> `auto`: Auto-approves tool calls with background safety checks that verify actions align with your request. Currently a research preview.

Auto mode is the mode users adopt when they want hands-off autonomous execution — the same use case that motivated #920. In `auto` today, Serena's MCP tool calls are not auto-approved by Serena's hook; they fall through to Claude Code's auto-mode classifier, which makes per-call probabilistic decisions rather than the deterministic "the user chose a permissive mode, extend it to Serena" semantic that #920 established.

The behavior is controlled by a single literal string check at [`src/serena/hooks.py:371`](https://github.com/oraios/serena/blob/main/src/serena/hooks.py#L371):

```python
_ACCEPT_EDITS_MODE = "acceptEdits"

def is_accept_edits_mode(self) -> bool:
    return self._permission_mode == self._ACCEPT_EDITS_MODE
```

**Reproduction**

Verified against `serena-agent==1.1.2` and main @ `14cd3ef`:

```bash
for mode in acceptEdits auto bypassPermissions default plan ""; do
  echo "=== mode: '$mode' ==="
  echo "{\"session_id\":\"t\",\"tool_name\":\"mcp__serena__replace_symbol_body\",\"permission_mode\":\"$mode\",\"tool_input\":{}}" \
    | serena-hooks auto-approve --client=claude-code
done
```

Only `acceptEdits` emits a decision:

```
=== mode: 'acceptEdits' ===
{"hookSpecificOutput": {"hookEventName": "PreToolUse", "permissionDecision": "allow", "permissionDecisionReason": "Auto-approved: Serena tool call while client is in acceptEdits mode.", "additionalContext": ""}}
=== mode: 'auto' ===
=== mode: 'bypassPermissions' ===
=== mode: 'default' ===
=== mode: 'plan' ===
=== mode: '' ===
```

All non-`acceptEdits` modes produce no output.

Extend `PreToolUseAutoApproveSerenaHook` so that Claude Code's `auto` mode is treated equivalently to `acceptEdits` for the purposes of Serena auto-approval. The minimal change is adding `"auto"` to the accepted set alongside `"acceptEdits"`.

How to make that configurable is the maintainers' call. Two shapes I can see working:

1. **Hard-code the expanded set.** Accept `acceptEdits` and `auto` by default. Simple, matches the #920 pattern.
2. **`--modes` option on the `auto-approve` subcommand.** Let users opt into any subset via something like `serena-hooks auto-approve --client=claude-code --modes=acceptEdits,auto`. Preserves today's default for existing users while letting new adopters opt into `auto`.

Happy to write the PR once direction is confirmed — per CONTRIBUTING.md I'm raising this as an issue first.

1. **User adds Serena tools to `autoMode.allow`.** Claude Code's auto mode has its own classifier with a prose-based allow list users can extend. In principle this works, but it's fragile (tool names evolve), per-tool (multiple entries), and crosses the boundary #920 was built to draw — Serena auto-approval should track the user's *permission mode choice*, not require maintaining a parallel allowlist.
2. **Shell wrapper around the hook.** A few lines of bash in `~/.claude/settings.json` can check `permission_mode` and shell out to `serena-hooks auto-approve`. Works today, but duplicates logic every Serena user has to author.
3. **Fix upstream in Claude Code.** #920's thread flagged that ideally Claude Code would natively extend permissive modes to MCP edit tools. That's the cleanest long-term fix, but #920 established the precedent that Serena provides the bridge today, and this issue asks to extend the existing bridge to cover the newer mode.

**Additional context**

- Verified against `serena-agent 1.1.2` (installed via `uv tool install -p 3.13 serena-agent@latest --prerelease=allow`) and main at commit `14cd3ef`.
- Claude Code's `auto` is labeled "research preview" in docs; the `permission_mode` hook field that exposes it is stable and canonically documented, so coupling to the field is low risk even if auto-mode semantics evolve.
- `bypassPermissions` and `dontAsk` are also silent in the current implementation. Those modes already suppress prompts at the Claude Code level, so the hook's silence is architecturally harmless — just noting for completeness in case maintainers want a while-you're-there fix.
- Relates to / follow-up on #920.

scarson:
## Summary

Closes #1386. Extends `serena-hooks auto-approve` to recognize Claude Code's `auto` permission mode in addition to the existing `acceptEdits` mode. Both are the user-facing permissive modes for hands-off execution; covering both lets the blanket approval reach Serena's MCP tools instead of falling through to per-call prompts (or, in `auto` mode, to Claude Code's per-call probabilistic classifier).

Per @MischaPanch's guidance in [#1386](https://github.com/oraios/serena/issues/1386#issuecomment-4312235251), this is the "hardcoded extended set" approach (option 1 from the issue) — no `--modes` flag, no opt-in surface — keeping the alpha-feature footprint small.

## Changes

- **`src/serena/hooks.py`** — replaced the single `_ACCEPT_EDITS_MODE` literal with `_AUTO_APPROVE_MODES: frozenset[str] = frozenset({"acceptEdits", "auto"})` and renamed `is_accept_edits_mode()` → `is_auto_approve_mode()`. The reason string now interpolates the actual mode (e.g. `"...while client is in auto mode."`) for unambiguous debug output now that two modes share the path. CLI `--help` text updated.
- **`test/serena/test_hooks.py`** — added 4 tests:
  - `test_approves_serena_tool_in_auto_mode` (positive, class level)
  - `test_auto_approve_command_in_auto_mode` (positive, CLI level)
  - `test_stays_silent_in_bypass_permissions_mode`, `test_stays_silent_in_dont_ask_mode` (negatives that pin the boundary, so future edits to `_AUTO_APPROVE_MODES` are deliberate)
- **`docs/02-usage/030_clients.md`** — updated the `auto-approve` bullet to describe both modes.
- **`CHANGELOG.md`** — entry under `* Hooks:` in `Unreleased`. Note the entry calls out the explicit tradeoff: in `auto` mode, Serena tool calls now bypass Claude Code's auto-mode safety classifier in favor of deterministic pre-approval, matching `acceptEdits`'s posture for built-in edits. This is exactly what #1386 asked for, but it's worth surfacing.

## Boundary: `bypassPermissions` and `dontAsk` are deliberately excluded

- `bypassPermissions` already approves everything before the hook would matter, so silence here is harmless.
- `dontAsk` is the user's deliberate deny-by-default posture (auto-deny unless an explicit allow rule matches); the hook honors that choice rather than blanket-overriding it.

The two new negative tests pin this boundary so any future expansion of `_AUTO_APPROVE_MODES` is intentional.

## Test plan

- [x] `poe format` — clean (ruff, no changes)
- [x] `mypy src/serena/hooks.py` — no issues
- [x] `pytest test/serena/test_hooks.py` — **48 passed** (44 existing + 4 new)
- [x] End-to-end repro from the issue — `auto` and `acceptEdits` now both produce `allow`; `bypassPermissions`, `default`, `plan`, and empty mode all stay silent as expected.
- [x] Manual review of docstring + docs + CLI `--help` for consistency with the new behavior.

```bash
$ for mode in acceptEdits auto bypassPermissions default plan ""; do
    echo "=== mode: '$mode' ==="
    printf '{"session_id":"t","tool_name":"mcp__serena__replace_symbol_body","permission_mode":"%s","tool_input":{}}' "$mode" \
      | serena-hooks auto-approve --client=claude-code
    echo ""
  done

=== mode: 'acceptEdits' ===
{"hookSpecificOutput": {"hookEventName": "PreToolUse", "permissionDecision": "allow", "permissionDecisionReason": "Auto-approved: Serena tool call while client is in acceptEdits mode.", "additionalContext": ""}}

=== mode: 'auto' ===
{"hookSpecificOutput": {"hookEventName": "PreToolUse", "permissionDecision": "allow", "permissionDecisionReason": "Auto-approved: Serena tool call while client is in auto mode.", "additionalContext": ""}}

=== mode: 'bypassPermissions' ===

=== mode: 'default' ===

=== mode: 'plan' ===

=== mode: '' ===
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,7 @@ Status of the `main` branch. Changes prior to the next official version change w
 
 * Hooks:
   - `serena-hooks auto-approve` now also emits an `allow` decision when Claude Code reports
-    `permission_mode == "auto"`, in addition to the existing `acceptEdits` behavior. The two
-    modes are the user-facing permissive modes; covering both lets blanket approvals reach
-    Serena's MCP tools instead of falling through to per-call prompts. Note that this means
-    Serena tool calls in `auto` mode no longer go through Claude Code's auto-mode safety
-    classifier — they are pre-approved deterministically, matching `acceptEdits`'s posture
-    for built-in edits. #1386
+    `permission_mode == "auto"`, in addition to the existing `acceptEdits` behavior. #1386
 
 * General:
   - Breaking change in mode definitions: Projects (project.yml) can no longer override `base_modes`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 Status of the `main` branch. Changes prior to the next official version change will appear here.
 
+* Hooks:
+  - `serena-hooks auto-approve` now also emits an `allow` decision when Claude Code reports
+    `permission_mode == "auto"`, in addition to the existing `acceptEdits` behavior. The two
+    modes are the user-facing permissive modes; covering both lets blanket approvals reach
+    Serena's MCP tools instead of falling through to per-call prompts. Note that this means
+    Serena tool calls in `auto` mode no longer go through Claude Code's auto-mode safety
+    classifier — they are pre-approved deterministically, matching `acceptEdits`'s posture
+    for built-in edits. #1386
+
 * General:
   - Breaking change in mode definitions: Projects (project.yml) can no longer override `base_modes`.
     Instead, they can define `added_modes` to add modes on top of base and default modes.  
--- a/docs/02-usage/030_clients.md
+++ b/docs/02-usage/030_clients.md
@@ -218,9 +218,9 @@ The hooks will:
   `grep` or `read_file` calls without using any Serena tools in between.
 - **`activate`**: Prompt the agent to activate the project and read Serena's instructions at session start.
 - **`cleanup`**: Clean up hook session data when the session ends.
-- **`auto-approve`**: Auto-approve Serena tool calls whenever Claude Code is in `acceptEdits` mode,
-  so blanket edit approvals cover Serena's destructive tools (e.g. `replace_symbol_body`,
-  `rename_symbol`) instead of prompting on every call.
+- **`auto-approve`**: Auto-approve Serena tool calls whenever Claude Code is in a permissive
+  permission mode (`acceptEdits` or `auto`), so blanket approvals cover Serena's destructive
+  tools (e.g. `replace_symbol_body`, `rename_symbol`) instead of prompting on every call.
 
 For more details on Claude Code's hook system, see the
 [Claude Code hooks documentation](https://code.claude.com/docs/en/hooks).
--- a/src/serena/hooks.py
+++ b/src/serena/hooks.py
@@ -359,29 +359,41 @@ class SessionEndCleanupHook(Hook):
 
 
 class PreToolUseAutoApproveSerenaHook(PreToolUseHook):
-    """Pre-tool-use hook that auto-approves Serena tool calls while the client is in accept-edits mode.
+    """Pre-tool-use hook that auto-approves Serena tool calls while the client is in a permissive permission mode.
 
-    Claude Code's ``acceptEdits`` permission mode only applies to its built-in editing tools;
-    Serena's destructive tools (e.g. ``replace_symbol_body`` or ``rename_symbol``) would still
-    prompt the user on every call. This hook emits an ``allow`` decision for any Serena MCP
-    tool call whenever the client reports ``acceptEdits`` as the active permission mode, so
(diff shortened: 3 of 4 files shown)


MischaPanch:
Thanks, merging

PR #1450, opened 2026-05-06
solidlsp: support more C++ file extensions

vilchanskyio-work:
Solidlsp needlessly filters out valid C++ file extensions. As far as I'm aware, there is no official list. Current list (of popular file extensions) was taken from the C++ Wikipedia page.

### DIFF
--- a/src/solidlsp/ls_config.py
+++ b/src/solidlsp/ls_config.py
@@ -285,7 +285,7 @@ class Language(str, Enum):
             case self.RUBY_SOLARGRAPH:
                 return FilenameMatcher(".rb")
             case self.CPP | self.CPP_CCLS:
-                return FilenameMatcher(".cpp", ".h", ".hpp", ".c", ".hxx", ".cc", ".cxx")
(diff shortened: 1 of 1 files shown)


vilchanskyio-work:
The issue that this solves (in my case): .hh headers (a rather popular extension) are completely ignored (despite being perfectly fine by clangd).

vilchanskyio-work:
Based on the failing tests ("install zig"), I don't think this PR is the reason for it. Please re-run yourself if needed, I don't know your CI (and its state).

MischaPanch:
Thanks, superseded by #1451 (pls correct me if I'm wrong), hence closing

vilchanskyio-work:
Yep, that's even better. I wasn't sure whether ccls backend supports all types of files as well. I'd add a justification in a comment why ccls extensions list differs. Currently the comment (source of the file extension list) is the same, so it's not clear why the list differs for clangd and ccls.

vilchanskyio-work:
That was regarding #1451

PR #1203, opened 2026-03-22
Allow symbol operations in dot-prefixed directories

Asher-:
## Problem

`is_ignored_dirname()` in the base `SolidLanguageServer` class returns
`dirname.startswith(".")`, which blanket-ignores all dot-prefixed
directories. This prevents `find_symbol` and `get_symbols_overview` from
working on files inside `.github/`, `.vscode/`, and any other legitimate
dot-prefixed directory.

The intent was to skip hidden/internal directories, but the filter is too
broad. Language-specific dot directories (`.build`, `.elixir_ls`,
`.terraform`, etc.) are already handled by individual LS overrides that
call `super().is_ignored_dirname()` — making the base class filter
redundant for those cases and harmful for everything else.

## Fix

Changed the base `is_ignored_dirname()` to only ignore VCS directories
(`.git`, `.svn`, `.hg`, `.bzr`) which should universally be excluded.
All language-specific dot directories continue to be handled by their
respective LS overrides.

## Note

This changes the default filter from "all dot-prefixed directories" to
"VCS directories only." Dot-prefixed directories that are listed in
`.gitignore` are still filtered by the gitignore-based exclusion at line
612. However, un-gitignored dot directories (e.g. `.github/`, `.vscode/`)
will now be visible to symbol operations — that is the intent, since those
directories contain legitimate code. If a broader default filter is
preferred, the `_VCS_DIRS` set can be extended.

## Related issues

Fixes #1187, fixes #853.

PR #1456, opened 2026-05-07
Add ignore_all_dot_files config to allow setting it to false; remove .serena from ALWAYS_IGNORED_DIRS

EarthmanWeb:
## Context

PR #1203 replaced the blanket `dirname.startswith(".")` with an explicit `_ALWAYS_IGNORED_DIRS` set — a solid improvement. However, two gaps remain:

1. **No user control over dot-directory indexing.** Some dot-prefixed directories (`.github/`, `.devcontainer/`) contain legitimate code that users may want symbol retrieval on. The current allowlist requires a code change to adjust.

2. **`.serena` is in `_ALWAYS_IGNORED_DIRS`, blocking symbol retrieval on Serena's own memory files.** Serena stores project memories as `.md` files inside `.serena/`. With `markdown` configured as a language, `get_symbols_overview` and `find_symbol` should work on these files — they're Serena's own content, authored and managed through its tools. Cache and state files within `.serena/` are non-`.md` and won't match the markdown LSP's source filter, so there's no noise risk.

### Commit 1: `ignore_all_dot_files` config option (rebased from #1101)
- Adds `ignore_all_dot_files` field to `ProjectConfig` (default: `true`, preserving current behavior)
- Wires it through `_from_dict` → `LanguageServerFactory` → `LanguageServerConfig` → `SolidLanguageServer`
- `is_ignored_dirname()` checks `_ALWAYS_IGNORED_DIRS` first, then applies the dot-prefix filter only when `ignore_all_dot_files=true`
- Pyright's `_get_initialize_params` respects the setting for exclude/include lists
- Documents the option in `project.template.yml`

### Commit 2: Remove `.serena` from `_ALWAYS_IGNORED_DIRS`
- `.serena/` memory files are valid symbol retrieval targets when markdown is a configured language
- The `ignore_all_dot_files` flag (default: `true`) still protects dot-prefixed directories in general
- Users who want symbol retrieval on `.serena/` memories set `ignore_all_dot_files: false` explicitly

## Supersedes

- #1101 (closed, same branch — now rebased on current `main` with the `_from_dict` wiring fix added)


MischaPanch:
Hi @EarthmanWeb . I am not sure this option is needed in practical applications, as we take into account the user's gitignore. Is that not enough?

I also don't see how the workspace/configuration handler in pyright specifically has something to do with this new setting - rather we should use the existing mechanisms to identify what has to be ignored and handle it there. Or am I missing something?

EarthmanWeb:
@MischaPanch 
The naming of this might be confusing, because it's not actually creating the ability to ignore, but rather the converse, creating the ability to unignore dot prefixed folders. 

Unignoring files in gitignore does NOT allow access to dot folders, that is the issue this is fixing, by providing an OPTIONAL ability to include dotfolders.

Skipping dotfolders and especially .serena is BAKED IN currently, so even when trying to unignore them, the hardcoded skips do not allow it.

Often I am working on files in dotfolders, including .serena itself, and want to use serena to index them. This fix allows that, and does not affect operations otherwise, since it defaults to True, which is the current default now

src/solidlsp/ls.py
```diff
@@ -555,7 +423,11 @@ def is_ignored_dirname(self, dirname: str) -> bool:
         A language-specific condition for directories that should always be ignored. For example, venv
         in Python and node_modules in JS/TS should be ignored always.
         """
-        return dirname in self._ALWAYS_IGNORED_DIRS
+        if dirname in self._ALWAYS_IGNORED_DIRS:
+            return True
+        if self._ignore_all_dot_files and dirname.startswith(".") and dirname not in (".", ".."):
+            return True
```
opcode81, line 429: > The naming of this might be confusing, because it's not actually creating the ability to ignore, but rather the converse, creating the ability to unignore dot prefixed folders.

@EarthmanWeb with this addition, you are reintroducing the concept of *ignoring* directories starting with `.`, not *unignoring* them.
This is a concept we want to gradually abandon, as ignored folders are - in most cases - fully handled by .gitignore patterns already in the repository. 
The only reason we didn't fully delete the entire `is_ignored_dirname` concept is that we were worried that some people might rely on this, so we decided to keep a list of names that most users probably do want to ignore.

PR #1460, opened 2026-05-08
Optimize glob_match, search_text, scan_directory

lotuseater:
Main changes:
  - Cache compiled glob regexes used by `glob_match`.
  - Thread optional `is_dir` hints through ignore/path matching to avoid redundant filesystem checks.
  - Make multiline `search_text` line-number calculation incremental instead of repeatedly recounting newlines.
  - Add regression tests for ignore matching, glob-cache behavior, and multiline search line numbers.

  Main benchmark results on the Serena repo corpus:
  - `glob_match`: 31.3 ms -> 3.2 ms, about 9.7x faster.
  - `scan_directory + GitignoreParser`: 771 ms -> 230 ms, about 3.4x faster.

### DIFF
--- a/src/serena/project.py
+++ b/src/serena/project.py
@@ -427,14 +427,23 @@ class Project(ToStringMixin):
             )
         return self.__ignored_patterns
 
-    def _is_ignored_relative_path(self, relative_path: str | Path, ignore_non_source_files: bool = True) -> bool:
+    def _is_ignored_relative_path(
+        self,
+        relative_path: str | Path,
+        ignore_non_source_files: bool = True,
+        is_dir: bool | None = None,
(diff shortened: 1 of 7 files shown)


MischaPanch:
Thx for the PR. I had a brief look, I'm not sure the perf. wins are worth the complexity. These things are hardly the bottleneck of anything. 

>glob_match: 31.3 ms -> 3.2 ms, about 9.7x faster

This statement is rather confusing, since caching can only help when one repeatedly compiles the same patterns, and that doesn't tend to happen in real applications.

MischaPanch:
No response, since the optimization doesn't seem worth it, closing

PR #1461, opened 2026-05-08
Add Ada / SPARK language support via AdaCore Ada Language Server

vkuprin:
Adds `Language.ADA` via AdaCore's [Ada Language Server](https://github.com/AdaCore/ada_language_server). One enum covers Ada and SPARK — both share `.ads`/`.adb`, and ALS distinguishes SPARK by source-level pragmas/aspects. Follows `.serena/memories/adding_new_language_support_guide.md`.

ALS auto-downloads from AdaCore's GitHub releases (linux-x64/arm64, darwin-x64/arm64, win32-x64) with per-platform SHA256 verification, pinned to `2026.2.202604091`. Users with an existing ALS (Alire, GNAT Studio, VS Code Ada extension) can point Serena at it via `ls_specific_settings.ada.ls_path`.

Test repo at `test/resources/repos/ada/test_repo/`: a minimal `default.gpr` GNAT project with `helper.ads`/`helper.adb` and a `main.adb` that calls `Helper.Greet`.

### Two Ada-specific test accommodations worth flagging

- `test_find_references_within_file` queries the `Helper` package (used in the `with` clause and as the qualifier of `Helper.Greet`) rather than a single-use local. Serena's `request_references` sets `includeDeclaration=False`, and ALS returns no other references for a single-use local constant.
- `test_bare_symbol_names` passes `whitespace_allowed=True, period_allowed=True` because ALS surfaces synthetic `"With clauses"` and dotted unit references like `"Ada.Text_IO"` — both legitimate Ada output, both supported by `has_malformed_name`'s existing flags.

### Test plan

- [x] `pytest test/solidlsp/ada/ -m ada` — 9/9 pass on macOS arm64
- [x] `ruff check`, `ruff format --check`, `mypy` — clean
- [x] CI matrix (ubuntu-latest, macos-latest, windows-latest) — running
- [x] Manual smoke test through the agent — `FindSymbolTool` resolves `Helper/Greet` in both spec and body; `FindReferencingSymbolsTool` returns the call site in `Main` (line 4 of `main.adb`)


src/solidlsp/language_servers/ada_language_server.py
```diff
@@ -0,0 +1,236 @@
...
+from solidlsp.settings import SolidLSPSettings
+
+from .common import RuntimeDependency, RuntimeDependencyCollection
+
+log = logging.getLogger(__name__)
+
+# Version pinning convention (see eclipse_jdtls.py for the full spec):
+#   INITIAL_* — frozen forever; legacy unversioned install dir is reserved for it.
+#   DEFAULT_* — bumped on upgrades; goes into a versioned subdir.
+INITIAL_ALS_VERSION = "2026.2.202604091"
```
MischaPanch: we don't need the initial version for new LS, it is in existing ones to correct for an earlier mistake. Pls remove

test/solidlsp/ada/test_ada_basic.py
```diff
@@ -0,0 +1,147 @@
...
+    @pytest.mark.parametrize("repo_path", [Language.ADA], indirect=True)
+    def test_find_definition_across_files(self, language_server: SolidLanguageServer, repo_path: Path) -> None:
+        # main.adb LSP line 4 / column 40 sits on the `G` of `Helper.Greet`:
+        #   `   Greeting : constant String := Helper.Greet ("Ada");`
+        #                                            ^ col 40
+        main_path = str(repo_path / "src" / "main.adb")
+        definitions = language_server.request_definition(main_path, 4, 40)
+
+        assert definitions, f"Expected non-empty definition list but got {definitions=}"
+        # ALS may return the spec only, or both spec and body.
```
MischaPanch: why is there non-deterministic behavior? Consider removing the if below

test/solidlsp/ada/test_ada_basic.py
```diff
@@ -0,0 +1,147 @@
...
+        helper_children = helper_symbol.get("children", [])
+        helper_child_names = [c.get("name") for c in helper_children]
+        assert helper_child_names, f"Helper package has no children — hierarchicalDocumentSymbolSupport is not working. Roots: {root_names}"
+        assert "Greet" in helper_child_names, f"Greet not nested under Helper. Children: {helper_child_names}"
+
+        # Greet must NOT appear at root level — that would indicate the flat fallback format.
+        assert "Greet" not in root_names, f"Greet should be a child of Helper, not at root level. Roots: {root_names}"
+
+    @pytest.mark.parametrize("language_server", [Language.ADA], indirect=True)
+    def test_bare_symbol_names(self, language_server: SolidLanguageServer) -> None:
```
MischaPanch, line 133: pls add a test for diagnostics, like for other LS

MischaPanch (changes requested):
Thanks for the contribution! Looks good, only minor comments to be addressed, then this can be merged

test/solidlsp/ada/test_ada_basic.py
vkuprin: Comment addressed

PR #1476, opened 2026-05-11
Add news item for v1.3.0 release

opcode81:


MischaPanch (approved):

MischaPanch (changes requested):
We should mention the changes in hooks in more detail, they are very user facing

PR #1477, opened 2026-05-12
Revamp Memory System

MischaPanch:
Apart from the memories stuff, the PR also adds the `dotall` kwarg and shortens the docstrins in the SearchForPattern an EditMemory tools. Done this here due to the extension of the CLI which mirrors the EditMemory interface


src/serena/memories/memory_reference_analysis.py
```diff
@@ -0,0 +1,780 @@
...
+
+if TYPE_CHECKING:
+    from serena.project import MemoriesManager
+
+
+# ---------------------------------------------------------------------------
+# Constants
+# ---------------------------------------------------------------------------
+
+MEMORY_REF_PREFIX: str = "mem:"
```
MischaPanch, line 32: Redefined here but it prevents circular imports. I suggest to leave like this for now and later consolidate when we split off memory management from project.py

opcode81 (approved):
@MischaPanch I have made some additional changes and we could now merge this.

Two things we could consider:
* Establish a different default case convention for memories, i.e. kebab-case (as used in Penpot)
* Establish `mem:critical-info` instead of `mem:core` as the root memory - though not sure about this one.

src/serena/memories/memory_reference_analysis.py
opcode81: Fixed.

PR #1494, opened 2026-05-16
refactor: enforce Language enum usage across all language servers

duducpp:
## Motivation
Raw string literals were used as language IDs (e.g. `"ada"`, `"go"`) across all language server constructors, making it easy to introduce typos and inconsistencies. The `Language` enum already existed but was underused.

## Summary
Replace all raw language ID strings with `Language` enum members across 64 language server files, and update the base class to accept the enum type directly.

## Changes
- Replace raw string language identifiers with `Language` enum members in all 64 language server files
- Update `SolidLanguageServer.__init__` signature to accept `Language` instead of `str`
- Remove redundant `Language(language_id)` cast in `ls.py` since callers now pass the enum directly
- Update `language_id` param docstring to reflect the enum type

## Breaking changes
`SolidLanguageServer.__init__` now requires a `Language` enum member for `language_id` instead of a plain `str`. Any third-party subclass must update its constructor call.

### DIFF
--- a/src/solidlsp/language_servers/ada_language_server.py
+++ b/src/solidlsp/language_servers/ada_language_server.py
@@ -16,7 +16,7 @@ from typing import cast
 from overrides import override
 
 from solidlsp.ls import LanguageServerDependencyProvider, LanguageServerDependencyProviderSinglePath, SolidLanguageServer
-from solidlsp.ls_config import LanguageServerConfig
+from solidlsp.ls_config import Language, LanguageServerConfig
 from solidlsp.lsp_protocol_handler.lsp_types import InitializeParams
 from solidlsp.settings import SolidLSPSettings
 
@@ -147,7 +147,7 @@ class AdaLanguageServer(SolidLanguageServer):
             config,
             repository_root_path,
             None,
-            "ada",
+            Language.ADA,
             solidlsp_settings,
         )
         self.server_ready = threading.Event()
--- a/src/solidlsp/language_servers/al_language_server.py
+++ b/src/solidlsp/language_servers/al_language_server.py
@@ -137,7 +137,7 @@ class ALLanguageServer(SolidLanguageServer):
         request fails, preventing repeated unsuccessful attempts.
         """
 
-        super().__init__(config, repository_root_path, ProcessLaunchInfo(cmd=cmd, cwd=repository_root_path), "al", solidlsp_settings)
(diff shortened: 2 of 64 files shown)


opcode81:
This is based on a misconception: While the language ID is an LSP concept, the `Language` enum is a SolidLSP concept.
* The language ID identifies the language of a source file in the context of a given language server.
* The `Language` enum identifies a language server in SolidLSP.  
   We originally weren't planning to support more than one language server per language, which explains the naming.

The docstrings should have made this clear:

https://github.com/oraios/serena/blob/1767a259ceee2fbf1f6857f110b453c762fe05af/src/solidlsp/ls_config.py#L45-L48

https://github.com/oraios/serena/blob/1767a259ceee2fbf1f6857f110b453c762fe05af/src/solidlsp/ls.py#L492-L495

Notice that, for example, the `OmniSharp` language server correctly passes `csharp` (instead of `csharp_omnisharp`) and `PhpActor` passes `php` (instead of `php_phpactor`), correctly reflecting the language rather than the language server.

So the error in the implementation is actually *not* that the `Language` enum is being underused but that the `Language` enum instance is derived from the language ID, while the correct way to obtain it is via `get_language_enum_instance`.

opcode81:
@duducpp we appreciate your efforts, but it is recommended to create an issue first in order to prevent unnecessary work (see our contribution guidelines).

duducpp:
@opcode81 so, is it possible to use a different language_id when instantiating SolidLanguageServer? At least in the current codebase, I couldn’t find any other usages.

OmniSharp still uses `Language.CSHARP`: [PR diff reference](https://github.com/oraios/serena/pull/1494/changes#diff-03081cc647310e20cae9db182ff08f652f2de7eca98ef5ddf539f07b050774e0L120)

And the config-based loading uses `config.code_language`, which also maps to the same Language enum: [Config loading reference](https://github.com/oraios/serena/pull/1494/changes#diff-3a33477d0f4f70065e7e275452b3a53d4bf64ca2b83746d2c81460e1709d8eafL125)

opcode81:
The language id used for `OmniSphere` is `csharp` (informing the LS of the lanugage content appropriately), while the `Language` enum item is `Language.CSHARP_OMNISHARP = "csharp_omnisharp"`.
The constructor would previously (before my fix, see branch `main`) assign `Language.CSHARP = "csharp"`, which is actually incorrect but works because the relevant enum item's properties for LS operation are equivalent.

Importantly, the match between language id `csharp` and the value of `Language.CSHARP` was basically incidental only. There need not be any connection at all, and my commit has fixed this.

PR #1504, opened 2026-05-19
Add CA65 (6502/65C02/65816 assembly, cc65 toolchain) language support

JC-000:
## Summary

Adds `Language.CA65` to Serena: support for CA65 macro assembly used across the 6502 family
(NES, C64, Apple II, Atari, etc.) via the [cc65](https://cc65.github.io/) toolchain.

Two commits:
1. **Add CA65 language server** — the language-specific work (new `Ca65LanguageServer`,
   enum entry, factory case, test corpus, tests).
2. **Surface LSP `detail` field in `LanguageServerSymbol.to_dict`** — a small,
   language-agnostic companion change (8 lines) that lets *any* LS publish per-symbol
   context via the standard `detail` field through to MCP clients. Happy to split into a
   separate PR if you'd prefer.

## How it's structured

Wraps [`ca65-ls`](https://github.com/JC-000/ca65-asm-serena-lsp) — a separately-installable
Python + pygls language server backed by the
[`pogyomo/tree-sitter-ca65`](https://github.com/pogyomo/tree-sitter-ca65) grammar. The
Serena-side shim follows the established `LanguageServerDependencyProviderSinglePath`
pattern (modeled on `PyrightServer`); ca65-ls runs as `python -m ca65_ls.server --stdio`
using the same interpreter Serena is running under.

**Architecture:** tree-sitter-ca65 is the primary source of truth for symbol parsing; cc65's
`ld65 --dbgfile` debug info is consumed as an opt-in enricher that supplies post-link
addresses, segment membership, and symbol sizes when present. Codebases without `.dbg`
still get full documentSymbol / workspace/symbol / definition / references / rename
functionality — the enrichment is purely additive. Diagnostics come from a background
`ca65 -g` invocation.

**Cache invalidation:** `Ca65LanguageServer._document_symbols_cache_fingerprint` includes
the ca65-ls package version plus a sha256 of the behavior-affecting source files. This
keeps editable-install dev iteration honest — changes to ca65-ls auto-invalidate Serena's
on-disk symbol cache without manual `rm`.

## Capabilities

- `textDocument/documentSymbol` — hierarchical, with nested cheap locals
- `workspace/symbol` — fuzzy/substring, with (declarator, definition) dedup
- `textDocument/definition` — implementation-kind preferred over declarators
- `textDocument/references` — **scope-aware**: cheap locals (`@name`) stay confined to
  their parent routine's body, `.proc`/`.scope`-nested labels resolve within their
  enclosing scope. Real-world impact: a `find_referencing_symbols` on `@loop` returns the
  local jump sites instead of every `@loop` across the workspace.
- `textDocument/hover` — kind, scope path, parent label, address/segment/size, and
  the immediately-preceding `;` comment block from the def site
- `textDocument/rename` + `textDocument/prepareRename` — same scope rules; refuses
  anonymous labels; auto-prefixes `@` when renaming a cheap local
- `textDocument/publishDiagnostics` — from `ca65 -g` on save

## Installation (for users)

```sh
pip install "ca65-ls @ git+https://github.com/JC-000/ca65-asm-serena-lsp@main#subdirectory=packages/ca65-ls"
```

PyPI release is coordinated with this PR — I can publish `ca65-ls` to PyPI prior to merge
if that matches your usual workflow.

## Validation

- **In-package tests** (in the companion `ca65-ls` repo): 83 unit tests across buffer
  parsing, workspace indexing, scope-aware references, hover content, and rename edits.
- **Serena integration tests**: `test/solidlsp/ca65/test_ca65_basic.py` — 5 tests
  exercising `request_document_symbols`, `request_definition`, `request_references` against
  a synthetic CA65 corpus that uses `.proc`, `.scope`, `.import`/`.export`, `.macro`,
  `.struct`, cheap locals, and anonymous labels. The corpus's `.dbg`/`.lbl`/`.map` fixtures
  are committed (regenerable via `tools/regen_fixtures.sh` in the ca65-ls repo).
- **Wire test against a real CA65 codebase** (~208 files / 9219 symbols): every M4 feature
  works end-to-end through Serena's MCP tools — scope-aware refs return one hit instead of
  4000+ for shared names like `@loop`, cross-file rename touches `.import` / `.export` /
  `jsr` sites atomically, document overviews surface post-link addresses inline.

## Notes for review

- **`detail` field commit (companion):** the change is 8 lines in `symbol.py` and is
  language-agnostic. I kept it in this PR because it's load-bearing for ca65-ls's UX
  (clients see per-symbol address/segment/size info without a hover round-trip) and the
  rationale comments inline reference the CA65 use case. If you'd prefer it as a separate
  PR, say the word and I'll split.
- **CI**: tests skip via `language_tests_enabled(Language.CA65)` when the `ca65` binary
  isn't on the PATH (mirroring the existing clojure / ccls handling in `test/conftest.py`).
  If you want them to run on CI, the runner needs `apt-get install cc65` or equivalent.
- **Reserved-word gotcha**: CA65 treats `a`/`x`/`y`/`s` as register names, so they can't
  be used as `.struct` field names. Documented inline in the test corpus README.
- **Grammar dependency**: ca65-ls pins
  [pogyomo/tree-sitter-ca65 @ b22ead1](https://github.com/pogyomo/tree-sitter-ca65/commit/b22ead1).
  Five grammar gaps are documented and worked around in ca65-ls; none affect the typical
  6502 / NES / C64 codebases we tested against.

## Test plan

- [ ] `pytest test/solidlsp/ca65/test_ca65_basic.py` after installing ca65-ls into Serena's
  venv (`uv pip install -e ...` from the ca65-ls repo, or the pip-from-git command above)
- [ ] Manual: activate a CA65 project, exercise `find_symbol` / `find_referencing_symbols` /
  `rename_symbol` from the MCP tools
- [ ] (optional) Manual: confirm Serena's other language servers still work — the only
  shared file touched is `symbol.py` and the change is additive


MischaPanch:
While we appreciate the contribution, this is too exotic and too dependent on unverified third-party code. Very few users would ever benefit from this addition, so in our opinion it doesn't cross the threshold justifying maintenance overhead and increased supply chain attack risk. Therefore closing

PR #1507, opened 2026-05-20
typescript_vts: add initialization_options pass-through

utrumo:
## Summary

Adds an optional `initialization_options` dict to `ls_specific_settings.typescript_vts`. The dict is forwarded to **vtsls** verbatim through all three LSP configuration channels:

- the `initialize` request (`initializationOptions` field),
- a `workspace/didChangeConfiguration` notification sent right after `initialized`,
- and as the response to `workspace/configuration` pull requests (per section, with dotted-path support).

**Why all three channels:** vtsls reads `typescript.tsdk` only via `workspace/configuration` pulls, while other settings come through `initializationOptions`. Forwarding on all three covers vtsls behaviour variations and stays consistent.

**Primary use case — Yarn Plug'n'Play projects.** After `yarn dlx @yarnpkg/sdks vscode`, users can point vtsls at the PnP-aware `tsserver.js`:

```yaml
ls_specific_settings:
  typescript_vts:
    initialization_options:
      typescript:
        tsdk: "<path-to>/.yarn/sdks/typescript/lib"
      vtsls:
        autoUseWorkspaceTsdk: true
```

`autoUseWorkspaceTsdk: true` is required because in a headless LSP there is no UI prompt to confirm switching to the workspace TypeScript. See [yioneko/vtsls#169](https://github.com/yioneko/vtsls/issues/169) for the recipe.

**Default behaviour is unchanged** — the field is optional; `null` (default) or `{}` skips all three channels.

## Implementation notes

- The validated dict is exposed as a `@functools.cached_property` (`_initialization_options`) so the underlying `CustomLSSettings.get` lookup and its INFO log run exactly once per instance, regardless of how many call sites read it.
- Non-dict values raise `ValueError` at server start.
- `_extract_section_value` supports dotted-path lookups (e.g. `typescript.tsdk`) for `workspace/configuration` pulls.

## Test plan

- [x] Unit tests added — `test/solidlsp/typescript/test_vts_initialization_options.py` (14 tests):
  - Pass-through through `initialize` (verbatim, unset, empty dict, non-dict raises)
  - Direct accessor contract (`_initialization_options`) covering both call sites
  - Section lookup (top-level, dotted, missing, empty user_settings)
  - Base initialize payload not disturbed when settings are added
- [x] `uv run pytest test/solidlsp/typescript/test_vts_initialization_options.py` — 14/14 pass
- [x] `uv run mypy` on both files — clean
- [x] `uv run ruff format --check` + `uv run ruff check` — clean
- [x] Tests marked `@pytest.mark.typescript` (parity with other typescript test files)
- [ ] **No integration test against a real vtsls process** — omitted because vtsls install + Yarn PnP setup is non-trivial for CI. Unit tests cover the LSP-payload contract (initialize / didChange / configuration handlers).

## Files

- `vts_language_server` source — accessor + forwarding logic
- new test file for initialization_options
- `docs/02-usage/050_configuration.md` — new setting entry, full Yarn PnP example, Troubleshooting subsection, links to vtsls schema and VS Code TS settings
- `CHANGELOG.md` — Unreleased entry under `Language Servers:`


src/solidlsp/language_servers/vts_language_server.py
```diff
@@ -114,12 +128,53 @@ def _setup_runtime_dependencies(cls, config: LanguageServerConfig, solidlsp_sett
...
+        if section:
+            for part in section.split("."):
+                if isinstance(value, dict) and part in value:
+                    value = value[part]
+                else:
+                    return {}
+        return value if value is not None else {}
+
+    @functools.cached_property
+    def _initialization_options(self) -> dict:
```
MischaPanch, line 130: no need for caching, the operations are all cheap

test/solidlsp/typescript/test_vts_initialization_options.py
```diff
@@ -0,0 +1,122 @@
...
+    """
+
+    SETTINGS = {
+        "typescript": {"tsdk": "project/.yarn/sdks/typescript/lib"},
+        "vtsls": {"autoUseWorkspaceTsdk": True},
+        "javascript": {},
+    }
+
+    def test_top_level_section_returns_subdict(self) -> None:
+        assert VtsLanguageServer._extract_section_value(self.SETTINGS, "typescript") == {"tsdk": "project/.yarn/sdks/typescript/lib"}
```
MischaPanch: never use private methods in tests. I think all these tests can be deleted

src/solidlsp/language_servers/vts_language_server.py
```diff
@@ -179,12 +239,14 @@ def register_capability_handler(params: dict) -> None:
...
+        init_options = self._initialization_options
+
+        def workspace_configuration_handler(params: dict) -> list[object] | object:
+            # vtsls pulls settings for sections like "typescript", "vtsls", "javascript".
+            # Return the matching sub-dicts from the user-provided initialization_options.
             if "items" in params:
-                return [{}] * len(params["items"])
-            return {}
+                return [self._extract_section_value(init_options, item.get("section", "")) for item in params["items"]]
+            return self._extract_section_value(init_options, "")
```
MischaPanch: this makes no sense, it is the same as returning init_options

The method `_extract_section_value` has weird and unintuitive behavior. Pls have a critical look at the code and make only the changes you definitely need

MischaPanch (changes requested):
Thx, did a first quick review

src/solidlsp/language_servers/vts_language_server.py
utrumo: Done — switched to a plain `@property`

src/solidlsp/language_servers/vts_language_server.py
utrumo: Agreed — removed _extract_section_value

test/solidlsp/typescript/test_vts_initialization_options.py
utrumo: Done — removed

MischaPanch:
Thanks, much better. A nitpick for the future - a property should never raise an error (you don't expect a value error on something that looks like attribute access). But it's fine, I will merge after CI runs through

PR #1516, opened 2026-05-26
fix: allow non-existent paths in ignore check

weiconghe:
## Problem
     
  \_is_ignored_relative_path\ raises \FileNotFoundError\ when the target path does not exist on disk. This blocks editing tools (\
eplace_content\, \insert_after_symbol\, \insert_before_symbol\) from 
  operating on newly created files that haven't been indexed yet.                                                                                                                                                  

  **Error traceback:**                                                                                                                                                                                             
  \\\
  FileNotFoundError: File .../BatchModifyConvert.java not found, the ignore check cannot be performed                                                                                                              
    at serena/project.py:449 (_is_ignored_relative_path)                                                                                                                                                           
    at serena/project.py:537 (validate_relative_path)
    at serena/tools/file_tools.py:220 (replace_content)                                                                                                                                                            
  \\\                                                                                                                                                                                                           
  
  This is a common scenario when using Serena as an MCP server alongside another agent (e.g., Claude Code) that creates files with its own Write tool.
  
  ## Solution

  Return \False\ (not ignored) for non-existent paths instead of raising. A path that doesn't exist on disk cannot be matched by ignore patterns, so this is semantically correct and safe.                      

  ## Changes                                                                                                                                                                                                       

  - \src/serena/project.py\: \_is_ignored_relative_path\ returns \False\ with a debug log for non-existent paths                                                                                             
  - \src/serena/project.py\: Updated \alidate_relative_path\ docstring to reflect new behavior                                                                                                                
  - \	est/serena/config/test_global_ignored_paths.py\: Added 3 test cases                                                                                                                                        

  ## Test Results                                                                                                                                                                                                  

  All 15 tests pass (12 existing + 3 new):                                                                                                                                                                         
  \\\                                                                                                                                                                                                           
  test_nonexistent_path_not_ignored PASSED                                                                                                                                                                         
  test_nonexistent_path_with_ignore_patterns PASSED                                                                                                                                                                
  test_validate_relative_path_allows_nonexistent PASSED                                                                                                                                                            
  \\\                                                                                                                                                                                                           

  ## Note                                                                                                                                                                                                          

  The downstream tools (\EditedFileContext\, \
eplace_content\) already handle file creation when the file doesn't exist, so no further changes are needed in the editing pipeline.

### DIFF
--- a/src/serena/project.py
+++ b/src/serena/project.py
@@ -191,8 +191,8 @@ class Project(ToStringMixin):
 
     def _is_ignored_relative_path(self, relative_path: str | Path, ignore_non_source_files: bool = True) -> bool:
         """
-        Determine whether an existing path should be ignored based on file type and ignore patterns.
-        Raises `FileNotFoundError` if the path does not exist.
+        Determine whether a path should be ignored based on file type and ignore patterns.
+        Returns False for non-existent paths since they cannot be matched by ignore patterns.
 
         :param relative_path: Relative path to check
         :param ignore_non_source_files: whether files that are not source files (according to the file masks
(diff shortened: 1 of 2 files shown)


MischaPanch:
Thanks, that's reasonable, merging

PR #1517, opened 2026-05-26
feat(hooks): expand _CODE_FILE_EXTENSIONS to include config/schema formats

joeblackwaslike:
## Summary

Adds structured non-source file types to `_CODE_FILE_EXTENSIONS` in `PreToolUseRemindAboutSymbolicToolsHook`:

- **Config/data:** `.yaml`, `.yml`, `.json`, `.jsonc`, `.toml`, `.hcl`, `.tfvars`, `.tf`
- **Schema/query:** `.graphql`, `.gql`, `.proto`, `.sql`

## Motivation

These extensions are ubiquitous in real-world projects. An agent doing codebase exploration will frequently read `pyproject.toml`, `package.json`, `.github/workflows/*.yml`, GraphQL schemas, SQL migrations, and Terraform configs — often in bursts just like source files.

The rationale behind the existing set applies equally here: `search_for_pattern` (which works for _all_ file types) is usually more targeted than a raw read, and `find_symbol` / `get_symbols_overview` applies where a language server supports the format (e.g. JSON Schema, some YAML LSPs). The nudge hook keeps agents honest about both.

A short note was added to the docstring clarifying that `search_for_pattern` is always a valid alternative regardless of extension.

## What is NOT included

- `.md` / `.rst` / `.xml` — documentation and markup formats where the symbolic angle is weakest; a full read is often the right call for prose content
- `.env` / binary formats — not meaningful targets for symbolic search

## Test plan

- [ ] Existing tests pass (`pytest`)
- [ ] Manual: `serena-hooks remind` fires on repeated reads of `.yaml` / `.json` / `.toml` files
- [ ] Manual: counter resets on a `search_for_pattern` call targeting one of these files

### DIFF
--- a/src/serena/dashboard.py
+++ b/src/serena/dashboard.py
@@ -15,7 +15,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any, Optional, Self
 
 import psutil
-from flask import Flask, Response, redirect, request, send_from_directory
+from flask import Flask, Response, abort, redirect, request, send_from_directory
 from PIL import Image
 from pydantic import BaseModel
 from sensai.util import logging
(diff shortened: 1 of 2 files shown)


MischaPanch:
Thanks. I had a typo and force-pushed on main some minutes ago, so you'll have to rebuild your branch. Please cherry-pick your commit and add an entry to the changelog. Then this can be merged

MischaPanch:
I did this myself now since I want to release a new version, your commit is on main. Closing

PR #1521, opened 2026-05-28
Fix host validation

opcode81:


CHANGELOG.md
```diff
@@ -8,15 +8,23 @@ Status of the `main` branch. Changes prior to the next official version change w
...
+Add meta-data for the GitHub MCP registry
+
 # v1.5.2 (2026-05-26)
 
 * General:
   - Not existing paths return `False` on is ignored checks (instead of raising an error)
   - Add `serena-agent` CLI command so that `uvx serena-agent` can be used as entrypoint.
   - Fortls and pyright are now installed on the fly instead of being bundled in the serena-agent package.
 
+Dashboard:
```
MischaPanch, line 25: I did not want to advertise DNS rebinding attacks on earlier versions before the CVE was issued, so I didn't include it on purpose

src/serena/dashboard.py
```diff
@@ -197,14 +197,16 @@ def __init__(
...
         self._tool_names = tool_names
         self._agent = agent
+        self._host = host
         self._app = Flask(__name__)
-
-        self._app.config["TRUSTED_HOSTS"] = ["127.0.0.1", "localhost"]
-
+        local_hosts = ["127.0.0.1", "localhost"]
+        if self._host in local_hosts:
+            self._app.config["TRUSTED_HOSTS"] = local_hosts
```
MischaPanch, line 209: Let's discuss this, maybe turning off verification entirely is not the best way to go, a user could configure trusted hosts

opcode81: It is currently completely broken. A user who is not connecting locally cannot connect at all!

MischaPanch: If you think this is urgent, we can merge it in current form (pls remove the entry from the changelog though). I suspected there to be approximately 0 users not connecting locally, so I didn't think it was urgent and that we could discuss it in the next days

CHANGELOG.md
opcode81: The fix is released. The CVE being issued is not a relevant event. The only thing it does is draw more attention to this.

PR #1527, opened 2026-05-29
fix: query_project rejected read-only tools excluded by the current context

bor:
## Summary

`QueryProjectTool.apply` gated forwarding on `tool.is_active()`, which reflects the **caller's** active toolset (context + global config + modes + active project). A read-only tool excluded merely by the current context — e.g. `read_file`, `find_file`, `list_dir`, or `search_for_pattern` in CLI/IDE contexts such as `claude-code`, `codex`, `copilot-cli`, `vscode`, `junie`, `antigravity`, `jb-*`, `ide` — was therefore blocked from being forwarded to another registered project, even though the tool is read-only and perfectly runnable against the target. The error users saw was `Tool <name> is not active.`

## Root cause

`Tool.is_active()` -> `agent.tool_is_active(name)` -> membership in `agent._active_tools`, which is the base toolset reduced by context + global config + modes + project. Context-level exclusions (client-cosmetic, e.g. Claude Code has a native grep so `search_for_pattern` is excluded for UX) were therefore conflated with "tool may not run at all." Forwarding is a different question — context-level exclusions should not block it.

Corroboration: the existing server-side executor `ProjectServer._query_project` already gates on **only `is_readonly()`**, so read-only is the intended gate; the over-strict check existed only on the caller side.

## Fix

- Replace `assert tool.is_active()` in `QueryProjectTool.apply` with a check against the **global** Serena configuration only, via a new `ToolInclusionDefinition.disables_tool(name)` helper that unifies `excluded_tools` (incremental mode) and `fixed_tools` (fixed mode) handling.
- Keep the `is_readonly()` requirement.
- All three gates in `apply()` (read-only, global-config, registration) now raise explicit `ValueError`s rather than `assert`s, so they cannot be stripped by `python -O`.

After the fix:
- Context-level (client-cosmetic) exclusions no longer block forwarding. ✓ (the reported bug)
- Globally disabled tools still are. ✓
- Non-read-only tools still are. ✓

## Scope

Intentionally minimal — fixes only the reported behavior. `ProjectServer._query_project` is unchanged; it already gates on `is_readonly()`, which remains correct.

## Test plan

- [x] `test/serena/test_query_project_tools.py` — 4 in-process integration tests (no language server required; `search_for_pattern` is non-symbolic + read-only, so `_is_project_server_required` returns `False`):
  - happy path: context-excluded read-only tool forwards successfully (this fails on `main` with `AssertionError: Tool search_for_pattern is not active.`)
  - read-only gate: forwarding `create_text_file` raises `ValueError` matching `"read-only"`
  - global-config gate: `serena_config.excluded_tools` blocks forwarding with `ValueError` matching `"global"`
  - registration gate: unknown `project_name` raises `ValueError` matching `"not registered"`
  - Rejection tests pin to `ValueError` (not `AssertionError`) so a regression back to `assert` would fail loudly.
- [x] `test/serena/config/test_serena_config.py::TestToolInclusionDefinitionDisablesTool` — 3 unit tests covering default / incremental / fixed modes of the new helper.
- [x] `poe format` clean.
- [x] `poe type-check` clean.
- [x] CHANGELOG entry added under `# Unreleased (main)` / `* Tools:`.

### DIFF
--- a/src/serena/tools/query_project_tools.py
+++ b/src/serena/tools/query_project_tools.py
@@ -53,14 +53,22 @@ class QueryProjectTool(Tool, ToolMarkerOptional, ToolMarkerDoesNotRequireActiveP
         :param tool_params_json: the parameters to pass to the tool, encoded as a JSON string
         """
         tool = self.agent.get_tool_by_name(tool_name)
-        assert tool.is_active(), f"Tool {tool_name} is not active."
-        assert tool.is_readonly(), f"Tool {tool_name} is not read-only and cannot be executed in another project."
+        # explicit ValueError (not assert): safety/policy gates must not be strippable via `python -O`.
+        if not tool.is_readonly():
+            raise ValueError(f"Tool {tool_name} is not read-only and cannot be executed in another project.")
+        # the previous `is_active()` check conflated client-cosmetic context exclusions
+        # (e.g. search_for_pattern in the claude-code context, where the client has a native grep)
+        # with intentional global disabling. Forwarding a read-only tool to another project is
+        # valid even if the current context hides it; only the global config acts as kill-switch.
+        if self.agent.serena_config.disables_tool(tool_name):
+            raise ValueError(f"Tool {tool_name} is disabled in the global Serena configuration and cannot be executed.")
         if self._is_project_server_required(tool):
             client = ProjectServerClient()
             return client.query_project(project_name, tool_name, tool_params_json)
         else:
             registered_project = self.agent.serena_config.get_registered_project(project_name)
-            assert registered_project is not None, f"Project {project_name} is not registered and cannot be queried."
+            if registered_project is None:
+                raise ValueError(f"Project {project_name} is not registered and cannot be queried.")
             project = registered_project.get_project_instance(self.agent.serena_config)
             with tool.agent.active_project_context(project):
                 return tool.apply(**json.loads(tool_params_json))  # type: ignore
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ Status of the `main` branch. Changes prior to the next official version change w
   - Fix: Host validation required a local host regardless of the listen address (regression introduced in v1.5.2),
     preventing remote connections
 
+* Tools:
+  - Fix: `query_project` rejected read-only tools that the active context hides (e.g.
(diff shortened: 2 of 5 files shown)


src/serena/tools/query_project_tools.py
```diff
@@ -53,14 +53,22 @@ def apply(self, project_name: str, tool_name: str, tool_params_json: str) -> str
...
-        assert tool.is_readonly(), f"Tool {tool_name} is not read-only and cannot be executed in another project."
+        # explicit ValueError (not assert): safety/policy gates must not be strippable via `python -O`.
+        if not tool.is_readonly():
+            raise ValueError(f"Tool {tool_name} is not read-only and cannot be executed in another project.")
+        # the previous `is_active()` check conflated client-cosmetic context exclusions
+        # (e.g. search_for_pattern in the claude-code context, where the client has a native grep)
+        # with intentional global disabling. Forwarding a read-only tool to another project is
+        # valid even if the current context hides it; only the global config acts as kill-switch.
+        if self.agent.serena_config.disables_tool(tool_name):
+            raise ValueError(f"Tool {tool_name} is disabled in the global Serena configuration and cannot be executed.")
```
opcode81, line 64: The introduction of the new method does not make sense: Whether or not `SerenaConfig` disables a tool is meaningless.

It can make sense to allow tools that are not currently enabled, but there is an important limitation: The LLM will be aware of their existence (and know how to use them) only if they are exposed by the MCP server.
So removing the `is_active` condition will have a practical effect only if the tool was previously exposed but not enabled.
In all other cases, you would need to manually inform the LLM about these tools.

There is no need to add any additional condition if we just want to soften the condition that needs to apply.

opcode81 (changes requested):
Thanks for your PR, but I will need to reject this.
* As indicated in my comment, the new logic that is introduced does not make sense.
* There is no point in replacing AssertionErrors with ValueErrors.
* We do not need tests for this.
* The simplest approach is simply to remove the assertion that the tool must be active (noting that there are limitations regarding LLM awareness). You are welcome to propose a new PR with this one-line change plus changelog entry.

bor:
Thanks for the thorough review and the clear guidance, @opcode81 - much appreciated. Closing this PR; I'll open a fresh one with just the minimal change (drop the `is_active()` assertion) plus a changelog entry, as you suggested.

PR #1528, opened 2026-05-30
Add --parallel to `serena project index` for concurrent indexing

QuantFunc:
## Summary

`serena project index` indexes files one at a time (one `request_document_symbols` per file). On large projects this is slow — the bottleneck is **language-server round-trip latency, not local CPU**, so concurrent requests pipeline well.

This PR adds **`--parallel N`** to `serena project index` (`click.IntRange(min=1)`, **default 1 = unchanged serial behaviour**). For `N > 1`, a `ThreadPoolExecutor` issues concurrent document-symbol requests.

## What changed

**`src/serena/cli.py`**
- New `--parallel N` option on `project index`.
- The per-file loop is split into a worker (`index_one` → returns `(language, exception)`) and a main-thread accumulator (`record`). **Workers do no shared-state mutation**; the main thread does all counting / failure-list building, so there is no worker-side data race.
- On the parallel path the periodic intermediate save is intentionally skipped — running `save_all_caches()` (which iterates each LS's cache dict) while workers still write keys could raise `dict changed size during iteration`. The single `save_all_caches()` after the pool joins covers it.

**`src/solidlsp/ls.py`** — makes `SolidLanguageServer` safe to drive from multiple threads:
- Add a per-LS re-entrant `_state_lock` guarding the `open_file_buffers` bookkeeping and the document-symbol cache writes.
- The lock wraps **only in-process dict operations — never a language-server round-trip**. `open_file` constructs a new buffer with `open_in_ls=False` under the lock and sends the `didOpen` via `ensure_open_in_ls()` *after* releasing it; teardown captures the buffer under the lock and calls `fb.close()` (`didClose`) *outside* it. So the lock neither serializes concurrent requests nor creates a `_state_lock` ↔ stdin-lock ordering hazard.
- `open_file`'s ref-count decrement moves into a `try/finally` (also fixes a pre-existing ref-count leak when the `yield` body raised); the buffer delete is guarded by an identity check against concurrent re-creation.

## Backwards compatibility

`--parallel` defaults to 1 → the serial branch, which reproduces the original loop exactly. `_index_project` gains `parallel: int = 1`, so the existing `create --index` caller is unaffected. The `open_file` refactor preserves single-threaded semantics (`didOpen` timing unchanged via the idempotent `ensure_open_in_ls`).

## Measured

52-file C++ subtree (clangd):

| mode | rate | result |
|---|---|---|
| `--parallel 1` (serial) | ~2.25 it/s | cpp=52, 0 failures |
| `--parallel 8` | ~9.45 it/s (~4x) | cpp=52, 0 failures, no deadlock |

Cold-cache `--parallel 8` (exercises the new-buffer + `didOpen` path) and the serial run produce **byte-identical symbol caches (52/52 entries)** — parallel indexing yields the same result, just faster.

## Note

Document-symbol cache *writes* are under `_state_lock`; the cache *reads* are not yet (a follow-up). Under CPython this is safe for the `--parallel` distinct-file case (each worker writes a distinct key), which is verified by the byte-identical result above. Locking the reads too (for arbitrary concurrent callers / free-threaded builds) is a small follow-up.

## Usage

```
serena project index --parallel 8
```

### DIFF
--- a/src/serena/cli.py
+++ b/src/serena/cli.py
@@ -782,7 +782,16 @@ class ProjectCommands(AutoRegisteringGroup):
         help="Log level for indexing.",
     )
     @click.option("--timeout", type=float, default=10, help="Timeout for indexing a single file.")
-    def index(project: str, name: str | None, language: tuple[str, ...], log_level: str, timeout: float) -> None:
+    @click.option(
+        "--parallel",
+        type=click.IntRange(min=1),
+        default=1,
+        help="Number of files to index concurrently (must be >= 1). The default of 1 indexes serially. Values > 1 issue "
+        "document-symbol requests to the language server(s) concurrently via a thread pool, which can substantially speed "
+        "up indexing of large projects (the bottleneck is language-server round-trip latency, not local CPU). A starting "
+        "value of 4-8 is safe; increase while watching language-server memory/CPU usage.",
+    )
+    def index(project: str, name: str | None, language: tuple[str, ...], log_level: str, timeout: float, parallel: int) -> None:
         serena_config = SerenaConfig.from_config_file()
         registered_project = serena_config.get_registered_project(project, autoregister=True)
         if registered_project is None:
@@ -793,10 +802,10 @@ class ProjectCommands(AutoRegisteringGroup):
             except Exception as e:
                 raise click.ClickException(str(e))
 
-        ProjectCommands._index_project(registered_project, log_level, timeout=timeout)
+        ProjectCommands._index_project(registered_project, log_level, timeout=timeout, parallel=parallel)
 
     @staticmethod
-    def _index_project(registered_project: RegisteredProject, log_level: str, timeout: float) -> None:
+    def _index_project(registered_project: RegisteredProject, log_level: str, timeout: float, parallel: int = 1) -> None:
         lvl = logging.getLevelNamesMapping()[log_level.upper()]
         logging.configure(level=lvl)
         serena_config = SerenaConfig.from_config_file()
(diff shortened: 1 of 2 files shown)


AmirF194 (commented):
I ran this against `931ebac6f4de5074c1d60187e5eb6a63c6e0553b` in a clean container.

`index_one()` calls `ls_mgr.get_language_server(f)` from every worker thread in the `--parallel` pool. That reaches `_ensure_functional_ls()` -> `restart_language_server()` -> `_create_and_start_language_server()` (src/serena/ls_manager.py:149-180), which writes `self._language_servers[language] = language_server` with no lock; `LanguageServerManager` has no lock at all. This is a different object than the `_state_lock` this PR adds to `SolidLanguageServer`, which does not cover the manager's restart path. Before this PR `get_language_server()` was only ever called from one thread at a time.

If a language server dies mid-index, every worker thread for that language sees `is_running() == False` at once and each starts its own replacement, each overwriting the dict entry. Only the last write survives; the rest are started, unstopped processes.

Reproduced by loading the real `LanguageServerManager` from this branch (stubbing only the unrelated `serena.config.serena_config` / `solidlsp.ls` import chains) and driving `get_language_server()` from 8 threads against a manager whose server reports not running:

```
language servers created: 9
final manager._language_servers[PYTHON] is created[8]
stopped (had .stop() called): []
leaked: [0, 1, 2, 3, 4, 5, 6, 7]
```

5/5 runs leaked 8 of 9 created instances. For a real language server, each leaked instance is a live subprocess.

A lock around the check-then-act in `_ensure_functional_ls()` / `restart_language_server()`, the same shape as `_state_lock`, held only around the dict write and not across `.start()`, would fix it.

PR #1530, opened 2026-05-30
Fix gopls replace_symbol_body corrupting type/var/const declarations

### LINKED ISSUE #1529: [LS] replace_symbol_body corrupts Go type/var/const declarations by duplicating the leading keyword
## Summary

`replace_symbol_body` corrupts single Go `type`, `var`, and `const` declarations by duplicating the leading keyword. Replacing the body of a struct type with a keyword-inclusive body turns

```go
type Resampler struct { ... }
```

into

```go
type type Resampler struct { ... }
```

which no longer parses. The tool returns `{"result": "OK"}` for the corrupting edit (no error is reported), and the resulting syntax error wipes gopls's symbol tree, so subsequent symbolic edits on the same file fail with `No symbol matching ... found` for symbols that clearly exist.

## Root cause

gopls reports the document-symbol `range` of a single `type`/`var`/`const` declaration starting at the **identifier**, i.e. after the keyword, whereas it reports a `func` declaration's range starting at the `func` keyword. Serena derives both the displayed body and the replacement range from that range, so for these declarations the keyword is excluded from both, while it is included for funcs. That asymmetry is the bug: a body that includes the declaration's signature line (as the docs describe, and as funcs already behave) cannot be round-tripped.

## Reproduction

```go
// resampler.go
package resampler

type Resampler struct {
    inputRate  int
    outputRate int
}
```

1. `find_symbol(name_path="Resampler", relative_path="resampler.go", include_body=True)` returns the body `Resampler struct { ... }` (no `type` keyword, unlike funcs which include `func`).
2. `replace_symbol_body(name_path="Resampler", relative_path="resampler.go", body="type Resampler struct {\n\tinputRate int\n}")` returns `{"result": "OK"}` and writes `type type Resampler struct { ... }`.
3. The file no longer parses; subsequent symbolic edits on it fail.

`var` and `const` single declarations are affected identically. Grouped declarations (`var ( ... )`) are not affected, because the keyword sits on its own line.

## Relationship to prior fixes

This is the same class of range-vs-replacement mismatch as #1029 (fixed by #1030) and #1484. #1030 handled the trailing `end_col` for ten language servers but did not touch `gopls`, and it addressed the END of the range; this issue is about the START of the range (the leading keyword) for Go declarations specifically.

## Environment

- gopls v0.22.0, go 1.26.1
- Serena `main`

I have a fix with regression tests ready and will open a PR referencing this issue.

tphakala:
Fixes #1529

## Problem

gopls reports the document-symbol range of single `type`/`var`/`const` declarations starting at the identifier (after the keyword), unlike `func` declarations whose range includes the `func` keyword. Serena derives both the displayed symbol body and the `replace_symbol_body` replacement range from that range, so the keyword is dropped from both for these declarations. A keyword-inclusive edit then corrupts the file (`type Foo` becomes `type type Foo`), the tool returns `OK`, and the resulting syntax error breaks subsequent symbolic edits on the file.

## Fix

Override `request_document_symbols` in `gopls.py` to extend the range start of a single `type`/`var`/`const` declaration back to the keyword when only that keyword (plus optional indentation) precedes the identifier on the start line, mirroring the existing `nixd`/`fortran` post-processing pattern. The body is recomputed so the displayed body stays consistent with the replacement range. `func` declarations (range already at the keyword) and grouped declarations (`var ( ... )`, keyword on a separate line) are left untouched, and `selectionRange` is preserved so references and rename continue to point at the identifier.

The override operates on copies so the cached symbols are not mutated.

## Tests

- `test/solidlsp/go/test_go_basic.py::TestGoLanguageServer::test_type_var_const_body_includes_leading_keyword`: asserts the body and range of `type`/`var`/`const` symbols include the keyword, that `selectionRange` still points at the identifier, and that grouped vars are unaffected.
- `test/serena/test_symbol_editing.py::test_go_symbol_replacement_no_double_keyword`: an end-to-end `replace_symbol_body` round-trip that produces `type type` on the old code and passes with the fix.

Both tests fail against the pre-fix code and pass with the fix. `poe format`, `poe type-check`, and the Go test suite are green (gopls v0.22.0, go 1.26.1).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,11 @@ Status of the `main` branch. Changes prior to the next official version change w
     and other JDK types reported as "cannot be resolved") for projects whose source/target level
     exceeds the bundled JDK 21 JRE JDT-LS registers by default; configured runtimes extend rather
     than replace that bundled default. #1478
+  - `gopls`: Fix `replace_symbol_body` corrupting single `type`/`var`/`const` declarations by
+    duplicating the leading keyword (e.g. `type Foo` becoming `type type Foo`). gopls reports the
+    symbol range of such declarations starting at the identifier rather than the keyword (unlike
+    `func` declarations); the range is now extended to include the keyword so the body and the
+    replacement range stay consistent.
 
 * Tools:
   - Fix: `search_for_pattern` marked one line too many as matched whenever a match ended with a line
--- a/src/solidlsp/language_servers/gopls.py
+++ b/src/solidlsp/language_servers/gopls.py
@@ -2,13 +2,15 @@ import hashlib
 import json
 import logging
 import os
+import re
 import subprocess
 from collections.abc import Hashable
 from typing import Any
 
 from overrides import override
 
-from solidlsp.ls import RawDocumentSymbol, SolidLanguageServer
+from solidlsp import ls_types
+from solidlsp.ls import DocumentSymbols, LSPFileBuffer, RawDocumentSymbol, SolidLanguageServer, SymbolBodyFactory
 from solidlsp.ls_config import LanguageServerConfig
 from solidlsp.lsp_protocol_handler.server import ProcessLaunchInfo
 from solidlsp.settings import SolidLSPSettings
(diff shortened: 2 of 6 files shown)


tphakala:
The failing CI check is unrelated to this change.

The only failing test is `test/solidlsp/svelte/test_svelte_basic.py::TestSvelteLanguageServer::test_diagnostics_in_typescript_file`, which asserts on the Svelte language server's diagnostic output:

```
AssertionError: ['Unexpected token\nhttps://svelte.dev/e/js_parse_error', 'Expected token }\nhttps://svelte.dev/e/expected_token']
```

This PR only touches the Go language server (`gopls.py`) and Go tests, so it cannot affect Svelte diagnostics. All Go tests pass on every runner, including the two added here (`test_type_var_const_body_includes_leading_keyword` and `test_go_symbol_replacement_no_double_keyword`). The per-OS summaries are "1 failed, 1472 passed" on ubuntu and "1 failed, 1464 passed" on macos, and in both cases the single failure is the Svelte test.

This looks like a CI environment issue rather than a regression: the same test fails the same way on other unrelated open PRs (for example #1528), and it was still passing on `main` on 2026-05-27. CI installs `svelte-language-server` at `latest`, so a newer version changing its diagnostic messages is the likely cause.

opcode81:
I rebased your changes onto main, updating the change log, and added a missing update of the cache fingerprint computation.

opcode81:
Thank you for your contribution.

PR #1531, opened 2026-05-30
fix(nix): Add Linux-support for the tray manager in AppIndicator trays

futile:
## What this means for users

With this PR, users of Serena's Nix package can use `web_dashboard_interface: tray_manager` on Linux desktops whose tray uses StatusNotifier/AppIndicator, including common Wayland setups.
Existing Xorg/XEmbed tray behavior is not removed, and non-Linux package outputs are unchanged.

This PR only touches `flake.nix`.

## Background

In Wayland-based environments, which usually use StatusNotifier/AppIndicator trays, `web_dashboard_interface: tray_manager` needs pystray's AppIndicator backend.
The current Nix package does not provide this, and pystray instead falls back to its Xorg/XEmbed backend.
The result is that no tray icon will appear for StatusNotifier/AppIndicator trays, even if the tray manager started successfully.
Explicitly setting `PYSTRAY_BACKEND=appindicator` exposes the failure, causing the tray manager to fail startup, because the Python GI bindings and GTK/AppIndicator typelibs are not available in the virtualenv/wrapper, and it is not allowed to fall back to other backends.

## Reproducing the Issue

Build Serena using Nix (under Linux), for example with `nix build .#serena --no-link --print-out-paths`, run:

```sh
serena_out=/nix/store/...-serena
serena_wrapper="$serena_out/bin/serena"
serena_entry="$(strings "$serena_wrapper" | grep '/serena-env/bin/serena$' | head -1)"
serena_python="$(head -1 "$serena_entry" | sed 's|^#!||')"

export PYSTRAY_BACKEND=appindicator
export GI_TYPELIB_PATH="$(strings "$serena_wrapper" | grep '^/nix/store/.*/lib/girepository-1.0' | head -1)"
export XDG_DATA_DIRS="$(strings "$serena_wrapper" | grep '^/nix/store/.*/share/gsettings-schemas' | head -1)"
export GIO_EXTRA_MODULES="$(strings "$serena_wrapper" | grep '^/nix/store/.*/lib/gio/modules' | paste -sd: -)"

"$serena_python" - <<'PY'
import pystray
print(f"pystray backend: {pystray.Icon.__module__}")
from gi.repository import Gtk, AyatanaAppIndicator3
print("gtk/appindicator typelibs: ok")
from serena.dashboard import SerenaDashboardTrayManager
manager = SerenaDashboardTrayManager()
print(f"constructed: {manager.__class__.__name__}")
PY
```

This reconstructs the relevant environment from the Nix-built `serena` wrapper and confirms that pystray selects the AppIndicator backend and that GTK/Ayatana typelibs are visible.

## Additional Environment Note

In sandboxed MCP hosts such as Codex, the Serena process may not inherit desktop session variables even when the parent desktop has them. For the tray manager to interact with the desktop session, the launcher may need to pass through variables such as:

- `DISPLAY`
- `WAYLAND_DISPLAY`
- `XDG_CURRENT_DESKTOP`
- `DESKTOP_SESSION`
- `XDG_SESSION_DESKTOP`
- `XDG_SESSION_TYPE`
- `XDG_RUNTIME_DIR`
- `DBUS_SESSION_BUS_ADDRESS`

That environment setup is separate from this Nix packaging fix.
The packaging change makes the AppIndicator backend importable and correctly wrapped; the launcher still has to provide access to the user's graphical/session D-Bus environment.
In my Nix config I provide this environment through a simple wrapper script: https://github.com/futile/nixos-config/blob/27b3f6bd35537cf1dbb755cef9e2547b3f4f13d1/bin/serena-mcp-with-desktop-env

This is NOT anything that needs to (or can) be handled in this repo, but it might be necessary to reproduce the issue and fix successfully, and might be relevant for other Linux/Codex users as well.

## Additional Option/Follow-Up Possibility

For non-Nix installs, we could also (in this PR or a follow-up) add a Python optional extra such as `tray-appindicator = ["pygobject"]`, although I'm not familiar with this at all to give more info/trade-offs on this.

While that could help document/install the Python-level dependency, it would not replace this Nix change because GTK/AppIndicator native libraries and typelibs still need to be present in the runtime environment.

## Validation

- `nix fmt flake.nix`
- `nix flake check`
- `nix build .#serena --no-link --print-out-paths`
- Reproduction from above succeeds, output:
  - `pystray backend: pystray._appindicator`
  - `gtk/appindicator typelibs: ok`
  - `constructed: SerenaDashboardTrayManager`

---

Assisted-by: gpt-5.5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Status of the `main` branch. Changes prior to the next official version change w
   - Decouple configuration computation from the agent's task queue by introducing events for agent config/status updates.
     This allows the dashboard to display the configuration while the project provided at startup is still initialising. #1064
   - Fix empty executions queue displaying "Loading..."
+  - Tray manager: Add NixOS-support for AppIndicator-based trays (e.g., most Wayland-trays) to the package in flake.nix.
 
 # v1.5.3 (2026-05-26)
 
--- a/docs/02-usage/060_dashboard.md
+++ b/docs/02-usage/060_dashboard.md
@@ -25,7 +25,7 @@ The dashboard is a web application, which is opened in one of three ways:
   * via a tray manager that aggregates multiple instances (on supported platforms) 
   * via your default web browser (supported on all platforms)
 
-Configure your preferred interface via the setting `web_dashoard_interface` in [Serena's global configuration file](global-config) (see below).
+Configure your preferred interface via the setting `web_dashboard_interface` in [Serena's global configuration file](global-config) (see below).
 
 By default, the dashboard can be accessed at `http://localhost:24282/dashboard/index.html`,
 but a higher port may be used if the default port is unavailable/multiple instances are running.
--- a/flake.nix
+++ b/flake.nix
@@ -50,27 +50,79 @@
         sourcePreference = "wheel"; # or sourcePreference = "sdist";
       };
 
-      pyprojectOverrides = final: prev: {
-        # Add setuptools for packages that need it during build
-        ruamel-yaml-clib = prev.ruamel-yaml-clib.overrideAttrs (old: {
-          nativeBuildInputs = (old.nativeBuildInputs or []) ++ [
-            final.setuptools
-          ];
-        });
-        proxy-tools = prev.proxy-tools.overrideAttrs (old: {
-          nativeBuildInputs = (old.nativeBuildInputs or []) ++ [
-            final.setuptools
-          ];
-        });
-        pywebview = prev.pywebview.overrideAttrs (old: {
-          nativeBuildInputs = (old.nativeBuildInputs or []) ++ [
-            final.setuptools
-            final.setuptools-scm
-          ];
-        });
-      };
-
       python = pkgs.python311;
+      pyprojectHacks = pkgs.callPackage pyproject-nix.build.hacks {};
+
+      pyprojectOverrides = final: prev:
+        {
+          # Add setuptools for packages that need it during build
+          ruamel-yaml-clib = prev.ruamel-yaml-clib.overrideAttrs (old: {
+            nativeBuildInputs =
+              (old.nativeBuildInputs or [])
+              ++ [
+                final.setuptools
+              ];
+          });
+          proxy-tools = prev.proxy-tools.overrideAttrs (old: {
+            nativeBuildInputs =
+              (old.nativeBuildInputs or [])
+              ++ [
+                final.setuptools
+              ];
+          });
+          pywebview = prev.pywebview.overrideAttrs (old: {
+            nativeBuildInputs =
+              (old.nativeBuildInputs or [])
+              ++ [
+                final.setuptools
+                final.setuptools-scm
+              ];
+          });
+        }
+        // lib.optionalAttrs pkgs.stdenv.isLinux {
+          # pyproject-nix's virtualenv resolver follows pyproject-style
+          # passthru.dependencies, while these GI bindings are already packaged
+          # by nixpkgs' Python infrastructure. nixpkgsPrebuilt is the upstream
+          # adapter for that case:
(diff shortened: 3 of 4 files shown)


futile:
I looked at the Ubuntu test failures, and they seem to be possibly unrelated? They are inside `solidlsp` tests for the Nix and Python LSPs, iiuc. Please let me know whether I should take a deeper look at them or not.

opcode81:
I am not a Nix user, so I cannot properly review this.
Can this cause potentially cause problems for other Nix users (e.g. due to assumptions being made that do not hold for their environment)?
If not, then I would probably just merge this. The only thing we should adjust is the documentation (in the config template and the dashboard documentation page), such that users are aware of this.

> In sandboxed MCP hosts such as Codex, the Serena process may not inherit desktop session variables even when the parent desktop has them.

This is a bit of a usability issue. I think it might make sense to add a Serena CLI command `serena start-tray-manager` that will start the manager (without terminating if there are no instances). Then users could start manually from an environment that has the necessary variables.

futile:
Thank you for taking a look!

I wrote a longer reply before this, but then lost that text, so I'm keeping things a bit shorter now, I hope that's ok :see_no_evil: 

> I am not a Nix user, so I cannot properly review this. Can this cause potentially cause problems for other Nix users (e.g. due to assumptions being made that do not hold for their environment)? If not, then I would probably just merge this.

Behavior should not break, but might change, for a specific set of users: Users with an AppIndicator tray that used a compatibility layer for the older/legacy XEmbed tray were already able to use the tray manager through that before this PR. With this PR, Serena (i.e., pystray) will start to use the AppIndicator tray integration directly, instead of falling back to XEmbed. So it should still work, but will take the more modern path, so it _should_ actually be an upgrade for such users. If somebody really needs to go back, using `PYSTRAY_BACKEND=<previous-fallback-backend>` should fully restore the old behavior for them locally.

While in Linux-land you can never 100% avoid breaking someone's setup ([relevant xkcd](https://xkcd.com/1172/)), I think the upside of this PR of making the tray manager and icon work for _many_ more people makes it a clear improvement.

> The only thing we should adjust is the documentation (in the config template and the dashboard documentation page), such that users are aware of this.

Agree, will let an LLM guide me to the parts of the docs you pointed out, and add it there :)

> > In sandboxed MCP hosts such as Codex, the Serena process may not inherit desktop session variables even when the parent desktop has them.
> 
> This is a bit of a usability issue. I think it might make sense to add a Serena CLI command `serena start-tray-manager` that will start the manager (without terminating if there are no instances). Then users could start manually from an environment that has the necessary variables.

Also agree, but I honestly would rather leave this as a follow-up, because I don't really want to work on it (now) :see_no_evil: While the implementation itself should probably be straight-forward for an LLM agent, the doc changes, ensuring cross-platform support, reviews, etc. are a bit more than I want to tackle right now.

futile:
Added docs changes, which mostly add a mention of NixOS-support in the config template, and fix a typo in the dashboard docs. The dashboard docs for the tray manager mention neither NixOS nor Linux explicitly so far, and instead use more general language like "on supported platforms", which I thought is still appropriate, so I didn't change it.

opcode81:
> Also agree, but I honestly would rather leave this as a follow-up

Yes, follow-up was intended. I mainly wanted to provide some input as to how the UX issue could be resolved.

> The dashboard docs for the tray manager mention neither NixOS nor Linux explicitly so far, and instead use more general language like "on supported platforms", which I thought is still appropriate, so I didn't change it.

Yes, this was the right call.

opcode81:
Thank you very much for your additions. The only thing missing for this to be merged is an entry in the change log.

futile:
Added a changelog entry, I hope it's in line with the style you like! :)

opcode81:
Thanks. Made a slight adjustment in the changelog. Merging.

> While in Linux-land you can never 100% avoid breaking someone's setup ([relevant xkcd](https://xkcd.com/1172/))

So true. :-)

PR #1540, opened 2026-06-03
fix: emit valid codex stop-hook cleanup output

### LINKED ISSUE #1533: Codex Stop hook: `serena-hooks cleanup --client=codex` emits invalid Stop-hook stdout
## Summary

```text
Stop hook (failed)
```

```json
{
  "hooks": {
    "Stop": [
      {
        "hooks": [
          {
            "type": "command"
          }
        ]
      }
    ]
  }
}
```

## Environment

- Client: Codex
- OS: macOS

```bash
```

Observed:

```text
# exit code: 0
# stdout: empty
```

```json
{"continue": true}
```

```bash
```

Observed traceback:

```text
    sys.exit(hook_commands())
             ~~~~~~~~~~~~~^^
           ~~~~~~~~~^^^^^^^^^^^^^^^^^
    rv = self.invoke(ctx)
```

## What Codex expects

```json
{"continue": true}
```

## Expected behavior

For example:

```json
{"continue": true}
```

## Actual behavior

## Local workaround

```sh
```

## Related

zerone0x:
fixes #1533.

for codex cleanup hooks, this now always emits valid stop-hook json (`{"continue": true}`) and treats a missing session id as a no-op instead of raising.

### DIFF
--- a/src/serena/hooks.py
+++ b/src/serena/hooks.py
@@ -516,8 +516,22 @@ class SessionStartActivateProjectHook(Hook):
 
 
 class SessionEndCleanupHook(Hook):
+    def __init__(self, client: HookClient):
+        raw = sys.stdin.read()
+        input_data = json.loads(raw) if raw else {}
+        self._input_data = input_data
+        self._client = client
+
+        session_id = input_data.get("session_id") or input_data.get("sessionId")
+        self._session_id = str(session_id) if session_id else None
+        self.session_persistence_dir = os.path.join(serena_home_dir, "hook_data", self._session_id) if self._session_id is not None else ""
+        self.triggered_at_timestamp = datetime.now()
+
     def execute(self) -> None:
-        shutil.rmtree(self.session_persistence_dir, ignore_errors=True)
+        if self._session_id is not None:
+            shutil.rmtree(self.session_persistence_dir, ignore_errors=True)
+        if self._client == HookClient.CODEX:
+            click.echo(json.dumps({"continue": True}))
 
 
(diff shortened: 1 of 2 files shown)


src/serena/hooks.py
```diff
@@ -516,8 +516,22 @@ def execute(self) -> None:
...
+        session_id = input_data.get("session_id") or input_data.get("sessionId")
+        self._session_id = str(session_id) if session_id else None
+        self.session_persistence_dir = os.path.join(serena_home_dir, "hook_data", self._session_id) if self._session_id is not None else ""
+        self.triggered_at_timestamp = datetime.now()
+
     def execute(self) -> None:
-        shutil.rmtree(self.session_persistence_dir, ignore_errors=True)
+        if self._session_id is not None:
+            shutil.rmtree(self.session_persistence_dir, ignore_errors=True)
+        if self._client == HookClient.CODEX:
```
MischaPanch, line 533: Is it a bad idea to always echo this?

src/serena/hooks.py
```diff
@@ -516,8 +516,22 @@ def execute(self) -> None:
 
 
 class SessionEndCleanupHook(Hook):
+    def __init__(self, client: HookClient):
```
MischaPanch, line 519: super init is missing. The hooks assume that session_id is always passed, is that not true for codex? Serena's hooks don't work without a session id.

MischaPanch (changes requested):
The issue describes the case "no session id sent" but this case should never occur. Does this actually happen with codex? If yes, none of our hooks would work for codex

opcode81:
No response from OP. Closing.

PR #1544, opened 2026-06-06
feat: forward user-provided initializationOptions to pasls

zen010101:
## What
Forward user-provided `initializationOptions` to `pasls`, so users can pass
arbitrary pasls options (e.g. `fpcOptions`) via Serena's
`ls_specific_settings["pascal"]["initializationOptions"]`.

## Why
Serena currently sends only a hard-coded subset of `initializationOptions`
to pasls. Editor LSP clients (Sublime LSP, VSCode) let users forward
arbitrary options. Without this, there is no way to tell pasls about
third-party unit/include search paths (e.g. mORMot), which breaks symbol
resolution for projects that depend on external FPC libraries.

## Change
- Read `initializationOptions` from `ls_specific_settings["pascal"]`.
- Merge them into the initialize params, user values taking precedence.
- `_get_initialize_params` becomes an instance method to access the override.

## Compatibility
Backward compatible: with no `initializationOptions` configured, behavior
is unchanged.

### DIFF
--- a/src/solidlsp/language_servers/pascal_server.py
+++ b/src/solidlsp/language_servers/pascal_server.py
@@ -177,6 +177,12 @@ class PascalLanguageServer(SolidLanguageServer):
         )
         self.server_ready = threading.Event()
 
+        # Generic pass-through of LSP initializationOptions, mirroring how editor
+        # LSP clients (Sublime LSP, VSCode) let users forward arbitrary options to
+        # pasls -- e.g. fpcOptions for third-party unit/include search paths such
+        # as mORMot. Configured via ls_specific_settings["pascal"]["initializationOptions"].
+        self._init_options_override: dict = pascal_settings.get("initializationOptions", {}) or {}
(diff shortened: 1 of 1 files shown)


MischaPanch:
@opcode81 we could in principle centralize this and allow it for all LS, at the user's own risk, of course. WDYT?

opcode81:
Yes, but it would, of course, require refactoring all language servers, since the abstraction doesn't currently support centralising this in any way.

opcode81:
@zen010101 thanks for your PR. This is on hold as we consider centralising this in order to support this for all language servers.

zen010101:
ok

opcode81:
This is now supported for all language servers, see commit 352b3d3fed7d4bb3201ad5c0b0f5bb1794945f7e.
https://oraios.github.io/serena/02-usage/050_configuration.html#overriding-language-server-initialization-options

PR #1550, opened 2026-06-07
Fix find_project_root hijacking git worktrees nested under a Serena project

tphakala:
## Problem

`find_project_root()` (used by `--project-from-cwd`, the default launch mode for the Claude Code / Codex / Gemini contexts) walks up from CWD in **two passes**: it returns the first ancestor containing `.serena/project.yml`, and only if none is found does it look for `.git`.

That ordering breaks git worktrees. A common layout puts worktrees inside the repo, e.g. `<repo>/.claude/worktrees/<name>` (Claude Code creates them there natively). The worktree's own `.git` is a pointer file, and `.serena/` is usually gitignored, so the worktree itself has no `.serena/project.yml`. The first pass then climbs past the worktree and matches the parent repo's `.serena/project.yml`, even though a `.git` boundary (the worktree) sits closer. The second pass, which already handles worktree `.git` files, never runs.

The result: when a session starts from inside the worktree (the MCP server's CWD is the worktree), Serena activates the **parent repo** instead. All relative-path file operations then resolve against the parent tree, so symbol reads return stale content and edits are written to the wrong working tree. Under the `claude-code` context `activate_project` is excluded, so this cannot be corrected at runtime.

## Fix

Collapse the two passes into a single walk-up so the **nearest** project boundary wins, whether it is a `.serena/project.yml` or a `.git` (directory, or a worktree/submodule pointer file).

Same-level behavior is unchanged: a directory containing both still resolves to itself. The only behavior change is the cross-level case this bug is about: a closer `.git` now wins over an ancestor's `.serena/project.yml`, which is the correct outcome, since crossing a git boundary means you have left the parent project.

## Tests

Adds `test_git_worktree_not_hijacked_by_ancestor_serena`, which reproduces the worktree-under-Serena-project layout (nested `.git` pointer file under an ancestor with `.serena/project.yml`) and asserts resolution to the worktree. It fails on the old two-pass code and passes with the fix. All existing `TestFindProjectRoot` cases stay green, including `test_serena_preferred_over_git`.

`poe format` and `poe type-check` are clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 Status of the `main` branch. Changes prior to the next official version change will appear here.
 
+* General:
+  - Fix `--project-from-cwd` hijacking git worktrees nested under a Serena project. `find_project_root`
+    now walks up in a single pass so the nearest project boundary wins (either a `.serena/project.yml`
+    or a `.git`, including worktree/submodule pointer files), instead of preferring an ancestor's
+    `.serena/project.yml` over a closer `.git`. This previously bound CLI agents (Claude Code, Codex,
+    Gemini) launched from inside a worktree to the parent repo, causing stale reads and misdirected edits.
+
 * Language Servers:
   - `typescript_vts`: Add `initialization_options` setting in `ls_specific_settings.typescript_vts`.
     The dict is forwarded to vtsls via `initializationOptions`, `workspace/didChangeConfiguration`,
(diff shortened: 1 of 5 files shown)


opcode81:
Thanks for your contribution. Please ensure that
* documentation is updated appropriately,
* the changes are documented in the changelog.

tphakala:
Done. Pushed a commit with both:

- Changelog: added a `General:` entry under *Unreleased (main)* describing the `--project-from-cwd` worktree fix.
- Docs: clarified the `--project-from-cwd` resolution rule in `docs/02-usage/020_running.md` (nearest project boundary wins) and added a note to the *Serena and Git Worktrees* section in `docs/02-usage/999_additional-usage.md` explaining that launching from inside a nested worktree now resolves to the worktree itself.

`poe format` and `poe type-check` remain clean.

opcode81:
Thanks again for your contribution! Merging ...

PR #1553, opened 2026-06-08
Fix: never persist runtime config overrides during project registration

MischaPanch:
## By Mischa:

The regression test added by claude uses monkeypatching which I don't like (a change to a private var name will make the test go rogue). We can delete it or keep it, I am ambivalent. We could also delete it once (if) it fails or creates a problem.
We can also make it use SERENA_HOME instead of monkeypatching, but since we might want to delete it, I haven't done that

## By Claude:

Bug: `serena start-mcp-server --language-backend X --project <newdir>` silently rewrote the global serena_config.yml's language_backend to X. Cause: a not-yet-registered --project is auto-registered, and SerenaConfig.save() rewrote the WHOLE file from the in-memory config — which by then carried the CLI --language-backend override (and would equally carry --log-level / --tool-timeout / etc.).

Fix: project (de)registration now persists ONLY the projects list. add_registered_project / remove_project go through a new _persist_projects(), which re-loads the on-disk config, copies in the current project list, and saves that — so a transient runtime override of any field can never be written back.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@ Status of the `main` branch. Changes prior to the next official version change w
     or a `.git`, including worktree/submodule pointer files), instead of preferring an ancestor's
     `.serena/project.yml` over a closer `.git`. This previously bound CLI agents (Claude Code, Codex,
     Gemini) launched from inside a worktree to the parent repo, causing stale reads and misdirected edits.
-  - Fix: some CLI flags on start-mcp-server modified the global config (notably, language-backend). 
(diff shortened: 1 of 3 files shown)


opcode81:
I will extend this PR.

opcode81:
While your commit fixed the problem of project list updates incorrectly re-saving the entire configuration (which potentially contained transient overrides), it didn't address the issue that `save` really must not be a public method.

* I made it private
* I refactored its use in the `init` CLI command
* I improved the documentation
* I removed the test from your commit
* I improved the changelog entry

MischaPanch:
Ok, great, thank you for that! Feel free to merge at your convenience

PR #1560, opened 2026-06-10
feat: support directory paths in get_symbols_overview with max_files safeguard (#1412)

### LINKED ISSUE #1412: Feature request: support directory paths in get_symbols_overview for batch retrieval
## Summary

`get_symbols_overview` currently accepts only file paths and returns `ValueError: Expected a file path` when called with a directory. Add support for directory inputs that returns top-level symbols for every file in the directory (recursively or one level deep).

## Use case

When exploring a large codebase via Serena (e.g. understanding a C# module that spans 25+ files in one directory), an agent currently has to:

1. Call `find_file` or list the directory contents some other way.
2. Loop over every `.cs` file and call `get_symbols_overview` per file.

This is N round-trips and a lot of agent reasoning to do something the LSP server could batch on its own. For agents using Serena to write architecture overviews or plan changes touching a whole module, this is the most common navigation pattern, and the current per-file requirement is the slowest part.

## Proposed behaviour

```
get_symbols_overview(relative_path="src/Core")
→ {
    "src/Core/SessionMonitor.cs": [...top-level symbols...],
    "src/Core/CpuSampler.cs": [...],
    ...
  }
```

Optional flags worth considering:
- `recursive: bool = False` — descend into subdirectories.
- `glob: str = None` — filter files (e.g. `"*.cs"`).
- `max_files: int = 100` — protect against accidental whole-repo dumps.

## Workaround today

```python
files = find_file("*.cs", relative_path="src/Core")
for f in files:
    get_symbols_overview(relative_path=f)
```

Works, but in real agent runs this either burns turns or pushes the agent to fall back to `Read` for the whole file, which defeats the purpose of using Serena for symbolic navigation.

## Notes

- LSP's `documentSymbol` is per-document, but `workspace/symbol` exists. A directory variant could either iterate `documentSymbol` server-side or use `workspace/symbol` with a path filter.
- Searched open and closed issues for `get_symbols_overview directory` / `get_symbols_overview folder` — no existing report. Issues #1187, #853, #1203 touched dot-prefixed path handling, which is unrelated.

weiconghe:
## Issue

Closes #1412

## Type of change

- [x] New feature

### `GetSymbolsOverviewTool` now supports directory paths, guarded by `max_files` (Issue #1412)

**Problem:** `get_symbols_overview` raised `ValueError: Expected a file path, but got a directory path` when given a directory. Agents had to list directory contents, then call the tool per file — N round-trips for a common navigation pattern.

**Root cause:** The underlying LSP layer (`request_overview` in `solidlsp/ls.py`) already handled directories via `request_dir_overview`, but the tool layer in `symbol_tools.py` explicitly rejected directory paths.

**Maintainer concern (from @MischaPanch on the issue):** A previous variant allowed directories, but it was deliberately restricted because models tended to start with the top-level directory, producing huge outputs. The maintainer hinted that a `max_files` option could mitigate this.

**Fix:**
1. Added `_apply_directory` method that delegates to the existing LSP directory support. When a directory path is given, the tool returns a per-file mapping of grouped symbols:
```json
{
  "test_repo/services.py": {"Function": ["create_service_container", ...]},
  "test_repo/models.py": {"Class": ["User"]}
}
```
Single-file behavior is completely unchanged.

2. Added a `max_files` parameter (default `20`). When a directory contains **more** analyzable files than the limit, the tool raises `ValueError` *before* serializing — the error message shows the actual count, a sample of file paths, and tells the agent to narrow the path or learn the layout from memories first. This directly addresses @MischaPanch's concern: the agent can't accidentally dump the top-level directory, and gets actionable guidance when it tries.

```python
ValueError: Directory src contains 240 analyzable files, which exceeds max_files=20.
Narrow the path to a more specific subdirectory, or learn the repository layout from
memories before asking for a broad overview. Sample files found: ['src/App.tsx', ...]
```

Also added a `CHANGELOG.md` entry per `CONTRIBUTING.md`.

## How did you verify your code works?

**Environment:** Windows 11, Python 3.13, uv 0.11.13

**New tests (all passing):**
1. `test_get_symbols_overview_directory_returns_per_file_symbols` — passes a directory path, asserts the result is a per-file mapping with `.py` file keys
2. `test_get_symbols_overview_file_returns_same_format` — regression test, single-file output format is unchanged (not wrapped in per-file mapping)
3. `test_get_symbols_overview_directory_raises_when_exceeds_max_files` — passes `max_files=1` against a directory with multiple files, asserts `ValueError` with `max_files=1` in the message
4. `test_get_symbols_overview_nonexistent_path_raises` — asserts `FileNotFoundError` for a nonexistent file and a nonexistent directory path alike

**Regression:** All Python-language tests in `TestSerenaAgent` pass. The Go / C# / Rust / Kotlin / PHP / PowerShell failures in my local environment are caused by missing language-server binaries (`gopls`, `csharp-ls`, `kotlin-language-server`, `intelephense`, …); running the same test set on a clean `main` checkout fails identically, so they are unrelated to this PR. The one additional local failure (`test_find_symbol[java_model_class]`, an LSP-indexing race) passed on re-run, and the full Java test group passes.

## Screenshots / recordings

N/A

## Checklist

- [x] I have not included unrelated changes in this PR

### DIFF
--- a/src/serena/tools/symbol_tools.py
+++ b/src/serena/tools/symbol_tools.py
@@ -35,25 +35,39 @@ class RestartLanguageServerTool(Tool, ToolMarkerOptional):
 
 class GetSymbolsOverviewTool(Tool, ToolMarkerSymbolicRead):
     """
-    Gets an overview of the top-level symbols defined in a given file.
+    Gets an overview of the top-level symbols defined in a given file or directory.
     """
 
     symbol_dict_grouper = LanguageServerSymbolDictGrouper(["kind"], ["kind"], collapse_singleton=True)
 
-    def apply(self, relative_path: str, depth: int = -1, max_answer_chars: int = -1) -> str:
+    def apply(self, relative_path: str, depth: int = -1, max_answer_chars: int = -1, max_files: int = 20) -> str:
         """
-        Use this tool to get a high-level understanding of the code symbols in a file.
+        Use this tool to get a high-level understanding of the code symbols in a file or directory.
         This should be the first tool to call when you want to understand a new file, unless you already know
         what you are looking for.
+        When given a directory path, returns top-level symbols for every analyzable file in the directory.
 
-        :param relative_path: the relative path to the file to get the overview of
-        :param depth: depth up to which descendants shall be retrieved.
-            Default (-1) results in a language specific choice: 1 for java and kotlin and 0 for other languages
+        :param relative_path: the relative path to the file or directory to get the overview of
+        :param depth: depth up to which descendants of top-level symbols shall be retrieved
+            (e.g. 1 retrieves immediate children). Default (-1) results in a language specific
+            choice: 1 for java and kotlin and 0 for other languages.
         :param max_answer_chars: if the overview is longer than this number of characters,
             no content will be returned. -1 means the default value from the config will be used.
             Don't adjust unless there is really no other way to get the content required for the task.
+        :param max_files: only used when relative_path is a directory. If the directory contains more
+            analyzable files than this limit, the tool raises ValueError instead of returning a partial
+            overview — narrow the path to a subdirectory, or learn the layout from memories first.
+            Default 20. Don't increase unless you really need a broad sweep and accept the token cost.
         :return: a JSON object containing symbols grouped by kind in a compact format.
+            For directories, returns a mapping of file paths to their grouped symbols.
         """
+        file_path = os.path.join(self.project.project_root, relative_path)
+        if not os.path.exists(file_path):
+            raise FileNotFoundError(f"File or directory {relative_path} does not exist in the project.")
+
+        if os.path.isdir(file_path):
+            return self._apply_directory(relative_path, depth=depth, max_answer_chars=max_answer_chars, max_files=max_files)
+
         if depth == -1:
             if relative_path.endswith((".java", ".kt")):
                 depth = 1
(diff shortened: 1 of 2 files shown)


weiconghe:
Hi, I noticed the CI checks failed on this PR, but the failures don't appear to be related to my changes:

- **Ubuntu**: Failed during `conftest.py` import — `clojure -Spath` returned a non-zero exit code. This is a Clojure CLI environment setup issue in the CI runner, occurring before any test runs.
- **Windows**: `test_find_symbol_references[typescript_helper_refs]` failed — the TypeScript LSP returned a reference in `index.ts` instead of the expected `use_helper.ts`. Out of 1386 tests, only this one failed, and it's unrelated to the directory path / `FileNotFoundError` fix in this PR.
- **macOS**: Still pending.

Could you kindly rerun the failed jobs? Thanks!

MischaPanch:
Hi @weiconghe . We have been thinking about the best way to implement this and will get back to you. Either building on your PR or implementing it in a different way. The topic is a bit more subtle than one might think first

weiconghe:
Rebased onto current `main` and resolved the conflicts (the `LanguageServerId` rename and the file-sync changes touched the same regions of `symbol_tools.py`). Also added a `CHANGELOG.md` entry per `CONTRIBUTING.md`, and an extra test covering nonexistent paths. The CLA is signed.

CI note: the only failing check is `niche (ubuntu-latest)`, where the two `test_r_basic.py` cross-file tests on R return empty results — the same job fails on `main` at the moment, so it's unrelated to this change.

Happy to adapt if you've settled on a different approach for this — no rush.

src/serena/tools/symbol_tools.py
```diff
@@ -91,6 +104,49 @@ def make_depth_0_result() -> str:
 
         return self._limit_length(result_json_str, max_answer_chars, shortened_result_factories=shortened_results)
 
+    def _apply_directory(self, relative_path: str, depth: int = 0, max_answer_chars: int = -1, max_files: int = 20) -> str:
+        symbol_retriever = self.create_language_server_symbol_retriever()
+        path_to_symbols = symbol_retriever.get_symbol_overview(relative_path)
```
AmirF194: I ran this against `67857c38b8cc9a70f9b7dc9b73cc8d4c4db9c55b` in a clean container, instrumenting `SolidLanguageServer.request_document_symbols` to count LSP calls.

`_apply_directory` builds `path_to_symbols` with the full `get_symbol_overview(relative_path)` call before checking `max_files` (`symbol_tools.py:109-112`). That call walks the whole directory tree and issues one `textDocument/documentSymbol` request per file, so `max_files` only bounds the size of the JSON that gets returned, not the LSP work done to build it. Against `max_files=1` on the `test_repo` fixture (11 files), the guard still let all 11 document-symbol requests happen before raising:

```
request_document_symbols call count with max_files=1: 11
```

That's the same cost the PR describes fixing: MischaPanch's concern on #1412 was an agent dumping the top-level directory, and a real repo directory with hundreds of files will issue hundreds of LSP requests before `max_files` ever fires, whatever the limit is set to.

One way to fix it: have `request_dir_overview`/`get_symbol_overview` take an optional file-count ceiling and raise as soon as it exceeds it while walking, instead of collecting everything first. If that plumbing is too invasive for the LSP layer, an alternative is a cheap pre-count (e.g. walk the ignore-filtered file list once) before calling `get_symbol_overview` at all.

weiconghe: Confirmed — the guard checked `len(path_to_symbols)` only after `get_symbol_overview(relative_path)` had walked the whole tree, so it bounded the returned JSON, not the per-file LSP work, which is the actual cost the #1412 concern was about.

Fixed in 8e8a5251 using the second approach: `_apply_directory` now pre-counts the files the walk would analyze — `os.walk` pruned with the language server's own `is_ignored_path`, i.e. the same filter `request_full_symbol_tree` applies per entry — and raises before `get_symbol_overview` is called at all. The counted set is exactly the files the walk issues `textDocument/documentSymbol` requests for, so `max_files` now bounds that request count directly. I went with the pre-count rather than plumbing a ceiling through `request_dir_overview`/`request_full_symbol_tree` to keep the change out of the shared LSP layer — happy to move it there instead if you prefer.

The max_files test now instruments `request_document_symbols` and asserts 0 requests when the guard fires on `test_repo` with `max_files=1` — the same scenario that produced the 11 requests you measured. All 4 overview tests pass locally on Windows (Python/pyright); CI runs the same suite on Linux.

AmirF194: Confirmed, 8e8a5251 fixes it. The pre-count walk applies the same is_ignored_path predicate to the same paths request_full_symbol_tree uses, so the count matches what the LSP walk would actually visit, and the new test asserting zero document-symbol calls when the guard fires is the right check.

PR #987, opened 2026-02-03
C/C++ language server fixes

MischaPanch:

PR #1564, opened 2026-06-10
Add Unreal Engine 5 fixture and tests for C/C++ (clangd)

Kiborgik:
Adds an Unreal Engine 5 fixture and test suite for the existing C/C++ (clangd) support, plus a setup guide for UE projects. This follows the pattern of #987 (cpp fixture and setup guide) as an isolated addition along existing lines; everything in the diff is test resources and documentation.

## Why

UE game code is written against a macro reflection layer (UCLASS, UFUNCTION, UPROPERTY, GENERATED_BODY) and ships generated reflection files that repeat every identifier from the hand-written sources. The current cpp fixture does not exercise this shape, so regressions in macro handling or generated-file leakage would go unnoticed.

## Contents

- A self-contained UE fixture under `test/resources/repos/cpp/ue_test_repo/` with its own compile_commands.json: a small game module, stub engine headers mirroring UE 5.7's ObjectMacros.h (the annotation macros expand to nothing in real UE compilation too, only UnrealHeaderTool parses them), and UnrealHeaderTool-style generated files that intentionally repeat the hand-written identifiers. The generated files act as a honeypot: a text search would surface them, a symbol-level tool must not.
- `test/solidlsp/cpp/test_cpp_unreal.py` with 12 tests under the existing `cpp` marker, running clangd over the fixture through a module-scoped session. They assert concrete symbol names and locations: UCLASS/USTRUCT types with their UFUNCTION/UPROPERTY members, UENUM enumerators, UINTERFACE pairs, macro-manufactured delegate types and log categories, cross-file references, go-to-definition, and a rename WorkspaceEdit. Negative assertions verify that nothing resolves into `Intermediate/`.
- Two document-symbol helpers in `test/solidlsp/conftest.py`, shared rather than private to one module.
- `docs/03-special-guides/unreal_engine_setup_guide_for_serena.md` covering the two UnrealBuildTool routes to a compile_commands.json, recommended ignored_paths, known behavior, and troubleshooting. Linked from the C/C++ setup guide and the languages page.
- CHANGELOG entry, and a codespell allowance for the UE abbreviation.

## Why a separate repository directory

The first version of this PR placed the fixture inside the shared `test_repo`, which made its stub headers the first standalone headers that a cpp_ccls session encounters. That surfaced two problems on CI: ccls answers documentSymbol only for files present in the compilation database, and worse, ccls 0.20240202 (the build that Ubuntu and Homebrew ship) crashes intermittently when its session covers these headers, leaving the run hanging on a dead connection. With the fixture in its own repository directory the shared `test_repo` and `test_cpp_basic.py` stay exactly as they are on main, ccls never encounters the UE files, and clangd remains the backend the setup guide documents for UE projects. I can share the reproduction details if useful.

## Known limitation

Blueprint and reflection-driven references are invisible to any LSP, so renames cannot update them. This is a property of UE, not of clangd, and the guide says so explicitly.

## Testing

Local run (Windows 11, clangd 19.1.2 auto-downloaded by Serena):

```
$ uv run pytest test/solidlsp/cpp/test_cpp_unreal.py -v
test_cpp_unreal.py::TestClangdUnrealEngine::test_symbol_tree_contains_reflected_types PASSED
test_cpp_unreal.py::TestClangdUnrealEngine::test_document_symbols_show_uclass_members PASSED
test_cpp_unreal.py::TestClangdUnrealEngine::test_references_resolve_to_handwritten_sources PASSED
test_cpp_unreal.py::TestClangdUnrealEngine::test_definition_resolves_to_source_not_generated PASSED
test_cpp_unreal.py::TestClangdUnrealEngine::test_symbol_locations_are_in_handwritten_sources PASSED
test_cpp_unreal.py::TestClangdUnrealEngine::test_uenum_and_enumerators_visible PASSED
test_cpp_unreal.py::TestClangdUnrealEngine::test_macro_generated_delegate_type_resolves_to_source PASSED
test_cpp_unreal.py::TestClangdUnrealEngine::test_interface_pattern_symbols PASSED
test_cpp_unreal.py::TestClangdUnrealEngine::test_log_category_symbol_in_source PASSED
test_cpp_unreal.py::TestClangdUnrealEngine::test_nondynamic_delegate_type_in_source PASSED
test_cpp_unreal.py::TestClangdUnrealEngine::test_character_members_with_ue_types_visible PASSED
test_cpp_unreal.py::TestClangdUnrealEngine::test_rename_edit_targets_only_source_files PASSED
======================== 12 passed, 1 warning in 3.61s ========================

$ uv run pytest test/solidlsp/cpp -m cpp -q
36 passed, 1 xfailed, 1 warning in 10.76s
```

The xfailed test is the pre-existing `test_find_references_in_newly_written_file`, unchanged by this PR. Lint, type-check, codespell, and the docs build are clean locally.

The full cpp suite, clangd and ccls both, also passed three consecutive runs in an Ubuntu 24.04 container with apt ccls (40 passed each), which is the environment where the earlier layout hung.

Beyond the stub fixture, the same setup was validated against a real UE 5.7.4 commercial-scale project (1099 translation units, GAS, UMG, Niagara, online subsystems, real engine headers). A harness drove the full tool surface through Serena's agent layer, 40 checks passing: symbol overviews of UCLASS headers, scoped find_symbol, declarations, implementations, cross-file references, go-to-definition, rename with call-site verification, symbol-level edits, and the project's CQTest-based automation test sources, where TEST_CLASS-manufactured classes appear as symbols. All symbol results resolved exclusively to hand-written sources.

Happy to address any feedback.


Kiborgik:
CI caught two real issues: a directory-traversal-order dependence in one of the new tests (macOS), and an intermittent crash of ccls 0.20240202 when its session covers the new standalone stub headers, which left the Ubuntu job hanging. The fixture now lives in its own repository directory served by clangd only, the shared test_repo and test_cpp_basic.py are back to their state on main, and the full cpp suite passed three consecutive runs in an Ubuntu 24.04 container with apt ccls. Reproduction details available if useful.

Kiborgik:
@MischaPanch folded the #1566 changes into this PR, as you suggested. Two additions, both C/C++ (clangd):

- A project with a `.uproject` at its root gets a default `ignored_paths` preset (`Binaries`, `DerivedDataCache`, `Intermediate`, `Saved`) when its configuration is autogenerated. An existing `project.yml` is left unchanged.
- When clangd starts in such a project without a `compile_commands.json`, it logs a warning that points to the setup guide.

Both are covered by unit tests. The setup guide and changelog are updated.

I also ran the actual generation path on a real UE 5.7 project. Pointing Serena at its `.uproject` produced this `project.yml`:

```yaml
languages:
- cpp
ignored_paths:
- Binaries
- DerivedDataCache
- Intermediate
- Saved
```

The gitignore-style match keeps the hand-written `Source/` files and excludes the `Intermediate/` build tree. Happy to share the full validation output if useful.

src/serena/config/serena_config.py
```diff
@@ -400,6 +409,19 @@ def autogenerate(
             config_with_comments["project_name"] = project_name
             config_with_comments["languages"] = languages_to_use
 
+            # Apply the UE ignored_paths preset for C++ projects with a .uproject (#1566);
+            # only at autogeneration time, so an existing project.yml is never modified.
+            cpp_language_values = {Language.CPP.value, Language.CPP_CCLS.value}
```
MischaPanch: Thanks for the contribution!

This logic doesn't belong here, we shouldn't start making exceptions for all situations here, and even if we did, it should be designed as a higher level concept.

This should be part of the language server startup logic, which you can share between the CPP and CCLS language servers by writing a util function. The LS can extend the ignored paths automatically instead of extending the saved project config. The simplest solution is just to always ignore these paths for these LS, or is there a downside for non-unity projects?

The LS solution also has the benefit that it will work for all users out of the box

src/solidlsp/language_servers/clangd_language_server.py
```diff
@@ -128,7 +132,16 @@ def _prepare_compile_commands(self) -> str | None:
         compile_db_path = os.path.join(self.repository_root_path, "compile_commands.json")
 
         if not os.path.exists(compile_db_path):
-            # No compile_commands.json, nothing to do
+            # No compile_commands.json. For Unreal Engine projects, surface the setup guide
+            # instead of failing silently (clangd needs the DB to resolve engine code).
+            if self._is_unreal_engine_project():
```
MischaPanch: The warning will likely not be seen by most users, maybe raising an error and failing would be better?

MischaPanch (changes requested):

src/serena/config/serena_config.py
Kiborgik: clangd and ccls now skip these directories in is_ignored_dirname. The list and the .uproject check live in common.py, shared by both. The config-time preset is removed, so project.yml is no longer modified.

I gated it on a .uproject being present rather than always skipping these names, because non-UE C++ projects can have real Binaries or Saved directories.

This runs in the language server, so it applies to symbol lookups. Text search and file listing still see these directories, since they use ignored_paths, which no longer lists them.

src/solidlsp/language_servers/clangd_language_server.py
Kiborgik: Changed it to raise SolidLSPException instead of logging a warning.

Raising aborts the language server startup, so a UE project without compile_commands.json can't start until that file exists. It only triggers when a .uproject is present. Non-UE C++ projects are unaffected, since clangd works without a database there.

Should it stay an error, or go back to a non-blocking warning?

MischaPanch: You know more about UE projects than I do, what do you think? If this limits important symbolic functionality and if UE projects can be expected to have compile_commands.json, raising seems like the better strategy to me

Kiborgik: > You know more about UE projects than I do, what do you think? If this limits important symbolic functionality and if UE projects can be expected to have compile_commands.json, raising seems like the better strategy to me

I think raising fits UE.

Without compile_commands.json clangd falls back to default flags and misses the engine
includes and defines like the *_API macros and WITH_EDITOR. UE code is mostly macros and
engine headers, so the index ends up near-empty and find_symbol/references/rename stop
working.

UE doesn't emit compile_commands.json by default, but it's the standard prerequisite for
clangd on UE and the error links the guide. A UE project without it can't do symbol work
anyway.

src/serena/config/serena_config.py
MischaPanch: The mixin-based design was suboptimal IMO, I adjusted it and will now merge.

PR #1567, opened 2026-06-11
refactor: Tighten initial instructions; exact + backend-aware CC tool table; list supported file types

paul-basanets:
## Summary

Refines the instruction texts coding agents receive from Serena, applying Anthropic's current Opus 4.x prompting guidance (cut redundancy, drop over-emphasis like `CRITICAL`/`FORBIDDEN`/`ALWAYS`, prefer positive framing), and enriches the project activation message.

## Changes

- **Laconic prompts** — `system_prompt`, the `claude-code` context, and the `editing` mode are made shorter (~37% / 53% / 59%) while preserving all load-bearing behavior. All Jinja conditionals and tool/config blocks are unchanged.
- **Exact, backend-aware CC tool table** — `cc_system_prompt_override`'s tool-mapping table now uses the real callable tool names, and gains a JetBrains-backend section (`jet_brains_*` navigation/refactoring tools) that renders only when the configured language backend is JetBrains.
- **Backend wiring** — regenerated the prompt factory for the new `jetbrains_backend` variable, added a `SerenaPromptFactory` override, and resolve the effective backend in `serena prompts print-cc-system-prompt-override` (global config, with an optional `--project` override).
- **Supported file types** — the project activation message now lists the supported file types (e.g. `.py, .pyi, .ts, ...`), derived from each language's source filename matcher via new `Language.get_source_file_extensions()` / `FilenameMatcher.file_extensions`.

## Conclusion

Now initial instructions are significantly shorter (~1k tokens reduction). 
My subjective impression is Claude Code works better with Serena now and calls its tools more ofter. 
I hope you'll find those changes useful.

### DIFF
--- a/src/serena/agent.py
+++ b/src/serena/agent.py
@@ -648,7 +648,7 @@ class SerenaAgent:
             log.info(f"Using language backend from global configuration: {self._language_backend.name}")
 
         # create the tool names mapping for prompts
-        self._prompt_tool_names_mapping = self._create_prompt_tool_names_mapping(self._language_backend)
+        self._prompt_tool_names_mapping = self.create_prompt_tool_names_mapping(self._language_backend)
 
         # create executor for starting the language server and running tools in another thread
         # This executor is used to achieve linear task execution
@@ -931,7 +931,7 @@ class SerenaAgent:
         return self._active_modes
 
     @staticmethod
-    def _create_prompt_tool_names_mapping(language_backend: LanguageBackend) -> dict[str, str]:
+    def create_prompt_tool_names_mapping(language_backend: LanguageBackend) -> dict[str, str]:
         """
         Creates a mapping from tool names to new tool names, which take into consideration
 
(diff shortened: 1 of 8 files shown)


MischaPanch:
Thank you! We have been running a dedicated effort to improve the instructions by building an in-house benchmark. I will take your proposal in consideration and soon merge the best of the formulations to commit the improved instructions.

Since the changes we already have will have conflicts with the changes here, merging the PR as is makes no sense. But I will keep it as reference until improved instructions are on main, then close it, ok?

paul-basanets:
@MischaPanch , of course, feel free to extract anything useful and close the PR when it's fine.
I just noticed the the instructions are too long and partially contradicts Anthropic guidelines. However i didn't tested it with other model families/vendors. 
Maybe it makes sense to have different intructions for different models (contexts) but it needs some extra research.

paul-basanets:
Looks like it's obsolete now

PR #1570, opened 2026-06-12
Support aliases of tool parameters, adding class SerenaFastMCPTool

opcode81:

PR #1571, opened 2026-06-12
Update dependencies for Python 3.13

MrLightspeed:
Updates the dependency lock on top of PR #1570 using Python 3.13.\n\nChanges:\n- Bumps the direct Starlette pin from 1.0.0 to 1.0.1, matching the open Dependabot security/update PR.\n- Regenerates uv.lock with Python 3.13.\n\nLocal verification:\n- uv sync --python 3.13 --extra dev --locked\n- uv run poe lint\n- uv run pytest test/serena/test_mcp.py test/serena/test_tool_parameter_types.py -q --tb=short\n- git diff --check

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,7 @@ dependencies = [
   # Exact pins because uvx installs from git, ignoring the lock file.
   "urllib3==2.7.0",
   "werkzeug==3.1.7",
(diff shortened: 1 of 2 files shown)


MischaPanch:
Dependabot already had a PR for that

PR #1575, opened 2026-06-13
fix: improve ignored_paths YAML guidance

### LINKED ISSUE #1561: Improve ignored_paths docs and validation for large C# repos
Serena's C# setup has two related discoverability issues around `ignored_paths`: large repositories can be expensive to diagnose when Roslyn opens unrelated nested projects, and common unquoted gitignore-style globs can crash Serena startup with a raw YAML error.

## Environment

- OS: Windows
- Serena package: `serena-agent 1.1.2`
- C# backend: LSP
- Roslyn language server: `roslyn-language-server.win-x64.5.5.0-2.26078.4`
- Client: Codex Desktop using Serena as a stdio MCP server
- Serena command shape:

```text
serena.exe start-mcp-server --context=codex --project <repo> --enable-web-dashboard false --open-web-dashboard false
```

Note: `serena --version` does not appear to be supported, so the Serena version above was read from the installed Python package metadata.

## What happened

While investigating why Serena appeared "heavy", the process tree showed that most memory was not the small `serena.exe` process itself, but the child Python process and the Roslyn language server.

The Roslyn server opened unrelated nested `.csproj` files from example/worktree-like directories under the repository. Redacted examples from the Serena log:

```text
LSP: [project/open] [LanguageServerProjectLoader] Warning while loading <repo>/examples/.../CompileCommandsJson.csproj: ...
LSP: [project/open] [LanguageServerProjectLoader] Error while loading <repo>/examples/.../NVAccess.NVDA.csproj: ...
error NETSDK1045: current .NET SDK does not support targeting .NET 10.0 [... <repo>/examples/.../GenerateZw.csproj]
```

This made the memory usage and diagnostics look related to the active project, while the expensive work was coming from nested projects that were not relevant to the current workspace.

The workaround was to add project-level `ignored_paths`, for example:

```yaml
ignored_paths:
- "examples/**"
- ".worktrees/**"
- "**/bin/**"
- "**/obj/**"
```

However, the first attempt used normal gitignore-looking patterns without quotes:

```yaml
ignored_paths:
- examples/**
- .worktrees/**
- **/bin/**
- **/obj/**
```

That caused Serena to fail at startup with a raw YAML parser error:

```text
ERROR Fatal exception: found undefined alias '**/bin/**'
  in "<project>/.serena/project.yml", line 63, column 3
```

This is technically valid YAML behavior because `*` starts an alias, but it is an easy trap when the config comments say the syntax is the same as gitignore and users naturally write `**/bin/**`.

## Why this is hard to discover

- Users may only see that Serena consumes a lot of memory, not that the C# language server is indexing unrelated nested projects.
- The generated `.serena/project.yml` leaves `ignored_paths: []`, but does not show quoted glob examples.
- The startup failure for `**/...` looks like an internal crash rather than a config validation problem.

## Additional context

In Codex, stdio MCP servers may be started when a project/thread is resumed or when server status is checked, before the user explicitly calls a Serena tool. This is not the root cause, but it makes the startup cost visible earlier.

## Suggested improvements

- Add quoted examples for common ignored paths in the generated `project.yml`, especially patterns that start with `*`, such as `"**/bin/**"` and `"**/obj/**"`.
- Validate `project.yml` and report a friendlier message when YAML aliases are accidentally used in `ignored_paths`.
- Consider documenting a C# / Roslyn troubleshooting section that explains how nested `.csproj` discovery affects memory and diagnostics.
- Clarify whether `ignored_paths` are expected to affect C# language server project discovery, and if so, how early they are applied.

## Expected behavior

Users configuring Serena for a large C# repository should have an obvious, safe path to exclude nested examples, worktrees, and build output directories, and a config typo should not look like an unexplained Serena crash.

zerone0x:
Fixes #1561.

This makes unquoted `*`-prefixed `ignored_paths` globs fail with a clear hint instead of a raw YAML alias error, and adds quoted examples to the project/global config docs and templates.

Tested with:
- `uv run --extra dev ruff format --check src/serena/util/yaml.py test/serena/config/test_serena_config.py`
- `uv run --extra dev ruff check src/serena/util/yaml.py test/serena/config/test_serena_config.py`
- `uv run --extra dev pytest test/serena/config/test_serena_config.py -vv`
- `uv run --with codespell codespell docs/02-usage/040_workflow.md src/serena/resources/project.template.yml src/serena/resources/serena_config.template.yml`

### DIFF
--- a/src/serena/resources/project.template.yml
+++ b/src/serena/resources/project.template.yml
@@ -71,6 +71,13 @@ additional_workspace_folders: []
 
 # list of additional paths to ignore in this project.
 # Same syntax as gitignore, so you can use * and **.
+# Important: quote patterns that start with `*`, otherwise YAML treats them as aliases.
+# Example:
+#   ignored_paths:
+#     - "examples/**"
+#     - ".worktrees/**"
+#     - "**/bin/**"
+#     - "**/obj/**"
 # Note: global ignored_paths from serena_config.yml are also applied additively.
 ignored_paths: []
 
--- a/src/serena/resources/serena_config.template.yml
+++ b/src/serena/resources/serena_config.template.yml
@@ -103,6 +103,7 @@ ls_specific_settings: {}
 # list of paths to ignore across all projects.
 # Same syntax as gitignore, so you can use * and **.
 # These patterns are merged additively with each project's own ignored_paths.
+# Quote patterns that start with `*`, e.g. `"**/bin/**"`.
 ignored_paths: []
 
 # list of regex patterns which, when matched, mark a memory entry as read‑only.
--- a/src/serena/util/yaml.py
+++ b/src/serena/util/yaml.py
@@ -68,7 +68,17 @@ def load_yaml(path: str, comment_normalisation: YamlCommentNormalisation = YamlC
     """
     with open(path, encoding=SERENA_FILE_ENCODING) as f:
         yaml = _create_yaml(preserve_comments=True)
-        commented_map: CommentedMap | None = yaml.load(f)
+        try:
+            commented_map: CommentedMap | None = yaml.load(f)
+        except Exception as e:
+            msg = str(e)
+            if "undefined alias" in msg:
+                raise ValueError(
+                    f"Invalid YAML in {path}: values that start with `*` must be quoted. "
+                    "This often happens in `ignored_paths` when using gitignore-style globs like "
+                    '`"**/bin/**"` or `"**/obj/**"`.'
+                ) from e
+            raise
     if commented_map is None:  # ruamel returns None for empty documents, but we want an empty CommentedMap
         commented_map = CommentedMap()
     normalise_yaml_comments(commented_map, comment_normalisation)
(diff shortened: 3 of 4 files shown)


docs/02-usage/040_workflow.md
```diff
@@ -69,6 +69,18 @@ You can specify local overrides for the settings in a `project.local.yml` file i
 (which, by default, is ignored by git). 
 Any keys defined therein will override the respective key in `project.yml`.
 
+For large C# / Roslyn repositories, `ignored_paths` is often the easiest way to keep nested examples,
```
MischaPanch: This is too specific for this top-level doc. Instead, we could add an entry `notes_on_ignore_settings` to 03-special guides and link to it here, mentioning that it's particularly relevant for certain language servers or in certain situations. Does the C# LS not respect .gitignore?

MischaPanch (changes requested):

MischaPanch:
@zerone0x ping. Do you want to finish this up?

docs/02-usage/040_workflow.md
opcode81: I removed this

Issue #900, opened 2026-01-04
Julia LS initialisation problems


Meng-Fei:
I have:

- [x] Read the readme and verified that the issue cannot be solved by adjusting configuration
- [x] Understood that Serena's dashboard can be disabled through the config
- [x] Understood that by default a client session will start a separate instance of a Serena server. 
- [x] Understood that for multi-agent setups, the SSE mode should be used.
- [x] Verified that non-project files are ignored using either gitignore or the corresponding setting in `.serena/project.yml`
- [x] Have looked for similar issues and discussions, including closed ones
- [x] Made sure it's an actual issue and not a question - those should be opened as discussion instead.

If you have encountered a real and new issue:

- [x] I performed `<uv invocation> serena project health-check`
- [x] I indexed the project as described in the readme
- [x] Added sufficient explanation of my setup: the MCP client, the OS, the programming language, any config adjustments or relevant project specifics
- [x] Explained how the issue arose, added instructions on how to reproduce it (if possible)
- [x] If the issue happens on an open source project, I have added the link
- [x] Wrote a meaningful title and description

## Setup

  - **MCP Client**: Claude Code 2.0.74
  - **OS**: Linux (openSUSE)
  - **Language**: Julia (LanguageServer.jl v4.5.1)
  - **Serena**: 0.1.4-603ef3f8

  ## Issue

  The Julia LanguageServer takes ~10 seconds to initialize (confirmed by health-check). In stdio mode, Claude Code closes the MCP connection after ~4 seconds of no response, before Serena can finish initializing the language server.

  **Health-check log**:
  Language server startup (language=julia) completed in 10.035 seconds

  **Current behavior**: Serena crashes with stdio transport when using slow-initializing language servers.

  **Workaround**: Use SSE transport mode, which works perfectly.

  ## Request

  The issue checklist mentions "for multi-agent setups, the SSE mode should be used." However, SSE is also required for **single-agent setups with slow-initializing language servers**.

  Please consider:
  1. **Documentation update**: Add guidance that SSE mode should be recommended for language servers with slow initialization (e.g. Julia and maybe others)
  2. **Or async initialization**: Initialize language servers asynchronously so MCP responds within timeout window

  This is primarily a documentation/UX issue rather than a bug.

MischaPanch:
Thanks for the report. I thought we already have async initialization, right @opcode81 ?

opcode81:
Yes, I suspect that there was another problem with using stdio mode, which is not related to LS startup times.

[...]

PR #1580, opened 2026-06-16
Prevent Julia LS setup from clobbering Serena's stdio MCP pipe

### LINKED ISSUE #1577: Enabling the `julia` language server kills the STDIO MCP server right after `initialize` ("tools fetch failed")
## Summary

With `julia` in a project's `languages:` list, Serena's **stdio** MCP server exits a few milliseconds after `initialize`, before it can answer `tools/list`. The MCP client therefore reports **"tools fetch failed"**, and because the whole process exits, the auto-opened web dashboard tab shows **"Error loading configuration"** (connection refused). Removing `julia` makes everything work.

This bites Claude Code specifically, since it launches Serena over stdio.

## Root cause

Serena's stdio transport uses the process's **stdin/stdout** for the JSON-RPC protocol. On startup, the Julia LS init runs a probe subprocess in `solidlsp/language_servers/julia_server.py::_setup_runtime_dependency()`:

```python
check_cmd = [julia_path, "-e", "using LanguageServer"]
result = subprocess.run(check_cmd, check=False, capture_output=True, text=True, timeout=10)
```

`capture_output=True` redirects the child's stdout/stderr but **not stdin**, so the Julia child **inherits Serena's stdin — the MCP pipe**. That disrupts/closes the stream, and Serena's stdio server then exits (`anyio.ClosedResourceError` observed on the client) right after `initialize`. The same applies to the `Pkg.add` path in `_install_language_server()`.

SSE / streamable-http transports are **not** affected, because they don't use stdin/stdout for the protocol (a standalone `--transport sse` Serena with Julia stays alive indefinitely) — which is a useful confirmation that the protocol stream is the thing being clobbered.

## Environment

- serena `1.5.3` (mcp `1.27.0`), installed via `uv tool`
- Julia `1.12` (LanguageServer.jl installed, but not required to trigger)
- macOS 26 (arm64); the mechanism is not platform-specific

## Minimal reproduction

A/B script: spins up Serena over stdio (via the bundled `mcp` client) against a throwaway one-file Julia project, once with `languages: ["julia"]` and once with `languages: []`, and reports whether `list_tools` succeeds.

Prereqs: `serena` and `julia` on PATH. Run with Serena's interpreter:
```bash
"$(dirname "$(readlink -f "$(command -v serena)")")/python" repro.py
```

<details>
<summary><code>repro.py</code></summary>

```python
#!/usr/bin/env python3
"""A/B: does Serena's stdio MCP server survive `initialize` with julia enabled?"""
import asyncio
import os
import shutil
import tempfile


def make_project(languages: str) -> str:
    root = tempfile.mkdtemp(prefix="serena-julia-stdio-")
    os.makedirs(os.path.join(root, ".serena"), exist_ok=True)
    with open(os.path.join(root, ".serena", "project.yml"), "w") as f:
        f.write(f'project_name: "serena-julia-stdio-repro"\nlanguages: {languages}\n')
    with open(os.path.join(root, "foo.jl"), "w") as f:
        f.write('module Foo\ngreet() = println("hello")\nend\n')
    return root


async def try_session(languages: str) -> str:
    from mcp import ClientSession, StdioServerParameters
    from mcp.client.stdio import stdio_client

    root = make_project(languages)
    params = StdioServerParameters(
        command="serena",
        args=["start-mcp-server", "--context", "claude-code", "--project", root],
    )
    try:
        async with stdio_client(params) as (read, write):
            async with ClientSession(read, write) as session:
                await asyncio.wait_for(session.initialize(), timeout=60)
                tools = await asyncio.wait_for(session.list_tools(), timeout=60)
                return f"OK — initialize + list_tools succeeded ({len(tools.tools)} tools); server stayed alive"
    except Exception as e:
        return f"FAILED — {type(e).__name__} (server exited after initialize -> 'tools fetch failed')"
    finally:
        shutil.rmtree(root, ignore_errors=True)


async def main() -> None:
    if shutil.which("serena") is None or shutil.which("julia") is None:
        raise SystemExit("need both `serena` and `julia` on PATH")
    print('languages: ["julia"] :', await try_session('["julia"]'))
    print("languages: []       :", await try_session("[]"))


if __name__ == "__main__":
    asyncio.run(main())
```
</details>

## Expected vs actual

Expected: both configurations connect and list tools.

Actual (unpatched serena 1.5.3):
```
languages: ["julia"] : FAILED — ExceptionGroup (server exited after initialize -> 'tools fetch failed')
languages: []        : OK — initialize + list_tools succeeded (20 tools); server stayed alive
```

## Suggested fix

Don't let the Julia probe/install subprocesses inherit Serena's stdin. Adding `stdin=subprocess.DEVNULL` to both `subprocess.run` calls in `julia_server.py` resolves it; verified locally:

```python
# _setup_runtime_dependency()
result = subprocess.run(check_cmd, check=False, capture_output=True, text=True, timeout=10, stdin=subprocess.DEVNULL)
# _install_language_server()
result = subprocess.run(install_cmd, check=False, capture_output=True, text=True, timeout=300, stdin=subprocess.DEVNULL)
```

After the patch, the same A/B prints `OK` for both, and Julia symbol tools (`find_symbol`, `get_symbols_overview`, …) work over stdio.

More generally, any `subprocess` Serena spawns while running the stdio transport should avoid inheriting the process stdin/stdout (e.g. default `stdin`/`stdout` to `DEVNULL`/a pipe), so a language-server runtime probe can never clobber the MCP channel. Other language backends that shell out during setup may have the same latent issue.

## Relationship to #900

Issue #900 reports that Julia LS slow initialization (~10 s) causes a **client timeout** in stdio mode. This is a **different bug**: here the server exits within milliseconds because the Julia probe subprocess **inherits and disrupts Serena's stdin pipe**, killing the JSON-RPC channel outright. The fix is also different — #900 asks for async init or docs; this bug needs `stdin=subprocess.DEVNULL` (or equivalent) on the probe subprocesses.

flin2:
## Problem

Fixes #1577.

With `julia` in a project's `languages:` list, Serena's **stdio** MCP server exits a few milliseconds after `initialize`, before it can answer `tools/list`. The client reports **"tools fetch failed"** and, because the process exits, the dashboard shows "Error loading configuration". Removing `julia` makes everything work. This bites Claude Code specifically, since it launches Serena over stdio.

## Root cause

The stdio transport uses the process's stdin/stdout for the JSON-RPC protocol. On startup, the Julia LS setup in `julia_server.py` runs probe/install subprocesses:

```python
subprocess.run(check_cmd, check=False, capture_output=True, text=True, timeout=10)
subprocess.run(install_cmd, check=False, capture_output=True, text=True, timeout=300)
```

`capture_output=True` redirects the child's stdout/stderr but **not stdin**, so the Julia child **inherits Serena's stdin — the MCP pipe**. That disrupts/closes the stream, and the stdio server exits (`anyio.ClosedResourceError` on the client) right after `initialize`.

SSE / streamable-http transports are not affected, since they don't use stdin/stdout for the protocol — a standalone `--transport sse` Serena with Julia stays alive indefinitely, confirming the protocol stream is what's being clobbered.

## Fix

Pass `stdin=subprocess.DEVNULL` to both `subprocess.run` calls in `julia_server.py` so the probe/install children can't inherit and disrupt the MCP channel. Added a comment explaining the stdio-transport hazard.

## Verification

A/B repro (full script in #1577) spins up Serena over stdio against a one-file Julia project, once with `languages: ["julia"]` and once with `languages: []`:

Before:
```
languages: ["julia"] : FAILED — server exited after initialize -> 'tools fetch failed'
languages: []        : OK — initialize + list_tools succeeded (20 tools)
```

After this patch both print `OK`, and Julia symbol tools (`find_symbol`, `get_symbols_overview`, …) work over stdio.

`ruff format --check` and `ruff check` pass on the changed file.

## Note / follow-up

More generally, any subprocess Serena spawns while running the stdio transport should avoid inheriting the process stdin/stdout, so a language-server runtime probe can never clobber the MCP channel. Other language backends that shell out during setup may have the same latent issue — left out of this PR to keep it focused, but happy to follow up.

This is distinct from #900 (Julia LS slow init causing a *client timeout*); here the server exits within milliseconds because the probe disrupts the stdin pipe, and the fix differs accordingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,10 @@ Status of the `main` branch. Changes prior to the next official version change w
     results in svelte-only mode (`languages: [svelte]`). They are now routed to the companion TS server,
     so symbols defined in plain `.ts`/`.js` files are again discoverable via `find_symbol`/
     `get_symbols_overview`, and `find_referencing_symbols` no longer fails to locate `.ts`/`.js` symbols. #1552
+  - `JuliaLanguageServer`: Fix the stdio MCP server exiting right after `initialize` ("tools fetch failed")
+    when `julia` is enabled. The runtime probe/install subprocesses now run with `stdin=subprocess.DEVNULL`,
+    so they can no longer inherit and clobber Serena's stdin, which is the JSON-RPC pipe under the stdio
+    transport. #1577
   - Improve quoting of arguments in shell executions
 
 * JetBrains:
--- a/src/solidlsp/language_servers/julia_server.py
+++ b/src/solidlsp/language_servers/julia_server.py
@@ -78,10 +78,14 @@ class JuliaLanguageServer(SolidLanguageServer):
                 f"Checked locations: {common_locations}"
             )
 
-        # Check if LanguageServer.jl is installed
+        # Check if LanguageServer.jl is installed.
+        # stdin=DEVNULL: when Serena runs over the stdio MCP transport, the JSON-RPC
+        # channel *is* the process stdin/stdout. Without this, the Julia child inherits
+        # Serena's stdin (the MCP pipe) and clobbers it, killing the server right after
+        # initialize ("tools fetch failed"). See https://github.com/oraios/serena/issues/1577
         check_cmd = [julia_path, "-e", "using LanguageServer"]
         try:
-            result = subprocess.run(check_cmd, check=False, capture_output=True, text=True, timeout=10)
+            result = subprocess.run(check_cmd, check=False, capture_output=True, text=True, timeout=10, stdin=subprocess.DEVNULL)
             if result.returncode != 0:
                 # LanguageServer.jl not found, install it
                 JuliaLanguageServer._install_language_server(julia_path)
@@ -99,7 +103,9 @@ class JuliaLanguageServer(SolidLanguageServer):
         install_cmd = [julia_path, "-e", 'using Pkg; Pkg.add("LanguageServer")']
 
         try:
-            result = subprocess.run(install_cmd, check=False, capture_output=True, text=True, timeout=300)  # 5 minutes for installation
+            result = subprocess.run(
+                install_cmd, check=False, capture_output=True, text=True, timeout=300, stdin=subprocess.DEVNULL
+            )  # 5 minutes for installation
 
             if result.returncode == 0:
                 log.info("LanguageServer.jl installed successfully!")


opcode81:
Please add this change to the change log in the "Language Servers" section; then this can be merged.

flin2:
Done — added an entry under the Unreleased "Language Servers" section in `CHANGELOG.md` (commit 1f5105fa). Ready for merge whenever you are. Thanks!

opcode81:
Thank you very much for this improvement!

PR #1582, opened 2026-06-16
Make tool call errors surface explicitly as errors at the MCP level

opcode81:
* Refactor `Tool.apply_ex` to either return a success response or raise `ToolCallError` (if `catch_exceptions`=False)
* Adapt `SerenaFastMCPTool` to catch `ToolCallError` and reraise as
  `fastmcp.exceptions.ToolError`, triggering an error response
  in the SDK


MischaPanch (approved):
LGTM, thank you

The CI pipeline already ran through (the new fast one, need to delete the old workflows)

PR #1589, opened 2026-06-18
Use validate_relative_path (not assert) for new-file path confinement in create_text_file

alex-schose:
`CreateTextFileTool.apply` guards the **new-file** branch with
`assert abs_path.is_relative_to(project_root)`, while the overwrite branch uses the
always-on `validate_relative_path`. Python strips `assert` statements under `-O` /
`PYTHONOPTIMIZE`, so when the server runs optimized, the new-file branch performs no
path-containment check and a `create_text_file` call with a `../` relative path can
write outside the project root (`Path.resolve()` collapses the `..`).

This change makes the new-file branch use the same always-on guard as the overwrite
branch — consistent behavior, not dependent on assertions being enabled. It also keeps
the new-file path consistent with Serena's documented symlink stance
(`validate_relative_path` → `is_path_in_project` is lexical), unlike the prior
`.resolve()`-based check.

Low severity (requires running under `-O`); filed as a defense-in-depth hardening fix.

### DIFF
--- a/src/serena/tools/file_tools.py
+++ b/src/serena/tools/file_tools.py
@@ -69,9 +69,10 @@ class CreateTextFileTool(EditingToolWithDiagnostics):
             if will_overwrite_existing:
                 self.project.validate_relative_path(relative_path, require_not_ignored=True)
             else:
-                assert abs_path.is_relative_to(self.get_project_root()), (
-                    f"Cannot create file outside of the project directory, got {relative_path=}"
-                )
+                # Use the always-on containment guard rather than a bare `assert`, which is
+                # compiled out under `python -O` / PYTHONOPTIMIZE and would otherwise leave the
+                # new-file branch with no path-containment check.
+                self.project.validate_relative_path(relative_path)
 
             # writing the file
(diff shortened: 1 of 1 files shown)


src/serena/tools/file_tools.py
```diff
@@ -69,9 +69,10 @@ def apply(self, relative_path: str, content: str) -> str:
             if will_overwrite_existing:
                 self.project.validate_relative_path(relative_path, require_not_ignored=True)
             else:
-                assert abs_path.is_relative_to(self.get_project_root()), (
-                    f"Cannot create file outside of the project directory, got {relative_path=}"
-                )
+                # Use the always-on containment guard rather than a bare `assert`, which is
```
MischaPanch, line 72: I am the co-author of Serena and still this comment is unintelligible to me. Pls rephrase (maybe it was written by AI?) and make an entry in the changelog

MischaPanch (changes requested):

MischaPanch:
Actually, I had one more look, and closing this now. We use asserts in several places, I don't expect any of our users to run with -O

PR #1596, opened 2026-06-23
Fix dangling registered project poisoning project activation

scytherswings:
## Problem

When a registered project's root directory is deleted out from under Serena — most commonly a removed git worktree (`git worktree remove ...`) that was registered as its own project — activating **any** project then fails with a `FileNotFoundError` referencing the *deleted* project's path:

```
FileNotFoundError - [Errno 2] No such file or directory:
'/…/ocean_game/.worktrees/biome-visibility-authoring'
```

The path in the error is not the project being activated; it's an unrelated, dangling entry. One dangling registered project breaks activation globally until the MCP server is restarted.

### Root cause

`get_registered_project()` resolves a project by path by iterating over **every** registered project and calling `RegisteredProject.matches_root_path()`. That method calls `self.project_root.samefile(...)`, and `Path.samefile` raises `FileNotFoundError` when the registered project's own root directory no longer exists. The first dangling entry encountered aborts the whole lookup — and therefore activation of the valid, requested project.

## Fix

`RegisteredProject.matches_root_path` now treats a non-existent directory as a non-match (catches `OSError`) rather than letting the filesystem error propagate. A directory that no longer exists cannot match an existing path, so the dangling entry is simply skipped instead of poisoning the lookup.

This isolates the failure to the offending project, as suggested, without changing behaviour for valid projects.

## Tests

Added `TestGetRegisteredProjectWithDanglingProject` (developed test-first):

- Registers a dangling project *before* a valid one, deletes the dangling directory, then looks up the valid project by path.
- Before the fix it reproduces the reported `FileNotFoundError`; after the fix it returns the valid project.

Full `test/serena/config/test_serena_config.py` suite passes (43 tests); `ruff` clean. CHANGELOG updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ Status of the `main` branch. Changes prior to the next official version change w
   - Improve quoting/escaping of arguments in shell executions on Windows (via `oslex` dependency)
   - Add tool parameter alias support, adding `name_path` as an alias for `name_path_pattern` in `find_symbol` tools
   - Make tool call errors surface explicitly as errors at the MCP protocol level
+  - Fix: a registered project whose root directory was deleted while Serena was already running could break
+    `activate_project`/project lookup, raising `FileNotFoundError` in `RegisteredProject.matches_root_path`
 
 * Language Servers:
   - `typescript_vts`: Add `initialization_options` setting in `ls_specific_settings.typescript_vts`.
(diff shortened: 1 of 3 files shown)


opcode81:
Thanks for your PR.

Note on impact: Only cases are affected where the project directory is deleted while Serena is already running. A restart of Serena would filter out the deleted project at startup.

scytherswings:
I regularly use work trees in an agentic loop so the ability to activate and deactivate projects in Serena is incredibly helpful however sometimes my workflow results in a deleted work tree and that can cause Serena to get into a deadlock state. This PR fixes said deadlock without me being forced to lose my context or restart the MCP server.

Issue #1598, opened 2026-06-24
Add LaTeX support via texlab


KoscheiiB:
**Is your feature request related to a problem? Please describe.**

I use Serena for research-based coding, where a project is rarely just code. The analysis spans a few languages (R, Python, MATLAB) and a LaTeX writeup ties the results together. Each of those already emits LaTeX-ready output that lands in the document: R's `knitr::kable`, pandas `DataFrame.to_latex` and `sympy.latex` in Python, MATLAB's `latex()` for symbolic results. So the `.tex` is where output from every language converges, and it is also the one file Serena cannot navigate: the code is fully symbol-aware, but the document that references those results is opaque text.

**Describe the solution you'd like**

Add LaTeX/BibTeX as a language backed by [texlab](https://github.com/latex-lsp/texlab). It would register like the other non-programming languages (markdown, json, yaml): explicit opt-in, gated experimental, `is_programming_language=False`. texlab provides hierarchical document symbols (the section tree) plus definitions and references over `\label`/`\ref` and `\cite`, so the writeup becomes navigable with the same tools as the code it reports on.

**Describe alternatives you've considered**

Leaving LaTeX unsupported keeps the writeup outside Serena. Treating `.tex` as plain text (markdown-style) gives no `\label`/`\cite` graph, which is the part that matters for navigation. texlab is the de-facto LaTeX language server, so there is no real second candidate.

**Additional context**

License: texlab is GPL-3.0; Serena is MIT. The integration downloads a prebuilt texlab binary at runtime and talks to it over LSP stdio as a separate process. Nothing is vendored or linked, so this is the same arrangement Serena already ships with for the proprietary Intelephense (default PHP server) and the MPL terraform-ls. I would state the license and the runtime download explicitly in the PR.

Status: I have a working branch: texlab 5.25.1, five platforms (macOS/Linux arm64+x64, Windows x64), sha256-verified download, ruff and pyright clean. Automated smoke tests cover document symbols. I have also verified cross-file reference resolution on a real multi-file project, where `\ref`/`\cite` resolve to their `\label` and bib definitions across `.tex` files; happy to land that as an automated test too.

Would you accept this? If so, any constraints before I open the PR (platform coverage, how you would like the experimental gating handled)?

opcode81:
Yes, we would accept this as described. Thanks for your interest in contributing to Serena.

MischaPanch:
Implemented in #1600

PR #1600, opened 2026-06-24
feat(solidlsp): add LaTeX language server via texlab

KoscheiiB:
## What

Adds experimental LaTeX / BibTeX language support via
[texlab](https://github.com/latex-lsp/texlab). Refs #1598.

- New `TexlabLanguageServer`: texlab v5.25.1 is downloaded at runtime as a
  prebuilt, SHA-256-verified binary for osx-arm64/x64, linux-arm64/x64, and
  win-x64. Nothing is vendored.
- `LATEX` enum + matcher + factory, registered as a non-programming language
  (`is_programming_language=False`, same tier as Markdown/JSON/YAML): experimental
  and explicit opt-in via language `latex`, covering `.tex`, `.bib`, `.sty`, `.cls`.
- Capabilities: sectioning document symbols, `\ref` to `\label` go-to-definition
  and references (within-file and cross-file), and `\cite` to BibTeX-entry
  resolution.

### Beamer note

Beamer presentations are supported: `\section`/`\subsection` and frames defined
with `\frametitle{...}` surface as document symbols (frames nest under their
section). Note that texlab indexes a frame only when it uses the `\frametitle`
command. The inline `\begin{frame}{Title}` form is not exposed as a symbol.

## License

texlab is GPL-3.0; Serena is MIT. texlab is downloaded at runtime and spoken to
over LSP stdio as a separate process. This is mere aggregation, the same
arrangement Serena already ships for the proprietary Intelephense (default PHP)
and MPL terraform-ls. No texlab source is vendored or linked.

## Tests

- `test/solidlsp/latex/`: document symbols, within-file and cross-file
  `\ref`/`\label` references, `\cite` to bib resolution, and a beamer fixture
  (frames + cross-reference).
- `test/serena/test_serena_agent.py`: LATEX find-symbol and cross-file
  reference cases.
- Fixtures are small, synthetic, and self-authored (consistent with the other
  language test repos).

### DIFF
--- a/test/solidlsp/latex/test_latex_basic.py
+++ b/test/solidlsp/latex/test_latex_basic.py
@@ -1,0 +1,37 @@
+"""Basic tests for the texlab-based LaTeX language server."""
+
+import pytest
+
+from serena.symbol import LanguageServerSymbol
+from solidlsp import SolidLanguageServer
+from solidlsp.ls_config import Language
+
+
+@pytest.mark.latex
+class TestLatexLanguageServerBasics:
+    """Basic functionality of the LaTeX (texlab) language server."""
+
+    @pytest.mark.parametrize("language_server", [Language.LATEX], indirect=True)
+    def test_document_symbols_are_sections(self, language_server: SolidLanguageServer) -> None:
+        """Sectioning commands should surface as document symbols."""
+        symbols, _roots = language_server.request_document_symbols("main.tex").get_all_symbols_and_roots()
+        assert len(symbols) > 0, "Should find at least some symbols in main.tex"
+
+        names = {s.get("name", "") for s in symbols}
+        for expected in ("Introduction", "Methods", "Conclusion"):
+            assert expected in names, f"Expected section '{expected}' among document symbols, got: {sorted(names)}"
+
+    @pytest.mark.parametrize("language_server", [Language.LATEX], indirect=True)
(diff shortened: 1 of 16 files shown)


test/solidlsp/latex/test_latex_basic.py
```diff
@@ -0,0 +1,28 @@
...
+
+from solidlsp import SolidLanguageServer
+from solidlsp.ls_config import Language
+
+
+@pytest.mark.latex
+class TestLatexLanguageServerBasics:
+    """Basic functionality of the LaTeX (texlab) language server."""
+
+    @pytest.mark.parametrize("language_server", [Language.LATEX], indirect=True)
```
MischaPanch, line 14: Thank you for the contribution!

Ideally, subsections would follow the name_path hierarchy, like in markdown. So the name_path of a subsection should be "supersection/subsection". Could you pls add that to the tests?

MischaPanch (commented):
Looks great, thanks! Just a minor comment on tests.

Really cool to have latex support in Serena

test/solidlsp/latex/test_latex_basic.py
KoscheiiB: Done. Added the test like asked, asserting the subsection's name_path nests as `Methods/Implementation Details`. And thanks for the review!

PR #1603, opened 2026-06-25
fix: normalize project_root in is_path_in_project to handle Windows path separator mismatch (#1602)

### LINKED ISSUE #1602: Windows: is_path_in_project false negative when project_root uses forward slashes (Git Bash/MSYS)
### Description

When `project_root` is stored with forward slashes (common when Serena runs in Git Bash/MSYS on Windows), the `is_path_in_project` method produces false negatives — paths that are **inside** the project root are incorrectly rejected as "outside of the repository root".

### Root cause

`os.path.commonpath()` always normalizes paths to use backslashes on Windows. But `self.project_root` may contain forward slashes (e.g. coming from Git Bash/MSYS environment). The equality check fails because:

```python
os.path.commonpath(["D:/JavaTool/github-pr/serena", "D:\JavaTool\github-pr\serena"])
# → "D:\JavaTool\github-pr\serena"     (backslash)

"D:\JavaTool\github-pr\serena" == "D:/JavaTool/github-pr/serena"
# → False  (mismatch!)
```

### Steps to reproduce

1. Run Serena MCP server from Git Bash on Windows
2. Set project root with forward slashes (e.g. `D:/repo/my-project`)
3. Call `list_dir` with the absolute path `D:\repo\my-project`
4. Get `ValueError: ... points to path outside of the repository root`

### Impact

All file tools that call `validate_relative_path` are affected — `list_dir`, `read_file`, `create_text_file`, `find_file`, `replace_text`, `delete_lines`, `insert`, etc.

### Expected behavior

Absolute paths that are inside the project root (including the project root itself) should be treated as valid paths.

### Environment

- Windows 11
- Python 3.13
- serena-agent 1.5.3
- Claude Code (running in Git Bash)

### Related

- #922 — related Windows path issue, but the underlying separator mismatch was not fixed at that time

weiconghe:
### Issue for this PR

Closes #1602

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation

### What does this PR do?

**Problem:** When `project_root` is stored with forward slashes (common in Git Bash/MSYS on Windows), `is_path_in_project` produces false negatives. `os.path.commonpath()` normalizes all paths to backslashes on Windows, but the equality check against `self.project_root` (still with forward slashes) fails — even though both refer to the same directory.

**Impact:** All file tools that call `validate_relative_path` are affected — `list_dir`, `read_file`, `create_text_file`, `find_file`, `replace_text`, `delete_lines`, `insert`, etc. Users on Windows who run Serena via Claude Code (Git Bash) hit "outside of the repository root" errors when passing valid absolute paths.

**Fix:** Apply `os.path.normpath()` to `self.project_root` before the comparison, ensuring consistent separator style on both sides. Also improve the error message to include the project root and suggest using relative paths.

1. Created a test project with forward-slash `project_root`, verified `is_path_in_project` returns `True` for:
   - Absolute path equal to project root (the original bug)
   - Relative paths within the project
   - Subdirectory paths (absolute)
2. Verified security boundaries still hold:
   - Sibling directory paths are correctly rejected
   - Paths with ".." escape attempts are rejected
   - Different drive paths are rejected
3. Ran existing test suite (15 tests in `test_global_ignored_paths.py`) — all pass
4. Environment: Windows 11, Python 3.13

### Screenshots / recordings

N/A

### Checklist

- [x] I have not included unrelated changes in this PR

### DIFF
--- a/src/serena/project.py
+++ b/src/serena/project.py
@@ -36,7 +36,7 @@ class Project(ToStringMixin):
         is_newly_created: bool = False,
     ):
         assert serena_config is not None
-        self.project_root = project_root
+        self.project_root = os.path.normpath(project_root)
         self.project_config = project_config
         self.serena_config = serena_config
         self._serena_data_folder = serena_config.get_project_serena_folder(self.project_root)
@@ -268,7 +268,7 @@ class Project(ToStringMixin):
         path = os.path.normpath(path)
 
         try:
-            return os.path.commonpath([self.project_root, path]) == os.path.normpath(self.project_root)
+            return os.path.commonpath([self.project_root, path]) == self.project_root
         except ValueError:
             # occurs, in particular, if paths are on different drives on Windows
             return False
--- a/src/serena/project.py
+++ b/src/serena/project.py
@@ -268,7 +268,7 @@ class Project(ToStringMixin):
         path = os.path.normpath(path)
 
         try:
-            return os.path.commonpath([self.project_root, path]) == self.project_root
+            return os.path.commonpath([self.project_root, path]) == os.path.normpath(self.project_root)
         except ValueError:
             # occurs, in particular, if paths are on different drives on Windows
             return False
@@ -295,7 +295,10 @@ class Project(ToStringMixin):
         :param require_not_ignored: if True, the path must not be ignored according to the project's ignore settings
         """
         if not self.is_path_in_project(relative_path):
-            raise ValueError(f"{relative_path=} points to path outside of the repository root; cannot access for safety reasons")
+            raise ValueError(
+                f"{relative_path=} points to path outside of the repository root "
+                f"(project root: {self.project_root}); please use a path relative to the project root"
+            )
 
         if require_not_ignored:
             if self.is_ignored_path(relative_path):
--- a/test/serena/test_project.py
+++ b/test/serena/test_project.py
@@ -1,0 +1,206 @@
+import contextlib
+import os
+import platform
+import shutil
+import tempfile
+from pathlib import Path
+
+import pytest
+
+from serena.config.serena_config import ProjectConfig, SerenaConfig
+from serena.project import Project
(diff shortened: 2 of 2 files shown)


MischaPanch (approved):
LGTM, @opcode81 wanna have a brief look?

opcode81:
See my comment in the issue.

src/serena/project.py
```diff
@@ -268,7 +268,7 @@ def is_path_in_project(self, path: str | Path) -> bool:
         path = os.path.normpath(path)
 
         try:
-            return os.path.commonpath([self.project_root, path]) == self.project_root
+            return os.path.commonpath([self.project_root, path]) == os.path.normpath(self.project_root)
```
opcode81: Calling `normpath` on `self.project_root` should never be required, because the path with which a Project instance is constructed is always normalised already.
If it is not, then this is where the problem should be solved, not here.

weiconghe:
Thanks for the careful review — you're right. I dug into this more and the original error was actually correct behavior.

My setup: a single Serena instance (streamable-http on a fixed port) shared across multiple coding agents. The active project was `self-agriculturebaogong-backend`, but an agent working in a different directory passed an absolute path from that directory to `list_dir`. Serena correctly rejected it — the path was genuinely outside the project root, not a separator mismatch.

So `project_root` was already normalized via `Path.resolve()`, exactly as you pointed out. The `normpath` change isn't needed. Closing this PR.

(For the record, the one thing that would've actually helped here is a clearer error message showing the active project root — useful when several agents share one instance — but that's a minor diagnostic tweak, not this PR.)

opcode81:
I added the project root to the error message in ff3c77d4522e5c015866dfef3d7d4cc4ae05bc58

weiconghe:
Thanks for adding the project root to the error message!

Issue #1609, opened 2026-06-25
clangd: re-check for compile_commands.json after a failed Unreal Engine startup


Kiborgik:
For an Unreal Engine project without a usable compile_commands.json, the clangd language
server raises at startup with instructions to generate the database. The check runs once, at
startup, and the failure is cached. After the user generates the database in the same
session, Serena keeps raising the stale "no usable compile_commands.json" error.

There is no restart_language_server or activate_project tool to re-run the check, so the only
way forward is to reconnect the MCP. Once clangd is running it watches compile_commands.json
and reloads on its own, so this affects only the first start, before the database exists.

This is not specific to Unreal. Any project that gains its compilation database mid-session
hits the same stale-startup behavior. Two directions:

- Re-check on demand: re-run the database preparation when a tool is called after a failed
  start, instead of caching the failure.
- A restart tool: a way to restart a language server without reconnecting the MCP.

A companion PR documents the empty or missing database error and tells the user to reconnect
after generating the database.

MischaPanch:
There is literally a tool called `RestartLanguageServerTool`, it's just optional since most users don't need it. You can enable it in the config. Closing this, feel free to reopen if I misunderstood

PR #1610, opened 2026-06-25
Unreal Engine: reorder the setup guide, fail on an empty compile_commands.json

Kiborgik:
Follow-up to #1564 from real use of the Unreal Engine support on a UE 5.7 project: an
installed Launcher engine, a C++ project outside the engine tree, MSVC, and no standalone
LLVM. It updates the setup guide, corrects the clangd startup check, and adds two unit tests.

## Why

On that setup the guide's primary route failed and the working route looked unavailable.

- The VSCode project generator, the route the guide led with, writes an empty `[]` for an
  out-of-tree project on an installed engine. UnrealBuildTool treats the project as foreign
  and emits no commands, so the output stays an empty array even after clearing
  `Intermediate/`. The file is also named `compileCommands_Default.json`, not
  `compileCommands_<Project>.json`.
- `GenerateClangDatabase` was labeled "requires LLVM installed", which reads as a separate
  LLVM download. It needs a clang toolchain, which the Visual Studio "C++ Clang tools"
  component provides, and it also runs without one through `-Compiler=VisualStudio2022`.
- The empty `[]` then passed Serena's existence check, so clangd started against a database it
  could not use and resolved nothing, with no error.
- The startup error pointed only at the guide path inside the install directory, which an
  agent driving Serena usually cannot open.

## Root cause

UnrealBuildTool, UE 5.7.

- `VSCodeProjectFileGenerator.cs`: the project root is the engine root unconditionally, so an
  out-of-tree project is always foreign. The foreign branch keeps only the game-name-matched
  project, then the per-target loop skips every target on `BuildTarget == null`, so no `.rsp`
  files are written and `WriteCompileCommands` emits `[]`.
- `Modes/GenerateClangDatabase.cs` injects `-Compiler=Clang` unless you pass your own
  `-Compiler`, and `VCEnvironment.cs` throws "Clang x64 must be installed" when it finds no
  LLVM directory. Passing `-Compiler=VisualStudio2022` avoids the forced Clang and builds an
  MSVC database with no toolchain install. clangd itself is the only LLVM tool needed for
  navigation, and Serena already bundles it.

## Changes

- Guide: recommend `GenerateClangDatabase` with the Visual Studio "C++ Clang tools" component,
  document the `-Compiler=VisualStudio2022` MSVC database for setups without a clang toolchain,
  and demote the VSCode generator with a step to verify its output is a non-empty array. Add
  notes for Rider users, when to regenerate, and how to read clangd's index logs.
- clangd: treat an empty `[]` database the same as a missing one for an Unreal Engine project,
  through a shared helper, and inline the `GenerateClangDatabase` command in the startup error.
- Two unit tests for the empty-database paths.

## Known limitation

The missing or empty database check runs only at language server startup and the failure is
cached. After you generate the database in the same session, Serena keeps raising the stale
error, and there is no restart tool, so you have to reconnect the MCP. Once clangd is running
it watches the database and reloads on its own.

This is language-server lifecycle behavior rather than an Unreal issue, so I kept it out of
this PR. I opened #1609 describing it with two possible fixes.

## Testing

Local, Windows 11, clangd 19.1.2 auto-downloaded.

On the real project, with a native-clang database from the Route 1 command, Serena resolves
the full symbol graph: get_diagnostics_for_file on a core gameplay class returns no errors,
and find_referencing_symbols on its base class resolves the subclass hierarchy, cross-module
call sites, and all method definitions. The MSVC database resolves the same symbols and logs
the _Failed to compile ... index may be incomplete lines the guide now explains_.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,19 +24,7 @@ Status of the `main` branch. Changes prior to the next official version change w
     `activate_project`/project lookup, raising `FileNotFoundError` in `RegisteredProject.matches_root_path`
 
 * Language Servers:
-  - C/C++ (clangd): add Unreal Engine 5 fixture and tests verifying that reflection-macro
-    code (`UCLASS`, `UFUNCTION`, `UPROPERTY`, `GENERATED_BODY`) yields correct symbols,
-    references, definitions, and rename edits in hand-written sources (never in
-    UHT-generated files); add an Unreal Engine setup guide.
-  - C/C++ (clangd, ccls): detect Unreal Engine projects by a `.uproject` file at the project
-    root and skip the engine build/cache directories (`Binaries`, `DerivedDataCache`,
-    `Intermediate`, `Saved`) during indexing. clangd fails with a pointer to the Unreal Engine
-    setup guide when started in such a project without a `compile_commands.json`. #1566
-  - C/C++ (clangd): for Unreal Engine projects, treat an empty `compile_commands.json`
-    the same as a missing one, since clangd cannot index against it, and inline the
-    `GenerateClangDatabase` command in the startup error instead of only pointing at the
-    setup guide. The Unreal Engine setup guide is reordered to put UnrealBuildTool's clang
-    database first.
+  - C/C++ (clangd): improve support and documentation for Unreal Engine 5 projects.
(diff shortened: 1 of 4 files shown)


MischaPanch:
Thanks for the fix!

PR #1614, opened 2026-06-26
Add pyrefly as alternative lsp for python

### LINKED ISSUE #1392: Support pyrefly as alternative lsp for python
[Pyrefly](https://pyrefly.org/) is faster than pyright and more mature than ty. https://pyrefly.org/blog/speed-and-memory-comparison/
Command to start the lsp server is `pyrefly lsp`

0x1997:
Close #1392. Related tests passed on my machine.

### DIFF
--- a/docs/01-about/020_programming-languages.md
+++ b/docs/01-about/020_programming-languages.md
@@ -125,6 +125,11 @@ Some languages require additional installations or setup steps, as noted.
   we also support [Phpactor](https://github.com/phpactor/phpactor) (language `php_phpactor`), which requires PHP 8.1+;
   and the experimental [PHPantom](https://github.com/PHPantom-dev/phpantom_lsp) backend (language `php_phpantom`)
 * **Python**
+  (by default, uses [Pyright](https://github.com/microsoft/pyright) (language `python`);
+  alternatives: [ty](https://github.com/astral-sh/ty) (language `python_ty`),
+  [pyrefly](https://github.com/facebook/pyrefly) (language `python_pyrefly`),
+  [Jedi](https://github.com/palotas/jedi-language-server) (language `python_jedi`);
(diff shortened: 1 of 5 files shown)


itxaiohanglover:
Nice addition! Pyrefly as an alternative Python LSP gives users more choices. The uvx-based startup with version pinning and workspace indexing limit is well-structured. Docs update listing all Python LSP options is helpful.

test/resources/repos/python/test_repo/pyrefly.toml
```diff
@@ -0,0 +1 @@
+search_path = ["."]
```
MischaPanch: Is this needed for pyrefly to function? If yes, that's a problem,  most users won't have that and shouldn't be expected to have it

0x1997: I removed the `search_path` line. A `pyrefly.toml` or `pyproject.toml` file is still needed for pyrefly to determine the project root. Otherwise some tests will fail.

MischaPanch: Do we have a guard against that? The LS should fail startup with a suitable error message if the condition is not met. Alternatively, a pyrefly.toml could be generated on the fly instead of raising, though I'm not sure how much sense it makes

MischaPanch (changes requested):
Overall this looks good and it's a nice contribution, thank you! Let's just make sure we either set up a guard against the missing toml or create it on the fly.

Seeing that an empty pyrefly.toml actually works, creating such an empty file on the fly seems like a reasonable default for me, though we should log a warning when the creation is triggered.

After addressing this, the PR can be merged

MischaPanch:
Thanks, merging (I will add a changelog entry on main). I am tinkering with CI atm, so the failing tests are not important

PR #1619, opened 2026-06-28
Replace mypy by ty

MischaPanch:
Waiting for mypy got really tiring... Now type check is less than a second.

I gave claude a /goal while doing some parallel work

### DIFF
--- a/.serena/memories/conventions.md
+++ b/.serena/memories/conventions.md
@@ -14,13 +14,8 @@
 - mccabe complexity cap: 20.
 
 ## Typing (ty)
-- Type checker is **ty** (Astral), configured under `[tool.ty]` in `pyproject.toml`. (Replaced mypy.)
-- `[tool.ty.environment].python-platform = "all"`: analyze for all platforms (not the OS ty runs on). Keeps the check deterministic across the CI matrix and lets platform-conditional stdlib members (`subprocess.CREATE_NO_WINDOW`, `ctypes.windll`, `pwd`, `os.getuid`) resolve without per-OS type-ignores. Do NOT pin a single platform — each OS view has its own gaps.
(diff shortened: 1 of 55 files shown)


opcode81:
This is not correctly rebased onto main. The branch contains alternate versions of commits on main.

.serena/memories/conventions.md
```diff
@@ -13,10 +13,13 @@
...
+- Excluded paths (`[tool.ty.src].exclude`): `build/`, `docs/`. NOTE: `test/resources/` is deliberately NOT excluded here — that config is also read by the ty language server (the `PYTHON_TY` diagnostics provider) when it analyzes the fixture repos under `test/resources`, and excluding them would suppress the diagnostics those tests rely on. `test/resources` is instead excluded from the `ty check test` CLI task via `--exclude test/resources` (and carved out of the test override via its `exclude`).
+- Test files (`[[tool.ty.overrides]]` on `test/**`, excluding `test/resources/`) relax type rules that ty mis-fires on for pytest/MagicMock-heavy code (e.g. `invalid-argument-type`, `unsupported-operator`, `too-many-positional-arguments`, `parameter-already-assigned`, `unresolved-attribute`); still checked for undefined names, syntax, unused ignores, etc.
+- Suppress a ty diagnostic with `# ty: ignore[<rule>]`; `# type: ignore` (blanket) is also respected. Unused ignores are reported (`unused-type-ignore-comment`).
```
opcode81: This is not relevant information on conventions. These are technical details on ty's configuration, which should be documented in the place where they are applied, i.e. pyproject.toml - which is already the case. So this is an unnecessary duplication that isn't helpful.

MischaPanch:
> This is not correctly rebased onto main. The branch contains alternate versions of commits on main.

We'll squash this

.serena/memories/conventions.md
MischaPanch: shortened to just one line (typical opus verbosity...)

PR #1622, opened 2026-06-29
ADE updates, part 2

opcode81:


opcode81:
@MischaPanch please decide whether to salvage or close.

Issue #1606, opened 2026-06-25
project.yml: add on_activate_command — a shell hook that runs before any backend initializes


jordancarney1:
**Problem**
When `activate_project` is called, Serena fully manages the LSP process lifecycle. It spawns the language server, waits for it to be ready, and shuts it down when the project is deactivated. The JetBrains backend has no equivalent. Serena simply calls `find_client()` and raises `ServerNotFoundError` if the IDE isn't already (manually) open. This creates an asymmetry that makes the JetBrains backend unsuitable for headless or automated workflows.

The gap isn't only about JetBrains. There's no general mechanism for a project to declare setup that must happen before any backend initializes. Common scenarios that are currently impossible without external tooling:

- Containerized LSPs: ensure a Docker container is running before spawning a language server that connects to it
- Monorepo prep: run npm install or codegen before a TypeScript LSP starts indexing
- Remote LSPs: establish an SSH tunnel before a remote language server is reachable
- JetBrains headless: start a Rider daemon against the right git worktree .sln before `find_client()` scans for it
- Environment gating: fail fast with a clear message if a required service isn't available

**Proposed Solution**
Add an optional `on_activate_command` field to `project.yml` that Serena executes as a shell command before any backend initialization:

```
# Optional shell command to run before the language backend (LSP or JetBrains)
# is initialized. Serena will execute this and wait for it to complete.
# A non-zero exit code aborts activation with the command's stderr as the error message.
on_activate_command: ""
```

The hook would fire `in _activate_project()` in `agent.py`, immediately before the backend-specific branch:

```
# (proposed addition)
if project.project_config.on_activate_command:
    self._run_on_activate_command(project)

# existing code — unchanged
if self.get_language_backend().is_lsp():
    self.issue_task(init_language_server_manager)
```

**Design notes**
- Backward compatible: field defaults to empty string / None, existing behavior unchanged
- Backend-agnostic: fires for both LSP and JetBrains, not tied to either concept
- Blocking: runs synchronously before the backend initializes, so the command can ensure preconditions are met
- Exit code semantics: non-zero exit aborts activation with a clear error, consistent with how other CLI hook systems work
- Scope: single field on `ProjectConfig`, one call site in `_activate_project`, no new dependencies

**Questions for maintainers**
1. Is this a direction you'd accept a PR for?
2. Any preference on the field names: `on_activate_command`, `pre_activation_command`, `setup_command`?

opcode81:
>  The JetBrains backend has no equivalent

[...]

PR #1623, opened 2026-06-29
Add activation_command to project config

jordancarney1:
## Summary

Adds two optional fields to `ProjectConfig` that allow a project to declare a shell command that runs **before the language backend (LSP or JetBrains) initialises**:

- `activation_command` - the shell command, run in the project root
- `activation_command_timeout` - safety backstop in seconds (default `180`); on expiry the process tree is killed and activation continues

Discussed and scoped in #1606. Follows the implementation path outlined by @opcode81.

## Design

- **Trust-gated**: only executes for trusted projects (`trusted_project_path_patterns`), matching the existing `ls_specific_settings` pattern. Untrusted projects log a warning and skip.
- **Exit code is the primary signal**: a non-zero exit is logged as an error but never aborts activation (no return channel to the LLM at that point, per the constraint in #1606).
- **Timeout is a safety backstop**: non-terminating commands (e.g. a watch task accidentally used as the command) would otherwise permanently wedge the serial task executor and make every subsequent tool call time out for the rest of the session. On expiry, `terminate_process_tree_with_kill_fallback` kills the process tree and activation continues.
- **Backend-agnostic**: fires before both LSP and JetBrains backend init.
- **`project.local.yml` override works for free**: both fields use the existing local-override mechanism, so per-machine timeout overrides are possible without touching the committed `project.yml`.

## Motivation / concrete use case

A TypeScript monorepo using NX with project references, where `dist/` is gitignored. On a fresh clone, cross-package imports are unresolvable until declaration files are generated:

```yaml
activation_command: "npx nx run-many -t build"
```

A global agent startup hook can technically dispatch per-project, but that puts project-specific knowledge in a global file that doesn't travel with the repo. The command belongs with the project config.

## Changes

- `src/serena/config/serena_config.py` - two new `ProjectConfig` fields, validation, `_from_dict` wiring
- `src/serena/agent.py` - `_run_activation_command()` method; called from `init_project_services` before `_init_active_project_language_backend`
- `src/serena/resources/project.template.yml` - documented fields (required for template-completeness test)
- `CHANGELOG.md` - appended to the trusted-projects "Current effects" sub-list
- `docs/02-usage/050_configuration.md` - new "Activation Command" section
- `test/serena/config/test_serena_config.py` - config parse tests (defaults, round-trip, validation)
- `test/serena/test_activation_command.py` - agent-level tests (trusted runs, untrusted skips, failure continues, timeout kills and continues)
- `scripts/smoke_test_activation_command.py` - end-to-end smoke test (no LLM required, follows the demo script pattern)

## Test plan

- [x] `poe format` - clean
- [x] `poe type-check` - clean  
- [x] 12 new tests passing (`test_serena_config.py::TestProjectConfigActivationCommand`, `test_activation_command.py`)
- [x] Existing `test_template_is_complete` passes
- [x] Smoke tested end-to-end: `activation_command` fired and sentinel written before JetBrains backend init on a real C# project


docs/02-usage/050_configuration.md
```diff
@@ -32,6 +32,7 @@ Some of the configurable settings include:
   * global ignore rules
   * logging settings
   * the set of trusted project paths
+  * a pre-backend [activation command](activation-command) run before the language server starts (per project, trusted projects only)
```
opcode81: The changes in this file are in the wrong place. This file is about the global configuration.
Move to `Project Configuration` section in 040_workflow.md

scripts/smoke_test_activation_command.py
```diff
@@ -0,0 +1,74 @@
+"""
+Smoke test for the activation_command feature.
```
opcode81: Please remove this file.

jordancarney1: Removed.

docs/02-usage/050_configuration.md
jordancarney1: Updated.

opcode81:
Thanks @jordancarney1!

PR #1629, opened 2026-07-01
Disable mcp tools by language

Tobbse:
Hey 👋 
Here's a PR for a new config option to disable specific tools for specific languages in a project.
Please let me know what you think! Happy to address any feedback.

## Why?
Sometimes I need to use `exclude_tools` as certain language server implementations are unreliable. That particularly bothered me after using it for more Haxe code recently (after we just added support for it a few weeks back).
I have a Haxe codebase to test this with (unusually large and complex for a Haxe project), and for some tools the language server simply takes too long, so things time out (_btw, this seems to be just the Haxe LSP implementation and not something we can fix in Serena, except maybe improving how to deal with and communicate timeouts and errors, etc._).

Disabling those tools for that project completely is not ideal. It also contains a lot of Java code and our PHP backend, so I don't want to "globally" disable stuff.

## Per-level tool disabling
Therefore, here's a PR that allows you to disable specific tools only for specific languages. I hope this doesn't go against your idea of how Serena should be used. My experience is that in reality not all language server implementations are super robust and fast, so this seems like a useful addition to me. I'm excited to hear your thoughts on it. :)

Basically, there's a new config option `excluded_tools_by_language`:

```
languages: [typescript, python, haxe]

excluded_tools_by_language:
  haxe:
    - find_symbol
    - replace_symbol_body
```

The tools should also clearly communicate that information when it's used. It is exposed as a summary on project activation, mentioning all disabled language-specific tools. Additionally, it's part of the description of each tool.

And when targeting, for example, only Haxe code on a specifically disabled tool, it will return an error like:
```
Error: Tool 'find_symbol' is disabled for haxe in this project; 'src/Main.hx' is a haxe file. Inspect haxe code with text-based tools (e.g. search_for_pattern) or read_file instead.
```

Or when targeting multiple languages, it runs against the "allowed" tools and prepends a one-line note for each language that was specifically disabled:
```
⚠ Coverage: skipped 12 haxe file(s) (tool disabled for haxe).
[...]
```

`excluded_tools` should still work as before.

## Checklist

### DIFF
--- a/.serena/project.yml
+++ b/.serena/project.yml
@@ -3,24 +3,27 @@ project_name: "serena"
 
 
 # list of languages for which language servers are started; choose from:
-#   al                  ansible             bash                clojure             cpp
-#   cpp_ccls            crystal             csharp              csharp_omnisharp    dart
-#   elixir              elm                 erlang              fortran             fsharp
-#   go                  groovy              haskell             haxe                hlsl
-#   java                json                julia               kotlin              lean4
-#   lua                 luau                markdown            matlab              msl
-#   nix                 ocaml               pascal              perl                php
-#   php_phpactor        powershell          python              python_jedi         python_ty
-#   r                   rego                ruby                ruby_solargraph     rust
-#   scala               solidity            swift               systemverilog       terraform
-#   toml                typescript          typescript_vts      vue                 yaml
-#   zig
+#   al                  angular             ansible             bash                clojure
+#   cpp                 cpp_ccls            crystal             csharp              csharp_omnisharp
+#   dart                elixir              elm                 erlang              fortran
+#   fsharp              go                  groovy              haskell             haxe
+#   hlsl                html                java                json                julia
+#   kotlin              lean4               lua                 luau                markdown
+#   matlab              msl                 nix                 ocaml               pascal
+#   perl                php                 php_phpactor        powershell          python
+#   python_jedi         python_ty           r                   rego                ruby
+#   ruby_solargraph     rust                scala               scss                solidity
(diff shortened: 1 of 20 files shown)


Tobbse:
And I hope this can be counted as "Isolated additions that don't change existing behaviour".

opcode81:
> And I hope this can be counted as "Isolated additions that don't change existing behaviour".

Unfortunately, it cannot, so we will have to reject this.
We really cannot skip the software design phase for changes like this.
And we must furthermore consider alternative ways of addressing the underlying problem before we settle on a solution.

That being said, it's not clear to me how your example can result in a useful configuration: `find_symbol` is arguably the most useful symbolic tool, yet you are disabling it. Is there really a point in keeping other LSP tools?
And why is `find_symbol` unbearably slow in the first place? Serena caches the language server's symbol information after initial indexing, so `find_symbol` should be fast unless you constantly switch branches or otherwise invalidate the (per-source file) cache.

Tobbse:
thanks for the reply. okay, I can understand that of course.

as for my example: it was just that, an example, but it was really badly chosen. I mixed up `find_symbol` and `find_referencing_symbols`. that is one of the compiler-backed tools causing trouble for me on the Haxe side with large projects, because the language server is just so slow to respond.

I will see if there's something I can fix in the Haxe implementation instead, but I can't really fix a slow language server. oh well, let's see. :)

opcode81:
If you end up concluding that there is a need to address this in Serena, we should go back to the drawing board and you can open an issue to discuss solutions.

Issue #1435, opened 2026-04-29
A broken mklink folder will trigger : "The ignore spec could not be computed"


friedikitty:
Easy to reproduce this problem:

- Windows
- Initialize a project with Serena.
- In the project, use mklink to link some_folder to an existing folder.
- Delete or rename the existing folder. The bug was not intentional; the link failed because it pointed to a cloud disk using S3, which cannot be avoided (network, what ever).
- serena’s read_file fails with: The ignore spec could not be computed.  **Notice, not a file in that failed folder**.
- I cleaned up all the .gitignore files, but I still got the same problem. I had to use Codex to locate the issue. It seems to have nothing to do with .gitignore
- By codex's instruction, remove the broken linked folder and restart the MCP. The problem is solved.

- here is what codex found in the local serena git, which I did not confirm, just for references -- (newest: 5dd1a410000142ce0d9f1acbbbe9dd9dd9d2fb13, 2026/04/28):
```
The failure is in file_system.py (line 167): os.scandir(abs_path) is not protected against FileNotFoundError. It only catches PermissionError around each entry. Then project.py (line 328) calls GitignoreParser, catches the exception, leaves __ignore_spec unset, and later read_file fails because it validates ignored paths first at file_tools.py (line 38).
```

I have:

- [x] read the relevant parts of the documentation and verified that the issue cannot be solved by adjusting configuration
- [x] understood that the Serena Dashboard can be disabled through the config
- [x] understood that, by default, a client session will start a separate instance of a Serena server. 
- [x] understood that, for multi-agent setups, the Streamable HTTP/SSE mode should be used.
- [x] understood that non-project files are ignored using either .gitignore or the corresponding setting in `.serena/project.yml`
- [x] looked for similar issues and discussions, including closed ones
- [x] made sure it's an actual issue, not a question (use GitHub Discussions instead).

If you have encountered an actual issue:

- If using language servers (not the JetBrains plugin), 
  - [x] I performed `<uv invocation> serena project health-check`
  - [x] I indexed the project as described in the documentation
- [x] I added sufficient explanation of my setup: the MCP client, the OS, the programming language(s), any config adjustments or relevant project specifics
- [x] I explained how the issue arose and, where possible, added instructions on how to reproduce it
- [x] If the issue happens on an open-source project, I have added the link
- [x] I provided a meaningful title and description

MischaPanch:
[...]

PR #1634, opened 2026-07-02
fix(file_system): handle PermissionError on scandir in gitignore walker

### LINKED ISSUE #1624: PermissionError on scandir crashes gitignore walker even when target path is in ignored_paths
## Summary

`serena project index` crashes with `PermissionError` when the project tree contains a subdirectory the current user can't read (e.g., a Docker-container-created dir owned by `root:root` mode `0700`), even when that path is explicitly listed in `.serena/project.yml` under `ignored_paths`.

Same code path as #1435 (which fixed `FileNotFoundError` on broken symlinks) but with `PermissionError` instead — needs identical guard.

## Environment

- Serena: `1.5.4.dev0` (commit `103a17072e9b`)
- Python: 3.13.13
- OS: Ubuntu Linux 6.17.0
- Project: Laravel/PHP, ~12k files, sibling Docker container produces root-owned artifacts inside the project tree

## Reproduction

1. Inside the project tree, create a directory the current user can't read:
   ```bash
   # Either run a Docker container that creates files in storage/framework/testing
   # as root (default behavior of laravel/sail-style stacks), or:
   sudo mkdir -p storage/framework/testing && sudo chmod 700 storage/framework/testing
   ```

2. List it in `.serena/project.yml`:
   ```yaml
   ignored_paths:
     - "storage/framework/testing/**"
   ```

3. Run `serena project index <path> --log-level INFO`

## Expected

Indexing skips the unreadable directory (gitignored + in `ignored_paths`) and completes successfully.

## Actual

Crashes at startup before any file is indexed:

```
File "serena/util/file_system.py", line 190, in _iter_gitignore_files
    yield from scan(next_abs_path)
File "serena/util/file_system.py", line 167, in scan
    for entry in os.scandir(abs_path):
                 ~~~~~~~~~~^^^^^^^^^^
PermissionError: [Errno 13] Permission denied:
  '/home/sejin/PhpstormProjects/master/storage/framework/testing'
...
File "serena/project.py", line 195, in _ignored_patterns
    raise ValueError("The ignored patterns could not be computed; ...")
```

## Root cause

`file_system.py:167` — `for entry in os.scandir(abs_path):` is not guarded against `PermissionError` on the scandir call itself. The per-entry guard added in #1435 handles `FileNotFoundError` but not `PermissionError` on directory access.

Secondary issue: the walker doesn't consult `ignored_paths` before recursing — it enumerates the whole tree first, then applies the filter. Even when the user explicitly asks to skip a path, Serena tries to read it.

## Workaround

Set `ignore_all_files_in_gitignore: false` and manually mirror the entire `.gitignore` in `ignored_paths`. Skips the recursive gitignore walker entirely. Tedious; loses inheritance of nested `.gitignore` files.

## Suggested fix

Wrap `os.scandir()` in `file_system.py:167` with both `PermissionError` and `FileNotFoundError`:

```python
try:
    entries = os.scandir(abs_path)
except (PermissionError, FileNotFoundError) as e:
    LOG.debug(f"Skipping unreadable directory {abs_path}: {e}")
    return
for entry in entries:
    ...
```

Or apply `ignored_paths` as a pre-filter on directory descent (more invasive but architecturally better — Serena's own config should win over filesystem accessibility).

weiconghe:
### Issue for this PR

Closes #1624

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

### What does this PR do?

**Problem:** `serena project index` crashes with `PermissionError` when the project tree contains a subdirectory the current user can't read (e.g., a Docker-container-created dir owned by `root:root` mode `0700`), even when that path is explicitly listed in `.serena/project.yml` under `ignored_paths`.

**Root cause:** `_iter_gitignore_files()` in `file_system.py` contains an inner `scan()` function that calls `os.scandir(abs_path)` without guarding against `PermissionError` on the directory itself. While the inner loop already handles `PermissionError` on individual entries (added by #1435 for `FileNotFoundError` on broken symlinks), the `os.scandir()` call itself crashes if the entire directory is unreadable.

**Fix:** Wrap the `os.scandir()` call in a `try/except PermissionError` block. If the directory can't be read, skip it with a debug log message.

- Added regression test `test_scandir_permission_error_on_subdirectory` that creates an unreadable subdirectory (`chmod 000`) and verifies `GitignoreParser` initializes without crashing and still functions correctly
- All 24 existing tests in `test_file_system.py` continue to pass
- Fix does not affect non-Linux platforms (where `os.chmod 000` is a no-op, the test still passes)

### Checklist

- [x] I have included no unrelated changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Status of the `main` branch. Changes prior to the next official version change w
   - Update prompts/instructions: Serena instructions manual, modes (editing, interactive) 
   - Allow structured tool output to be configured on a per-context basis, disabling it for Claude Code
     (which does not correctly unpack structured output) #1042
+  - Fix: File system permission errors during gitignore scanning were not caught #1624 
 
 * CLI:
   - Fix `--project-from-cwd` hijacking git worktrees nested under a Serena project. `find_project_root`
--- a/src/serena/util/file_system.py
+++ b/src/serena/util/file_system.py
@@ -164,7 +164,12 @@ class GitignoreParser:
         queue: list[str] = [self.repo_root]
 
         def scan(abs_path: str | None) -> Iterator[str]:
-            for entry in os.scandir(abs_path):
+            try:
+                entries = os.scandir(abs_path)
+            except PermissionError as ex:
+                log.debug(f"Skipping unreadable directory {abs_path}: {ex}")
+                return
+            for entry in entries:
                 try:
                     if entry.is_dir(follow_symlinks=follow_symlinks):
                         queue.append(entry.path)
--- a/test/serena/util/test_file_system.py
+++ b/test/serena/util/test_file_system.py
@@ -3,7 +3,6 @@ import shutil
 import tempfile
 from pathlib import Path
 
-# Assuming the gitignore parser code is in a module named 'gitignore_parser'
 from pathspec import PathSpec
 
 from serena.util.file_system import GitignoreParser, GitignoreSpec, match_path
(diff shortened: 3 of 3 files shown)


opcode81:
Thanks for your PR!

PR #1635, opened 2026-07-02
feat: add QML language support via Qt's qmlls

### LINKED ISSUE #1381: RFE: QML language support
This is a missing feature. so no existing problem.

QML is a language used to build declarative, cross-platform user interfaces using the Qt UI libraries. Qt ships with a LSP server (`qmlls` or `qmlls6` depending on the installation) which could technically be used by Serena.
I'm not, however, sure if it's all that would be needed or if additional plumbing (aside wiring support in) would be needed.

Existing solutions for Claude Code at least are too complicated, they implement far more than what I'd like to have (which is what Serena does).

weiconghe:
### Issue for this PR

Closes #1381

### Type of change

- [x] New feature (non-breaking change which adds functionality)

### What does this PR do?

Adds QML (Qt Modeling Language) language support to Serena using Qt's official `qmlls` (or `qmlls6`) language server.

**Implementation details:**
- New `QmlLanguageServer` class in `src/solidlsp/language_servers/qml_language_server.py` that discovers the `qmlls6` or `qmlls` binary from PATH (same pattern as Crystal/Crystalline)
- Registered `QML` language in `ls_config.py` with `.qml` file extension
- The server looks for `qmlls6` first (Qt 6), then falls back to `qmlls`

**User setup:**
Users need to have Qt 6 installed with `qmlls` (or `qmlls6`) available on PATH. Qt ships this binary with the Qt 6 SDK. The language server is a standard LSP implementation, so all standard Serena features (document symbols, go-to-definition, references, etc.) should work.

- Verified language registration: `Language.QML` resolves correctly, file extension matching works, and `get_ls_class()` returns `QmlLanguageServer`
- Added test repository with QML source files (`src/main.qml`, `src/shapes.qml`)
- Added a basic test suite validating server startup and document-symbol retrieval; ran all 3 tests locally against a real `qmlls` binary — **3 passed**. qmlls names QML objects by their *type*, so the assertions check `ApplicationWindow`/`Button` (main) and `Rectangle`/`Text` (shapes)
- CI now installs `qmlls` on the Ubuntu `other-langs` batch. `textDocument/documentSymbol` needs qmlls from Qt 6.6+, but Ubuntu noble's apt only ships Qt 6.4.2 (whose qmlls returns -32601); so CI uses the official standalone build (`TheQtCompanyRnD/qmlls-workflow` 0.7, linked from the Qt docs). Verified that Linux 0.7 binary advertises `documentSymbolProvider` and returns symbols via `textDocument/documentSymbol`. Tests skip on Windows/macOS CI and locally when `qmlls` is not on PATH (guarded centrally by `language_tests_enabled`)
- All 24 existing `test_file_system.py` tests continue to pass

### Checklist

- [x] I have included no unrelated changes

### Screenshots / recordings

N/A

### DIFF
--- a/src/solidlsp/language_servers/qml_language_server.py
+++ b/src/solidlsp/language_servers/qml_language_server.py
@@ -1,0 +1,128 @@
+"""
+Provides QML specific instantiation of the LanguageServer class using Qt's qmlls.
+"""
+
+import logging
+import shutil
+
+from solidlsp.ls import (
+    LanguageServerDependencyProvider,
+    LanguageServerDependencyProviderSinglePath,
+    SolidLanguageServer,
+)
+from solidlsp.ls_config import LanguageServerConfig
+from solidlsp.settings import SolidLSPSettings
+
+log = logging.getLogger(__name__)
+
+
+class QmlLanguageServer(SolidLanguageServer):
+    """
+    Provides QML specific instantiation of the LanguageServer class using qmlls.
+
+    qmlls is the official QML language server shipped with Qt 6.
+    It must be installed separately; see https://doc.qt.io/qt-6/qtqml-tool-qmlls.html
+    for installation instructions.
+
+    The dependency provider looks for ``qmlls6`` first, then falls back to ``qmlls``.
+    Users can override the executable via the ``ls_path`` entry in ``ls_specific_settings.qml``.
+    """
+
+    def __init__(self, config: LanguageServerConfig, repository_root_path: str, solidlsp_settings: SolidLSPSettings):
+        super().__init__(config, repository_root_path, None, "qml", solidlsp_settings)
+
+    def _create_dependency_provider(self) -> LanguageServerDependencyProvider:
+        return self.DependencyProvider(self._custom_settings, self._ls_resources_dir)
+
+    class DependencyProvider(LanguageServerDependencyProviderSinglePath):
+        def _get_or_install_core_dependency(self) -> str:
+            """
+            Discover the qmlls executable on PATH.
+
+            Tries ``qmlls6`` first (Qt 6+), then falls back to ``qmlls``.
+
+            :return: path to the qmlls executable
+            :raises FileNotFoundError: if qmlls is neither on PATH nor provided via ``ls_path``
+            """
+            qmlls_binary = shutil.which("qmlls6") or shutil.which("qmlls")
+            if qmlls_binary is None:
+                raise FileNotFoundError(
+                    "qmlls (QML language server) is not installed or not in PATH.\n"
+                    "Please install Qt 6 and ensure 'qmlls' (or 'qmlls6') is available on your PATH.\n"
+                    "See: https://doc.qt.io/qt-6/qtqml-tool-qmlls.html"
+                )
+            return qmlls_binary
+
+        def _create_launch_command(self, core_path: str) -> list[str]:
+            # qmlls communicates via stdio by default; no extra flags are required.
+            return [core_path]
+
+    def _create_base_initialize_params(self) -> dict:
+        """
+        Return the language-specific initialize params for the QML language server.
+
+        ``processId``, ``rootPath``, ``rootUri`` and ``workspaceFolders`` are populated by the
+        default ``InitializeParamsBuilder`` and must not be set here.
+        """
+        initialize_params = {
+            "locale": "en",
+            "capabilities": {
+                "textDocument": {
+                    "synchronization": {"didSave": True, "dynamicRegistration": True},
+                    "definition": {"dynamicRegistration": True, "linkSupport": True},
+                    "references": {"dynamicRegistration": True},
+                    "documentSymbol": {
+                        "dynamicRegistration": True,
+                        "hierarchicalDocumentSymbolSupport": True,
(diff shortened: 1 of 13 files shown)


src/solidlsp/language_servers/qml_language_server.py
```diff
@@ -0,0 +1,135 @@
...
+        if qmlls_binary is None:
+            raise RuntimeError(
+                "qmlls (QML language server) is not installed or not in PATH.\n"
+                "Please install Qt 6 and ensure 'qmlls' (or 'qmlls6') is available on your PATH.\n"
+                "See: https://doc.qt.io/qt-6/qtqml-tool-qmlls.html"
+            )
+        return qmlls_binary
+
+    @staticmethod
+    def _get_initialize_params(repository_absolute_path: str) -> InitializeParams:
```
opcode81: The interface for this has now changed on `main` (see also updated memory).
Please rebase your branch onto `main` and adapt it.

opcode81 (changes requested):
Thanks for your PR!

Please address the changed interface (see comment) and add an entry to the changelog.

src/solidlsp/language_servers/qml_language_server.py
```diff
@@ -0,0 +1,135 @@
...
+    The server looks for ``qmlls6`` first, then falls back to ``qmlls``.
+    """
+
+    def __init__(self, config: LanguageServerConfig, repository_root_path: str, solidlsp_settings: SolidLSPSettings):
+        qmlls_path = self._find_qmlls()
+
+        super().__init__(
+            config,
+            repository_root_path,
+            ProcessLaunchInfo(cmd=qmlls_path, cwd=repository_root_path),
```
opcode81: Passing `ProcessLaunchInfo` is deprecated; new implementations should not use it. 
Please implement `LanguageServerDependencyProviderSinglePath` instead. `_find_qmlls` belongs there.
See our memory on adding a language server.

weiconghe:
Update: I verified the server against a real `qmlls` locally (qmlls 0.6 standalone) and it exposed two wrong assertions in my own test suite — qmlls names QML objects by their **type** (`ApplicationWindow`, `Button`, `Rectangle`, `Text`), not by their `id`. The original assertions (`mainWindow`, `customRect`, ...) were checking `id` values that qmlls never surfaces as symbol names. Fixed in `7955076c`; all 3 tests now pass against the real binary (server start + document symbols for both files).

Heads-up: `qmlls` is not installed in CI (no Qt setup in `.github/workflows`), so these tests currently **skip** in CI rather than running — they only run where `qmlls`/`qmlls6` is on PATH. Per the add-language guide ("tests should run in CI"), it'd be worth wiring up a CI install step (qmlls ships standalone, so a small download would suffice). Happy to add that here or in a follow-up — let me know your preference.

opcode81:
> qmlls is not installed in CI

I am aware of this. However, CI was recently restructured by @MischaPanch and I would leave it up to him to decide whether QML should be tested and, if so, in which part of the matrix.

test/conftest.py
MischaPanch, line 404: @weiconghe pls add the QML download and marker in the section "other-langs" in the GH workflow pytest.yml. If it's only possible on ubuntu, restrict the installation and marker to ubuntu. Then move the disabled check to this section, as the QML support will be tested in CI

MischaPanch (changes requested):

test/conftest.py
MischaPanch: Thanks! Good that we added it, caught some issues (see failed action)

```
FAILED test/solidlsp/qml/test_qml_basic.py::TestQmlLanguageServer::test_document_symbols_main[qml] - solidlsp.ls_exceptions.SolidLSPException: Error processing request textDocument/documentSymbol with params:
{'textDocument': {'uri': 'file:///home/runner/work/serena/serena/test/resources/repos/qml/test_repo/src/main.qml'}}
(caused by There was no handler registered with registerDocumentSymbolHandler to handle a requests with method textDocument/documentSymbol (-32601))
FAILED test/solidlsp/qml/test_qml_basic.py::TestQmlLanguageServer::test_document_symbols_shapes[qml] - solidlsp.ls_exceptions.SolidLSPException: Error processing request textDocument/documentSymbol with params:
{'textDocument': {'uri': 'file:///home/runner/work/serena/serena/test/resources/repos/qml/test_repo/src/shapes.qml'}}
(caused by There was no handler registered with registerDocumentSymbolHandler to handle a requests with method textDocument/documentSymbol (-32601))
```

MischaPanch:
@weiconghe in principle ready to merge, thanks for the changes. But I noticed that the tests don't cover reference search (neither within nor cross-file), the thing that tends to have problems most often. Also diagnostics are not tested. If you feel like adding those tests before the merge, it would be nice. Otherwise we could merge but I wouldn't be too confident that everything works with these tests missing. For cross-file reference search it's important to test that references are found in files that were not opened by other tests.

MischaPanch:
@xormania I appreciate the enthusiasm but pls let the maintainers review and call for action in PRs that are not written by yourself, unless we specifically ask for help.

There's nothing to think through here, almost all other languages have this test coverage which can be used as basis for the extension of qml coverage

Issue #1251, opened 2026-03-30
Memory path traversal, shell command timeout, and cross-project memory poisoning (security findings)


AgentSeal:
## Body:                                                                   
                                                                                                                                                                                                   
  **Checklist:**                                                             
  - [x] read the relevant parts of the documentation and verified that the issue cannot be solved by adjusting configuration
  - [x] understood that the Serena Dashboard can be disabled through the config
  - [x] understood that, by default, a client session will start a separate instance of a Serena server                                                                                            
  - [x] understood that, for multi-agent setups, the Streamable HTTP/SSE mode should be used
  - [x] understood that non-project files are ignored using either .gitignore or the corresponding setting in `.serena/project.yml`                                                                
  - [x] looked for similar issues and discussions, including closed ones     
  - [x] made sure it's an actual issue, not a question                                                                                                                                             
                                                                                                                                                                                                   
  ---                                          
                                                                                                                                                                                                   
  We scanned Serena as part of our MCP security registry at [AgentSeal](https://github.com/AgentSeal/agentseal). Score: **35/100 (Risky)**. Found 10 issues across the codebase, 4 of which are  high-impact and have clear fixes. Sharing here for your evaluation.
                                                                                                                                                                                                   
  ### Summary                                                                                                                                                                                      
                 
  | # | Finding | Severity | Fix complexity |                                                                                                                                                      
  |---|---------|----------|----------------|                                
  | 1 | Memory path traversal via unsanitized name | Critical | One-line fix |
  | 2 | Shell execution with no timeout | Critical | Small fix |                                                                                                                                   
  | 3 | Cross-project poisoning via global memory | High | Medium |                                                                                                                                
  | 4 | Arbitrary directory activation | High | Small fix |                                                                                                                                        
  | 5 | Prompt injection to RCE via file content | High | Medium |                                                                                                                                 
  | 6 | System prompt extraction via initial_instructions | Medium | Small fix |                                                                                                                   
                                                                                                                                                                                                   
  ### 1. Memory path traversal via unsanitized name                                                                                                                                                
                                                                             
  `MemoriesManager.get_memory_file_path()` in `src/serena/project.py` constructs paths by splitting the name on `/` without checking for traversal:                                                
                                                                             
  ```python                                                                                                                                                                                        
  parts = name.split("/")                                                    
  filename = f"{parts[-1]}.md"                 
  if len(parts) > 1:                      
[...]

Issue #1235, opened 2026-03-27
Duplicate Serena instances and unsafe concurrent writes when the same project directory is open in multiple MCP clients simultaneously


cperalt:
## Checklist

  - [x] read the relevant parts of the documentation and verified that the issue cannot be solved by adjusting configuration
  - [x] understood that the Serena Dashboard can be disabled through the config
  - [x] understood that, by default, a client session will start a separate instance of a Serena server
  - [ ] understood that, for multi-agent setups, the Streamable HTTP/SSE mode should be used
  - [x] understood that non-project files are ignored using either .gitignore or the corresponding setting in `.serena/project.yml`
  - [x] looked for similar issues and discussions, including closed ones (see #805 — related but distinct)
  - [x] made sure it's an actual issue, not a question

  ## Setup

  - **OS:** macOS
  - **MCP Clients:** Claude Code (terminal) and Cursor IDE (with Claude extension), both configured with Serena at user scope via `~/.claude/plugins`
  - **Language:** TypeScript (large Nx monorepo)
  - **Serena launch command:** `uvx --from git+https://github.com/oraios/serena serena start-mcp-server`

  ## Description

  When the same project directory (or git worktree) is open simultaneously in two different MCP clients — e.g. a terminal Claude Code session and Cursor IDE — each client independently spawns its own Serena process. There is no mechanism for a new client to discover and reuse an already-running instance for the same
   directory.

  This leads to two concrete problems:

  ### 1. Redundant resource usage

  Each instance starts its own language server (e.g. `tsserver`) against the same codebase. For large monorepos this means double the memory and CPU with no benefit — both instances are doing the same indexing work independently.

  ### 2. Unsafe concurrent writes to `.serena/` with no file locking

  Both instances read the cache and memories from `.serena/` on startup into memory, then write back to disk on shutdown. With no file locking:

  - **Cache (`.serena/cache/`):** whichever instance shuts down last overwrites the other's work — last-writer-wins with no merge.
  - **Memories (`.serena/memories/`):** if both instances call `write_memory` on the same file at the same time, the result is undefined.

  ## Reproduction

  1. Open a project directory in a terminal Claude Code session — Serena starts (instance A).
  2. Open the same directory in Cursor IDE with Claude extension — Serena starts again (instance B).
  3. Verify with `ps aux | grep "serena start-mcp-server"` — two processes, same project cwd.
  4. Close both simultaneously — both attempt to write cache to the same `.serena/cache/` path.

  ## Suggested solution

[...]

PR #1637, opened 2026-07-02
Fix: serialize memory writes to prevent corruption under concurrent access

bymebyu:
## Problem

Memory write operations (`save_memory` / `delete_memory` / `move_memory`) perform unguarded filesystem writes. When the same project is operated on concurrently — several threads in one process, or two Serena processes at once (e.g. the same project opened in two MCP clients, cf. #1235) — writes can interleave and leave torn or inconsistent memory files.

## Fix

- Serialize all memory writes behind a `threading.Lock` (in-process) plus an advisory `filelock.FileLock` (cross-process), stored next to the memory directories as `.memory_write.lock`.
- The file lock is created lazily, so constructing a `MemoryManager` has no filesystem side effects — `MemoryManager(serena_data_folder=None)` stays FS-free, as the existing unit tests rely on.
- `filelock` is already a project dependency; no new dependencies.

## Tests

- New `TestConcurrentWrites`: 8 threads × 20 writes to the same memory → the final file is exactly one writer's payload (never a mix); a parallel save/delete loop completes without errors.
- Full `test/serena/test_memories_manager.py` passes (105 passed).
- `ruff check` / `ruff format` clean; CHANGELOG entry added under *Memories*.

Related: #1235 (multi-client concurrent usage), #1251 (memory safety hardening).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,9 @@ Status of the `main` branch. Changes prior to the next official version change w
 
 * Memories:
   - Make memory iteration follow symbolic links 
+  - Fix: serialize memory write operations (save/delete/rename) with a thread lock and an advisory
+    file lock, preventing torn or interleaved memory files when the same project is used from
+    multiple threads or Serena processes concurrently
 
 * Dependencies:
   - Add dependency `oslex`
--- a/src/serena/memories/memory_manager.py
+++ b/src/serena/memories/memory_manager.py
@@ -2,9 +2,13 @@ import logging
 import os
 import re
 import shutil
+import threading
 from collections.abc import Iterator, Sequence
+from contextlib import contextmanager
 from pathlib import Path
 from typing import Literal
+
+from filelock import BaseFileLock, FileLock
 
 from serena.config.serena_config import (
     SerenaPaths,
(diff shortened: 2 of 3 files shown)


opcode81:
> several threads in one process, or two Serena processes at once (e.g. the same project opened in two MCP clients

* Several threads in one process cannot edit the same memory at the same time. Serena's tool executions are fully serialized through the task queue.
* If you have several agents simultaneously working on the same project/files without being centrally orchestrated to work on separate things, you need to make them work on separate worktrees/clones of the project and later merge the changes appropriately via git/the VCS.

So you are essentially solving a problem that only exists when you're most definitely doing something you should not be doing - and issues with memory writes are unlikely to be your only problem in this case.

bymebyu:
Thanks — that's fair, and digging into it you're right on the core:

- Within a single process, tool executions are fully serialized through the task queue, so the in-process lock is redundant.
- The only remaining case is two separate Serena processes on the same project root (e.g. two clients / two sessions on the same folder). A write lock there prevents torn files, but it doesn't make the memory store safe under genuinely concurrent, uncoordinated agents — read-modify-write updates can still be lost, and as you said, memory writes wouldn't be the only casualty in that setup. Separate worktrees/clones with VCS merge is the right model for that.

Since this would only be a partial guard for a usage pattern that has deeper problems anyway, I'll close it rather than ship a half-fix. Thanks for taking the time to explain the reasoning.

PR #1638, opened 2026-07-02
Fix: language composition percentages diluted by unrecognised files

bymebyu:
## Problem

`determine_programming_language_composition` computes percentages over the **total** file count. Files that belong to no supported language (images, plain text, licenses, lock files, binary assets) therefore dilute the numbers: a repository with 2 Python files and 23 assets reports Python at 8%, even though Python is 100% of its source code. In documentation/asset-heavy repositories this makes the reported composition misleading.

## Fix

- Percentages are now computed relative to **files matching at least one supported language** (the denominator is "recognised source files").
- The counting also runs in a single pass over the file list instead of one pass per language (~40 languages), which is noticeably faster on large repositories.

The only consumer (`ProjectConfig.autogenerate`) sorts languages by percentage; a uniform rescaling of the denominator does not change that ranking, so auto-detection behavior is unchanged — only the reported numbers become meaningful.

## Tests

- New `test/serena/util/test_inspection.py` covering: single-language repo, dilution case (2 `.py` + 23 unrecognised files → Python 100%), mixed-language ratios, no-recognised-files and empty repos.
- `test/serena/config` suite passes (63 passed); `ruff check` / `ruff format` clean; CHANGELOG entry added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,10 @@ Status of the `main` branch. Changes prior to the next official version change w
     (which does not correctly unpack structured output) #1042
   - Fix: Project-specific filtering of files for source files ignored the language backend. 
     The check is really only possible for LSP. 
-  - Fix: File system permission errors during gitignore scanning were not caught #1624 
+  - Fix: File system permission errors during gitignore scanning were not caught #1624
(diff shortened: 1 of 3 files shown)


opcode81:
Thanks for your PR!

PR #1639, opened 2026-07-02
TypeScript/VTS: event-based indexing wait and no automatic typing acquisition

bymebyu:
## Problem

1. Before answering cross-file reference requests, the TypeScript language server integration waits a **fixed 2 seconds** (`_get_wait_time_for_cross_file_referencing`). On large projects 2 seconds is often not enough — `find_referencing_symbols` returns incomplete results; on small projects the fixed sleep just wastes time on the first request.
2. During initialization, tsserver's automatic typing acquisition may **download type packages from the network**, slowing startup and introducing a network dependency where none is expected.

## Fix

- Replace the fixed sleep with event-based indexing tracking via the `$/progress` machinery that the codebase already uses (`expect_indexing` / `wait_for_indexing`): before the first cross-file reference request, wait until active indexing progress tokens drain, with a configurable `indexing_timeout` (`ls_specific_settings`, default 30s) as a backstop. This extends the existing progress-tracking mechanism along existing lines rather than introducing new machinery.
- Set `disableAutomaticTypingAcquisition: true` in `initializationOptions` for both TypeScript and VTS servers.

## Tests

- `test/solidlsp/typescript` passes (10 passed).
- `ruff check` clean; CHANGELOG entry added under *Language Servers*.

### DIFF
--- a/src/solidlsp/language_servers/typescript_language_server.py
+++ b/src/solidlsp/language_servers/typescript_language_server.py
@@ -7,6 +7,7 @@ import os
 import pathlib
 import shutil
 import threading
+import time
 from typing import Any, cast
 
 from overrides import override
@@ -73,6 +74,7 @@ class TypeScriptLanguageServer(SolidLanguageServer):
     You can pass the following entries in ls_specific_settings["typescript"]:
         - typescript_version: Version of TypeScript to install (default: "5.9.3")
         - typescript_language_server_version: Version of typescript-language-server to install (default: "5.1.3")
+        - indexing_timeout: float, timeout in seconds for project indexing (default: 30.0)
     """
 
     @classmethod
@@ -81,7 +83,7 @@ class TypeScriptLanguageServer(SolidLanguageServer):
 
     # Safety timeout for $/progress-based indexing wait. Normally the event fires
     # well within this window; the timeout is only hit if the server never sends progress.
-    INDEXING_PROGRESS_TIMEOUT = 15.0 if os.name == "nt" else 10.0
+    INDEXING_PROGRESS_TIMEOUT = 30.0
 
     def __init__(self, config: LanguageServerConfig, repository_root_path: str, solidlsp_settings: SolidLSPSettings):
         """
@@ -231,6 +233,11 @@ class TypeScriptLanguageServer(SolidLanguageServer):
         root_uri = pathlib.Path(repository_absolute_path).as_uri()
         initialize_params = {
             "locale": "en",
+            "initializationOptions": {
+                "preferences": {
+                    "disableAutomaticTypingAcquisition": True,
+                },
+            },
             "capabilities": {
                 "textDocument": {
                     "synchronization": {"didSave": True, "dynamicRegistration": True},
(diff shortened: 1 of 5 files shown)


bymebyu:
CI caught a real defect in the first version — thanks for the thorough matrix. Root cause and fix:

**Why `test_find_symbol_references[typescript_helper_refs]` failed on all three OSes:** the first version replaced the fixed 2-second wait with `wait_for_indexing()`, but the indexing-complete event starts in the *set* state and nothing armed it before the request — so the wait returned instantly and cross-file references ran with **no wait at all**, returning incomplete results.

**Fix (pushed):** the complete mechanism now has three parts:
1. a `_pre_open_for_cross_file_references()` hook, called before `didOpen` in `SymbolLocationRequest.execute()`, which arms the tracking (`expect_indexing()`) so the subsequent wait can observe the progress triggered by opening the file;
2. a 2-second grace period (identical to the previous fixed wait) for tsserver to *start* reporting `$/progress` — if no progress starts, the request proceeds, so the fast path is never slower than before;
3. if progress does start, the request waits until it drains, bounded by `indexing_timeout` (default 30s).

**Validation:** on a machine where 2 seconds is not enough, this exact test fails even on current `main` (incomplete refs); with this change it passes in ~7.5s. `test/solidlsp/typescript` (10) and the typescript subset of `test_serena_agent.py` (10) pass locally, ruff clean.

src/solidlsp/language_servers/typescript_language_server.py
```diff
@@ -393,13 +400,14 @@ def progress_handler(params: dict) -> None:
         # typescript-language-server may send $/progress for "Initializing JS/TS
         # language features…" after initialized. If no progress is sent,
         # _indexing_complete stays SET and wait() returns immediately.
+        indexing_timeout = self._custom_settings.get("indexing_timeout", self.INDEXING_PROGRESS_TIMEOUT)
```
opcode81, line 407: New LS-specific settings should be documented in 050_configuration.md

src/solidlsp/language_servers/typescript_language_server.py
```diff
@@ -231,6 +233,11 @@ def _get_initialize_params(self, repository_absolute_path: str) -> InitializePar
         root_uri = pathlib.Path(repository_absolute_path).as_uri()
         initialize_params = {
             "locale": "en",
+            "initializationOptions": {
+                "preferences": {
+                    "disableAutomaticTypingAcquisition": True,
```
opcode81, line 242: Please document why you are setting this option for completeness.

opcode81 (changes requested):
Thank you for this PR!

Minor change request only.

bymebyu:
Thanks for the review! Addressed both points in `04ec37d0`:

- **`indexing_timeout` docs** — added a row to the TypeScript settings table in `docs/02-usage/050_configuration.md` (default `30.0`s; notes that Serena logs a warning and proceeds if indexing doesn't finish in time).
- **`disableAutomaticTypingAcquisition` rationale** — added a comment explaining why it's set: with ATA enabled, `tsserver` fetches `@types/*` from npm in the background during indexing, which makes startup slow, network-dependent, and nondeterministic (and can hang on offline/locked-down machines); Serena relies on the types already installed in the project instead.

Ready for another look when you have a chance.

PR #1642, opened 2026-07-03
Perl: expose file_filter/ignore_dirs via ls_specific_settings (#1449)

### LINKED ISSUE #1449: Perl: file extensions hardcoded to .pm/.pl — exclude legitimate Perl source like .cgi
The Perl language server integration hardcodes the file extension filter to `.pm` and `.pl`:

```python
# solidlsp/language_servers/perl_language_server.py:165
"fileFilter": [".pm", ".pl"],
```

Same value is also hardcoded at line 201 inside `didChangeConfiguration`. This is not exposed via project config or `ls_specific_settings`, so users cannot extend it without forking and patching the Python source.

## Why this matters

`.cgi` has been a standard extension for Perl web applications for 30+ years (mod_perl, Apache CGI, every classic Perl web framework). For mature Perl codebases the `.cgi` files often outnumber the `.pm` files significantly.

Concrete real-world example: a 1,450-file Perl game backend I tested against has:
- 340 `.pm`
- 285 `.pl`
- 825 `.cgi`

Indexing with Serena+Perl::LanguageServer+Perl 5.38 (after solving the Coro/perlbrew install path) succeeds technically, but only 343 files get indexed — the entire `.cgi` half of the codebase is invisible. Find-symbol/find-references on any sub defined in `.cgi` returns nothing, even though those subs are core to the application.

Other extensions affected by the same hardcoding: `.cgi`, `.t` (tests), `.psgi` (PSGI handlers), `.ph` (Perl headers), `.pod` (POD-only files), and any project-specific extensions.

## Suggested fix

Surface the file filter through `ls_specific_settings` so projects can extend it via `project.yml`:

```yaml
ls_specific_settings:
  perl:
    file_filter: [".pm", ".pl", ".cgi", ".t", ".psgi"]
    ignore_dirs: [".git", ".svn", "blib", "local", ".carton", "vendor", "_build", "cover_db"]
```

The default could remain `[".pm", ".pl"]` for backward compatibility; projects with non-standard layouts opt in. Perl::LanguageServer itself accepts an arbitrary `perl.fileFilter` array via `didChangeConfiguration`, so the LSP side already supports this — it's only the Serena wrapper that pins the value.

## Patch shape

In `solidlsp/language_servers/perl_language_server.py`, accept `file_filter` and `ignore_dirs` from the LS config, defaulting to the current values, and use those in both the `workspace/configuration` response (line 165) and the `didChangeConfiguration` notification (line 201).

weiconghe:
Fixes #1449

## What & why

`Perl::LanguageServer` only indexes files whose extension is in its `perl.fileFilter` and skips directories in its `perl.ignoreDirs`. Both lists were hardcoded in `PerlLanguageServer`, so projects whose Perl source uses non-standard extensions — e.g. `.cgi` / `.psgi` web handlers, which can outnumber `.pm`/`.pl` in mature Perl backends (the reporter's example: 825 `.cgi` vs 340 `.pm` + 285 `.pl`) — had no way to make those files visible to `find_symbol` / `find_referencing_symbols` without forking Serena.

`860c5841` added `.t` to the default `fileFilter` as a surface-level fix (one more extension). This PR addresses the root configurability gap that the maintainer acknowledged in #1449:

> MischaPanch: "The proposed patch works, would you like to open a PR with it?"

The reporter's proposed approach is exactly what's implemented here. Opening it myself since the issue has been idle for ~2 months after that invitation, with no PR or follow-up from the reporter. All diagnosis credit for the configurability direction goes to @aaronbamblett.

## Change

Surface the two lists through `ls_specific_settings["perl"]`, mirroring the established pattern in `groovy_language_server.py`:

```yaml
ls_specific_settings:
  perl:
    file_filter: [".pm", ".pl", ".t", ".cgi", ".psgi"]
    ignore_dirs: [".git", "blib", "local", "vendor", "cover_db"]
```

- `PerlLanguageServer.__init__` now resolves `self._file_filter` / `self._ignore_dirs` from `solidlsp_settings.get_ls_specific_settings(Language.PERL)`, defaulting to the existing values, so behaviour is unchanged for projects that don't set the override.
- Both LSP handshake sites (`workspace/configuration` handler and `didChangeConfiguration` notification) read the instance attributes instead of the old class constant / inline literals.
- The old `_FILE_FILTER` class constant is replaced by module-level `_DEFAULT_FILE_FILTER` / `_DEFAULT_IGNORE_DIRS`, with the "keep in sync with `Language.PERL.get_source_fn_matcher()`" note preserved on the default.

This follows serena's CONTRIBUTING.md scope ("small bug fixes" / "extend along existing lines" via PR without prior discussion), and the trusted-project gating added recently to `ls_specific_settings` applies automatically.

## Verification

Added `test/solidlsp/perl/test_perl_config.py` — 6 unit tests against the pure `_resolve_filter_settings` helper, deliberately not requiring a Perl runtime (existing Perl integration tests are skipped when `Perl::LanguageServer` isn't installed; these run everywhere):

- defaults returned when `ls_specific_settings` is absent
- defaults returned when only another language is configured (no cross-language leak)
- custom `file_filter` makes extra extensions (`.cgi`, `.psgi`) visible — the #1449 scenario
- custom `ignore_dirs` applied
- both overrides apply independently
- default module-level lists aren't mutated across instances (the resolver copies defaults)

Local checks on Windows / Python 3.13:

```
$ uv run --extra dev pytest test/solidlsp/perl/test_perl_config.py -v
6 passed
$ ruff check src/solidlsp/language_servers/perl_language_server.py test/solidlsp/perl/test_perl_config.py
All checks passed!
$ ruff format --check <same files>
2 files already formatted
$ ty check src/solidlsp        # _ty_core
All checks passed!
$ ty check test --exclude test/resources   # _ty_test
All checks passed!
```

`CHANGELOG.md` updated under "Language Servers".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,12 @@ Status of the `main` branch. Changes prior to the next official version change w
     (e.g. JDTLS reporting Lombok-generated classes under `target/classes`). Missing paths are now
     classified by the usual ignore rules, and symbol locations resolving to non-existent files are skipped.
   - Solidity: fix diagnostics intermittently coming back empty on slow or cold environments (macOS/Windows CI).
+  - Perl: expose `file_filter` and `ignore_dirs` via `ls_specific_settings["perl"]`, so projects with
+    non-standard extensions (e.g. `.cgi`, `.psgi`) can make those files visible to Perl::LanguageServer.
+    Configured extensions are also synced into the Perl source-file matcher (now a cached singleton),
+    keeping `find_symbol` / symbol indexing consistent with the LS; the matcher is reset on every
+    language server activation so one project's reconfiguration does not leak into the next.
+    `FilenameMatcher` gains `add_extensions` / `reset` methods. Defaults are unchanged. #1449
   - C/C++ (clangd): improve support and documentation for Unreal Engine 5 projects.
   - HLSL (`shader-language-server`): pass `--locked` to `cargo install` when building from source
     on macOS (and in the manual-install instructions), honoring the crate's packaged `Cargo.lock`.
--- a/src/solidlsp/ls.py
+++ b/src/solidlsp/ls.py
@@ -699,6 +699,11 @@ class SolidLanguageServer(ABC):
         """
         identifies the language server (not to be confused with the language id passed to the language server)
         """
+        # The source filename matcher is a @cache'd per-language singleton. A previous project may
+        # have extended it (e.g. Perl's file_filter adding .cgi); reset it here so every activation
+        # starts from the language's default extensions, then language-server subclasses re-apply
+        # their own settings during the rest of __init__.
+        self.language.get_source_fn_matcher().reset()
         self._published_diagnostics: dict[str, list[ls_types.Diagnostic]] = {}
         self._published_diagnostics_generation_by_uri: dict[str, int] = {}
         self._published_diagnostics_generation = 0
--- a/docs/02-usage/050_configuration.md
+++ b/docs/02-usage/050_configuration.md
@@ -871,6 +871,34 @@ Notes:
 - Use the FPC compiler driver (`fpc`/`fpc.exe`), not backend compilers like `ppc386.exe`.
 - These settings are passed as environment variables to the pasls process.
 
+#### Perl
+
+Serena uses [Perl::LanguageServer](https://metacpan.org/pod/Perl::LanguageServer) for Perl support. Install Perl and the server with `cpanm Perl::LanguageServer`; Linux and macOS only (the server does not run on Windows).
+
+Perl::LanguageServer only indexes files whose extension is in its `perl.fileFilter` and skips directories listed in `perl.ignoreDirs`. Both are exposed below so projects with non-standard extensions (e.g. `.cgi` / `.psgi` web handlers) can make those files visible (#1449).
+
+**Configuration:**
+
+Configure the language server via `ls_specific_settings.perl` in `serena_config.yml`:
+
+| Setting        | Default                                                                                     | Description                                                                                                    |
+| -------------- | ------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------- |
+| `file_filter`  | `[".pm", ".pl", ".t"]`                                                                      | File extensions (with leading dot) that Perl::LanguageServer should index, e.g. `[".pm", ".pl", ".t", ".cgi"]`. |
+| `ignore_dirs`  | `[".git", ".svn", "blib", "local", ".carton", "vendor", "_build", "cover_db"]`             | Directory names Perl::LanguageServer should skip when indexing.                                               |
+
+Example configuration:
+
+```yaml
+ls_specific_settings:
+  perl:
+    file_filter: [".pm", ".pl", ".t", ".cgi", ".psgi"]
+    ignore_dirs: [".git", "blib", "local", "vendor", "cover_db"]
+```
+
+Notes:
+- Extensions added via `file_filter` are also synced into Serena's Perl source-file matcher, so `find_symbol` and symbol indexing treat the same files as the language server. Defaults are unchanged when these keys are omitted.
+- The matcher is reset on every language server activation, so one project's `file_filter` does not leak into another.
+
 #### PHP (`Intelephense`)
 
 Serena uses Intelephense for the `php` language key.
--- a/src/solidlsp/language_servers/perl_language_server.py
+++ b/src/solidlsp/language_servers/perl_language_server.py
@@ -12,7 +12,7 @@ from typing import Any
 from overrides import override
 
 from solidlsp.ls import SolidLanguageServer
-from solidlsp.ls_config import LanguageServerConfig
+from solidlsp.ls_config import Language, LanguageServerConfig
 from solidlsp.ls_utils import PlatformId, PlatformUtils
 from solidlsp.lsp_protocol_handler.lsp_types import DidChangeConfigurationParams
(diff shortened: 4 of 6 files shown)


opcode81:
The problem is that Serena's symbol index is exclusively determined by the `FilenameMatcher` returned by `Language.PERL.get_source_fn_matcher()`. If the matcher is out of sync with what is passed to the language server, this leaves you with inconsistent behaviour:
* The language server will return symbols for a file with a different extension (e.g. a `.cgi` file) when *explicitly* asked to do so.
* But `find_symbol` will not find symbols that don't use one of the extensions in the `FilenameMatcher`.

Therefore, the out-of-sync case is a problem, and changing the extensions in only one place won't give you the desired behaviour. We would need to modify the `FilenameMatcher` as well, and we could easily do so if we turn the instances into per-language singletons.

weiconghe:
Good catch — you're right that `find_symbol` relies on `Language.PERL.get_source_fn_matcher()`, so extending only the LS-side `fileFilter` would leave the matcher (and thus the symbol index) out of sync. Fixed in the updated push.

What I did:
- Added `FilenameMatcher.add_extensions(*exts)` (idempotent, respects `case_sensitive`) — a small mutator intended for the now-`@cache`d per-language singletons.
- In `PerlLanguageServer`, after resolving `file_filter` from `ls_specific_settings["perl"]`, I call `Language.PERL.get_source_fn_matcher().add_extensions(*file_filter)`, so the configured extensions reach every consumer of the matcher: `project.is_ignored_path` (symbol index traversal), `ls.is_ignored_path`, and the language-composition detection.

The default `file_filter` (`[".pm", ".pl", ".t"]`) matches the existing Perl matcher, so behaviour is unchanged for projects that don't set the override; only added extensions (e.g. `.cgi`) propagate. The sync is a separate `_sync_source_fn_matcher` step so `_resolve_filter_settings` stays a pure, unit-testable function.

Tests added (`test/solidlsp/perl/test_perl_config.py`): the resolver cases, `FilenameMatcher.add_extensions` (new extension, idempotency across repeated calls, case-insensitive normalisation), and the matcher sync (custom `file_filter` extends the Perl matcher; default `file_filter` leaves it unchanged — both restore the singleton afterwards). 11 tests pass; `ruff`/`ty` clean.

opcode81:
There are further subtleties to consider: If a new project is activated following the activation of a project that reconfigured the file extensions, then the original extensions must apply; the reconfiguration must be undone.
Therefore, 
* the matcher needs a `reset` method (and it needs to store the original configuration for this work) 
* the `reset` method must be called upon language server initialization. The natural place for this would be in `SolidLanguageServer.__init__` after `self.language = self.get_language_enum_instance()`.

weiconghe:
Good catch on the cross-project leak — that's a real defect in my add-only approach. Once project A had added `.cgi`, the singleton would have kept it for every subsequently activated project. Thanks for spotting it.

Implemented exactly as you suggested:
- `FilenameMatcher` now snapshots its initial extensions at construction and exposes `reset()`, which restores that snapshot.
- `SolidLanguageServer.__init__` calls `self.language.get_source_fn_matcher().reset()` right after `self.language` is determined, so each activation starts from the language's defaults before the subclass re-applies its own `file_filter` (Perl's `_sync_source_fn_matcher` runs after `super().__init__()`, so the ordering is reset → re-apply).
- Added `test_reset_prevents_cross_project_leak` in `test_perl_config.py`, which reproduces the scenario directly: project A syncs `.cgi`, project B activates (reset + default config), and asserts `.cgi` is gone while `.pm` survives.

15 tests pass; `ruff` / `ty` clean. Pushed as `de4b78ed`.

src/solidlsp/ls.py
```diff
@@ -696,6 +696,11 @@ def __init__(
...
         self.open_file_buffers: dict[str, LSPFileBuffer] = {}
         self.language = self.get_language_enum_instance()
+        # The source filename matcher is a @cache'd per-language singleton. A previous project may
+        # have extended it (e.g. Perl's file_filter adding .cgi); reset it here so every activation
+        # starts from the language's default extensions, then language-server subclasses re-apply
+        # their own settings during the rest of __init__.
+        self.language.get_source_fn_matcher().reset()
         """
         identifies the language server (not to be confused with the language id passed to the language server)
         """
```
opcode81: The docstring is now in the wrong place.

CHANGELOG.md
```diff
@@ -48,6 +48,12 @@ Status of the `main` branch. Changes prior to the next official version change w
   - Make tool call errors surface explicitly as errors at the MCP protocol level
 
 * Language Servers:
+  - Perl: expose `file_filter` and `ignore_dirs` via `ls_specific_settings["perl"]`, so projects with
```
opcode81, line 59: The new options should also be documented in the "Perl" section of 050_configuration.md

opcode81: Add new section in lexicographical position if not present.

weiconghe:
Thanks — moved the reset below the `self.language` attribute docstring so it stays attached to the attribute it documents. Pushed as `30bb6b87`.

weiconghe:
There was no Perl section yet, so I added `#### Perl` in lexicographical position (between Pascal and PHP). It documents both `file_filter` and `ignore_dirs` under `ls_specific_settings.perl` — defaults, an example `serena_config.yml` snippet, and notes that configured extensions are also synced into Serena's source-file matcher (and reset on each activation so they don't leak across projects). Pushed as `967472e4`.

opcode81:
Thanks for your contribtion!

PR #1648, opened 2026-07-03
Svelte + TypeScript: fail loudly on companion TS timeouts; deterministic test fixture

xormania:
Make the Svelte companion TypeScript server raise on readiness and indexing timeouts instead of silently proceeding with a partial index. This turns flaky/wrong cross-file results into clear failures.

## Changes

- `TypeScriptLanguageServer`: timeout handling factored into overridable hooks (`_handle_server_ready_timeout`, `_handle_project_indexing_timeout`); the base server keeps its historical permissive behavior, so plain TypeScript setups are unchanged.
- `SvelteTypeScriptServer` (the companion owned by `SvelteLanguageServer`): overrides the hooks to raise `TimeoutError` with a diagnostic indexing-state summary (`describe_indexing_state()`) instead of serving requests from a cold or partially indexed program.
- Configurable `server_ready_timeout` / `indexing_timeout` via `ls_specific_settings`.
- `SvelteCompanionPreparationError` (raised when `.svelte` files cannot be opened on the companion) reports a count-first file listing capped at 10 entries and chains the first underlying error, so the cause is visible without DEBUG logging.
- Svelte test fixture: `npm ci` with a committed `package-lock.json`, removing dependency-resolution non-determinism from CI runs.
- Test fixture bootstrap: run `svelte-kit sync` explicitly and fail loudly if `.svelte-kit/tsconfig.json` cannot be generated. The fixture's `prepare` script masks sync failures (`svelte-kit sync || echo ''`), and a known npm bug (npm/cli#4828) can silently skip platform-specific optional dependencies — the result is unresolvable `$lib` path aliases and cross-file tests failing with confusing partial results instead of a clear, actionable error.
- codespell config: skip third-party fixture repositories (anchored pattern `./test/resources/repos`; codespell matches skip globs against the `./`-prefixed paths it walks, so an anchor-less pattern never matches).
- Regression tests (`test/solidlsp/test_typescript_timeout_policy.py`): pin the timeout-policy split directly — base server permissive, Svelte companion strict — plus the settings plumbing (including svelte-over-typescript precedence for `indexing_timeout`) and the internal start-or-completion indexing-wait semantics. No language-server processes are spawned and the policy tests use no mocks (a small hand-written fake drives the companion guard); the tests are fully deterministic.
- Docs: `server_ready_timeout` / `indexing_timeout` rows in the TypeScript and Svelte settings tables in `docs/02-usage/050_configuration.md`, including the companion's settings inheritance and strict behavior.

## Checklist

### DIFF
--- a/src/solidlsp/language_servers/typescript_language_server.py
+++ b/src/solidlsp/language_servers/typescript_language_server.py
@@ -73,6 +73,7 @@ class TypeScriptLanguageServer(SolidLanguageServer):
         - typescript_version: Version of TypeScript to install (default: "5.9.3")
         - typescript_language_server_version: Version of typescript-language-server to install (default: "5.1.3")
         - indexing_timeout: float, timeout in seconds for project indexing (default: 30.0)
+        - server_ready_timeout: float, timeout in seconds for the server-ready signal (default: 10.0)
     """
 
     @classmethod
@@ -82,6 +83,7 @@ class TypeScriptLanguageServer(SolidLanguageServer):
     # Safety timeout for $/progress-based indexing wait. Normally the event fires
     # well within this window; the timeout is only hit if the server never sends progress.
     INDEXING_PROGRESS_TIMEOUT = 30.0
+    SERVER_READY_TIMEOUT = 10.0
 
     def __init__(self, config: LanguageServerConfig, repository_root_path: str, solidlsp_settings: SolidLSPSettings):
         """
@@ -114,6 +116,35 @@ class TypeScriptLanguageServer(SolidLanguageServer):
         """
         return self._indexing_complete.wait(timeout=timeout)
 
+    def _wait_for_indexing_start_or_completion(self, timeout: float, start_grace: float | None = None) -> bool:
+        """Wait until TypeScript indexing has started and drained, or provably never started.
+
+        :param timeout: Maximum seconds to wait once active indexing progress is observed.
+        :param start_grace: Maximum seconds to wait for progress to begin after opening files.
+        :return: True if indexing completed or no progress began within the grace period, False on timeout.
+        """
+        grace = self._INDEXING_START_GRACE_S if start_grace is None else start_grace
+
+        # wait for progress to begin
+        progress_deadline = time.monotonic() + grace
+        while time.monotonic() < progress_deadline:
+            with self._progress_lock:
+                if self._active_progress_tokens:
+                    break
+                if self._indexing_complete.is_set():
+                    return True
+            time.sleep(0.05)
+
+        # treat absent progress as ready
+        with self._progress_lock:
+            has_active_progress = bool(self._active_progress_tokens)
+            if not has_active_progress:
+                self._indexing_complete.set()
+                return True
+
+        # wait for active progress to drain
+        return self.wait_for_indexing(timeout=timeout)
+
     def expect_indexing(self) -> None:
         """Signal that new files are about to be opened and async indexing should be awaited.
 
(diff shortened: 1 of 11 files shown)


xormania:
I looked into how you do things. I'll update PRs. BTW - I like what you're doing, you've answered some useful context issues. I get it now.

xormania:
Ready

src/solidlsp/language_servers/typescript_language_server.py
```diff
@@ -114,6 +116,35 @@ def wait_for_indexing(self, timeout: float) -> bool:
         """
         return self._indexing_complete.wait(timeout=timeout)
 
+    def wait_for_indexing_after_open(self, timeout: float, start_grace: float | None = None) -> bool:
```
MischaPanch: this has nothing to do with opening files and it is in fact called at most once, in preparation for the reference search. The logic was previously inlined there, and I think it should still be inlined. Any reason you pulled this out as method? For sure it shouldn't be a public method

test/solidlsp/svelte/test_svelte_references.py
```diff
@@ -8,19 +8,24 @@
 pytestmark = pytest.mark.svelte
 
 
+def _sorted_reference_paths(refs: list[dict]) -> list[str]:
```
MischaPanch: I feel like there are several unrelated changes in this PR that clutter the diff and make review harder than necessary. Is this change really needed, for example?

test/solidlsp/test_typescript_timeout_policy.py
```diff
@@ -0,0 +1,211 @@
+"""Regression tests for the TypeScript/Svelte timeout policy split.
```
MischaPanch, line 1: I propose to delete this, there's too much mocking for this to be useful or maintainable

MischaPanch (changes requested):
Thanks for the PR, rather minor comments

src/solidlsp/language_servers/typescript_language_server.py
xormania: Good points — made it protected and renamed it off "open": it's now `_wait_for_indexing_start_or_completion`, since you're right that the mechanism is "wait for indexing to start, then drain (or conclude it never started)," not opening. Single underscore so the Svelte companion can still call it on `self._ts_server`.

On inlining: I kept it as one small helper because there are two call sites in two classes — the base path in `_wait_for_cross_file_references_if_needed` (`typescript_language_server.py`) and the Svelte companion path in `_ensure_svelte_files_indexed_on_ts_server` (`svelte_language_server.py`). Each fires at most once per server, as you say. But before this PR the Svelte site was a bare `wait_for_indexing(timeout)` with no start-grace, so it would sit out the whole timeout and then report failure whenever tsserver decided there was nothing to index — one of the flake modes this PR is closing. Extracting gives the Svelte path the same grace-then-drain the base path already had, and it reads `_active_progress_tokens`/`_indexing_complete` under `_progress_lock` (the lock the `$/progress` handler holds when it mutates that set) instead of lock-free as the old inline loop did. Inlining would mean either duplicating that block in both classes or regressing the Svelte path. Happy to factor it differently if you'd prefer.

test/solidlsp/svelte/test_svelte_references.py
xormania: Agreed, the helper is overkill. Dropped `_sorted_reference_paths` and the redundant `sorted_ref_paths` local (it recomputed the set `ref_paths` already holds) and folded the diagnostic straight into the asserts as `assert "…" in ref_paths, sorted(ref_paths)`. I kept the one-line message only because these are the cross-file reference tests this PR de-flakes and a sorted path list makes a CI failure legible at a glance — but it's a nicety, not the fix, so happy to drop it entirely if you'd prefer the leaner diff.

test/solidlsp/test_typescript_timeout_policy.py
xormania: standby

xormania: Fair — I've reworked the file to remove the objectionable parts entirely. As of 9580a6ae:

  - No mocking or patching of any kind remains — no unittest.mock, no monkeypatch. The four guard tests now run the real _find_all_svelte_files against a pytest tmp_path tree (its only dependency is repo_path), including its
  node_modules/dot-dir filtering.
  - The companion is a small hand-written fake that records calls — it stands in for a subprocess-owning server, which a unit test can't run for real. I kept one ordering assertion deliberately: expect_indexing() must be armed before
  the first open_file, because opening files before progress tracking is armed is exactly the race this guard exists to pin. That's a behavioral invariant, not mock bookkeeping.
  - The internal-state assert (_svelte_files_indexed) is gone, replaced by its observable consequence: a second _ensure_svelte_files_indexed_on_ts_server() call is a complete no-op.

  On keeping the file: the first three classes use no test double at all — they drive the real timeout methods on object.__new__ shells with real threading primitives (the same constructor-bypass used in test_rename_didopen.py,
  test_scala_stale_lock_handling.py, test_jdtls_path_resolution.py, test_cpp_unreal_unit.py), and they're the only deterministic coverage of the permissive-base vs strict-companion split. Reproducing the raise paths through the live
  fixture would mean forcing real timeouts with tiny values and hoping the server is slow enough — the kind of timing dependence this PR is removing. The file is deliberately unmarked, so it runs in the catch-all batch on all three
  OSes with no toolchain.

  If you'd still rather delete the file outright, I'll drop it — your call.

MischaPanch:
Thanks for addressing the comments, merging. I am not too happy with the bare-server approach in the tests but if it gets in the way, the tests can be removed later

Issue #944, opened 2026-01-16
Serena MCP consumes all system memory (~30GB) causing Claude Code to freeze


kmgallahan:
# Serena MCP consumes all system memory (~30GB) causing Claude Code to freeze

I have:

- [x] Read the readme and verified that the issue cannot be solved by adjusting configuration (no memory limit settings exist)
- [x] Understood that Serena's dashboard can be disabled through the config
- [x] Understood that by default a client session will start a separate instance of a Serena server.
- [x] Understood that for multi-agent setups, the SSE mode should be used. (Switched to SSE mode - issue persists)
- [x] Verified that non-project files are ignored using either gitignore or the corresponding setting in `.serena/project.yml`
- [x] Have looked for similar issues and discussions, including closed ones
- [x] Made sure it's an actual issue and not a question - those should be opened as discussion instead.

If you have encountered a real and new issue:

- [ ] I performed `<uv invocation> serena project health-check`
- [ ] I indexed the project as described in the readme
- [x] Added sufficient explanation of my setup: the MCP client, the OS, the programming language, any config adjustments or relevant project specifics
- [x] Explained how the issue arose, added instructions on how to reproduce it (if possible)
- [ ] If the issue happens on an open source project, I have added the link
- [x] Wrote a meaningful title and description

---

## Environment

- **OS:** Windows 11
- **MCP Client:** Claude Code CLI
- **Serena Version:** Latest (1/16/2026)
- **Project Language:** C#/C++/Rust

## Description

Serena MCP has caused complete system memory exhaustion **three times** now. Each time, Serena consumed approximately 30GB of RAM, using all available free system memory. This caused Claude Code to become completely unresponsive and unable to proceed.

**Important update:** The third occurrence happened with **SSE mode enabled**, ruling out the multiple-instance theory. The python.exe process running Serena ballooned to ~30GB even when using the recommended SSE transport mode (`http://localhost:8000/sse`).

## Steps to Reproduce

1. Start Claude Code with Serena MCP configured
2. Use Claude Code normally with Serena tools
3. Over time, Serena's memory usage grows unbounded
4. Eventually all system memory (~30GB) is consumed
5. Claude Code freezes and cannot continue

## Expected Behavior

Serena should maintain reasonable memory usage and not consume all available system memory.

## Actual Behavior

Serena consumes all available system memory (~30GB), causing:
- System slowdown
- Claude Code becoming unresponsive
- Inability to continue working without killing the Serena process

## Additional Context

[...]

PR #1649, opened 2026-07-03
fix(solidlsp): cap LSP message and header size to prevent unbounded memory growth

RasmusKD:
## Problem

`solidlsp` frames LSP messages by reading a `Content-Length` header and then allocating **exactly that many bytes, with no upper bound**, and reads the header line with an unbounded `readline()`. A language server that dies mid-frame or emits a corrupt / oversized / negative `Content-Length` can drive an unbounded allocation until the host runs out of memory. In practice the reader catches the resulting error and the supervisor restarts the language server, which re-triggers it, so the `python` process climbs to the RAM ceiling and parks there. This matches the behavior in #944 (process reaching ~30 GB even in single-instance mode).

## Root cause

In `ls_process.py`, both `StdioLanguageServer._read_ls_process_stdout` (via `_read_bytes_from_process`) and `TCPLanguageServer._read_loop` do:

- `line = stream.readline()` with no maximum length (an unbounded header line), and
- read exactly `num_bytes = content_length(line)` bytes with no sanity cap on `num_bytes`.

`content_length()` parses the value with a bare `int()`, so any size the server declares is trusted.

## Fix

- Add `MAX_LSP_MESSAGE_SIZE` (default **512 MiB**, overridable via `SOLIDLSP_MAX_LSP_MESSAGE_SIZE`; a misconfigured value falls back to the default with a warning) and `MAX_LSP_HEADER_LINE` (64 KiB).
- Validate the declared message size immediately after parsing `Content-Length`. A length that is **negative or exceeds the limit** raises `LanguageServerTerminatedException`, so the reader aborts and the language server is torn down / restarted instead of the process being OOM-killed.
- Pass `MAX_LSP_HEADER_LINE` to every header `readline()` so a malformed newline-less header can't balloon memory either.
- Applied symmetrically to the stdio and TCP read paths.

Failing fast on a pathological frame is strictly better than exhausting host RAM: a clean language-server restart vs. a machine-wide OOM. With the default limit no legitimate message is rejected; the env override exists for the rare server that genuinely needs larger frames.

## Tests

`test/solidlsp/test_ls_process_message_size.py`:

- unit coverage of the size guard (accepts values up to and including the limit; rejects oversized and negative lengths), and
- an integration test that drives the real `_read_ls_process_stdout` loop with an oversized frame and asserts it aborts (via `_cancel_pending_requests`) **before** attempting the body read.

`ruff check` / `ruff format` clean.

## Scope

Intentionally focused on the unbounded-allocation issue. A separate, lesser concern in the same loop (a server that floods endless header lines could spin the header-drain loop) is a *hang* rather than an OOM and is left out to keep this change reviewable; happy to follow up separately if useful.

Addresses #944.

### DIFF
--- a/src/solidlsp/ls_process.py
+++ b/src/solidlsp/ls_process.py
@@ -52,6 +52,46 @@ class LanguageServerTerminatedException(Exception):
         return f"LanguageServerTerminatedException: {self.message}" + (f"; Cause: {self.cause}" if self.cause else "")
 
 
+def _positive_int_from_env(name: str, default: int) -> int:
+    """Read a positive int from the environment, falling back to ``default`` on absence or
+    an invalid/non-positive value (so a misconfigured override can never break startup).
+    """
+    raw = os.environ.get(name)
+    if not raw:
+        return default
+    try:
+        value = int(raw)
+    except ValueError:
+        log.warning("Ignoring invalid %s=%r (not an integer); using default %d", name, raw, default)
+        return default
+    if value <= 0:
+        log.warning("Ignoring non-positive %s=%d; using default %d", name, value, default)
+        return default
+    return value
+
+
+# Upper bounds on inbound LSP wire data. A misbehaving language server (one that dies
+# mid-frame, or emits a corrupt / oversized / negative "Content-Length") can otherwise drive
+# an unbounded read allocation until the host runs out of memory. A message that exceeds the
+# limit (or declares a negative length) is treated as a fatal protocol error: the reader
+# stops and the language server is torn down (and later restarted) instead of the whole
+# process being OOM-killed. The message limit defaults to 512 MiB and can be raised via
+# SOLIDLSP_MAX_LSP_MESSAGE_SIZE for the rare server that legitimately needs larger frames.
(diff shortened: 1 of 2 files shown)


opcode81:
> solidlsp frames LSP messages by reading a Content-Length header and then allocating exactly that many bytes, with no upper bound, and reads the header line with an unbounded readline(). A language server that dies mid-frame or emits a corrupt / oversized / negative Content-Length can drive an unbounded allocation until the host runs out of memory.

Has this actually happened to you?
#944 had a different cause, which was fixed a long time ago.

opcode81:
No response from OP, therefore closing.

RasmusKD:
No, I haven't hit a real oversized-frame OOM, our own high-RAM case turned out to be tsserver indexing gitignored sibling repos, a config issue, not this path. This PR is defensive hardening: content_length() trusts any declared size via a bare int() and allocates it unbounded, so a server that dies mid-frame or emits a corrupt/negative length can drive allocation to the host ceiling. Bounded framing (with an env override) fails fast to a clean restart instead. Happy to have it judged purely on that merit; if you consider it out of scope without a concrete repro, no hard feelings.

PR #1651, opened 2026-07-05
Add optional persisted log trimming

nerclid:
## Summary

Adds optional cleanup for persisted Serena logs under the Serena home directory.

New global config options:

- `persisted_log_retention_days`: delete dated persisted log folders older than the configured number of days
- `persisted_log_max_files`: keep only the newest persisted log files

Both default to `null`, preserving current behavior.

The trimming logic is centralized in `SerenaPaths`, where persisted log paths are already owned. Startup config loading invokes trimming after configuration is read. Count-based trimming excludes the currently active log file so startup cannot delete or unlink the log file that Serena is already writing to.

Also makes the macOS `shader_language_server` cargo install use `--locked`. The first CI run failed in `other-langs (macos-latest)` because `cargo install shader_language_server --version 1.3.1` was resolving dependencies without the crate's published lockfile after a newer `shader-sense` release. Using `--locked` keeps that install reproducible.

## Testing

- `uv run python -m pytest test\serena\config\test_serena_config.py -q`
- `uv run poe lint`
- `uv run poe type-check`
- Manual docs build equivalent to `poe doc-build` on Windows
- `uv run --with codespell codespell CHANGELOG.md docs\02-usage\065_logs.md src\serena\config\serena_config.py src\serena\resources\serena_config.template.yml test\serena\config\test_serena_config.py`
- `git diff --check`
- `uv run ruff format --check src\solidlsp\language_servers\hlsl_language_server.py src\serena\config\serena_config.py test\serena\config\test_serena_config.py`
- `uv run ruff check src\solidlsp\language_servers\hlsl_language_server.py src\serena\config\serena_config.py test\serena\config\test_serena_config.py`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ Status of the `main` branch. Changes prior to the next official version change w
   - During project creation, language composition percentages are now computed relative to the total number 
     of recognised source files instead of all files, i.e. unrecognised files are ignored in the percentage 
     computation.
+  - Add optional persisted log trimming by retention days and maximum number of log files.
 
 * CLI:
   - Fix `--project-from-cwd` hijacking git worktrees nested under a Serena project. `find_project_root`
--- a/docs/02-usage/065_logs.md
+++ b/docs/02-usage/065_logs.md
@@ -12,3 +12,7 @@ Additionally, logs are persisted in the Serena home directory, which, by default
 
 You can adjust the log level via the [global configuration](global-config).
 You additionally have the option of enabling full tracing of language server communication (mostly for development purposes).
+
+Persisted logs can be trimmed automatically on startup by setting `persisted_log_retention_days` in the global configuration.
+Persisted logs can also be limited by count by setting `persisted_log_max_files`.
+Both defaults are `null`, which disables persisted log trimming.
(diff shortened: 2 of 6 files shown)


opcode81:
Thanks for your interest in contributing to Serena, but this does not adhere to our contribution guidelines:
* No prior discussion of new features. (We do not need two settings to limit logs.)
* Unrelated changes in a single PR.

nerclid:
Thanks for looking at it. Understand, and sorry for not realizing there was a prior discussion rule. 
I appreciate you making the project public, it has been very helpful for me.

opcode81:
@nerclid I now realise that you added the unrelated change only because of a CI issue (but it has been fixed via another PR in the meantime).

Our prior discussion rule serves to ensure an optimal outcome from the get-go. Serena is not a vibe-coded project, and we care about doing things right, first and foremost.
If you are interested in contributing a log trimming feature, here's my input:

* One option `log_retention_days` is sufficient (default `null`).
* This is no longer a concern that would be well-placed in `SerenaPaths`; add a new component `LogFileManager` in `serena.util.logging` instead, which handles both the name provision and the trimming (singleton).
* An invokation in `SerenaConfig` is not appropriate. Instead, it should conditionally run (only if value is set) during agent startup, probably `SerenaAgent.__init__`.
* It should run in a separate thread to not delay agent startup. (This also sidesteps the problem in your solution where this could raise an exception and thereby make agent startup fail completely.) This should not raise exceptions and log errors instead, as log cleansing is not critical at all.
* We don't need extensive tests for this.

If we had discussed this beforehand, we would have saved time, so that's why the rule makes sense. I hope you understand.

PR #1652, opened 2026-07-05
HLSL: build shader_language_server with cargo --locked to fix macOS CI

xormania:
Without --locked, cargo install freshly resolves shader-sense ^1.3.1, which since 2026-07-04 picks shader-sense 1.4.0. That release removed the public field GlslCompilationParams.preamble (split into preamble_path/preamble_content), which shader_language_server 1.3.1's server_config.rs still constructs, so the from-source build fails with E0560 and cargo exits 101. macOS is the only platform that builds the server from source (Linux/Windows download prebuilt release binaries), so only the macOS CI jobs fail: all 15 test/solidlsp/hlsl fixtures error at setup (run 28726267725, job 85184119276).

--locked makes cargo honor the Cargo.lock packaged with the published crate, which pins shader-sense 1.3.1 - the dependency set the 1.3.1 release was built and tested from - and immunizes the build against future lockstep releases of the two crates. Verified locally: the unlocked build reproduces the exact E0560/exit-101 failure; the locked build compiles cleanly and yields a working binary reporting v1.3.1.

Also adds --locked to the manual-install help text, matching the existing convention in taplo_server.py.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,10 @@ Status of the `main` branch. Changes prior to the next official version change w
 
 * Language Servers:
   - C/C++ (clangd): improve support and documentation for Unreal Engine 5 projects.
+  - HLSL (`shader-language-server`): pass `--locked` to `cargo install` when building from source
+    on macOS (and in the manual-install instructions), honoring the crate's packaged `Cargo.lock`.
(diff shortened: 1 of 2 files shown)


xormania:
yee or nee?

opcode81:
Thanks for your contribution!

xormania:
Thank you. My first!

PR #1660, opened 2026-07-06
Java (eclipse_jdtls): don't hard-ignore packages named dist/lib/classes/out (#1645)

### LINKED ISSUE #1645: Java (eclipse_jdtls): is_ignored_dirname hard-ignores packages named dist/lib/classes/out, hiding legitimate source from symbol tools
## Checklist

- [x] I have read the relevant parts of the documentation and verified that the issue cannot be solved by adjusting configuration (`ignored_paths` only **adds** patterns; there is no config hook to remove an entry from a language server's hardcoded `is_ignored_dirname` list).
- [x] I understood that the Serena Dashboard can be disabled through the config.
- [x] I understood that, by default, a client session will start a separate instance of a Serena server.
- [x] I understood that non-project files are ignored using either `.gitignore` or the corresponding setting in `.serena/project.yml`.
- [x] I looked for similar issues/discussions, including closed ones — this is the **Java/eclipse_jdtls counterpart** of #1523 / #1524 (TypeScript `coverage`); see also #731/#743 (PHP `vendor`), #1187/#1203/#853 (dot-dirs), and #1449/#1642 (Perl configurability).
- [x] I made sure it's an actual issue, not a question.

## Problem

`solidlsp/language_servers/eclipse_jdtls.py` hard-ignores any path segment named `dist`, `lib`, `classes`, or `out`:

```python
def is_ignored_dirname(self, dirname: str) -> bool:
    # Ignore common Java build directories from different build tools:
    # ...
    return super().is_ignored_dirname(dirname) or dirname in [
        "target",   # Maven
        "build",    # Gradle
        "bin",      # Eclipse
        "out",      # IntelliJ IDEA
        "classes",  # General
        "dist",     # General
        "lib",      # General
    ]
```

The intent is to skip build **output**, but the match is on a bare directory name, so it also excludes legitimate **source** packages. `dist`, `lib`, `classes`, and `out` are all valid Java package identifiers and are commonly used as package names (e.g. `com.example.dist`, `...lib`, `...out`). Any Java source file whose package path contains such a segment becomes invisible to the symbol tools.

This is exactly the same failure mode as #1523 (TS `coverage`), and the symptom is identical:

- `find_symbol` returns `[]`.
- `get_symbols_overview` / passing the file explicitly to `find_symbol` fails with `Cannot extract symbols from file <path>. Active languages: ['java']`.
- `request_full_symbol_tree` logs `You passed a file explicitly, but it is ignored. This is probably an error.`
- ...even though the file is **not** gitignored (`git check-ignore` reports nothing) and `Project.GitignoreParser.should_ignore()` returns `False`.

## Reproduction

Minimal example — a Gradle project whose application package is `com.example.app.dist` (a business-domain name, unrelated to build output):

```
src/main/java/com/example/app/dist/service/OrderProcessor.java   -> [] / Cannot extract symbols
src/main/java/com/example/app/dist/Application.java              -> [] / Cannot extract symbols
src/main/java/com/example/app/core/Application.java              -> OK  (no `dist` segment)
src/main/java/com/example/common/util/Helper.java               -> OK
```

Every file under a package segment named `dist` (or `lib` / `classes` / `out`) disappears from `find_symbol` / `find_referencing_symbols` / `get_symbols_overview`, while sibling packages without those segments resolve fine. In a real multi-module repo this can silently hide a large fraction of the source tree. Because `build/` and `bin/` are already covered by `.gitignore`, the four "General" entries (`dist`, `lib`, `classes`, `out`) add little protection against real build output while actively breaking legitimate source.

serena-agent version: `1.5.3`.

## Suggested fix

Two options, matching precedent already in the repo:

1. **Remove the over-broad "General" entries** (`dist`, `lib`, `classes`, `out`) from the eclipse_jdtls list, the same way #1524 removes `coverage` from the TS servers. Genuine build output (`target`, `build`, `bin`) stays ignored and is also normally gitignored anyway.
2. **Make the list configurable** via `ls_specific_settings` (`ignore_dirs`), the root-cause approach taken for Perl in #1642 — so projects can override per-language without forking.

Happy to open a PR for whichever direction the maintainers prefer.

AmirF194:
## Root cause

`EclipseJDTLS.is_ignored_dirname` hard-ignored directories named `target`/`build`/`bin`/`out`/`classes`/`dist`/`lib`. Four of these (`dist`, `lib`, `classes`, `out`) are valid Java package identifiers, and Java packages map directly to directories, so a source file whose package path contained one of those segments (e.g. `com.example.app.lib.Foo`) was pruned during traversal (`ls.py`, `if self.is_ignored_dirname(part)`) and became invisible to `find_symbol`, `get_symbols_overview`, and `find_referencing_symbols`, despite `git check-ignore` reporting nothing.

## Fix

Remove the override entirely, per your call in #1645. The base `is_ignored_dirname` already skips `_ALWAYS_IGNORED_DIRS` (VCS/venv/cache/IDE internals), and real build output is already excluded via `.gitignore`, so the hardcoded list was redundant and harmful. Java now uses the base behavior.

## Verification

Added `TestIsIgnoredDirname` (offline, no Java/JDTLS process, matching the file's existing unit tests): it fails on `main` and passes with this change.

```
without the fix: 7 failed, 5 passed   (lib/dist/classes/out/target/build/bin wrongly ignored)
with the fix:    12 passed
```

Whole file: `pytest test/solidlsp/java/test_jdtls_path_resolution.py` gives 55 passed. `ruff check`, `ruff format --check`, and `ty check` are clean. CHANGELOG updated under Language Servers.

## Notes

- @canweed you found and diagnosed this, and opcode81 offered to take your PR. If you already have one in progress, I'm glad to close this in favour of it, just say so.
- Out of scope here, but the same `... or dirname in [<build dirs>]` pattern appears in about a dozen other language servers (gopls, typescript, svelte, sourcekit, and others list `dist`/`build`). I kept this PR to the Java case you flagged; happy to open a follow-up if you'd like the pattern reviewed more broadly.

Fixes #1645

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,10 @@ Status of the `main` branch. Changes prior to the next official version change w
   - `Java`: invalidate JDTLS workspace cache when Java import settings change #1576
   - `Java`: use `JAVA_HOME` for Gradle import when `use_system_java_home` is enabled and `gradle_java_home`
     is unset. #1657
+  - `Java`: stop hard-ignoring directories named `target`/`build`/`bin`/`out`/`classes`/`dist`/`lib` in
+    `EclipseJDTLS`. These are all valid Java package identifiers, so ignoring them by name hid legitimate
+    source from the symbol tools even when they were not gitignored. Removed the hardcoded override; real
+    build output is already excluded via `.gitignore`. #1645
   - Improve quoting of arguments in shell executions
   - Add **LaTeX** support (experimental) via [texlab](https://github.com/latex-lsp/texlab).
   - PHP: add support for PHPantom as alternative to the already supported PHP LS #1554.
--- a/src/solidlsp/language_servers/eclipse_jdtls.py
+++ b/src/solidlsp/language_servers/eclipse_jdtls.py
@@ -254,24 +254,6 @@ class EclipseJDTLS(SolidLanguageServer):
         ls_resources_dir = self.ls_resources_dir(self._solidlsp_settings)
         return self.DependencyProvider(self._custom_settings, ls_resources_dir, self._solidlsp_settings, self.repository_root_path)
 
-    @override
-    def is_ignored_dirname(self, dirname: str) -> bool:
-        # Ignore common Java build directories from different build tools:
-        # - Maven: target
-        # - Gradle: build, .gradle
-        # - Eclipse: bin, .settings
-        # - IntelliJ IDEA: out, .idea
-        # - General: classes, dist, lib
-        return super().is_ignored_dirname(dirname) or dirname in [
-            "target",  # Maven
-            "build",  # Gradle
-            "bin",  # Eclipse
-            "out",  # IntelliJ IDEA
-            "classes",  # General
-            "dist",  # General
-            "lib",  # General
-        ]
-
     class DependencyProvider(LanguageServerDependencyProvider):
         def __init__(
             self,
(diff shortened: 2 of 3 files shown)


opcode81:
Thanks for your PR!

PR #1665, opened 2026-07-07
fix(memory): reject absolute and escaping memory names in get_memory_file_path

AmirF194:
## What

`MemoryManager.get_memory_file_path` guards against `..` segments "for security reasons", but not against **absolute** memory names. Because it builds the path with `base_dir / "/".join(parts[:-1])`, an absolute name discards the base directory: `Path(project_memory_dir) / "/etc/cron.d"` resolves to `/etc/cron.d`, escaping the `.serena/memories` sandbox. Since `save_memory` / `load_memory` / `delete_memory` / `move_memory` all route through this method, a crafted (e.g. prompt-injected) memory name could read, write, or delete arbitrary `*.md` files outside the sandbox. Both the project and `global/` branches were affected.

Reproduced on current `main`: `get_memory_file_path("/etc/cron.d/backdoor")` returns `/etc/cron.d/backdoor.md` and even attempts to `mkdir` that directory (only stopped by OS permissions), while the existing control `get_memory_file_path("../../etc/x")` correctly raises.

## Fix

- Reject absolute names and empty path segments up front, alongside the existing `..` check (`os.path.isabs(name) or "" in parts`). This runs on the full name, so it covers both branches before any directory is created.
- Route both branches through a new `_resolve_memory_path` helper that adds a **lexical** containment backstop (`os.path.normpath` + `is_relative_to`, before any `mkdir`) enforcing the invariant that every returned path stays inside its base memories directory.

The containment check is deliberately lexical rather than `Path.resolve()`-based: serena intentionally supports directory symlinks inside `memories/` (e.g. a monorepo symlinking each submodule's memory dir), and `resolve()` would follow those symlinks out of the base and wrongly reject them. The existing `TestListMemoriesFollowsSymlinks` behavior is preserved.

## Verification

- Added `TestGetMemoryFilePathContainment` (offline): absolute project name, absolute system path, absolute global name, `..` still rejected, plus containment assertions that accepted names resolve inside their base dir.
- The three absolute-rejection tests fail on `main` and pass on this branch; full `test/serena/test_memories_manager.py` is 112 passed (including the symlink feature test).
- `uv run poe lint` (ruff format + check) and `uv run poe type-check` (ty) are clean.

## Checklist

- [x] This PR follows the guidelines in `CONTRIBUTING.md` regarding the scope of PRs (single logical change).

Implementation done with AI assistance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,6 +97,8 @@ Status of the `main` branch. Changes prior to the next official version change w
 
 * Memories:
   - Make memory iteration follow symbolic links 
+  - Fix: a memory name that was absolute (e.g. `/etc/cron.d/backdoor`) or contained empty path
+    segments could write outside the `memories` folder.
 
 * Dependencies:
   - Add dependency `oslex`
--- a/src/serena/memories/memory_manager.py
+++ b/src/serena/memories/memory_manager.py
@@ -147,11 +147,41 @@ class MemoryManager:
         # use a callable replacement to avoid backreference interpretation of characters in ref_new
         return re.subn(pattern, lambda _m: ref_new, content)
 
+    def _resolve_memory_path(self, base_dir: Path, parts: Sequence[str]) -> Path:
+        """
+        Builds the ``*.md`` path for ``parts`` under ``base_dir``, creating any parent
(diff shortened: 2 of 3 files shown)


opcode81:
Thanks for your contribution!

PR #1667, opened 2026-07-08
search_for_pattern: add a snippet stage to the overflow shortening chain

### LINKED ISSUE #1640: search_for_pattern overflow fallback: agents time out on bare line numbers — proposal: intermediate snippet stage
## Context

When `search_for_pattern` results exceed `max_answer_chars`, the shortening chain currently falls back from full match content directly to bare line numbers (`Match lines per file: {...}`). The agent then knows *where* matches are but not *what* matched, and has to re-read files to recover that information.

## Evidence

We ran an agent-level A/B benchmark (headless Claude agents, 8 coding tasks × 5 trials, deterministic checkers, built-in Read/Grep disabled so the serena tool surface is isolated). One task required finding a specific match among ~1,300 matches of a pattern in a large generated file (full result ≈ 195K chars > limit).

With the bare line-number fallback, the agent **timed out (300 s) in 2/5 trials** and answered one line off in the remaining 3. With an intermediate snippet stage (below), it succeeded **5/5 with exact line numbers**, at similar cost.

## Proposal

Add one stage to the shortening chain:

```
full matches → {"line": N, "text": "<matched text, truncated to 60 chars>"} → line numbers → per-file counts → summary
```

The snippet stage fits within the response limit whenever matched lines are long (which is exactly when full results overflow), and preserves enough context for the agent to act without re-reading files.

A prepared branch with the change exists at `bymebyu:feat/search-overflow-snippets` (19 insertions in `SearchForPatternTool`). Happy to open a PR if this direction is acceptable — or feel free to take the idea and implement it your way.

bymebyu:
Resolves #1640 (direction approved by @opcode81 in the issue).

## Summary

When a `search_for_pattern` result exceeds `max_answer_chars`, the shortening chain
previously fell back from full match content **directly to bare line numbers**, so the
agent learned *where* the matches were but not *what* matched — and had to re-read files
to recover that. This adds one intermediate stage to the chain:

```
full matches → {"line": N, "text": "<matched text, truncated to 60 chars>"} → bare line numbers → per-file counts → summary
```

The snippet stage fits within the limit precisely when the matched lines are long (which
is exactly the case that makes full results overflow), and it degrades gracefully: if even
the snippets exceed the limit, the chain continues on to bare line numbers, per-file counts
and finally a summary, as before.

## What the agent actually receives (deterministic)

Reproducible tool-level demonstration — a search with ~1,300 matches whose full result is
196,274 chars, against the default `max_answer_chars` of 150,000. The task is to find the one
match belonging to a specific class (`Message777`). This is the first fallback returned in
each case (the `The answer is too long (…)` banner is added by the framework in both):

**Before** — bare line numbers:
```
Match lines per file:
{"generated/pb_models.py": [23, 46, 69, …, 17848, 17871, 17894, 17917, 17940, … (~1300 numbers)]}
```
Line `17894` is indistinguishable from the other ~1,300 — to tell which match is `Message777`,
the agent has to open the file.

**After** — line number + matched text:
```
Match locations per file (line + matched text):
{"generated/pb_models.py": [ …,
  {"line": 17894, "text": "return cls(field_id=data.get('field_id', 777), field_name=da"}, … ]}
```
The distinguishing text (`777`) is inline, so the agent can pick the right match and issue a
targeted `read_file` — without re-reading the file just to discover what matched.

## Changes

- `src/serena/tools/file_tools.py` — snippet stage in `SearchForPatternTool` (adds a
  `{line, text}` intermediate and keeps a dedicated bare-line-numbers stage after it).
- `CHANGELOG.md` — entry under *Tools*.

The motivating agent-level A/B measurement is described in #1640.

### DIFF
--- a/src/serena/tools/file_tools.py
+++ b/src/serena/tools/file_tools.py
@@ -611,15 +611,27 @@ class SearchForPatternTool(Tool):
             file_to_matches[match.source_file_path].append(match.to_display_string())
 
         # capture lightweight match data for shortening before serialization
-        match_lines_by_file: dict[str, list[int]] = defaultdict(list)
+        match_lines_by_file: dict[str, list[dict[str, int | str]]] = defaultdict(list)
         for match in matches:
             assert match.source_file_path is not None
-            match_lines_by_file[match.source_file_path].append(match.matched_lines[0].line_number)
+            first = match.matched_lines[0]
+            match_lines_by_file[match.source_file_path].append({"line": first.line_number, "text": first.line_content.strip()})
 
         # shortened result closures, from least to most aggressive shortening
+        _TEXT_TRUNCATE = 60
+
         def make_lines_only() -> str:
-            """Match locations without surrounding context"""
-            return f"Match lines per file:\n{self._to_json(match_lines_by_file)}"
+            """Match locations with line number and truncated matched text."""
+            compact = {
+                path: [{"line": m["line"], "text": str(m["text"])[:_TEXT_TRUNCATE]} for m in lines]
+                for path, lines in match_lines_by_file.items()
+            }
+            return f"Result too long. Use read_file with line ranges below:\n{self._to_json(compact)}"
+
+        def make_line_numbers_only() -> str:
+            """Match locations as bare line numbers (no text)."""
+            numbers = {path: [m["line"] for m in lines] for path, lines in match_lines_by_file.items()}
+            return f"Match lines per file:\n{self._to_json(numbers)}"
 
         def make_per_file_counts() -> str:
             counts = {path: len(lines) for path, lines in match_lines_by_file.items()}
@@ -630,5 +642,7 @@ class SearchForPatternTool(Tool):
 
         result = self._to_json(file_to_matches)
         return self._limit_length(
-            result, max_answer_chars, shortened_result_factories=[make_lines_only, make_per_file_counts, make_summary]
+            result,
+            max_answer_chars,
+            shortened_result_factories=[make_lines_only, make_line_numbers_only, make_per_file_counts, make_summary],
         )
--- a/src/serena/tools/file_tools.py
+++ b/src/serena/tools/file_tools.py
@@ -626,7 +626,7 @@ class SearchForPatternTool(Tool):
                 path: [{"line": m["line"], "text": str(m["text"])[:_TEXT_TRUNCATE]} for m in lines]
                 for path, lines in match_lines_by_file.items()
             }
-            return f"Result too long. Use read_file with line ranges below:\n{self._to_json(compact)}"
+            return f"Match locations per file (line + matched text):\n{self._to_json(compact)}"
 
         def make_line_numbers_only() -> str:
             """Match locations as bare line numbers (no text)."""
--- a/src/serena/tools/file_tools.py
+++ b/src/serena/tools/file_tools.py
@@ -620,13 +620,34 @@ class SearchForPatternTool(Tool):
         # shortened result closures, from least to most aggressive shortening
         _TEXT_TRUNCATE = 60
 
-        def make_lines_only() -> str:
-            """Match locations with line number and truncated matched text."""
+        def render_first_lines(truncate: bool) -> str:
+            """Render each match's first line, either in full or truncated to a fixed length."""
+
+            def entry_text(text: str) -> str:
+                if truncate and len(text) > _TEXT_TRUNCATE:
+                    return text[:_TEXT_TRUNCATE] + "..."
+                return text
+
             compact = {
-                path: [{"line": m["line"], "text": str(m["text"])[:_TEXT_TRUNCATE]} for m in lines]
+                path: [{"line": m["line"], "text": entry_text(str(m["text"]))} for m in lines]
                 for path, lines in match_lines_by_file.items()
             }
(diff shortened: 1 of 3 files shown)


src/serena/tools/file_tools.py
```diff
@@ -616,15 +616,27 @@ def apply(
...
+
         def make_lines_only() -> str:
-            """Match locations without surrounding context"""
-            return f"Match lines per file:\n{self._to_json(match_lines_by_file)}"
+            """Match locations with line number and truncated matched text."""
+            compact = {
+                path: [{"line": m["line"], "text": str(m["text"])[:_TEXT_TRUNCATE]} for m in lines]
+                for path, lines in match_lines_by_file.items()
+            }
+            return f"Match locations per file (line + matched text):\n{self._to_json(compact)}"
```
opcode81: Instead of adding a new stage to the shortening chain which adds truncation, this replaces the previous first stage, where the full first line is retained.

Also, for the new shortening stage, the LLM is not informed 
* that lines are truncated to `_TEXT_TRUNCATE` characters in general
* which lines are truncated (could add an indicator like `...` at the end of affected lines)

opcode81 (changes requested):

src/serena/tools/file_tools.py
bymebyu: Thanks — addressed both points in be924c3c.

**Point 2 (signal truncation): done.** Truncated lines now get a trailing `...`, and the header states it, e.g.
`Matched lines (text over 60 chars is truncated, marked with a trailing '...'); use read_file with the line numbers for full content:`

**Point 1 (retain the full first line): adopted, but as a first stage rather than a straight replacement.** Replacing the truncating stage with a full-first-line stage re-overflows in the case that motivates the feature: on the ~1,300-match search I measured (full result ≈196K), the full-first-line stage is ≈206K > the 150K default `max_answer_chars`, so `_limit_length` skips it and falls straight through to bare line numbers — i.e. no matched text exactly when the result set is largest.

So the chain now tries the full first line first, and only falls back to a signalled truncated stage when the full lines don't fit:

```
full matches → full first lines → truncated first lines (marked) → line numbers → counts → summary
```

That yields full first lines for the common overflow (a few hundred matches, where they fit) and signalled truncation (≈119K, fits) for the pathological large-result case, before finally degrading to bare line numbers. Happy to tweak the wording/threshold further.

AmirF194 (commented):
Nice addition. Falling straight from full matches to ~1300 bare line numbers really did force a re-read to tell matches apart, so the snippet rung earns its place.

One thing I checked while reading, in case it saves anyone else the trip: the entry text stays correct even when `context_lines_before > 0`. `matched_lines` is filtered to `LineType.MATCH` in `MatchedConsecutiveLines.__post_init__` (src/serena/util/text_utils.py:75-82, with the `len(matched_lines) > 0` assert), so `matched_lines[0].line_content` is always a matched line here, never a leading context line.

Optional, not blocking: the `search_for_pattern` shortening chain has no test coverage today (the only `_limit_length` mention in the suite is the ReadFileTool bypass at test/serena/test_file_tools.py:20). That leaves the new stage, and in particular its ORDER in the chain, free to regress silently. Your PR description already has the ideal deterministic case, so it is cheap to pin. This passes on your branch (be924c3c):

```python
from unittest.mock import MagicMock

from serena.project import Project
from serena.tools.file_tools import SearchForPatternTool


def test_search_for_pattern_snippet_stage(tmp_path):
    lines = []
    for i in range(60):
        lines += ["filler above", "filler above",
                  f"MATCHME item number {i:04d} " + "payload " * 6,
                  "filler below", "filler below"]
    (tmp_path / "data.txt").write_text("\n".join(lines) + "\n", encoding="utf-8")

    project = Project.load(str(tmp_path), serena_config=SerenaConfig(gui_log_window=False, web_dashboard=False))
    agent = MagicMock()
    agent.get_active_project_or_raise.return_value = project
    tool = SearchForPatternTool(agent)

    def run(cap: int) -> str:
        return tool.apply(substring_pattern="MATCHME", context_lines_before=2,
                          context_lines_after=2, restrict_search_to_code_files=False,
                          max_answer_chars=cap)

    # wide but overflowing cap: the new snippet stage (line + matched text)
    snippet = run(7000)
    assert "The answer is too long" in snippet
    assert '"text":' in snippet and "MATCHME item number 0000" in snippet
    assert "Match lines per file" not in snippet  # not the bare-numbers stage

    # tighter cap: chain degrades past the snippet stage to bare line numbers
    bare = run(1000)
    assert "Match lines per file" in bare and '"text":' not in bare
```

Trim it however you like; the two assertions that matter are that a wide-but-overflowing cap returns the `{"line", "text"}` snippet, and a tighter cap drops past it to bare numbers. Reviewed with AI assistance; the test above was run in a clean container against this branch.

bymebyu:
Thanks for the careful read, @AmirF194 — and for confirming the `matched_lines[0]` / `context_lines_before` case; you're right that `MatchedConsecutiveLines.__post_init__` filters to `LineType.MATCH`, so the snippet always reflects a matched line rather than a leading context line.

Added your test as `test_search_for_pattern_snippet_stage` in 93780efe — it pins both the snippet stage and the fall-through to bare line numbers, which was exactly the untested part of the chain. Credited you in the commit. 🙏

opcode81:
Thanks for your contribution!

PR #1672, opened 2026-07-09
fix: Prevent hooks from hanging on unclosed stdin pipes

dutch2005:
Claude Code does not close stdin after piping JSON payloads for hooks, causing sys.stdin.read() to block indefinitely waiting for EOF. This replaces the blocking read with an incremental parser and a timeout thread, fixing hook deadlocks.

### DIFF
--- a/src/serena/hooks.py
+++ b/src/serena/hooks.py
@@ -3,6 +3,8 @@ import os
 import pickle
 import shutil
 import sys
+import threading
+import queue
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


MischaPanch:
Thanks for the report. I have never seen problems with this in claude code and there were no other reports on this. The proposed solution is also very hacky. Are you sure with the root cause analysis?

AmirF194 (commented):
Dug into the root-cause question from the comment above with a differential test, and there's a concrete correctness gap in the fix itself, independent of whether Claude Code actually leaves stdin open.

Setup: spawned the hook (`SessionStartActivateProjectHook`, which only needs `session_id`) as a subprocess and fed it a valid `{"session_id": ...}` payload in 3-byte chunks with a 0.2s delay between writes (about 1.2s total to fully arrive), then closed stdin normally, so this is a slow pipe, not a hung one.

- On `main` (e08e964, blocking `sys.stdin.read()`): succeeds after 2.7s with the correct session_id.
- On this branch (65b6637): the `q.get(timeout=1.0)` in `Hook.__init__` fires before the payload finishes arriving, `input_data` falls back to `{}`, and the hook raises `ValueError: Session ID is required in the hook input data` at 1.0s, before the writer even finished sending.

So the fix trades "hangs forever on a genuinely stuck pipe" for "hard-fails within 1 second on any pipe slower than that," and the resulting error is misleading: it reports a missing session ID when the real cause is a timeout mid-read.

The reader thread keeps running in the background after the 1s deadline, so one option is to reset the deadline on every chunk actually read (bail only when nothing arrives for 1s), instead of bounding the whole read to a single 1s window from the start. That would still recover from a truly stuck pipe without penalizing a slow-but-alive one.

Verify command: `python parent_test.py` (subprocess writes the payload in 3-byte chunks, 0.2s apart, then closes stdin), run unmodified against both commits above.

MischaPanch:
I'm going ahead and closing this, thanks for the analysis @AmirF194

PR #1673, opened 2026-07-09
SolidLSP: classify missing paths in is_ignored_path instead of raising

xormania:
Language servers can report locations of generated files that are not on disk. JDTLS does this for Lombok-generated types, reporting e.g. target/classes/test_repo/LombokModel$LombokModelBuilder.class while indexing; is_ignored_path then raised FileNotFoundError, turning an intermittent indexing state into a hard test failure (test_bare_symbol_names[java]; Tests run #4240, job jvm (macos-latest)).

Changes:
- is_ignored_path classifies missing paths with the usual lexical rules instead of raising: ignored-dirname components, the unsupported-extension rule, then the configured pathspec. A missing path is treated as a file if its name has a suffix and as a directory otherwise (so the leaf of a missing suffixless path is also checked against ignored dirnames). Behavior for existing paths is unchanged: the checks were reordered (ignored-dirname before unsupported-extension), but each check is pure, so the outcomes are identical.
- SymbolLocationRequest.convert_location_item skips locations whose absolute path does not exist (log + return None), so phantom build artifacts never reach consumers regardless of ignore configuration.
- New unit tests (test double, no language server process; they run in the catch-all CI batch) pin the missing-path classification — including the exact Lombok .class path shape from the CI failure — and the skipping of symbol locations that are not on disk.

Notes:
- The removed raise arrived verbatim with the multilspy import (2f607ff2) and was never a deliberate guard.
- The Project layer's equivalent check already tolerates missing paths (9c19d5f7), but returns False unconditionally. The LS layer instead classifies lexically so that LS-reported build artifacts still classify as ignored: with #1645 having removed the JDTLS build-dirname ignores, it is the unsupported-extension rule that catches *.class paths.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,10 @@ Status of the `main` branch. Changes prior to the next official version change w
   - Make tool call errors surface explicitly as errors at the MCP protocol level
 
 * Language Servers:
+  - Fix: `SolidLanguageServer.is_ignored_path` raised `FileNotFoundError` for paths that are not present
+    on disk, crashing symbol tools when a language server reported locations of generated files
+    (e.g. JDTLS reporting Lombok-generated classes under `target/classes`). Missing paths are now
+    classified by the usual ignore rules, and symbol locations resolving to non-existent files are skipped.
   - C/C++ (clangd): improve support and documentation for Unreal Engine 5 projects.
   - HLSL (`shader-language-server`): pass `--locked` to `cargo install` when building from source
     on macOS (and in the manual-install instructions), honoring the crate's packaged `Cargo.lock`.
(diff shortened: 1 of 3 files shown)


MischaPanch (approved):
LGTM, but @opcode81 should have a quick look too

CHANGELOG.md
```diff
@@ -30,6 +30,10 @@ Status of the `main` branch. Changes prior to the next official version change w
   - During project creation, language composition percentages are now computed relative to the total number 
     of recognised source files instead of all files, i.e. unrecognised files are ignored in the percentage 
     computation.
+  - Fix: `SolidLanguageServer.is_ignored_path` raised `FileNotFoundError` for paths that are not present
+    on disk, crashing symbol tools when a language server reported locations of generated files
+    (e.g. JDTLS reporting Lombok-generated classes under `target/classes`). Missing paths are now
+    classified by the usual ignore rules, and symbol locations resolving to non-existent files are skipped.
```
opcode81: This belongs in the "Language Servers" section; otherwise LGTM

PR #1674, opened 2026-07-09
Solidity: extend test_file_diagnostics xfail to all non-Linux platforms

xormania:
The test was already xfailed on macOS ("passes on Ubuntu but consistently fails on macOS CI"). The same failure signature -- the Solidity language server returns no diagnostics (AssertionError: []) -- also occurs on Windows: Tests run #4236, job catch-all (windows-latest), on main at e08e964d. Broaden the condition from sys.platform == "darwin" to sys.platform != "linux".

strict=False keeps a green run green when diagnostics do come through, and the marker is inert on Linux, so Ubuntu still enforces the test. The root cause (why the LS returns empty diagnostics on macOS/Windows runners) is not yet identified. Test-only quarantine, no changelog entry.

## Checklist

### DIFF
--- a/test/solidlsp/solidity/test_solidity_diagnostics.py
+++ b/test/solidlsp/solidity/test_solidity_diagnostics.py
@@ -10,8 +10,9 @@ from test.solidlsp.util.diagnostics import assert_file_diagnostics
 @pytest.mark.solidity
 class TestSolidityDiagnostics:
     @pytest.mark.xfail(
-        sys.platform == "darwin",
-        reason="Unknown — passes on Ubuntu but consistently fails on macOS CI; root cause not yet identified.",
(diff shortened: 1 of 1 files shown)


xormania:
I'll get to root cause - this keeps CI working for now with minimal downside

xormania:
Hold up - I think I found it - its a time out issue (2.5s - ouch) - i'm working on a fix PR

xormania:
Closing in favor of #1675, which fixes the root cause (await the language server's validation-completion signal) instead of quarantining. Its CI run — [Tests #4249](https://github.com/oraios/serena/actions/runs/29063798905) — is green across all 13 matrix jobs, including catch-all (macos-latest) and catch-all (windows-latest) with the xfail removed.

PR #1675, opened 2026-07-10
Solidity: await validation completion for diagnostics instead of a fixed wait

xormania:
## Summary

- Fix Solidity diagnostics intermittently returning an empty list on slow or cold environments
  (macOS/Windows CI). The Nomic language server validates asynchronously (debounce → forge/solc
  compile) and signals completion with `custom/validation-job-status`, which Serena ignored
  ("Unhandled method") while waiting a fixed 2.5s for a non-empty `publishDiagnostics` — returning
  `[]` whenever validation finished later (on CI runners this includes downloading solc, which the
  CI cache does not cover).
- Diagnostics requests keep the 2.5s non-empty publish wait as the fast path and, when it misses,
  await the completion signal against a 60s deadline (aligned with the existing indexing wait),
  then take the publication made for the request URI — accepting a validated empty result. The
  signal carries no document URI, so a completion belonging to another document re-arms the wait.
  If the signal never arrives, a warning is logged and the previous empty-result behavior applies.
- Remove the macOS xfail from `test_file_diagnostics` and add regression tests for the fallback
  delivery and for the never-signalled bound.

This mirrors #1639 (TypeScript/VTS: event-based indexing wait replacing a fixed sleep) and
supersedes the quarantine approach of #1674, which can be closed if this lands.

## Root cause evidence

- Shrinking the publish wait reproduces the exact CI failure signature (`AssertionError: []`) on
  Linux; the pinned server bundle (0.8.4) emits `custom/validation-job-status` on every completion
  path and publishes the document's diagnostics before it (verified in the packaged source).
- The historical macOS xfail (bd7e64e8) and the Windows failure (Tests run #4236,
  `catch-all (windows-latest)`) share the same empty-diagnostics signature.

## Test plan

- [x] Full solidity suite locally: 11 passed (includes the three new regression tests, one
  pinning the re-arm on spurious completions from other documents)
- [x] New tests verified in isolation on a fresh server instance
- [x] `poe format` / `poe type-check` / `poe lint` clean
- [x] CI on this PR: [Tests run #4249](https://github.com/oraios/serena/actions/runs/29063798905) — all 13 matrix jobs green, including `catch-all (macos-latest)` and `catch-all (windows-latest)`, the first runs of this test un-quarantined on those platforms.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ Status of the `main` branch. Changes prior to the next official version change w
   - Make tool call errors surface explicitly as errors at the MCP protocol level
 
 * Language Servers:
+  - Solidity: fix diagnostics intermittently coming back empty on slow or cold environments (macOS/Windows CI).
   - C/C++ (clangd): improve support and documentation for Unreal Engine 5 projects.
   - HLSL (`shader-language-server`): pass `--locked` to `cargo install` when building from source
     on macOS (and in the manual-install instructions), honoring the crate's packaged `Cargo.lock`.
--- a/src/solidlsp/language_servers/solidity_language_server.py
+++ b/src/solidlsp/language_servers/solidity_language_server.py
@@ -10,7 +10,7 @@ import pathlib
 import platform
 import shutil
 import threading
-from time import sleep
+from time import monotonic, sleep
 from typing import Any
 
 from overrides import override
@@ -40,6 +40,14 @@ class SolidityLanguageServer(SolidLanguageServer):
     Requires Node.js and npm to be installed.
     """
 
+    _VALIDATION_COMPLETION_TIMEOUT = 60.0
+    """upper bound for awaiting the validation-completion signal (``custom/validation-job-status``),
+    aligned with the indexing wait; on cold environments validation includes downloading the pinned
+    solc version"""
+
+    _POST_VALIDATION_PUBLISH_WAIT = 2.5
+    """grace period for the diagnostics publication that accompanies the completion signal"""
+
     @staticmethod
     def _determine_log_level(line: str) -> int:
         """Suppress known non-critical stderr output from the Solidity language server."""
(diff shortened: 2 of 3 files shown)


xormania:
This is solid (no pun intended)

src/solidlsp/language_servers/solidity_language_server.py
```diff
@@ -328,12 +352,43 @@ def request_text_document_diagnostics(
...
+                while diagnostics is None:
+                    remaining = deadline - monotonic()
+                    if remaining <= 0 or not self._validation_completed.wait(timeout=remaining):
+                        log.warning(
+                            f"Solidity validation completion was not signalled within "
+                            f"{self._VALIDATION_COMPLETION_TIMEOUT:.0f}s for {uri} "
+                            f"(last status: {self._last_validation_status}); returning without diagnostics"
+                        )
+                        break
+                    diagnostics = self._wait_for_published_diagnostics(
```
MischaPanch, line 386: we should add a debug log so something visible is happening in the potential 60s wait

src/solidlsp/language_servers/solidity_language_server.py
```diff
@@ -66,6 +74,11 @@ def __init__(self, config: LanguageServerConfig, repository_root_path: str, soli
             "solidity",
             solidlsp_settings,
         )
+        self._validation_completed = threading.Event()
+        """set when the language server reports an asynchronous validation run as finished
+        (``custom/validation-job-status``); cleared at the start of each diagnostics request"""
+        self._last_validation_status: Any = None
```
MischaPanch:
```suggestion
        self._last_validation_status: Any | None = None
```

CHANGELOG.md
```diff
@@ -48,6 +48,11 @@ Status of the `main` branch. Changes prior to the next official version change w
   - Make tool call errors surface explicitly as errors at the MCP protocol level
 
 * Language Servers:
+  - Solidity: fix diagnostics intermittently coming back empty on slow or cold environments
+    (macOS/Windows CI). The language server validates asynchronously via forge/solc and signals
+    completion with `custom/validation-job-status`; diagnostics requests now fall back to awaiting
+    that signal (60s bound) instead of returning an empty result after a fixed 2.5s publish wait.
+    The diagnostics test no longer carries a platform xfail.
```
MischaPanch:
```suggestion
  - Solidity: fix diagnostics intermittently coming back empty on slow or cold environments (macOS/Windows CI).
```

MischaPanch (changes requested):

MischaPanch:
Thanks, merging

PR #1678, opened 2026-07-11
fix(editing): delete last line of a file without a trailing newline

AmirF194:
### Root cause

`delete_lines` / `replace_lines` address the position one line past the last line (line == number of lines, col 0). `get_index_from_line_col` walks the text counting newlines to reach that line, but when the file has **no trailing newline** there is no closing newline to count, so it hits EOF and raises `InvalidTextLocationError`. `delete_text_between_positions` had no guard for this, so deleting the last line of such a file — a very common shape — crashed instead of deleting.

`ReadFileTool` numbers lines with `str.splitlines()`, which reports the same N lines whether or not the file ends in `\n`, so the model reads N lines and asks to delete line N-1, and the delete path disagrees.

### Fix

When the end position is the one-past-final-line position (line == number of lines, col 0) and cannot be resolved, clamp it to end-of-file in `delete_text_between_positions` — the exact guard `insert_text_at_position` already has for the same past-EOF position. A start position that is genuinely out of range still raises.

### Verified (clean Docker container, python:3.12-slim; serena deps installed, src selected via PYTHONPATH, `__file__` provenance-checked)

- On `main`: `delete_text_between_positions("a\nb\nc", 2, 0, 3, 0)` raises `InvalidTextLocationError` (`ls_utils.py:65`).
- On the branch: returns `("a\nb\n", "c")` — identical to the trailing-newline variant `"a\nb\nc\n"`.

Three `TextUtils` unit tests (delete-last-line, trailing-newline parity, and out-of-range still raises) fail/pass appropriately: the two fix cases fail on main and pass on the branch; the out-of-range guard test passes on both. `ruff check` + `ruff format` clean; CHANGELOG entry added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,10 @@ Status of the `main` branch. Changes prior to the next official version change w
   - Fix: Use LSP-compliant line splitting ("\n", "\r\n" and "\r" can define line breaks); previously only "\n" considered
   - Fix: Apply consistent line splitting across tools, uniformly applying the LSP splitting semantics; 
     affects `search_text` used by `search_for_pattern` tool (reported in #1684)
+  - Fix in `TextUtils` (used by editors): Deleting up to the end of the file, referencing the line one past 
+    the end of the file (particularly for files with no newline at end of file) raised `InvalidTextLocationError`
+    instead of accepting the deletion (change in `TextUtils.delete_text_between_positions`,
+    which now accepts the end position similar to `insert_text_at_position`).
   - Fix: glob pattern expansion in `expand_braces` did not terminate with empty or unbalanced braces #1690
 
 * CLI:
--- a/src/solidlsp/ls_utils.py
+++ b/src/solidlsp/ls_utils.py
@@ -237,7 +237,21 @@ class TextUtils:
         :return: a tuple containing the modified text and the deleted text
         """
         del_start_idx = TextUtils.get_index_from_line_col(text, start_line, start_col)
-        del_end_idx = TextUtils.get_index_from_line_col(text, end_line, end_col)
+        try:
+            del_end_idx = TextUtils.get_index_from_line_col(text, end_line, end_col)
+        except InvalidTextLocationError:
+            # Deleting through the final line addresses the position one line past the
+            # last line (line == number of lines, col 0). When the file has no trailing
+            # newline there is no closing newline to count, so get_index_from_line_col
+            # cannot resolve it; that position means "end of file". Clamp to len(text).
+            # (insert_text_at_position handles the same past-EOF position.)
+            text_stepper = TextStepper(text)
+            text_stepper.process_all()
+            num_lines_in_text = text_stepper.line + 1
+            if end_line == num_lines_in_text and end_col == 0:
+                del_end_idx = len(text)
+            else:
+                raise
 
         deleted_text = text[del_start_idx:del_end_idx]
         new_text = text[:del_start_idx] + text[del_end_idx:]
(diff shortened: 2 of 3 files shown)


AmirF194:
The red check is the `jvm (windows-latest)` job, which is flaky independently of this PR. `test_safe_delete_symbol_succeeds_when_no_references[java_model_user]` failed there with the JDTLS process throwing `org.eclipse.core.runtime.AssertionFailedException` (a language-server desync), and that same job failed on several recent `main` runs too (e08e964d, 0d4b70bc, d6fdac2c). The three Linux jobs are green.

This change only adds a branch inside the existing `except InvalidTextLocationError` path in `delete_text_between_positions`. When `get_index_from_line_col` resolves normally, every in-file symbol delete, including that test, the behavior is byte-identical to before. And the observed failure is a wrong-range edit with the symbol still present, not the `InvalidTextLocationError` this PR guards against, so it isn't something the change can produce. Happy to rebase if a fresh CI run would help.

AmirF194:
The only red job is `jvm (windows-latest)`, on `test_safe_delete_symbol_succeeds_when_no_references[java_model_user]`. In that run the Java language server logged an internal `org.eclipse.core.runtime.AssertionFailedException` in its stderr and the symbol was left in the file. The same test passes on `jvm (ubuntu-latest)` and `jvm (macos-latest)` in this run, and `jvm (windows-latest)` is green on the other current branches, so this looks like a Windows JDT flake rather than the change.

For what it's worth, the diff only adds an `except InvalidTextLocationError` branch in `delete_text_between_positions`; the success path that computes the delete range is unchanged, and that is the path this test exercises (no exception is raised here). Happy to re-trigger CI if that would help.

MischaPanch:
Thanks for the report. How did you run into this? What tool call actually led to `InvalidTextLocationError`?

If it's in delete symbol, the cause should be addressed there

opcode81:
@MischaPanch if language servers can have different notions of how the last line should be counted, then the proposed approach could be the most pragmatic. I suspect that they can return edits where the last line can be off by one.

AmirF194:
It comes from the line-based edit path (delete_lines / replace_lines), not delete_symbol. ReadFileTool numbers lines with splitlines(), so a file with no trailing newline still reports N lines. The model reads those N lines and asks to delete the last one, which addresses (line=N, col=0). get_index_from_line_col walks newlines to reach that one-past-final-line position, hits EOF because there is no closing newline to count, and delete_text_between_positions raises InvalidTextLocationError. With a trailing newline the same delete resolves fine, so the crash is specific to the no-trailing-newline shape.

delete_symbol reaches it only if the range it computes ends at that same one-past-final position; the fix is in the shared primitive so it covers both. The clamp targets exactly that past-EOF end position (the guard insert_text_at_position already applies for the same position), and a start position that is genuinely out of range still raises.

MischaPanch:
It's somewhat related to #1684 , let's land that one first.

Most our users never use the line-based editing tools since agent harnesses can do that and thus the tools are disabled for them. May I ask what your use case is for them? Are you using Serena through Claude Desktop or something like that?

AmirF194:
Not through Claude Desktop, no. I came across these while reading through Serena's editing and search utilities, and reproduced each with a targeted unit test rather than hitting them live in a harness. So I have no strong stake in the line-based tools staying enabled by default.

The #1684 fix holds up either way, since it is really about search_text and from_file_contents agreeing on line numbers regardless of which tools are exposed. Landing #1684 first makes sense to me; this one can follow it or be dropped as you prefer.

AmirF194:
Thanks for landing this, and for the sequencing behind #1684. Agreed that the shared line-numbering fix there was the right thing to land first; this guard just covers the no-trailing-newline edge on the same path.

PR #1679, opened 2026-07-11
fix(solidlsp): correct inverted case-folding in FilenameMatcher

chntllh:
## What

`FilenameMatcher.string_contains_relevant_filename` folds the input string to lower case when
`case_sensitive=True` (and leaves it unfolded when `False`) — the exact **inverse** of the rule its
sibling `is_relevant_filename` uses. The result is wrong in both modes: a case-insensitive matcher
misses upper-case input, and case-significant extensions (e.g. R's `.R` vs `.r`) are mismatched.

The fix folds only when `case_sensitive=False`, making the two sibling methods consistent.

This is a latent correctness bug — the method has no hot-path caller today, but it is a public method
on a public class, so the fix keeps the two matchers behaving identically.

## Testing

Added `test/solidlsp/test_ls_config.py` — pure unit tests (no language toolchain required) covering
both methods across both case-sensitivity modes, including the R `.R`/`.r` case. Verified the tests
**fail on the pre-fix code** and pass after. Full CI matrix (ubuntu/windows/macOS) green on my fork.

## Checklist

- [x] This PR follows the guidelines in `CONTRIBUTING.md` regarding the scope of PRs (small bug fix).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 Status of the `main` branch. Changes prior to the next official version change will appear here.
 
 * General:
+  - Fix: `FilenameMatcher.string_contains_relevant_filename` applied case-folding with an inverted
+    condition (folded the input when `case_sensitive=True` instead of when `False`), so it could
+    mismatch in both modes. It now folds consistently with `is_relevant_filename`. Added unit tests.
   - Add notion of trusted projects via new global configuration setting `trusted_project_path_patterns`.
     Current effects:
     - `ls_specific_settings` defined in project configurations will only be applied for trusted projects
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 Status of the `main` branch. Changes prior to the next official version change will appear here.
 
 * General:
+  - Fix: `FilenameMatcher.string_contains_relevant_filename` applied case-folding with an inverted
```
MischaPanch: Since this was not user-facing, let's remove the changelog entry

test/solidlsp/test_ls_config.py
```diff
@@ -0,0 +1,48 @@
+from __future__ import annotations
```
MischaPanch: Remove the import from future? I don't think it's needed

MischaPanch (changes requested):
Thanks for the fix! Minor comments, then this can be merged

PR #1557, opened 2026-06-08
rust-analyzer: lightweight profile (disable cachePriming + cargo.auto…

davidcforbes:
Fixes #1556.

Serena launches rust-analyzer with VS Code's full heavyweight `initializationOptions`, hardcoded. On a large, actively-built Rust workspace these drive rust-analyzer to **40+ GB resident** and trigger a full re-index on every external `cargo build`, even with no editor running. Serena's symbol tools don't need the two eager/auto re-indexing drivers:

- `cachePriming.enable`: `True` → `False` (no eager whole-workspace + all-deps index — the dominant contributor to the 40 GB resident set)
- `cargo.autoreload`: `True` → `False` (no re-index storm on `target/` / Cargo metadata changes)

### `checkOnSave` is intentionally left enabled

It is tempting to also disable `checkOnSave` for a lightweight profile, but doing so **breaks `get_diagnostics_for_file` for Rust**: rust-analyzer's diagnostics for ordinary errors (e.g. an unresolved name → `E0425`) come from flycheck (`cargo check`), and with `checkOnSave: false` Serena's pull-diagnostics request comes back empty — `test/solidlsp/rust/test_rust_diagnostics.py::test_file_diagnostics` fails with `[]`. `checkOnSave` only spawns transient `cargo check` child processes and is **not** a driver of rust-analyzer's resident RSS, so keeping it on preserves diagnostics at no meaningful memory cost.

Ideally these would be exposed as a config toggle (per-language `language_server_options`, or a global `lightweight` / `symbol_only` flag) rather than hardcoded, so users who want the full RA feature set can opt in. Happy to rework toward that if preferred.

## Testing

`uv run --extra dev pytest test/solidlsp/rust/` → **26 passed, 5 skipped** on Windows 11 (Rust stable 1.95, rust-analyzer from rustup). Bisected to confirm `cachePriming`/`autoreload` are the memory drivers and `checkOnSave` is required for the diagnostics test.

Issue #1556, opened 2026-06-08
rust-analyzer: unbounded memory (40+ GB) on large Rust workspaces — hardcoded heavyweight init options


davidcforbes:
Preconditions:

- [x] I have made sure it's an actual issue, not a question (use [GitHub Discussions](https://github.com/oraios/serena/discussions) instead).
- [x] I have consulted the [user guide](https://oraios.github.io/serena/02-usage/) and verified that the issue cannot be resolved by adjusting configuration/following recommended workflows.
- [x] I have looked for similar issues and discussions, including closed ones.

Issue details:

- [x] I have provided a meaningful title and description.
- [x] I have explained how the issue arose and, where possible, added instructions on how to reproduce it.
- [x] I have added details on my setup: Serena version, MCP client, OS, the programming language(s), relevant configuration adjustments and project specifics.
- [ ] If the issue relates to an application of Serena to an open-source project, I have added the link.

---

## Summary

On a large, actively-built Rust workspace, the `rust-analyzer` process Serena launches repeatedly grew to **40–44 GB RAM** — with **no editor (VS Code/JetBrains) running**. Serena's MCP server was the only thing keeping rust-analyzer engaged.

## Environment

- Serena `1.5.4.dev0` (git `main`), Windows 11, MCP server launched via `uvx --from git+https://github.com/oraios/serena serena`.
- Workspace: a multi-crate Rust workspace (~10 crates) with a path-dependency on a vendored fork of a large crate and a sibling UI crate — i.e. a large dependency graph.
- The workspace is **built continuously** from a terminal (`cargo build` / `cargo test`) while Serena is connected.

## Root cause

`src/solidlsp/language_servers/rust_analyzer.py` launches rust-analyzer with VS Code's full heavyweight `initializationOptions`, hardcoded, with no knob to tune them down. Two of these are not needed by Serena's symbol tools (`find_symbol`, `find_referencing_symbols`, `get_symbols_overview`, …) and are exactly what drives the resident memory on a big workspace:

| Option | Value | Effect |
|---|---|---|
| `cachePriming.enable` | `true` | Eagerly indexes the **entire workspace + all dependencies** at startup — the dominant contributor to the 40 GB resident set |
| `cargo.autoreload` | `true` | **Re-indexes on every `target/` / Cargo metadata change** — so each external `cargo build` triggers a re-index storm |

## Impact

- 40–44 GB rust-analyzer processes with no editor present.
- Continuous re-indexing whenever the project is built outside Serena.
[...]

Issue #1543, opened 2026-06-05
JetBrains backend crashes on startup: assert in init_jetbrains_ide makes documented-optional jetbrains_launch_command mandatory


Mohammad-Alavi:
## Summary

After PR #1539 (commit `1a3943b`, merged 2026-06-02), the JetBrains language backend fails to initialize on every startup unless `jetbrains_launch_command` is set. The new `init_jetbrains_ide()` task asserts the value is non-`None`, but the field defaults to `None` and is documented as optional — so any user on the default config with `language_backend: JetBrains` hits an `AssertionError` before the plugin connection is ever attempted.

## Reproduction

1. `language_backend: JetBrains` (global or project), `jetbrains_launch_command` unset (default).
2. JetBrains IDE already open with the Serena plugin running and the project loaded.
3. Start the MCP server.

Result — task fails immediately:

```
ERROR [Task-1:init_jetbrains_ide] Error during execution of Task-1:init_jetbrains_ide:
Traceback (most recent call last):
  File ".../serena/agent.py", line 1172, in init_jetbrains_ide
    assert self.serena_config.jetbrains_launch_command is not None
AssertionError
```

The IDE/plugin are healthy (plugin server responds with the correct `projectRoot`); the assert fires before `from_project()` is reached, so the already-running plugin is never used.

## Root cause

`src/serena/agent.py` (~line 1171):

```python
def init_jetbrains_ide() -> None:
    assert self.serena_config.jetbrains_launch_command is not None   # <-- crashes on default
    try:
        client = jetbrains_plugin_client.JetBrainsPluginClient.from_project(project, log_warning=False)
        log.info("Found Serena JetBrains Plugin server: %s", client)
    except jetbrains_plugin_client.ServerNotFoundError:
        log.info("Serena JetBrains Plugin server not found for project %s", project.project_name)
        if self.serena_config.jetbrains_launch_command:   # <-- launch already guarded here
            ...
```

`src/serena/config/serena_config.py`:

```python
jetbrains_launch_command: str | None = None
```

The launch path is already correctly guarded by the `if self.serena_config.jetbrains_launch_command:` at the end. The top-level `assert` is redundant and contradicts both the default (`None`) and the docs, which state launching an IDE is optional and that you can instead open the IDE manually.

## Expected

With the IDE already open and `jetbrains_launch_command` unset, Serena should find the running plugin server and proceed — no auto-launch needed.

## Suggested fix

[...]

PR #1680, opened 2026-07-11
fix(rust-analyzer): prevent indefinite startup hang and harden init

chntllh:
## What

Harden rust-analyzer startup against an indefinite hang and brittle initialization checks. All three
changes are confined to `_start_server`.

- **Indefinite hang:** `server_ready.wait()` had **no timeout**. If rust-analyzer crashes or never
  emits `experimental/serverStatus {quiescent: true}`, Serena startup hangs forever. Every other
  server that waits for readiness (kotlin, matlab, powershell, bash, …) uses a timeout and proceeds
  with a warning — rust-analyzer and clangd were the only two without one. Now waits 120s and proceeds
  with a warning on expiry.
- **Unguarded notification payloads:** the `language/status` and `experimental/serverStatus` handlers
  indexed `params[...]` directly, raising `KeyError` on an unexpected payload — and a `KeyError` in the
  `quiescent` handler could itself prevent `server_ready` from ever being set, feeding the hang above.
  Now use `params.get(...)`, matching the pyright server.
- **Over-strict capability assertion:** an exact-dict-equality assertion on `completionProvider`
  crashed initialization whenever rust-analyzer changed/added a field. These values are not used
  downstream, so a mismatch is now logged as a warning instead (same spirit as #1543).

Orthogonal to the memory/indexing work in #1556 / #1557, which tune `_get_initialize_params` options;
this only touches the readiness path in `_start_server`.

## Testing

All 34 Rust integration tests pass against real rust-analyzer (`poe test -m rust`); full CI matrix
(ubuntu/windows/macOS × all batches) green on my fork.

## Checklist

- [x] This PR follows the guidelines in `CONTRIBUTING.md` regarding the scope of PRs (bug fix).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,8 +47,13 @@ Status of the `main` branch. Changes prior to the next official version change w
   - Allow `query_project` tool to access read-only tools that are not enabled in the current configuration
   - Make tool call errors surface explicitly as errors at the MCP protocol level
 
-* Language Servers:
-  - C/C++ (clangd): improve support and documentation for Unreal Engine 5 projects.
+ * Language Servers:
+   - Rust (`rust-analyzer`): prevent an indefinite startup hang by adding a timeout to the readiness
+     wait (proceeds with a warning if the server never signals `quiescent`); harden the
+     `language/status` and `experimental/serverStatus` notification handlers against missing keys; and
+     relax the over-strict `completionProvider` capability check (an exact-equality assertion) to a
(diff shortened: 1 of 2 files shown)


MischaPanch:
Thanks for the contribution. We can merge this, but I'd like to know - is this a problem that you have actually encountered, or preemptive hardening?

chntllh:
I found it by code inspection while auditing LSP startup robustness, we did not reproduce an actual hang.

PR #1682, opened 2026-07-12
fix(search): glob '?' matches one character, not two

AmirF194:
## What

`glob_to_regex` (`src/serena/util/text_utils.py`) translated a glob `?` to `..`, i.e. two "any" characters, instead of a single `.`. As a result `search_text(..., is_glob=True)` treated `?` as "exactly two characters". The pattern `test_??.py` from `search_text`'s own docstring, for example, required four characters where two were intended.

Invariant: a glob `?` matches exactly one character. That is the behavior already documented for the sibling `glob_match` in the same file ("? matches a single character except /") and implied by the `test_??.py` docstring example.

## Fix

Emit `.` for `?` instead of `..` (one line). No other behavior changes; `*`, escapes and literal characters are untouched.

## Verification

Reproduced in a clean Docker container (uv, Python 3.11) against current `main` (`6018bf4`):

- On this branch, `test/serena/test_text_utils.py` passes (76 tests), including two added regression tests.
- Reverting only the one-character source change back to `..` makes both new tests fail:
  - `glob_to_regex("a?c")` returns `a..c`, so `re.fullmatch("a..c", "abc")` is `None`;
  - `search_text("c?t", is_glob=True)` matches `value_coat_end` (two characters) instead of `value_cat_end`.

Command:

    uv sync --extra dev
    uv run --no-sync pytest test/serena/test_text_utils.py -o addopts="" -k question

Scope note: this corrects the public `glob_to_regex` / `search_text(is_glob=True)` behavior. I did not find a first-party tool that currently passes `is_glob=True` with a `?` pattern, so this is not a user-facing tool regression; it fixes the glob utility and its documented `test_??.py` example.

## Checklist

- [x] This PR follows the guidelines in `CONTRIBUTING.md` regarding the scope of PRs (small bug fix, single logical change).

Done with AI assistance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 Status of the `main` branch. Changes prior to the next official version change will appear here.
 
 * General:
+  - Fix: in `glob_to_regex` / `search_text(is_glob=True)`, a `?` wildcard matched two characters instead
+    of one (it emitted `..` rather than `.`); it now matches a single character, consistent with the
+    `?` semantics documented for `glob_match` and the `test_??.py` example in `search_text`'s docstring.
   - Add notion of trusted projects via new global configuration setting `trusted_project_path_patterns`.
     Current effects:
     - `ls_specific_settings` defined in project configurations will only be applied for trusted projects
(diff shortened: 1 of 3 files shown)


MischaPanch:
Thank you, merging. I am surprised that it went unnoticed for so long

PR #1684, opened 2026-07-12
fix(search): newline-consistent line splitting in search_text so line content matches line number

AmirF194:
## What

`search_text` (`src/serena/util/text_utils.py`), which backs the `search_for_pattern` tool, decomposes the content into lines with `str.splitlines()` but computes each match's line number with `content[:start_pos].count("\n")` and then indexes back into that list with `lines[line_num]`.

`str.splitlines()` also breaks on `\r`, `\v`, `\f` (form feed), `\x1c`-`\x1e`, `\x85`, and the Unicode line separators `U+2028`/`U+2029`, whereas `count("\n")` counts only `\n`. So for any file that contains one of those characters before a match, the two decompositions disagree: `lines[line_num]` returns a different line than the one whose `\n`-based number is reported. The match's displayed content and its reported line number then refer to different lines, and both diverge from the `\n`-based convention used elsewhere in the same file (`MatchedConsecutiveLines.from_file_contents`, `MultiFileContentReplacer`) and by the line-based edit tools.

Invariant: the line list used to fetch `lines[line_num]` and the index reported as `line_number` must come from the same decomposition, and that decomposition is `\n` (the one the edit and read tools use).

## Fix

Build the line list with `content.split("\n")` instead of `content.splitlines()` (one line, plus a comment). This makes `lines[count("\n")]` line up with the reported number and with the rest of the file. No other behavior changes.

## Verification

Reproduced in a clean Docker container (uv, Python 3.11, `uv sync --extra dev --locked`) against current `main` (`9bbbfea`):

- Added regression test `test_search_text_line_content_aligns_with_line_number_over_special_whitespace`: searching `"TARGET_here"` in `"alpha\nbeta\x0cgamma\ndelta\nTARGET_here\nomega\n"` must return the match at line 3 (0-based) with content `"TARGET_here"`.
- On this branch the full `test/serena/test_text_utils.py` suite passes (77 tests, including the new one).
- Reverting only the source change back to `splitlines()` makes the new test fail: the match is reported at line 3 but with content `"delta"`, because the form feed after `beta` shifts the `splitlines()` index by one.
- `uv run poe lint` is clean.

Command:

    uv sync --extra dev --locked
    uv run pytest test/serena/test_text_utils.py -q
    uv run poe lint

Scope note: this path is reachable from the `search_for_pattern` tool (via `Project.search_project_files_for_pattern` and `search_files`) for any real file containing one of the above characters. The trigger is uncommon (form feeds, vertical tabs, lone carriage returns, or Unicode line separators), so severity is low, but the search result it produces in that case is incorrect. Scope is limited to the line decomposition in `search_text`; no other function is touched.

## Checklist

- [x] This PR follows the guidelines in `CONTRIBUTING.md` regarding the scope of PRs (small bug fix, single logical change).

Done with AI assistance.

### DIFF
--- a/src/serena/util/text_utils.py
+++ b/src/serena/util/text_utils.py
@@ -169,7 +169,12 @@ def search_text(
         raise ValueError("Pass either content or source_file_path")
 
     matches = []
-    lines = content.splitlines()
+    # Split on "\n" only, to stay consistent with the line numbering below
+    # (start_line_num = content[:start_pos].count("\n")). str.splitlines() also
+    # breaks on \r, \v, \f, \x1c-\x1e, \x85 and the Unicode line separators, which
+    # would desync lines[line_num] from the count("\n")-based index, and diverges
+    # from the "\n" convention used by from_file_contents and the edit tools.
+    lines = content.split("\n")
     total_lines = len(lines)
 
     # Convert pattern to a compiled regex if it's a string
--- a/src/serena/util/text_utils.py
+++ b/src/serena/util/text_utils.py
@@ -100,8 +100,8 @@ class MatchedConsecutiveLines:
     def from_file_contents(
         cls, file_contents: str, line: int, context_lines_before: int = 0, context_lines_after: int = 0, source_file_path: str | None = None
     ) -> Self:
-        # Same "\n" line decomposition as search_text, via the shared helper, so both
-        # agree on line content for a given line number regardless of CRLF/other breaks.
+        # Same "\n" line decomposition as search_text, via the shared helper, so both index
+        # a given line number the same way (unaffected by form feeds / other non-"\n" breaks).
         line_contents = TextUtils.split_lines(file_contents)
         start_lineno = max(0, line - context_lines_before)
         end_lineno = min(len(line_contents) - 1, line + context_lines_after)
@@ -174,7 +174,10 @@ def search_text(
     # Decompose with the shared TextUtils.split_lines helper so the content indexed here
     # stays consistent with the count("\n")-based line numbers below and with the same
     # convention used by from_file_contents and the edit tools (see TextUtils.split_lines).
-    lines = TextUtils.split_lines(content)
+    # Trim a trailing "\r" per line so CRLF content renders without carriage returns, matching
+    # the display str.splitlines() used to give here; this strip is search-display-specific and
+    # deliberately not baked into the shared helper.
+    lines = [line[:-1] if line.endswith("\r") else line for line in TextUtils.split_lines(content)]
     total_lines = len(lines)
 
     # Convert pattern to a compiled regex if it's a string
--- a/src/serena/util/text_utils.py
+++ b/src/serena/util/text_utils.py
@@ -100,7 +100,9 @@ class MatchedConsecutiveLines:
     def from_file_contents(
         cls, file_contents: str, line: int, context_lines_before: int = 0, context_lines_after: int = 0, source_file_path: str | None = None
     ) -> Self:
-        line_contents = file_contents.split("\n")
+        # Same "\n" line decomposition as search_text, via the shared helper, so both
+        # agree on line content for a given line number regardless of CRLF/other breaks.
+        line_contents = TextUtils.split_lines(file_contents)
         start_lineno = max(0, line - context_lines_before)
         end_lineno = min(len(line_contents) - 1, line + context_lines_after)
         text_lines: list[TextLine] = []
@@ -169,15 +171,10 @@ def search_text(
         raise ValueError("Pass either content or source_file_path")
 
     matches = []
-    # Split on "\n" only, to stay consistent with the line numbering below
-    # (start_line_num = content[:start_pos].count("\n")). str.splitlines() also
-    # breaks on \r, \v, \f, \x1c-\x1e, \x85 and the Unicode line separators, which
-    # would desync lines[line_num] from the count("\n")-based index, and diverges
-    # from the "\n" convention used by from_file_contents and the edit tools.
-    # Drop a single trailing "\r" per line so Windows CRLF (and lone-CR) endings
-    # render without it; this leaves the line count unchanged, so line numbers stay
-    # "\n"-based and keep matching the count("\n") index.
-    lines = [line[:-1] if line.endswith("\r") else line for line in content.split("\n")]
+    # Decompose with the shared TextUtils.split_lines helper so the content indexed here
+    # stays consistent with the count("\n")-based line numbers below and with the same
+    # convention used by from_file_contents and the edit tools (see TextUtils.split_lines).
+    lines = TextUtils.split_lines(content)
     total_lines = len(lines)
 
     # Convert pattern to a compiled regex if it's a string
--- a/src/serena/util/text_utils.py
+++ b/src/serena/util/text_utils.py
@@ -174,7 +174,10 @@ def search_text(
     # breaks on \r, \v, \f, \x1c-\x1e, \x85 and the Unicode line separators, which
     # would desync lines[line_num] from the count("\n")-based index, and diverges
     # from the "\n" convention used by from_file_contents and the edit tools.
-    lines = content.split("\n")
+    # Drop a single trailing "\r" per line so Windows CRLF (and lone-CR) endings
+    # render without it; this leaves the line count unchanged, so line numbers stay
+    # "\n"-based and keep matching the count("\n") index.
+    lines = [line[:-1] if line.endswith("\r") else line for line in content.split("\n")]
     total_lines = len(lines)
 
     # Convert pattern to a compiled regex if it's a string
--- a/src/solidlsp/ls_utils.py
+++ b/src/solidlsp/ls_utils.py
@@ -70,6 +70,19 @@ class TextUtils:
         return idx
 
     @staticmethod
+    def split_lines(text: str) -> list[str]:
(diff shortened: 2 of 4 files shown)


src/serena/util/text_utils.py
```diff
@@ -169,7 +169,12 @@ def search_text(
         raise ValueError("Pass either content or source_file_path")
 
     matches = []
-    lines = content.splitlines()
+    # Split on "\n" only, to stay consistent with the line numbering below
+    # (start_line_num = content[:start_pos].count("\n")). str.splitlines() also
+    # breaks on \r, \v, \f, \x1c-\x1e, \x85 and the Unicode line separators, which
+    # would desync lines[line_num] from the count("\n")-based index, and diverges
+    # from the "\n" convention used by from_file_contents and the edit tools.
+    lines = content.split("\n")
```
MischaPanch: This is not OS independent, right? On Windows `\r\n` may be used

AmirF194: You're right, thanks. split("\n") kept a trailing "\r" on CRLF (and lone-CR) lines, so on Windows the reported match content carried it even though the line number was correct.

641a640 drops a single trailing "\r" per line. The line count is unchanged, so line numbers stay newline-based and keep matching the count("\n") index, and the content renders without the carriage return. I added a CRLF test asserting the reported line has no trailing "\r". The newline-count numbering and from_file_contents already split on "\n", so search stays consistent with them; the carriage return was the one gap.

MischaPanch: I did not look into this in detail yet, but if we count and split lines in different places we need a central place (e.g. a central abstraction) to do this. Everything else is umaintainable. Wdyt @AmirF194 (since you already had a look). I agree that using splitlines and then using count("\n") is suboptimal, we need a consistent way to do that and centralize the logic accordingly

AmirF194: Agreed, that is a fair concern, and it is exactly what let this bug through. The "\n" line convention is already assumed in several places (`from_file_contents` splits on "\n", `search_text` counts "\n" for the numbers, `TextUtils.get_index_from_line_col` and `_get_updated_position_from_line_and_column_and_edit` count "\n"), while a couple of spots still reach for `str.splitlines()` (this function before the fix) or `splitlines(keepends=True)` (`TextUtils.get_text_in_lines_range`). So the decomposition can drift depending on which helper you land in.

`TextUtils` in `solidlsp/ls_utils.py` looks like the natural home: it already owns the line/column indexing on the "\n" convention, and `text_utils.py` already imports it. One helper there, say `TextUtils.split_lines(content)` that splits on "\n" and trims a single trailing "\r" per line, could back both `search_text` and `from_file_contents`, so the split and the count cannot diverge again.

That is wider than this bugfix, so I did not want to expand the PR without your call. Two ways to go:

1. Land this PR as the minimal `search_text` fix, and I open a follow-up that adds `TextUtils.split_lines` and routes the existing call sites through it.
2. Expand this PR now to add the helper and switch `search_text` and `from_file_contents` over to it.

I am glad to do either. My mild preference is (1), so the correctness fix can land on its own and the centralization gets its own reviewable diff, but I will follow whichever you prefer.

MischaPanch: Thanks for the quick response. Let's to this right from the start, a review won't be a problem. There's no point in merging a patch and then do the proper refactoring which adjusts it again. I agree, TextUtils is the natural place

AmirF194: Done, pushed 6b91e0b. Added `TextUtils.split_lines(text)` next to the other line/col helpers in `ls_utils.py` (splits on `"\n"`, matching the `count("\n")` convention that `get_index_from_line_col`/`get_line_col_from_index` already use, and drops a single trailing `"\r"` for Windows CRLF). Both `search_text` and `MatchedConsecutiveLines.from_file_contents` now go through it, so the split and the line count share one definition and cannot drift apart again.

I left `get_text_in_lines_range` on `splitlines(keepends=True)`, since it reconstructs the exact text rather than numbering lines; happy to fold it in too if you'd prefer one entry point for everything.

Tests: unit tests for `split_lines` plus a consistency test asserting `search_text` and `from_file_contents` return the same content for the same line number (it fails if either bypasses the helper). Full `test_text_utils` suite, `poe lint`, and `poe type-check` pass in a clean container.

src/solidlsp/ls_utils.py
```diff
@@ -69,6 +69,19 @@ def get_index_from_line_col(text: str, line: int, col: int) -> int:
...
+        r"""
+        Splits text into lines using the "\n" convention that this class' line/column indexing relies on
+        (get_index_from_line_col, get_line_col_from_index), so a line index computed as ``prefix.count("\n")``
+        always addresses the matching entry. The number of returned lines equals ``text.count("\n") + 1``.
+
+        Unlike ``str.splitlines()``, this does not treat "\r", form feed, vertical tab or the Unicode line
+        separators as line breaks (which would desync the content from the newline-based line number). A single
+        trailing "\r" per line is dropped so Windows CRLF endings do not leak into the returned content.
+        """
+        return [line[:-1] if line.endswith("\r") else line for line in text.split("\n")]
```
opcode81: The global helper doesn't make sense in this form. It is applied only in `search_text` where the removal of "\r" perhaps makes sense. It wouldn't make sense in most other contexts, however.

opcode81 (changes requested):

src/solidlsp/ls_utils.py
AmirF194: You're right, thanks. 7d70273 keeps `split_lines` as just the `\n` split (the shared convention that `search_text` and `from_file_contents` both index against) and moves the trailing-`\r` trim into `search_text`, where it belongs as you said. `from_file_contents` is back to its prior `\n`-only behavior, so nothing else picks up the carriage-return handling. 82 tests pass, lint and type-check clean.

src/serena/util/text_utils.py
```diff
@@ -169,7 +171,13 @@ def search_text(
...
 
     matches = []
-    lines = content.splitlines()
+    # Decompose with the shared TextUtils.split_lines helper so the content indexed here
+    # stays consistent with the count("\n")-based line numbers below and with the same
+    # convention used by from_file_contents and the edit tools (see TextUtils.split_lines).
+    # Trim a trailing "\r" per line so CRLF content renders without carriage returns, matching
+    # the display str.splitlines() used to give here; this strip is search-display-specific and
+    # deliberately not baked into the shared helper.
+    lines = [line[:-1] if line.endswith("\r") else line for line in TextUtils.split_lines(content)]
```
MischaPanch, line 180: that's one more loop, potentially slow. I also feel like it doesn't belong here, split_lines should take care of that

src/solidlsp/ls_utils.py
```diff
@@ -69,6 +69,20 @@ def get_index_from_line_col(text: str, line: int, col: int) -> int:
...
         return idx
 
+    @staticmethod
+    def split_lines(text: str) -> list[str]:
+        r"""
+        Splits text into lines using the "\n" convention that this class' line/column indexing relies on
+        (get_index_from_line_col, get_line_col_from_index), so a line index computed as ``prefix.count("\n")``
+        always addresses the matching entry. The number of returned lines equals ``text.count("\n") + 1``.
+
+        Unlike ``str.splitlines()``, this does not treat "\r", form feed, vertical tab or the Unicode line
```
MischaPanch, line 79: The recommended procedure is to use the inbuilt splitlines, so I think we should use that and adjust the counts that don't use it instead. Is there something preventing this from happening?

Maybe it makes sense if @opcode81 and I take this over instead of the ping pong in PR comments :)

AmirF194: The counts use "\n" because the LSP position math is newline-based: get_index_from_line_col / get_line_col_from_index in ls_utils walk "\n" only. So if search_text numbers lines with splitlines(), it agrees with ReadFileTool but desyncs from that position math on any "\r", form feed, vertical tab, or Unicode line separator, which is the same class of divergence this PR closes, just moved to the position/edit side. Standardizing everything on splitlines() would mean changing get_index_from_line_col too, which is a larger change to the LSP layer.

That said, the seam is your call. Happy for you and @opcode81 to take it over, and just as happy to implement whichever direction you pick. I only care that search and edit report the same line.

opcode81: @MischaPanch the problem with the built-in `splitlines` is that it does not match the logic in `TextUtils`. It splits at many characters, whereas the logic elsewhere in Serena assumes that only "\n" induces a new line.
We mostly compute line numbers based on "\n" occurrences, so just using `splitlines` will not achieve consistency. As the OP pointed out, its logic is inconsistent with the other logic.
Notably, LSP defines that line changes are to be induced by three character sequences ("\n", "\r\n" and "\r" on its own), which is at odds with both our "\n"-only logic and `splitlines`.

MischaPanch: If only I could travel back in time and force a single convention... @AmirF194 thank you for your work here, we'll take this over, all the splitting and counting need to be handled in a single place in a consistent way

AmirF194: Thanks, that works. Given the LSP position math is "\n"-based, consolidating the splitting and counting into one place is the right call. It's yours now; if it's useful I can implement whichever convention you settle on.

opcode81:
Superseded by #1691

PR #1687, opened 2026-07-12
fix(hooks): emit valid Codex cleanup output

### LINKED ISSUE #1533: Codex Stop hook: `serena-hooks cleanup --client=codex` emits invalid Stop-hook stdout

Bortlesboat:
## Summary
- emit the Codex Stop-hook success object after cleanup
- keep cleanup diagnostics on stderr so stdout remains parseable JSON
- return valid JSON even when Codex input lacks a session ID
- preserve existing behavior for all other hook clients

Closes #1533

## Verification
- uv run pytest test/serena/test_hooks.py::TestHookCli::test_cleanup_command test/serena/test_hooks.py::TestHookCli::test_codex_cleanup_emits_valid_stop_hook_json test/serena/test_hooks.py::TestHookCli::test_codex_cleanup_without_session_id_still_emits_valid_json -q (3 passed)
- uv run ruff check src/serena/hooks.py test/serena/test_hooks.py
- uv run ruff format --check src/serena/hooks.py test/serena/test_hooks.py
- uv run ty check src/serena/hooks.py
- git diff --check

### DIFF
--- a/src/serena/hooks.py
+++ b/src/serena/hooks.py
@@ -591,7 +591,15 @@ class HookCommands(AutoRegisteringGroup):
     @click.command("cleanup", help="Set this as hook at session end all hook data for the current session")
     @_client_option
     def cleanup(client: str) -> None:
-        SessionEndCleanupHook(HookClient(client)).execute()
+        hook_client = HookClient(client)
+        try:
+            SessionEndCleanupHook(hook_client).execute()
+        except Exception as exc:
+            if hook_client != HookClient.CODEX:
+                raise
+            click.echo(f"Serena cleanup failed: {exc}", err=True)
+        if hook_client == HookClient.CODEX:
+            click.echo(json.dumps({"continue": True}))
 
     @staticmethod
     @click.command(
--- a/test/serena/test_hooks.py
+++ b/test/serena/test_hooks.py
@@ -693,6 +693,27 @@ class TestHookCli:
         assert result.exit_code == 0
         assert not session_dir.exists()
 
+    def test_codex_cleanup_emits_valid_stop_hook_json(self, tmp_path: Path):
+        runner = CliRunner()
+        stdin_json = json.dumps({"session_id": "codex-cleanup"})
+        with patch("serena.hooks.serena_home_dir", str(tmp_path)):
+            result = runner.invoke(hook_commands, ["cleanup", "--client", "codex"], input=stdin_json)
+
+        assert result.exit_code == 0
+        assert json.loads(result.stdout) == {"continue": True}
+
+    def test_codex_cleanup_without_session_id_still_emits_valid_json(self):
+        runner = CliRunner()
+        result = runner.invoke(
+            hook_commands,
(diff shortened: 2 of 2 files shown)


Bortlesboat:
CI note: the Ubuntu catch-all failure is an unrelated timeout in test_dynamically_activated_mode_is_provided_once_per_session[python] after 1,000 tests passed; the tool returned TimeoutError after two four-minute waits. The changed Codex cleanup-hook tests pass, and the other completed platform/language shards plus build, spelling, and CodeQL are green.

Bortlesboat:
The remaining two red Ubuntu lanes are infrastructure/flaky rather than branch failures: 
ative was cancelled after exactly six hours without completing, and catch-all reached 1,000 passing tests before one unrelated TestPromptProvision call timed out after ~15 minutes. All 15 other executed matrix jobs pass. I tried rerunning failed jobs, but GitHub restricts workflow reruns here to repository admins. Could a maintainer rerun failed jobs for workflow 29207580519 when capacity permits?

AmirF194 (commented):
This resolves both scenarios from #1533 for the cleanup/Stop hook, confirmed against this branch (4231098) in a clean container: `cleanup --client=codex` with a session_id now prints `{"continue": true}`, and with a missing session_id it prints the same JSON to stdout while the ValueError goes to stderr, exit 0 either way.

One related gap this PR doesn't touch: `activate` (the SessionStart hook, hooks.py:586-588) has the identical failure shape as scenario 2. It has no try/except, so a missing session_id raises uncaught through `Hook.__init__`, same as cleanup did before this fix. Reproduced on this branch:

```
printf '%s' '{"stop_hook_active":false}' | serena-hooks activate --client=codex
```

exits nonzero with a raw traceback, no JSON on stdout, for every client including codex. If Codex's SessionStart hook has the same "must be valid JSON" contract as Stop, this would trip the same class of error this PR fixes for cleanup. Worth a follow-up PR applying the same guard to `activate`, or is SessionStart's contract different enough that it's a non-issue there?

Bortlesboat:
Closing in favor of #1738. It covers the same #1533 cleanup-hook fix on the current codebase, is mergeable, and its full cross-platform matrix is green. Keeping this conflicting branch open would duplicate that active work.

PR #1691, opened 2026-07-14
Apply consistent, LSP-compliant line counting

opcode81:
* Improve line/column computations, adhering to the [LSP standard](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#textDocuments)

   * Add helper class `TextStepper` to facilitate computations
   * Reimplement `TextUtils` methods based on `TextStepper`
   * Adhere to LSP definition of newlines in text documents: either "\n", "\r\n" or "\r" on its own.
     Previously, we only considered "\n".
   * Add util method `TextUtils.split_lines`, allowing usages of Python's `splitlines` to be replaced
     * in `search_text` and
     * in `MatchedConsecutiveLines.from_file_contents`,
     
     both of which affect the `replace_content` tool.

### DIFF
--- a/src/solidlsp/ls_utils.py
+++ b/src/solidlsp/ls_utils.py
@@ -31,6 +31,114 @@ class InvalidTextLocationError(Exception):
     pass
 
 
+class TextStepper:
+    def __init__(self, chars: str):
+        self._chars = chars
+        self._len = len(chars)
+        self.line = 0
+        """
+        the current 0-based line index
+        """
+        self.col = 0
+        """
+        the current 0-based column index
+        """
+        self.idx = 0
+        """
+        the current 0-based index in the full text.
+        The index specifies the next character to be processed, i.e. if this is
+        the length of the text, then the end of the text has been reached.
+        It specifies a location in the same way as a cursor insertion position:
+        cursor at the very beginning (idx=0) means insert before the first character.
+        """
+        self._is_newline = False
+        """
+        whether the last step was a line break
+        """
+        self._prev_line_start_idx = 0
+        """
+        start index of the last fully processed line (inclusive)    
+        """
+        self._prev_line_end_idx = 0
+        """
+        end of the last fully processed line (exclusive), excluding newline characters
+        """
+        self._line_start_idx = 0
+        """
+        start of the current, not yet completed line (inclusive)
+        """
+
+    def _get_char(self, idx: int) -> str | None:
+        if idx < 0 or idx >= self._len:
+            return None
+        return self._chars[idx]
+
+    def step(self) -> bool:
+        """
+        Processes the next character/newline sequence in the text
+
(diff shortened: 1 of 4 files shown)


src/solidlsp/ls_utils.py
```diff
@@ -41,46 +145,37 @@ def get_line_col_from_index(text: str, index: int) -> tuple[int, int]:
...
-                c = 0
-            else:
-                c += 1
-            idx += 1
-
-        return l, c
+        text_stepper = TextStepper(text)
+        while text_stepper.idx < index:
+            if not text_stepper.step():
+                raise InvalidTextLocationError
```
MischaPanch: we could include the passed location and the final index from the stepper into the error message

src/solidlsp/ls_utils.py
```diff
@@ -41,46 +145,37 @@ def get_line_col_from_index(text: str, index: int) -> tuple[int, int]:
...
     def get_index_from_line_col(text: str, line: int, col: int) -> int:
         """
         Returns the index of the given zero-indexed line and column number in the given text
         """
-        idx = 0
-        while line > 0:
-            if idx >= len(text):
+        text_stepper = TextStepper(text)
+        while text_stepper.line < line:
+            if not text_stepper.step():
```
MischaPanch: also here. I guess the uninformative errors are still from the multilspy aera

MischaPanch (commented):
I am a bit worried about performance here, since now we loop over characters for splitting lines, though I guess it doesn't matter. If you think it does matter (we do perform quite some text operations), this likely could be optimized with numba, I suspect that an agent might find a good optimization (the best one would probably be a c routine, if numba won't do the trick). I think I already know your answer (not needed), but still wanted to point it out, feel free to ignore the optimization aspect.

Optimization aside - are these really all the places where we count lines and columns? Is the issue about deletion of lines solved by this? #1678

opcode81:
> I am a bit worried about performance here

The updated algorithm now steps line by line in Python, leaving the character steps to the regex library.
This has led to a ~20-100x speedup in my test.

src/solidlsp/ls_utils.py
opcode81: We could, but nobody actually reads this error.

opcode81:
> Is the issue about deletion of lines solved by this? https://github.com/oraios/serena/pull/1678

No, it is not, because in

     delete_text_between_positions("a\nb\nc", 2, 0, 3, 0)

the position (3, 0) is actually invalid. It designates the start of the line one beyond the actual content (as if trailing newline was present). The creator of the ticket simply wants this position to be allowed, which is reasonable.

src/solidlsp/ls_utils.py
opcode81: Added

PR #1692, opened 2026-07-14
Clean up LanguageServerInterface creation

opcode81:


src/solidlsp/ls.py
```diff
@@ -786,15 +782,25 @@ def _create_dependency_provider(self) -> LanguageServerDependencyProvider:
...
-        env = self._dependency_provider.create_launch_command_env()
+        dependency_provider = self._get_dependency_provider()
+        cmd = dependency_provider.create_launch_command()
+        env = dependency_provider.create_launch_command_env()
         return ProcessLaunchInfo(cmd=cmd, cwd=self.repository_root_path, env=env)
 
-    def _create_language_server_interface(
-        self, process_launch_info: ProcessLaunchInfo, logging_fn: Callable[[str, str, StringDict | str], None] | None
-    ) -> LanguageServerInterface:
+    def _get_process_launch_info(self) -> ProcessLaunchInfo:
```
MischaPanch, line 796: I feel like we don't need the _create method anymore, it could be inlined into the getter. Wdyt?

MischaPanch (approved):
Minor proposal, but feel free to merge without addressing

src/solidlsp/ls.py
opcode81: We need it because an implementation overrides it.

opcode81: (I would have removed it from the get-go if that wasn't the case)

MischaPanch: I see, makes sense

PR #1694, opened 2026-07-14
fix: reject malformed brace globs

### LINKED ISSUE #1690: expand_braces() never terminates on malformed brace globs (empty {} or unbalanced {)
Preconditions:

- [x] I have consulted the user guide and verified that the issue cannot be resolved by adjusting configuration/following recommended workflows.
- [x] I have looked for similar issues and discussions, including closed ones (none cover brace expansion; #470 is a separate gitignore-parsing hang).

## Summary

`expand_braces()` in `src/serena/util/text_utils.py` never terminates when its input contains a `{` that is not part of a well-formed, non-empty `{a,b,...}` group. An empty pair `{}` or an unbalanced `{` (no closing brace) spins a tight CPU-bound loop forever.

This is reachable from any tool that accepts a glob. `paths_include_glob` / `paths_exclude_glob` flow straight into `expand_braces` with no validation:

- `ReplaceInFilesTool` -> `_collect_files`: `src/serena/tools/file_tools.py:354-355`
- `SearchForPatternTool` -> `Project.search_*_for_pattern` -> `search_files` -> `GlobMatcher(...)`: `src/serena/util/text_utils.py:290` (`GlobMatcher.__init__`)

So an agent that passes a typo'd glob such as `src/{utils,models` (a common mistake: forgetting the closing brace) hangs the process. The loop has no `yield`/`sleep`/exception, so a thread-based tool timeout cannot interrupt it.

## Root cause

`src/serena/util/text_utils.py:219` (function spans 212-232 on `98966c6`):

```python
patterns = [pattern]
while any("{" in p for p in patterns):          # loop condition: any pattern still has "{"
    new_patterns = []
    for p in patterns:
        match = re.search(r"\{([^{}]+)\}", p)   # requires a closing "}" AND >=1 char inside
        if match:
            ...                                  # progress: expands the group
        else:
            new_patterns.append(p)               # no progress: p (with its "{") is kept as-is
    patterns = new_patterns
```

The loop continues while any pattern contains `{`, but the body only makes progress when the regex `\{([^{}]+)\}` matches. That regex needs a closing `}` and at least one non-brace character inside. For an empty `{}` or an unbalanced `{`, no pattern matches, every pattern is re-appended unchanged, `patterns` is rebuilt identically, the `{` never leaves, and the condition stays true forever.

The invariant that is violated: `expand_braces` must terminate for every input string. A stray or empty brace that cannot form a group should be treated as a literal (matching how `fnmatch` treats it downstream), not spun on.

## Reproduction

Pure-Python, no language server or MCP client needed (I ran the real `serena.util.text_utils.expand_braces` from HEAD in a clean container, each call bounded by a `SIGALRM`):

```python
from serena.util.text_utils import expand_braces

expand_braces("**/*.{js,ts}")   # OK   -> ['**/*.js', '**/*.ts']
expand_braces("a{}b")           # HANGS forever (empty braces)
expand_braces("src/{a,b")       # HANGS forever (unbalanced brace)
```

Observed: the two malformed calls never return (with a 4s alarm bound both time out; the two well-formed controls return immediately). Expected: `expand_braces` terminates. Reasonable behavior would be to treat the stray/empty braces as literal characters (so `a{}b` yields `['a{}b']`), or to raise a clear "malformed glob pattern" error.

## Suggested fix

Make the loop stop when a full pass expands nothing, leaving any remaining brace as a literal:

```python
patterns = [pattern]
while True:
    new_patterns = []
    changed = False
    for p in patterns:
        match = re.search(r"\{([^{}]+)\}", p)
        if match:
            changed = True
            prefix, suffix = p[: match.start()], p[match.end() :]
            for option in match.group(1).split(","):
                new_patterns.append(f"{prefix}{option}{suffix}")
        else:
            new_patterns.append(p)
    patterns = new_patterns
    if not changed:
        break
return patterns
```

This is identical for well-formed input and terminates on malformed input (stray/empty braces pass through as literals). I am happy to send a PR with this fix and a regression test if you'd prefer that approach; if you'd rather reject malformed globs with an explicit error, I can do that instead.

## Setup

- Serena version: 1.5.4.dev0 (HEAD `98966c6`)
- Reproduced by calling the tool-facing function directly (no MCP client required); the same input reaches it through `ReplaceInFilesTool` / `SearchForPatternTool`.
- OS: Linux (container). The bug is in language-agnostic glob handling, so it is independent of OS/language server.

(Disclosure: this report was prepared with AI assistance; the reproduction above was run and verified against current HEAD.)

zerone0x:
## Summary
Fixes #1690.

Reject malformed brace globs instead of leaving `expand_braces()` in a non-terminating loop when it sees empty braces or unmatched braces.

## Changes
- Raise `ValueError` for empty `{}` and unmatched `{`/`}` brace expressions.
- Keep existing valid brace expansion behavior, including empty elements such as `{a,,b}`.
- Add regression coverage for malformed brace patterns.

## Test Plan
- `uv run --extra dev pytest test/serena/test_text_utils.py -q`
- `uv run --extra dev ruff check src/serena/util/text_utils.py test/serena/test_text_utils.py`
- `uv run --extra dev ruff format --check src/serena/util/text_utils.py test/serena/test_text_utils.py`

Co-Authored-By: OpenAI Codex <noreply@openai.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ Status of the `main` branch. Changes prior to the next official version change w
   - Fix: Use LSP-compliant line splitting ("\n", "\r\n" and "\r" can define line breaks); previously only "\n" considered
   - Fix: Apply consistent line splitting across tools, uniformly applying the LSP splitting semantics; 
     affects `search_text` used by `search_for_pattern` tool (reported in #1684)
+  - Fix: glob pattern expansion in `expand_braces` did not terminate with empty or unbalanced braces #1690
 
 * CLI:
   - Fix `--project-from-cwd` hijacking git worktrees nested under a Serena project. `find_project_root`
--- a/src/serena/util/text_utils.py
+++ b/src/serena/util/text_utils.py
@@ -216,18 +216,22 @@ def expand_braces(pattern: str) -> list[str]:
     Handles multiple brace sets as well.
     """
     patterns = [pattern]
-    while any("{" in p for p in patterns):
+    while any("{" in p or "}" in p for p in patterns):
         new_patterns = []
         for p in patterns:
-            match = re.search(r"\{([^{}]+)\}", p)
+            match = re.search(r"\{([^{}]*)\}", p)
             if match:
+                options_expr = match.group(1)
+                if not options_expr:
+                    raise ValueError(f"Invalid glob brace expression in {pattern!r}: empty braces are not allowed")
+
                 prefix = p[: match.start()]
                 suffix = p[match.end() :]
-                options = match.group(1).split(",")
+                options = options_expr.split(",")
                 for option in options:
                     new_patterns.append(f"{prefix}{option}{suffix}")
             else:
-                new_patterns.append(p)
+                raise ValueError(f"Invalid glob brace expression in {pattern!r}: unmatched brace")
         patterns = new_patterns
     return patterns
 
--- a/test/serena/test_text_utils.py
+++ b/test/serena/test_text_utils.py
@@ -606,6 +606,21 @@ class TestExpandBraces:
 
         assert sorted(expand_braces(pattern)) == sorted(expected)
 
+    @pytest.mark.parametrize(
+        "pattern",
+        [
+            "src/{}.py",
+            "src/{utils,models",
+            "src/utils,models}.py",
+        ],
+    )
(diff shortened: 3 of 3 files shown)


MischaPanch:
Thanks, pls add a short entry to the changelog, then this can be merged

PR #1654, opened 2026-07-05
Svelte test fixture: remove unused demo dependencies

xormania:
Removes SvelteKit demo-template dependencies that no test exercises: `@fontsource/fira-mono`, `@neoconfetti/svelte`, `@tailwindcss/vite`, and `tailwindcss`, along with their only usages (the `layout.css` imports, the sverdle page's confetti effect, and the tailwind vite plugin).

## Why

- Smaller, more maintainable fixture: a regenerated lockfile drops from 119 to 91 package entries (−24%).
- Less dependency surface for the fixture bootstrap — fewer packages that can break `npm install` / `svelte-kit sync`.
- `rolldown` / `lightningcss` remain: they are vite's own dependencies, not tailwind's.

## Test safety

- Tests reference the sverdle page only as an expected member of reference/rename result sets (no position queries into that file), and its `$lib` imports (`game.ts`, `Words.svelte`) — which put it in those result sets — are untouched.
- Verified locally: clean `npm install` + `svelte-kit sync` + full svelte suite (16/16) against the trimmed fixture.

## Sequencing (why this is a draft)

#1648 commits a fixture `package-lock.json` generated from the previous dependency set; whichever change lands second must regenerate it. This PR stays a draft until #1648 merges — then it will be rebased, the regenerated (trimmed) lockfile committed here, and marked ready for review.

## Checklist

- [ ] This PR follows the guidelines in `CONTRIBUTING.md` regarding the scope of PRs.
- [ ] For changes that add features or fix problems, I have added an entry to `CHANGELOG.md`, which concisely describes the change. *(Not applicable: test-fixture-only change with no user-facing behavior — no entry added.)*

PR #1698, opened 2026-07-15
CI: cache npm dependencies for web test fixtures

xormania:
## Problem

The Angular test repo's `npm install` is the only setup step in `pytest.yml` without a cache, and the Svelte fixture runs `npm ci` at test time — so every catch-all job pulls from the npm registry, putting it on the CI critical path (same failure class as the GH-download rate limits already worked around in CI). Both fixtures have committed lockfiles, so both installs are exactly reproducible.

## Fix

- Cache the Angular fixture's `node_modules` keyed on `runner.os` + lockfile hash (same shape as the uv-venv cache); skip the install step on an exact hit (same gating as go-binaries / R-packages). No restore-keys: a hit is byte-identical to what the install would produce, any lockfile change misses and reinstalls.
- Enable setup-node's npm download cache for catch-all only (the conditional renders an empty string in other batches, which disables it — they run no npm installs). This warms the Svelte fixture's test-time `npm ci` and the elm/web-LS installs.
- CACHING section of the header comment updated.

## Measured (rehearsed via a fork-internal PR: [cold run + warm re-run of the same merge ref](https://github.com/xormania/serena/actions/runs/29422420682))

- Cold: caches saved on all 3 OSes (`angular-npm-*` 8–9 MB, npm cache 42–46 MB; ~160 MB total against the 10 GB budget).
- Warm: `Cache restored` on both caches on all 3 OSes; Angular install step skipped.
- Honest sizing: the Angular install itself is only 2–8 s (the fixture is small since #1654), so the win is removing the registry dependency from every catch-all job, not wall-clock.

No changelog entry — CI-workflow-only, matching prior `CI:` commits.

## Checklist

- [x] For changes that add features or fix problems, I have added an entry to `CHANGELOG.md`, which concisely describes the change. *(n/a — CI-only workflow change.)*


xormania:
The `catch-all (windows-latest)` failure is unrelated to this change — evidence:

- **Failing test:** `test_component_class_diagnostics[angular]` — the diagnostics pull returned `[]` (`AssertionError: []`).
- **This run couldn't have been affected by the caching:** it was a cold miss (`Cache not found for input keys: angular-npm-Windows-…` in the job log), so `npm install` ran the pre-existing path. The other Angular tests in the same job passed, so the fixture was healthy.
- **Flake history:** 27/27 passes for this test across the last 12 runs of this workflow — a rare race, not a regression.

Root cause looks like the same class as the Solidity fix in #1675 (empty diagnostics on slow CI): the Angular companion TS path has no analysis-completion gate before pull diagnostics, so on a slow runner the request can race the analyzer and win an empty list. Happy to file that separately.

xormania:
@MischaPanch ready for review — CI is green (the windows flake discussed above didn't recur on the fresh run).

MischaPanch:
@xormania sorry for the delayed review, there was a lot going on and CI optimizations were downprioritized.

Looks good, but there's a small conflict and since I don't have permissions to push on your fork, I can't fix it. Could you resolve it pls?

xormania:
@MischaPanch 

I understand - in fact, the last week or so, I've been so busy - I haven't had time to check any of my PRs or messages. I finally did an hour ago and the timing is perfect. I've learned some cool things this last week. Anyway - I updated PR - sorry about maintainer edit permissions - i'll avoid from now on :-)

PR #1699, opened 2026-07-15
fix(tools): make read_file report the line numbers the editing tools use

ebarkhordar:
## Problem

`delete_lines` and `replace_lines` both document that the same range of lines must have been read with `read_file` first, so the line indices `read_file` reports have to be the ones the editing tools act on.

They are not always the same. `ReadFileTool.apply` decomposes the file with Python's `splitlines`, which treats `\f`, `\v`, `\x1c`, `\x1d`, `\x1e`, `\x85`, U+2028 and U+2029 as line breaks in addition to `\n`, `\r\n` and `\r`. The editing path (`code_editor.delete_lines` -> `TextUtils` / `TextStepper`) counts only the LSP-compliant line breaks, per #1691.

So for a file containing any of those characters, the two disagree, and every line after the character is shifted. A form feed is the realistic case: it is a page-break convention that appears inside source files. For `"line0\n\x0cline1\nline2\n"`:

- `read_file` splits it into `line0`, `''`, `line1`, `line2`, so it reports index 2 as `line1`
- `delete_lines(2, 2)` removes `line2`

The agent reads a line, asks to delete that index, and a different line is deleted. There is no error, so the corruption is silent.

## Root cause

`src/serena/tools/file_tools.py:46` used `result.splitlines()`, which is exactly the usage #1691 replaced elsewhere ("Add util method `TextUtils.split_lines`, allowing usages of Python's `splitlines` to be replaced"). `read_file` was the remaining content-splitting `splitlines` in the tool layer.

## Fix

Use `TextUtils.split_lines`, the same splitter the editing tools count with. Since `split_lines` yields a trailing empty line for newline-terminated files and `splitlines` does not, the trailing empty line is dropped so `read_file`'s output is unchanged for files that do not contain the exotic separators.

The invariant: the line index `read_file` reports refers to the same physical line that `delete_lines`/`replace_lines` operate on.

## Verification

Run in a clean `python:3.11-slim` container against this branch (module provenance asserted via `__file__` -> `/build/src/serena/tools/file_tools.py`):

```
python -m pytest test/serena/test_file_tools.py -q
```

- on this branch: `20 passed`
- with only `file_tools.py` reverted to `main` and the same new test: `6 failed, 14 passed`, failing exactly the `\f`, `\v`, `\x1c`, `\x85` and U+2028 cases

The 14 that pass on both are the no-change cases: for LF, CRLF, lone-CR, no-trailing-newline and empty files the new decomposition is identical to `splitlines`, which is what keeps this from changing `read_file` for ordinary files.

Also run: `test/serena/test_text_utils.py`, `test/solidlsp/test_text_utils.py`, `test/serena/test_edit_marker.py` (109 passed together with the new test), `ruff format --check` and `ruff check` clean on both changed files, and `ty check` reports the same 39 pre-existing diagnostics on this branch as on `main`.

What I did not verify: no end-to-end run through a live MCP client or a language server, since the change is confined to the line decomposition in `read_file`. The four `ruff check` errors under `test/solidlsp/java` and `test/solidlsp/test_typescript_timeout_policy.py` are pre-existing on `main` with the ruff version I used and are untouched here.

I have no evidence of this being hit in the wild; I found it by tracing the read/edit contract after #1691, so treat the severity as theory-plus-repro rather than a reported incident.

This contribution was prepared with AI assistance. The bug, the fix and the tests were verified as described above.

### DIFF
--- a/src/serena/tools/file_tools.py
+++ b/src/serena/tools/file_tools.py
@@ -21,6 +21,7 @@ from serena.util.text_utils import (
     expand_braces,
     glob_match,
 )
+from solidlsp.ls_utils import TextUtils
 
 
 class ReadFileTool(Tool):
@@ -42,8 +43,10 @@ class ReadFileTool(Tool):
         """
         self.project.validate_relative_path(relative_path, require_not_ignored=True)
 
+        # read lines, using the same (LSP-compliant) notion of line breaks as the line-based editing tools
         result = self.project.read_file(relative_path)
-        result_lines = result.splitlines()
+        result_lines = TextUtils.split_lines(result)
+
         if end_line is None:
             result_lines = result_lines[start_line:]
         else:
--- a/test/serena/test_file_tools.py
+++ b/test/serena/test_file_tools.py
@@ -1,0 +1,77 @@
+from pathlib import Path
+from unittest.mock import MagicMock
+
+import pytest
+
+from serena.config.serena_config import SerenaConfig
+from serena.constants import DEFAULT_SOURCE_FILE_ENCODING
+from serena.project import Project
+from serena.tools import ReadFileTool
+from solidlsp.ls_utils import TextUtils
+
+
+@pytest.fixture
+def read_file_tool(tmp_path: Path) -> ReadFileTool:
+    project = Project.load(str(tmp_path), serena_config=SerenaConfig(gui_log_window=False, web_dashboard=False))
+    agent = MagicMock()
+    agent.get_active_project_or_raise.return_value = project
+    tool = ReadFileTool(agent)
+    # bypass the length limit, which would otherwise depend on the agent configuration
+    tool._limit_length = lambda result, max_answer_chars: result  # type: ignore[method-assign]
+    return tool
+
+
+def _deleted_by_delete_lines(content: str, line: int) -> str:
+    """
+    :return: the text that `delete_lines(line, line)` removes from the given content, which is
+        computed via the same primitive the tool uses (`code_editor.delete_lines` deletes the text
+        between (line, 0) and (line + 1, 0)).
+    """
+    _, deleted_text = TextUtils.delete_text_between_positions(content, start_line=line, start_col=0, end_line=line + 1, end_col=0)
+    return deleted_text
+
+
+class TestReadFileToolLineNumbering:
+    """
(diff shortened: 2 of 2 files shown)


ebarkhordar:
The `catch-all (ubuntu-latest)` failure looks unrelated to this PR. It is `test/serena/test_hooks.py::TestHookCli::test_remind_command`, which does `json.loads(result.output)` and hits "Extra data: line 2": a dashboard news-fetch log line ("Remote news updated from https://oraios-software.de/serena_news.json") lands on stdout inside the captured CLI output, so the output is no longer just JSON.

This branch only touches `file_tools.py` and its test, the same job is green on main, and `catch-all (windows-latest)` and `catch-all (macos-latest)` both pass here. I do not have re-run rights on the repo, so it needs a re-run whenever suits you.

src/serena/tools/file_tools.py
```diff
@@ -43,7 +44,13 @@ def apply(self, relative_path: str, start_line: int = 0, end_line: int | None =
...
 
         result = self.project.read_file(relative_path)
-        result_lines = result.splitlines()
+        # NOTE: Must use the same (LSP-compliant) notion of line breaks as the line-based editing tools;
+        # Python's `splitlines` would additionally break on \f, \v, \x1c-\x1e, \x85, U+2028 and U+2029,
+        # yielding line indices that do not match the ones `delete_lines`/`replace_lines` operate on.
+        result_lines = TextUtils.split_lines(result)
+        # drop the empty line induced by a trailing line break (not reported as a line by `read_file`)
+        if result_lines and result_lines[-1] == "":
+            result_lines.pop()
```
opcode81, line 53: FYI I removed this in the updated version. The empty line should be part of the result

opcode81:
Thanks for your PR; I [updated it](https://github.com/oraios/serena/compare/e8c260cce388083f62d717133865edffe40e2718..fe91abd407bad6e919a39399640db16efd6cc06a) and will merge it once the tests confirm that everything's OK.
I had searched for `splitlines` when implementing the other PR, but somehow I must have missed that one.

ebarkhordar:
Thanks for the fast turnaround, and for picking it up directly. Good call dropping the trailing-empty-line pop: I had kept it so that full-file reads stayed byte-identical to the old `splitlines` output, but leaving `split_lines` intact is the cleaner model and makes the split/rejoin round-trip exact.

PR #1701, opened 2026-07-16
fix(ls): clamp out-of-bounds symbol ranges before constructing SymbolBody

### LINKED ISSUE #1498: GDScript/Godot LSP: clamp out-of-bounds symbol ranges and harden Windows TCP reader
## Summary

While setting up Serena's unreleased GDScript/Godot LSP support on a Windows Godot 4.6 project, I hit two issues that required local patches before `serena project health-check` could complete reliably:

1. Godot can return `textDocument/documentSymbol` ranges whose end line is past EOF, which can crash Serena when extracting a `SymbolBody`.
2. `TCPLanguageServer` currently reads the TCP socket through `socket.makefile("rb")` + `readline()` / `read()`, which was unstable during shutdown on Windows in this setup.

I did not find an existing issue for either exact failure mode. Related context: GDScript support landed in #1475, and #1493 covers a related broken-connection/reconnect case, but it does not appear to clamp symbol body ranges or replace the buffered socket reader.

## Environment

- Serena: `1.3.1.dev0`, installed from commit `7c7d5eef56b47d8ab0835dd9227ba4b80b90c50b`
- OS: Windows
- Godot: `4.6.2-stable`
- Project language: GDScript
- Godot editor LSP: observed listening on `127.0.0.1:6005`
- Client context: Codex MCP / `serena project health-check`

## Issue 1: Godot document symbol range can end past EOF

In `src/solidlsp/ls.py`, `SymbolBodyFactory.create_symbol_body()` currently passes the LSP range directly into `SymbolBody`:

```py
start_line = symbol["location"]["range"]["start"]["line"]
end_line = symbol["location"]["range"]["end"]["line"]
start_col = symbol["location"]["range"]["start"]["character"]
end_col = symbol["location"]["range"]["end"]["character"]
return SymbolBody(self._lines, start_line, start_col, end_line, end_col)
```

For at least one GDScript class symbol returned by Godot, the reported range ended beyond the file's final line. Later, `SymbolBody.get_text()` indexes `self._lines[self._end_line]`, which raised `IndexError`.

Local workaround that made the health check pass: clamp the start/end lines and columns to the available file buffer before constructing `SymbolBody`.

Expected behavior: Serena should tolerate LSP ranges that are slightly outside the local file buffer, especially for language servers that report whole-class ranges imperfectly.

## Issue 2: Windows shutdown instability with `socket.makefile()` reader

In `src/solidlsp/ls_process.py`, `TCPLanguageServer` uses a socket-backed buffered reader:

```py
self._file = sock.makefile("rb")
...
line = f.readline()
...
body = f.read(num_bytes)
```

On Windows, this reader could hang/crash during shutdown in the Godot TCP LSP setup, even with the existing "close socket before closing BufferedReader" ordering.

Local workaround that made shutdown reliable: avoid `socket.makefile()` for TCP reads and use a socket byte-buffer parser around `recv()`, with a short socket timeout so stop/shutdown can be observed cleanly.

Expected behavior: `TCPLanguageServer` should shut down cleanly on Windows when connected to Godot's editor-hosted LSP.

## Validation after local patches

After applying the two local patches, `serena project health-check` completed successfully against the Godot project. It connected to the Godot LSP, exercised symbol overview, symbol lookup, referencing-symbol lookup, and pattern search, then exited cleanly.

## Suggested next step

Happy to raise a PR. I suspect these may be best as two small changes:

1. Defensive range clamping in `SymbolBodyFactory` / `SymbolBody` creation.
2. A Windows-safe TCP read loop for `TCPLanguageServer`, coordinated with #1493 if that lands first.

AmirF194:
## Root cause

`SymbolBody.get_text` (`src/solidlsp/ls.py:219`) indexes the line buffer directly:

```python
last_line = self._lines[self._end_line]
```

An LSP server can report a symbol whose range ends past the end of the file. The common convention for a range that covers whole lines ends it at the start of the *following* line, which for the last line of a file is one line past EOF. `SymbolBodyFactory.create_symbol_body` passed the raw range endpoints straight into `SymbolBody` with no bounds check, so `get_text` raised `IndexError: list index out of range` on such a symbol. That is issue #1498 (GDScript/Godot surfaced it, but the fault is in generic range handling, not anything language-specific).

## Fix

`create_symbol_body` now clamps each range endpoint into the buffer before constructing the `SymbolBody`. A position past EOF is pulled to the end of the last line, so `get_text` returns the symbol through the end of the file instead of crashing. Clamping only the line and leaving `character` at 0 would instead silently drop the last line, because the whole-line convention puts the end at column 0 of the next line, so the column is clamped as well. The clamp is a no-op for in-bounds ranges.

Invariant: the `(line, character)` endpoints a `SymbolBody` is built with must address existing positions in its line buffer.

## Verification

Clean `python:3.11-slim` container, `pip install -e .`, HEAD `465c5eb`, `__file__` provenance confirmed (`solidlsp.ls -> /repo/src/solidlsp/ls.py`):

- New unit test `test/solidlsp/test_symbol_body.py` drives the real `SymbolBodyFactory` plus `get_text` with no language server and no Godot. Against unmodified `main` the two past-EOF cases FAIL with `IndexError` at `ls.py:219`; against this branch all four cases pass.
- `ruff format --check` and `ruff check` are clean on both changed files, and `ty check src/solidlsp` plus the new test file are clean (this mirrors `poe lint` and `poe type-check`).

I verified only the generic range path above. I did not run the Godot language server itself, so I cannot confirm whether Godot also emits an out-of-range `end.character` on an otherwise valid line; the clamp covers that case, but that specific trigger is unverified. The second item in #1498 (the Windows TCP reader) needs a Windows plus Godot setup and is left out of this PR, as @thoraxe suggested splitting it.

## Checklist

Fixes #1498

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ Status of the `main` branch. Changes prior to the next official version change w
     reached Serena's in-memory file contents, where the rest of the code assumes LF-normalized text and
     the `line_ending` setting is meant to be the single point of line-ending translation on write. The
     fallback now normalizes line endings to LF, consistently with the primary path.
+  - Fix: a symbol whose LSP range ended exactly one line past EOF, at column 0 (the convention for
+    a range covering whole lines through the end of the file), raised `IndexError` in
+    `SymbolBody.get_text`. That one well-defined case is now corrected to end at the actual last
+    line; any other out-of-range end position now raises `InvalidTextLocationError` instead,
+    rather than guessing at a body that could be wrong #1498
 
 * Tools:
   - Fix: `search_for_pattern` marked one line too many as matched whenever a match ended with a line
--- a/src/solidlsp/ls.py
+++ b/src/solidlsp/ls.py
@@ -32,7 +32,7 @@ from solidlsp.dependency_provider import (
 )
 from solidlsp.initialize_params import DefaultInitializeParamsBuilder, InitializeParamsBuilder
 from solidlsp.ls_config import FilenameMatcher, Language, LanguageServerConfig
-from solidlsp.ls_exceptions import SolidLSPException
+from solidlsp.ls_exceptions import InvalidTextLocationError, SolidLSPException
 from solidlsp.ls_process import LanguageServerInterface, StdioLanguageServer
 from solidlsp.ls_types import UnifiedSymbolInformation
 from solidlsp.ls_utils import FileUtils, PathUtils, TextUtils
@@ -209,17 +209,38 @@ class SymbolBody(ToStringMixin):
         return ["_lines"]
 
     def get_text(self) -> str:
+        end_line = self._end_line
+        end_col = self._end_col
+        if end_line >= len(self._lines):
+            if end_line == len(self._lines) and end_col == 0:
+                # LSP convention: a range covering whole lines through EOF sometimes ends
+                # at the start of the following, non-existent line (exactly one line past
+                # the last valid index, at column 0). That is well-defined: it means
+                # "through EOF", so treat it as ending at the end of the actual last line.
+                end_line = len(self._lines) - 1
+                end_col = len(self._lines[end_line])
+            else:
+                # Any other out-of-range end position (further past EOF, or exactly one
+                # line past EOF but not at column 0) is not the well-defined convention
+                # above; applying the same correction there would silently assume that
+                # a column meant for a nonexistent line still applies to the corrected
+                # one, which can produce a garbage body. Reject it instead of guessing.
+                raise InvalidTextLocationError(
+                    f"Symbol range end (line {self._end_line}, col {self._end_col}) is out of bounds "
+                    f"for a file with {len(self._lines)} lines"
+                )
+
         # extract relevant lines
-        symbol_body = "\n".join(self._lines[self._start_line : self._end_line + 1])
+        symbol_body = "\n".join(self._lines[self._start_line : end_line + 1])
 
         # remove leading content from the first line
         symbol_body = symbol_body[self._start_col :]
 
         # remove trailing content from the last line
-        last_line = self._lines[self._end_line]
-        trailing_length = len(last_line) - self._end_col
+        last_line = self._lines[end_line]
+        trailing_length = len(last_line) - end_col
         if trailing_length > 0:
-            symbol_body = symbol_body[: -(len(last_line) - self._end_col)]
+            symbol_body = symbol_body[: -(len(last_line) - end_col)]
 
         return symbol_body
 
--- a/src/solidlsp/ls_exceptions.py
+++ b/src/solidlsp/ls_exceptions.py
@@ -52,6 +52,15 @@ class SolidLSPException(Exception):
         return s
 
 
+class InvalidTextLocationError(SolidLSPException):
+    """
+    Raised when a symbol's LSP range refers to a text location that does not exist
+    in the file's current line buffer, other than the well-defined whole-line-through-EOF
+    convention (end line exactly one past EOF at column 0), which is corrected rather
+    than rejected.
+    """
+
+
 class MetalsStaleLockError(SolidLSPException):
     """
     Raised when a stale Metals H2 database lock is detected and the user
--- a/src/solidlsp/ls_utils.py
+++ b/src/solidlsp/ls_utils.py
@@ -21,14 +21,10 @@ from urllib.parse import urlparse
 import charset_normalizer
 import requests
 
-from solidlsp.ls_exceptions import SolidLSPException
+from solidlsp.ls_exceptions import InvalidTextLocationError, SolidLSPException
 from solidlsp.ls_types import UnifiedSymbolInformation
 
 log = logging.getLogger(__name__)
-
-
-class InvalidTextLocationError(Exception):
-    pass
 
 
 class TextStepper:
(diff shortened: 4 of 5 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ Status of the `main` branch. Changes prior to the next official version change w
 * General:
   - Speed up MCP startup when auto-creating projects at startup (e.g. using `--project-from-cwd`) by
     determining the project's languages in a background thread #1683
+  - Fix: a symbol whose LSP range ended past the end of the file (e.g. a range covering whole lines,
```
MischaPanch: Pls merge or rebase on main and move this into the upper section of the changelog (we released 1.6.0) yesterday

MischaPanch (changes requested):
LGTM except the minor comment.

Though we never had reports on that, do some LS actually return lines/cols that are out of range? It doesn't hurt to harden this, @opcode81 do you agree?

AmirF194:
Rebased onto main and moved the changelog line into the new "Unreleased (main)" section; v1.6.0 had already been cut when I opened this, so the rebase's auto-merge had carried it into that frozen section by content match. Reran the full suite in test/solidlsp/test_symbol_body.py and ruff check/format on the changed files, both clean.

On your question: yes, per the original report in #1498, maitlandmarshall saw Godot's GDScript LSP return a documentSymbol range whose end.line sat past EOF on a real Windows project, that is what crashed serena project health-check for them. My repro isolates the generic range-handling bug without needing Godot, but the real trigger observed in the wild is Godot's LSP for end.line specifically. I have not confirmed an out-of-range end.character in practice, so that part of the clamp is defensive rather than independently observed.

src/solidlsp/ls.py
```diff
@@ -240,11 +240,25 @@ def create_symbol_body(self, symbol: UnifiedSymbolInformation) -> SymbolBody:
...
+            if line < 0:
+                return 0, 0
+            if line > len(lines) - 1:
+                last = len(lines) - 1
+                return last, len(lines[last])
+            return line, min(max(col, 0), len(lines[line]))
+
+        start_line, start_col = clamp(rng["start"]["line"], rng["start"]["character"])
+        end_line, end_col = clamp(rng["end"]["line"], rng["end"]["character"])
+        return SymbolBody(lines, start_line, start_col, end_line, end_col)
```
opcode81: This is problematic for two reasons:
1. The clamping makes the invalid assumption that even if a line is off, the column being specified still applies to the "corrected" line. The truth is that if the line that was specified does not actually exist, there is a fundamental problem. So if this ever actually occurs, this could very well produce garbage.
2. `SymbolBody` is a cached representation, so "fixing" this here rather than in `SymbolBody` itself does not take effect until the cache is invalidated.

The question is: Where has this actually occurred?

* If there is a fundamental problem where a language server produces coordinates that do not align with Serena's computations, then I would like to know about it and figure out who's at fault (Serena vs. LS) - and then fix it rather than to produce a potentially incorrect body.
* If there is actually only a more specific problem (e.g. LS assumes additional empty line at end of file), then I would like to solve that specifically rather than to apply a generic "fix" that doesn't produce well-defined results. (And probably solve it `SymbolBody`)

opcode81 (changes requested):

src/solidlsp/ls.py
opcode81: Having read the PR description, it seems that we have the second case, i.e. issue pertaining to what the EOF coordinate is only, so I would propose to solve that specifically. Assuming that it is only ever off by one line (additional line break at EOF assumed), this has a well-defined solution where the result is guaranteed to be exactly as intended.

AmirF194: Agreed on both points. Moved the correction into SymbolBody.get_text itself instead of the factory, so it applies to a cached body too, not just a freshly constructed one. And narrowed it to exactly the case you're asking about: end_line one line past EOF, the whole-line-range convention, corrected to end at the actual last line. Any other out-of-range end_line still raises now (added test_get_text_end_line_far_past_eof_still_raises to keep it that way), and start_line/column clamping is gone entirely since I have no confirmed case for those.

To your question: the only case I've actually seen is the reporter's Godot GDScript symbol, end.line exactly one past EOF. No evidence of an LS returning out-of-range columns or a start_line off by itself, so not guessing at those. Pushed to the same branch.

src/solidlsp/ls.py
```diff
@@ -209,17 +209,29 @@ def _tostring_excludes(self) -> list[str]:
         return ["_lines"]
 
     def get_text(self) -> str:
+        end_line = self._end_line
+        end_col = self._end_col
+        if end_line == len(self._lines):
```
opcode81: I would propose the `>=` condition here.
Then, inside the if block, differentiate between `end_line == len(self._lines) && end_col == 0` and other cases. The correction applies in the positive case, and the else case should raise `InvalidTextLocationError`.

AmirF194: Agreed, pushed. Split the condition as you proposed: `end_line >= len(self._lines)`, then inside that only `end_line == len(self._lines) and end_col == 0` gets the correction (the well-defined whole-line-through-EOF case), everything else raises. Added `InvalidTextLocationError` in `ls_exceptions.py` for that path and a new test (`test_get_text_end_line_past_eof_with_nonzero_col_raises`) so a nonzero column at EOF stays rejected rather than silently clamped. Verified in a clean container: 6/6 tests pass, ruff and codespell clean.

opcode81:
Thanks for your contribution.

AmirF194:
Thanks for merging, and for pushing on the out-of-range question, that back and forth made the fix more defensible.

Issue #1139, opened 2026-03-06
Configurable line ending convention for file writes


cek:
**Is your feature request related to a problem? Please describe.**
On Windows, file-writing tools (`replace_content`, `create_text_file`, `replace_symbol_body`, etc.) produce CRLF line endings because `CodeEditor._save_edited_file()` opens files in text mode using the platform-default line ending. Projects that use LF line endings on Windows end up with mixed line endings after edits.

**Describe the solution you'd like**
A `line_ending` option in `project.yml` (with possible values: `"lf"`, `"crlf"`, `"native"`; default `"native"` for backwards compatibility, or similar).

**Describe alternatives you've considered**
Git-based workarounds: `core.autocrlf=true` or `.gitattributes` with `* text=auto`. These work but create friction (staging noise, hook conflicts, inconsistent behavior across tools).

**Additional context**
I believe the changes required would be quite small in scope:
- `CodeEditor._save_edited_file()` (`code_editor.py:76`): pass `newline=` to `open()`
- `CreateTextFileTool.apply()` (`tools/file_tools.py:76`): pass `newline=` to `Path.write_text()`

opcode81:
So far, we have been recommending that people use `autocrlf` (or `.gitattributes`) on Windows, which works well enough for the majority of cases.

Your request is reasonable for some workflows though, where LF needs to be used throughout, but this topic is not a priority for us. Your assessment is probably accurate, i.e. you will probably not need to change many more places than the ones you've already mentioned. To add this feature, the setting should be added to `SharedConfig`, as this is something one might want to configure both globally and on a per-project level (override). So if you want to explore this and issue a PR, feel free to do so.

PR #1706, opened 2026-07-17
fix: normalize line endings in read_file's charset_normalizer fallback

ebarkhordar:
## Checklist

## Problem

`FileUtils.read_file` has two return paths with different newline semantics:

- **Primary** (`ls_utils.py:390`): `open(file_path, encoding=encoding).read()` uses the default `newline=None`, so universal newlines apply and CR never reaches callers.
- **Fallback** (`ls_utils.py:399`, added in #682): `match.raw.decode(match.encoding)` decodes raw bytes, which performs no newline translation.

So a file that fails to decode with the project's configured `encoding` re-enters Serena carrying its CR characters, while an otherwise identical file that decodes cleanly does not. The rest of the code assumes in-memory content is LF-normalized and that the `line_ending` setting (#1139) is the single point of line-ending translation on write. Both writers depend on that: `code_editor.py:89` (`open(..., newline=self.newline)`) and `file_tools.py:87` (`write_text(..., newline=...)`).

The path is reachable through normal editing: `ls.py:164` fills `LSPFileBuffer._contents` from `read_file`, `code_editor.py:265` returns that buffer as the edited file's contents, and `_save_edited_file` writes it back with `newline=self.newline`. Editing a symbol in a non-UTF-8 file therefore rewrites the whole file through a second translation. The affected population (legacy Windows/cp1252 codebases with mixed encodings) is the one that motivated both #682 and #1139.

## Fix

Normalize to LF at the fallback's return, so both paths agree. One line at one return path, no new helper and no change to encoding detection or to the write side.

## Verification

Run in a clean `python:3.11-slim` container against HEAD `063f872a`, with the imported module's sha256 checked against the repo file (provenance match confirmed):

```
python -m pytest test/solidlsp/util/test_ls_utils.py -q
```

- On this branch: `4 passed`.
- With only `src/solidlsp/ls_utils.py` reverted to main and the new tests kept: `2 failed, 2 passed`. The two failures are the new fallback tests; `test_read_file_primary_path_normalizes_crlf` passes on main by design, since it pins the primary path's existing behaviour that the fallback is being aligned to.
- `ruff check` and `ruff format --check` clean on both changed files.
- The fixture is deliberately a realistic cp1252 file rather than a few bytes, because encoding detection is unreliable on very short inputs, and the test asserts the fixture is undecodable as UTF-8 so it cannot silently stop exercising the fallback.

Measured in the same container before the fix, composing the fallback's output with the real writer call shape and the actual `LineEnding.*.newline_str` values: with `line_ending: lf` the 22 CRLF pairs survived unchanged (the setting had no effect), and with `line_ending: crlf` they became `\r\r\n`. Those two outcomes are what the fix prevents, but they are not pinned by a shipped test: the tests here pin the read-side normalization only.

Not verified: behaviour on Windows. `line_ending: native` resolves to `os.linesep`, so on Windows it should double identically to the `crlf` case, but this box is Linux (`os.linesep == '\n'`) and I did not test that.

Disclosure: this patch was prepared with AI assistance. The reproduction, the differential test run and the reachability trace above were executed, not inferred.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 Status of the `main` branch. Changes prior to the next official version change will appear here.
 
+* General:
+  - Fix: `FileUtils.read_file`'s `charset_normalizer` fallback (used when a file cannot be decoded with
+    the project's configured `encoding`) decoded the raw bytes directly and therefore skipped the
+    universal-newline translation that the primary read path applies. CR characters from disk thus
+    reached Serena's in-memory file contents, where the rest of the code assumes LF-normalized text and
+    the `line_ending` setting is meant to be the single point of line-ending translation on write. The
+    fallback now normalizes line endings to LF, consistently with the primary path.
+
 # v1.6.0 (2026-07-16)
 
 * General:
--- a/src/solidlsp/ls_utils.py
+++ b/src/solidlsp/ls_utils.py
@@ -380,6 +380,9 @@ class FileUtils:
         Reads the file at the given path using the given encoding and returns the contents as a string.
         If decoding fails, tries to detect the encoding using charset_normalizer.
 
+        Line endings are normalized to LF (universal newlines), irrespective of the encoding
+        used to decode the file.
+
         Raises FileNotFoundError if the file does not exist.
         """
         if not os.path.exists(file_path):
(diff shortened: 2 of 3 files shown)


ebarkhordar:
Pushed 7cc80f7 to fix the spelling check: the fixture comments used "Autor" and "Alle", which codespell flags as misspellings of "Author" and "All". Renamed the first and dropped the German sentence. The accented characters that make the bytes undecodable as UTF-8, and therefore force the fallback under test, are unchanged.

The `native (windows-latest)` failure is not from this branch. `test_find_symbol[rust_add_function]` fails the same way on main at 063f872, the base this PR merges into: same `textDocument/hover` request, same `content modified (-32801)`, 1 failed / 123 passed (run 29502472168). I have not investigated what causes it, so I cannot say more than that.

opcode81:
Thanks!

ebarkhordar:
Thanks for the quick merge. Unrelated but in the same file: line 213 raises `InvalidTextLocationError("{line=}, {col=}")` without the f-prefix, so it reports literal braces instead of the values, where line 189 gets it right. Happy to send a one-liner if that is useful.

opcode81:
> line 213

Thanks for the pointer. Fixed it.

PR #1711, opened 2026-07-18
PHP: expose file_filter via ls_specific_settings (#1710)

### LINKED ISSUE #1710: PHP: file extensions hardcoded to .php - Drupal sources (.module, .install, .inc, .theme, .profile, .engine) invisible to find_symbol

Serena hardcodes the PHP source filename filter to `.php`. In `Language.get_source_fn_matcher()` (`src/solidlsp/ls_config.py`):

```python
case self.PHP | self.PHP_PHPACTOR | self.PHP_PHPANTOM:
    return FilenameMatcher(".php")
```

This matcher gates which files the symbol machinery sees at all: index traversal, `find_symbol`, `get_symbols_overview`, reference finding. PHP ecosystems that keep ordinary PHP in other extensions lose symbol visibility for those files, and Drupal is the heaviest case: hook implementations live in `.module`, install/update hooks in `.install`, shared procedural code in `.inc`, theme preprocess logic in `.theme`, installation profiles in `.profile`, theme engines in `.engine`. All of these are plain PHP files that Intelephense handles fine once told about them.

Counts from two production Drupal codebases I work on:

| Extension | Site A | Site B |
|---|---|---|
| `.module` | 343 | 612 |
| `.install` | 134 | 254 |
| `.inc` | 161 | 226 |
| `.theme` | 17 | 22 |
| `.profile` | 3 | 3 |
| `.engine` | 5 | 3 |

That is 660 to 1,100+ PHP files per project invisible to `find_symbol` / `find_referencing_symbols`, and they are disproportionately important ones: hooks are where Drupal's control flow lives.

A smaller pre-existing inconsistency in the same area: Intelephense's default `intelephense.files.associations` is `["*.php", "*.phtml"]`, so the language server indexes `.phtml` templates while Serena's matcher hides them from the symbol tools.

v1.6.0 introduced exactly the right mechanism in #1449 / #1642 (`FilenameMatcher.add_extensions()` / `reset()`, a generic matcher reset on every LS activation in `SolidLanguageServer.__init__`, a `file_filter` key under `ls_specific_settings`), but it is currently wired up only for Perl.

Mirror the Perl wiring (#1642) for PHP / Intelephense, defaults unchanged:

```yaml
# serena_config.yml
ls_specific_settings:
  php:
    file_filter: [".php", ".phtml", ".module", ".install", ".inc", ".theme", ".profile", ".engine"]
```

1. `intelephense.py`: module-level `_DEFAULT_FILE_FILTER = [".php"]`; a pure resolver reading `file_filter` from `ls_specific_settings["php"]` (unit-testable without a language server, like `PerlLanguageServer._resolve_filter_settings`); sync the configured extensions into `Language.PHP.get_source_fn_matcher()` via `add_extensions()`. The generic reset in `SolidLanguageServer.__init__` already prevents cross-project leakage, so nothing to do in `ls.py`.
2. A second, Intelephense-specific layer is equally required: extending Serena's matcher alone does not make the LS index the files, because Intelephense only indexes what matches its `files.associations`. The configured extensions therefore also need to be pushed to the server as globs, e.g. `{"settings": {"intelephense": {"files": {"associations": ["*.php", "*.phtml", "*.module", ...]}}}}` via `workspace/didChangeConfiguration` after `initialized` (Serena currently sends Intelephense no `didChangeConfiguration` at all; it only declares the client capability).
3. A `file_filter` row in the PHP (`Intelephense`) settings table in `docs/02-usage/050_configuration.md` (plus the same two notes the Perl section has), a CHANGELOG entry, and tests mirroring `test/solidlsp/perl/test_perl_config.py`: defaults preserved when the setting is absent, no cross-language leakage, matcher sync makes `.module` files visible, reset undoes the extension. On top of those runtime-free unit tests, a small Drupal-style `.module` file in `test/resources/repos/php/test_repo` with a symbol-visibility assertion in the existing marker-gated PHP suite.

Open questions where maintainer input would help before I open the PR:

1. Scope: wire only Intelephense (the default PHP backend), or `php_phpactor` / `php_phpantom` as well? The three enum members share one `FilenameMatcher(".php")` construction, but `get_source_fn_matcher()` is `@cache`d per member, so each backend needs its own sync and would read its own settings key. I would start with Intelephense only unless you prefer all three at once.
2. Should `.phtml` join the default `file_filter` to match Intelephense's own default associations? It would resolve the inconsistency above at the cost of a small default-behaviour change; I would leave it opt-in unless you say otherwise.
3. Intelephense persists its index across runs; after `files.associations` changes, a stale cache could keep newly associated files invisible. If that turns out to be the case, the implementation could set `clearCache` in `initializationOptions` when the effective filter differs from the previous one. I would investigate as part of the PR unless you already know whether it is needed.

- The generic `initializationOptions` passthrough under `ls_specific_settings` (also new in v1.6.0) can feed custom options to Intelephense, but even if the LS could be persuaded to index extra extensions that way, Serena's source matcher would still filter those files out of `find_symbol` / `get_symbols_overview`. A Serena-side `file_filter` is needed either way, and it should drive both layers so they cannot disagree.
- An `ignore_dirs` companion setting (which #1642 added for Perl) is deliberately not proposed: the Intelephense wrapper already has `ignore_vendor`, hardcodes `node_modules` / `cache`, and project-level ignores cover the rest.
- Renaming Drupal files to `.php` is not an option; the extensions are load-bearing for Drupal's module system.

**Additional context**

I am happy to submit the PR myself, modelled closely on #1642 (implementation, runtime-free config tests, fixture-based visibility test, docs, CHANGELOG), and to validate against real Drupal codebases.

vboychev:
Closes #1710

## What & why

Serena hardcodes the PHP source filename filter to `.php` (the PHP arm of `Language.get_source_fn_matcher()`), so ecosystems that keep ordinary PHP in other extensions lose symbol visibility for those files. Drupal is the heaviest case: hook implementations live in `.module`, install/update hooks in `.install`, plus `.inc` / `.theme` / `.profile` / `.engine` – 660 to 1,100+ files per project on the two production sites measured in #1710, and disproportionately important ones, since hooks are where Drupal's control flow lives.

v1.6.0 introduced the right mechanism in #1449 / #1642 (`FilenameMatcher.add_extensions()` / `reset()`, the generic matcher reset on every LS activation, a `file_filter` key under `ls_specific_settings`), but wired it only for Perl. This PR mirrors the accepted pattern for PHP / Intelephense.

## Change

`ls_specific_settings["php"]` accepts `file_filter`; defaults are unchanged (`[".php"]`) when the key is omitted:

```yaml
ls_specific_settings:
  php:
    file_filter: [".php", ".phtml", ".module", ".install", ".inc", ".theme", ".profile", ".engine"]
```

- `Intelephense._resolve_file_filter` (pure, unit-testable, mirroring `PerlLanguageServer._resolve_filter_settings`) reads the key; `_sync_source_fn_matcher` adds the extensions to `Language.PHP.get_source_fn_matcher()`. The generic reset in `SolidLanguageServer.__init__` already prevents cross-project leakage, so `ls.py` is untouched.
- Second, Intelephense-specific layer: a customized filter is also pushed to the server as `intelephense.files.associations` globs via `workspace/didChangeConfiguration` (Serena previously sent none), because the server only indexes what matches its associations; without the push, cross-file references and definitions skip the extra files. Unlike Perl's unconditional push, this one is deliberately skipped for the default filter: the pushed list replaces Intelephense's built-in `["*.php", "*.phtml"]`, so deriving it from `[".php"]` would silently drop `.phtml` indexing. The decision lives in the pure `_resolve_files_associations` helper (unit-tested), and user-configured `maxMemory` / `maxFileSize` are carried in the pushed section so it stays self-consistent regardless of the server's merge semantics.
- Docs: `file_filter` row, Drupal example, and the same two notes the Perl section has; CHANGELOG entry under "Language Servers".

Scope decisions, following the open questions in #1710: Intelephense only for now (`php_phpactor` / `php_phpantom` have separate cached matchers and would each need their own wiring – happy to follow up if wanted); `.phtml` stays opt-in rather than joining the default; `clearCache` turned out unnecessary for the fresh-workspace case the integration test exercises and is deferred pending maintainer input.

## Verification

- `test/solidlsp/php/test_php_config.py`: 11 runtime-free unit tests mirroring `test/solidlsp/perl/test_perl_config.py` (defaults preserved, no cross-language / cross-backend leakage, matcher sync, reset-prevents-leak, push decision), plus one live integration test (gated by the `php` marker) against a Drupal-style `drupal_module.module` fixture. The integration test pins both layers: cross-file references from `helper.php` land in the never-opened `.module` file (so the reference can only come from the association-driven background index), and the full symbol tree contains its symbols (matcher sync). The fixture is inert for every test running with default settings.
- Local runs on macOS / Python 3.13: `poe format`, `poe type-check`, the config tests (twice, checking the reference assertion is stable), and the full `php`-marked suite: 40 passed, 1 skipped (Phpactor unavailable locally), no regressions.

## Checklist

### DIFF
--- a/src/solidlsp/language_servers/intelephense.py
+++ b/src/solidlsp/language_servers/intelephense.py
@@ -10,7 +10,7 @@ from time import sleep
 from overrides import override
 
 from solidlsp.ls import LanguageServerDependencyProvider, LanguageServerDependencyProviderSinglePath, SolidLanguageServer
-from solidlsp.ls_config import Language, LanguageServerConfig
+from solidlsp.ls_config import LanguageServerConfig
 from solidlsp.ls_utils import PlatformId, PlatformUtils
 from solidlsp.lsp_protocol_handler.lsp_types import Definition, DefinitionParams, DidChangeConfigurationParams, LocationLink
 from solidlsp.settings import SolidLSPSettings
@@ -26,13 +26,6 @@ log = logging.getLogger(__name__)
 INITIAL_INTELEPHENSE_VERSION = "1.14.4"
 DEFAULT_INTELEPHENSE_VERSION = "1.14.4"
 
-# Default file extensions treated as PHP sources. Exposed for override via
-# `ls_specific_settings["php"]` (key `file_filter`). Extensions added via `file_filter` are also
-# synced into Language.PHP.get_source_fn_matcher() (a cached singleton) and, when customized,
-# pushed to Intelephense as `intelephense.files.associations` globs, so Serena's symbol index and
-# the language server agree on which files are PHP sources (#1710).
-_DEFAULT_FILE_FILTER: list[str] = [".php"]
-
 
 class Intelephense(SolidLanguageServer):
     """
@@ -42,11 +35,11 @@ class Intelephense(SolidLanguageServer):
         - maxMemory: sets intelephense.maxMemory
         - maxFileSize: sets intelephense.files.maxSize
         - ignore_vendor: whether or ignore directories named "vendor" (default: true)
-        - file_filter: list of file extensions (with leading dot) to treat as PHP sources,
-          e.g. [".php", ".phtml", ".module"]. Defaults to [".php"]. The extensions are added to
-          Serena's PHP source-file matcher and a custom list is pushed to Intelephense as
-          intelephense.files.associations globs, so find_symbol and the language server treat
-          the same files as PHP sources (see #1710).
+        - file_filter: list of additional file extensions (with leading dot) to treat as PHP
+          sources, e.g. [".module", ".inc"]. The extensions are added to the defaults (".php",
+          ".phtml") of Serena's PHP source-file matcher and are pushed to Intelephense via
+          files.associations, such that the symbol tools and the language server treat the
+          same files as PHP sources (see #1710).
     """
 
     @override
@@ -105,42 +98,6 @@ class Intelephense(SolidLanguageServer):
         def _create_launch_command(self, core_path: str) -> list[str]:
             return [core_path, "--stdio"]
 
-    @staticmethod
-    def _resolve_file_filter(solidlsp_settings: SolidLSPSettings) -> list[str]:
-        """Resolve the file extensions to treat as PHP sources.
-
-        Reads an optional override from ``ls_specific_settings["php"]`` (key ``file_filter``);
-        falls back to the default otherwise. Extracted as a pure function so the configuration
-        plumbing can be unit-tested without starting the language server.
-        """
-        php_settings = solidlsp_settings.get_ls_specific_settings(Language.PHP)
-        return php_settings.get("file_filter", list(_DEFAULT_FILE_FILTER))
-
-    @staticmethod
-    def _sync_source_fn_matcher(file_filter: list[str]) -> None:
-        """Keep Serena's PHP source-file matcher in sync with the configured ``file_filter``.
-
-        ``Language.PHP.get_source_fn_matcher()`` is a ``@cache``d per-language singleton, so adding
-        the configured extensions here propagates to every consumer of the matcher - symbol index
-        traversal, ignore checks, and language composition detection. Without this, ``find_symbol``
-        would not surface symbols in files whose extensions were added to ``file_filter`` (#1710).
-        """
-        Language.PHP.get_source_fn_matcher().add_extensions(*file_filter)
-
-    @staticmethod
-    def _resolve_files_associations(file_filter: list[str]) -> list[str] | None:
-        """The ``intelephense.files.associations`` globs to push for ``file_filter``, or ``None``.
-
-        Returns ``None`` for the default filter: the pushed list replaces Intelephense's built-in
-        associations (``["*.php", "*.phtml"]``), so pushing globs derived from ``[".php"]`` would
-        silently drop ``*.phtml`` indexing and change default behavior (unlike Perl::LanguageServer,
-        whose defaults match Serena's and allow an unconditional push). A customized filter is
-        pushed and fully defines what the server indexes (#1710).
-        """
-        if file_filter == _DEFAULT_FILE_FILTER:
-            return None
-        return [f"*{ext}" for ext in file_filter]
-
     def __init__(self, config: LanguageServerConfig, repository_root_path: str, solidlsp_settings: SolidLSPSettings):
         super().__init__(config, repository_root_path, None, "php", solidlsp_settings)
         self.request_id = 0
@@ -154,10 +111,11 @@ class Intelephense(SolidLanguageServer):
             self._ignored_dirnames.add("vendor")
         log.info(f"Ignoring the following directories for PHP projects: {', '.join(sorted(self._ignored_dirnames))}")
 
-        self._file_filter = self._resolve_file_filter(solidlsp_settings)
-        # Sync Serena's source-file matcher with the configured extensions so find_symbol and the
-        # language server agree on which files are PHP sources (see #1710).
-        self._sync_source_fn_matcher(self._file_filter)
+        # extend the source-file matcher with any additional extensions configured via `file_filter`,
+        # such that Serena's symbol tools treat the added extensions as PHP sources as well (#1710)
+        file_filter = self._custom_settings.get("file_filter")
+        if file_filter:
+            self.language.get_source_fn_matcher().add_extensions(*file_filter)
 
     def _create_dependency_provider(self) -> LanguageServerDependencyProvider:
         return self.DependencyProvider(self._custom_settings, self._ls_resources_dir)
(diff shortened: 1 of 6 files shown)


src/solidlsp/language_servers/intelephense.py
```diff
@@ -184,6 +237,25 @@ def do_nothing(params: dict) -> None:
...
+            max_file_size = self._custom_settings.get("maxFileSize")
+            if max_file_size is not None:
+                files_settings["maxSize"] = max_file_size
+            intelephense_settings: dict = {"files": files_settings}
+            max_memory = self._custom_settings.get("maxMemory")
+            if max_memory is not None:
+                intelephense_settings["maxMemory"] = max_memory
+            intelephense_config: DidChangeConfigurationParams = {"settings": {"intelephense": intelephense_settings}}
+            log.info(f"Sending workspace/didChangeConfiguration with file associations: {associations}")
+            self.server.notify.workspace_did_change_configuration(intelephense_config)
```
opcode81: Why handle this with a configuration change?
Can't we pass the right configuration from the get-go in `_create_base_initialize_params`?

src/solidlsp/language_servers/intelephense.py
```diff
@@ -93,6 +105,42 @@ def _get_or_install_core_dependency(self) -> str:
...
+    @staticmethod
+    def _sync_source_fn_matcher(file_filter: list[str]) -> None:
+        """Keep Serena's PHP source-file matcher in sync with the configured ``file_filter``.
+
+        ``Language.PHP.get_source_fn_matcher()`` is a ``@cache``d per-language singleton, so adding
+        the configured extensions here propagates to every consumer of the matcher - symbol index
+        traversal, ignore checks, and language composition detection. Without this, ``find_symbol``
+        would not surface symbols in files whose extensions were added to ``file_filter`` (#1710).
+        """
+        Language.PHP.get_source_fn_matcher().add_extensions(*file_filter)
```
opcode81: Please inline this. Note that you can reference the applicable enum item via `self.language` in a non-static setting.

src/solidlsp/language_servers/intelephense.py
```diff
@@ -93,6 +105,42 @@ def _get_or_install_core_dependency(self) -> str:
...
+    @staticmethod
+    def _resolve_file_filter(solidlsp_settings: SolidLSPSettings) -> list[str]:
+        """Resolve the file extensions to treat as PHP sources.
+
+        Reads an optional override from ``ls_specific_settings["php"]`` (key ``file_filter``);
+        falls back to the default otherwise. Extracted as a pure function so the configuration
+        plumbing can be unit-tested without starting the language server.
+        """
+        php_settings = solidlsp_settings.get_ls_specific_settings(Language.PHP)
+        return php_settings.get("file_filter", list(_DEFAULT_FILE_FILTER))
```
opcode81: Inline this.
And instead of using a static method, use an instance method. The LS-specific settings are then accessible via `self._custom_settings`.

src/solidlsp/language_servers/intelephense.py
```diff
@@ -93,6 +105,42 @@ def _get_or_install_core_dependency(self) -> str:
...
+
+        Returns ``None`` for the default filter: the pushed list replaces Intelephense's built-in
+        associations (``["*.php", "*.phtml"]``), so pushing globs derived from ``[".php"]`` would
+        silently drop ``*.phtml`` indexing and change default behavior (unlike Perl::LanguageServer,
+        whose defaults match Serena's and allow an unconditional push). A customized filter is
+        pushed and fully defines what the server indexes (#1710).
+        """
+        if file_filter == _DEFAULT_FILE_FILTER:
+            return None
+        return [f"*{ext}" for ext in file_filter]
```
opcode81: Especially if we move the configuration to the initialisation, we do not need to differentiate the case where the default filter is not changed.
If the notion of the default filter is required, it should come from the source filename matcher, not a separate constant. Intelephense's default is irrelevant.

Since `.phtml` is a standard (yet outdated) extension used for PHP, we can add it to the default `FilenameMatcher` for all PHP language servers.

opcode81 (changes requested):
Thanks for your contribution. Please see my comments.

src/solidlsp/language_servers/intelephense.py
vboychev: I tried moving this into `_create_base_initialize_params` first, but Intelephense ignores configuration passed via `initializationOptions`: its `onInitialize` only reads `storagePath`, `globalStoragePath`, `clearCache` and `licenceKey`, and settings such as `files.associations` are only applied from `workspace/didChangeConfiguration` (or pulled via `workspace/configuration`, a capability our client does not declare). The integration test confirms this: with the associations in `initializationOptions`, the never-opened `.module` file does not enter the server's index. I kept the push right after `initialized`, but per your other comments it is now unconditional and derives the globs from the source filename matcher, so the default-filter special-casing is gone. Side note: the pre-existing `intelephense.maxMemory` / `intelephense.files.maxSize` forwards in `initializationOptions` appear to be ignored by the server for the same reason (pre-dates this PR).

src/solidlsp/language_servers/intelephense.py
vboychev: Done: inlined into `__init__`, reading the setting via `self._custom_settings`.

src/solidlsp/language_servers/intelephense.py
vboychev: Done: inlined, using `self.language.get_source_fn_matcher()`.

src/solidlsp/language_servers/intelephense.py
vboychev: Done: the helper and `_DEFAULT_FILE_FILTER` are removed. `.phtml` is now part of the default `FilenameMatcher` for all PHP language servers, `file_filter` is additive, and the `files.associations` globs are derived from the source filename matcher and pushed unconditionally (no default-case differentiation).

src/solidlsp/language_servers/intelephense.py
opcode81: OK, thanks for looking into it.

opcode81:
Looks great now. Merging.

PR #1715, opened 2026-07-19
Java (JDT-LS): allow registering additional JRE/JDK runtimes via config

### LINKED ISSUE #1478: Java backend has no config key to register additional JREs with JDT-LS
_I checked most of the boxes. This is a Claude Code writeup of a bug - or missing feature - we believe we found. I am drive-by submitting it in hope it is valuable - delete it if not. We've fixed my install. Some workarounds/fixes are listed below._

# Bug/Feature Request: Java backend has no config key to register additional JREs with JDT-LS

**Component:** `solidlsp/language_servers/eclipse_jdtls.py` — the `initializationOptions` Serena sends to JDT-LS at LSP startup.
**Severity:** Java diagnostics + cross-module resolution are silently broken for any project whose source level exceeds Serena's bundled JDT-LS JRE version (currently JDK 21). Every JDK type appears as "cannot be resolved" — including `java.lang.Object`, `java.util.*`, `java.lang.String`. Symbol-navigation tools (`find_symbol`, `find_implementations`, etc.) still mostly work because they answer from the source AST without needing JDK type resolution. `get_diagnostics_for_file` and any compile-state-dependent tool returns garbage.

The breakage is silent: nothing in the Serena dashboard or logs warns that the JRE container is unresolvable. The user sees "Java is up and running" but every type lookup against JDK classes fails. Until I traced it to JDT-LS's `.classpath` content I assumed I'd misconfigured something at the project level.

## Reproduction

1. Have a Gradle (or Maven) Java project with `sourceCompatibility = JavaVersion.VERSION_25` (or any version newer than Serena's bundled JDT-LS JRE — currently 21).
2. Open the project in Serena.
3. Call `get_diagnostics_for_file` on any source file.
4. Result: ~50-100 "cannot be resolved" errors per file, dominated by `java.lang.Object`, `java.util.*`, `java.lang.String` — all standard JDK types.

The proximate cause is visible in the JDT-LS-generated `.classpath` for each imported module, in the workspace's `data_dir/.metadata/.plugins/org.eclipse.core.resources/.projects/<module>/.classpath`:

```xml
<classpathentry kind="con"
  path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-25/"/>
```

Buildship correctly extracted the source level from `build.gradle` and asked Eclipse JDT for a `JavaSE-25` JRE container. But Eclipse JDT has no `JavaSE-25` VM registered — the only JRE it knows about is its own running JVM (Serena's bundled JDK 21, `extension/jre/21.0.10-linux-x86_64`). So the container resolves to nothing, and every type that comes from rt.jar / `java.base` is unresolvable.

## Root cause

Eclipse JDT's "Installed JREs" list is normally populated by user input (in the Eclipse UI), or by VS Code's Java extension via the LSP `initializationOptions.settings.java.configuration.runtimes` map:

```json
"java.configuration.runtimes": [
  { "name": "JavaSE-21", "path": "/path/to/jdk21" },
  { "name": "JavaSE-25", "path": "/path/to/jdk25", "default": true }
]
```

JDT-LS reads that on startup and registers each entry as an Eclipse VM via `org.eclipse.jdt.launching.IVMInstallType`. Subsequent JRE container resolution then picks the entry whose `name` matches.

Serena currently exposes (under `ls_specific_settings.java`):
- `java_home` — JDK that JDT-LS itself is launched with
- `gradle_java_home` — JDK Buildship invokes Gradle with
- `use_system_java_home`, `gradle_wrapper_enabled`, `gradle_user_home`, `maven_user_settings`, `jdtls_xmx`, `jdtls_xms`, `jdtls_path`, `lombok_path`

There is no key that maps to `java.configuration.runtimes`. So users on projects that target a Java version newer than Serena's bundled JRE have no way to tell JDT-LS about an installed JDK that satisfies the `JavaSE-NN` container.

`java_home` and `gradle_java_home` look like they should help, but they don't:
- `java_home` controls the JVM JDT-LS itself runs on. Setting it to JDK 25 launches JDT-LS on Java 25, which incidentally makes its own embedded JRE register as JavaSE-25 in the auto-detected VM list — but only if `java_home` is actually used. In my testing, even with `java_home` set in `serena_config.yml`, the spawned JDT-LS process command line continued to reference `extension/jre/21.0.10-linux-x86_64/bin/java` — i.e. the bundled JRE, ignoring the config. (Possibly `java_home` only applies to "upstream JDT-LS mode" — `jdtls_path` + `lombok_path` — and is no-op in the default bundled mode.)
- `gradle_java_home` controls only the JDK Gradle inherits when Buildship invokes it via the Tooling API. It doesn't register a VM with JDT-LS's `IVMInstallType`. The Gradle daemon runs fine on JDK 25, generates the correct `.classpath` referencing `JavaSE-25`, but the container still can't be resolved because JDT-LS itself doesn't know about it.

## Workaround for affected users

**Symlink Serena's bundled JRE to a newer JDK.** Crude, but works perfectly until Serena's JDT-LS bundle gets bumped (at which point you redo it in the new bundle dir):

```bash
cd ~/.serena/language_servers/static/EclipseJDTLS/vscode-java-1.54.0-923/extension/jre/
mv 21.0.10-linux-x86_64 21.0.10-linux-x86_64.ORIG
ln -s /home/endre/Java/jdk25 21.0.10-linux-x86_64
```

(Adjust the vscode-java version dir to whatever is current — Serena's `DEFAULT_VSCODE_JAVA_VERSION` setting determines the dir name. Adjust the JDK path to your own JDK 25+ install.)

Then kill any running JDT-LS and wipe the broken workspace so Buildship redoes the project import on next launch:

```bash
pkill -9 -f EclipseJDTLS
rm -rf ~/.serena/language_servers/static/EclipseJDTLS/workspaces/<the-broken-one>/
```

Effect: the launcher path `extension/jre/21.0.10-linux-x86_64/bin/java` now resolves through the symlink to `/home/endre/Java/jdk25/bin/java`, so JDT-LS boots on JDK 25. Eclipse JDT's StandardVMType auto-detects the running JVM as Java 25, registers it with name `JavaSE-25`, and every project's `.classpath` reference to `JavaSE-25/` resolves to its own runtime. `java.lang.Object` resolves; diagnostics work; cross-module references work.

This is global — one symlink covers every Serena project on the machine. It breaks the next time Serena's `DEFAULT_VSCODE_JAVA_VERSION` is bumped (new bundle dir, new bundled JRE 21, symlink needs to be redone in the new path).

## Suggested fix

Expose a `runtimes` (or `configuration_runtimes`) key under `ls_specific_settings.java` that gets passed through into the JDT-LS `initializationOptions.settings.java.configuration.runtimes` map.

```yaml
ls_specific_settings:
  java:
    runtimes:
      - name: JavaSE-21
        path: /usr/lib/jvm/java-21-openjdk
      - name: JavaSE-25
        path: /home/endre/Java/jdk25
        default: true
```

The Serena Python side just needs to:
1. Validate the keys (`name`, `path`, optional `default`, optional `sources`, optional `javadoc`).
2. Inject the list into `initializationOptions.settings.java.configuration.runtimes` when constructing the LSP init payload sent to JDT-LS.

JDT-LS already handles the rest — registering each entry as an `IVMInstallType` so `JRE_CONTAINER` resolution finds them.

Bonus: with this in place, the `java_home` setting becomes purely "the JVM JDT-LS itself is launched on" (which can stay JDK 21 — that's fine for running JDT-LS), and the project's compile-target JVM is decoupled and explicit.

## Related observation — silent failure mode

The breakage is invisible from Serena's user-facing surface. The dashboard says everything is fine. No warning is logged. JDT-LS's own log dir (`workspaces/<hash>/data_dir/.metadata/.log`) shows zero errors. Only by examining the generated `.classpath` files inside the workspace metadata, OR by calling `get_diagnostics_for_file` and seeing the absurd "java.lang.Object cannot be resolved" message, does the user have any signal that something is wrong.

Worth considering a startup check: after Buildship's import completes, scan each project's generated `.classpath` for `JRE_CONTAINER/.../JavaSE-NN/` entries and verify that the named container is resolvable via `IVMInstallType.findVMInstall(name)`. If not, log a clear warning naming the missing container, the project that needs it, and the documented `runtimes` config key.

## Environment

- Linux 6.8 (Ubuntu)
- uvx (Astral UV via snap) auto-fetching Serena from `git+https://github.com/oraios/serena`
- Serena's bundled JDT-LS: `vscode-java-1.54.0-923` with bundled JRE `21.0.10-linux-x86_64` (Java 21)
- Project: Gradle 9.3.1 multi-module, `sourceCompatibility = VERSION_25`, JDK 25 (Temurin-25+36) installed at `/home/endre/Java/jdk25`
- Date discovered: 2026-05-12

arimu1:
## Problem

JDT-LS resolves each imported module's `JRE_CONTAINER/.../JavaSE-NN/` classpath entry against its own "Installed JREs" list. Serena hardcodes that list to a single `JavaSE-21` entry pointing at the bundled JRE. For any project whose source/target level exceeds that (e.g. `sourceCompatibility = JavaVersion.VERSION_25`), the requested `JavaSE-25` container never resolves, and every JDK type (`java.lang.Object`, `java.util.*`, `java.lang.String`, ...) comes back "cannot be resolved" — silently breaking `get_diagnostics_for_file` and cross-module resolution, with no warning anywhere in Serena's logs or dashboard.

`ls_specific_settings.java` had no key that maps to JDT-LS's `java.configuration.runtimes`, so users had no way to register an installed newer JDK. The only known workaround was symlinking over Serena's bundled JRE directory (documented in the issue), which is global, fragile, and breaks on every `vscode_java_version` bump.

## Root cause

In `src/solidlsp/language_servers/eclipse_jdtls.py`, `_create_base_initialize_params()` unconditionally overwrote `initializationOptions.settings.java.configuration.runtimes` with a single hardcoded `JavaSE-21` entry, discarding any way for a user to add another registered runtime.

## Fix

- Added an optional `runtimes` list under `ls_specific_settings.java`, mirroring the shape VS Code's Java extension sends via `java.configuration.runtimes` (`name`, `path`, optional `default`/`sources`/`javadoc`), documented in the class docstring alongside the other `ls_specific_settings.java` keys.
- Added `_resolve_configured_runtimes()`, which validates the setting (must be a list; each entry needs at least `name` + `path`; `path` must exist on disk) and normalizes each entry.
- The bundled `JavaSE-21` runtime is no longer unconditionally overwritten: configured runtimes now *extend* it. An entry that reuses the `JavaSE-21` name overrides the bundled one; if any configured entry claims `default`, the bundled runtime's own `default` flag is cleared so JDT-LS never sees two runtimes claiming to be the default.
- Added `runtimes` to the JDTLS workspace-hash inputs (`_compute_workspace_hash`'s `workspace_setting_keys`), so changing the setting lands in a fresh workspace instead of silently reusing a stale Buildship/Maven import from before the runtime was registered — consistent with how the other Java-import-affecting settings (`gradle_java_home`, `java_home`, etc.) are already handled.

```yaml
ls_specific_settings:
  java:
    runtimes:
      - name: JavaSE-21
        path: /usr/lib/jvm/java-21-openjdk
      - name: JavaSE-25
        path: /home/user/Java/jdk25
        default: true
```

## Testing

Ran locally (`uv venv -p 3.13 && uv sync --extra dev`):
- `poe format` — clean (one pre-existing formatting normalization applied by ruff on save).
- `poe type-check` — passes (`ty check src/serena src/solidlsp` and `ty check test`).
- Added 14 unit tests to `test/solidlsp/java/test_jdtls_path_resolution.py` (following the existing pattern used for `_resolve_gradle_java_home` / `_compute_workspace_hash` in that file — no real JDTLS process or JDK required):
  - `TestResolveConfiguredRuntimes`: empty when unset, valid entry normalization, `sources`/`javadoc` pass-through, rejects non-list values, rejects malformed entries (missing `name`/`path`, non-dict entries), rejects nonexistent `path`.
  - `TestConfiguredRuntimesInitializeSettings`: bundled runtime is sole default when nothing is configured, a configured runtime extends (not replaces) the bundled one, a same-named configured runtime overrides the bundled one, a configured `default: true` entry clears the bundled runtime's own default, and a malformed configured entry surfaces its `ValueError` through `_create_base_initialize_params()`.
- `pytest test/solidlsp/java -m "not java"` → 69 passed, 8 deselected (the deselected tests are pre-existing and require a real JDK/JDTLS process, unrelated to this change).

I don't have a JDK 25 install handy to exercise this end-to-end against a live JDT-LS process, so the above is unit-level verification of the settings-merging logic; the actual `java.configuration.runtimes` payload construction and merge/override/default-handoff behavior is what's covered.

Fixes #1478
https://github.com/oraios/serena/issues/1478

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,14 @@ Status of the `main` branch. Changes prior to the next official version change w
     line; any other out-of-range end position now raises `InvalidTextLocationError` instead,
     rather than guessing at a body that could be wrong #1498
 
+* Language Servers:
+  - Java (JDT-LS): add `runtimes` to `ls_specific_settings.java`, a list of extra JRE/JDK entries
+    (`name`, `path`, optional `default`/`sources`/`javadoc`) passed through to JDT-LS's
+    `java.configuration.runtimes`. Fixes silently broken JDK type resolution (`java.lang.Object`
+    and other JDK types reported as "cannot be resolved") for projects whose source/target level
+    exceeds the bundled JDK 21 JRE JDT-LS registers by default; configured runtimes extend rather
+    than replace that bundled default. #1478
+
 * Tools:
   - Fix: `search_for_pattern` marked one line too many as matched whenever a match ended with a line
     break, because the match's exclusive end index was mapped to a line number directly and therefore
--- a/docs/02-usage/050_configuration.md
+++ b/docs/02-usage/050_configuration.md
@@ -697,6 +697,7 @@ The following settings are supported for the Java language server:
 | `gradle_wrapper_enabled` | `false` | Use the project's Gradle wrapper (`gradlew`) instead of the bundled Gradle distribution. Enable this for projects with custom plugins or repositories. |
 | `gradle_java_home` | `null` | Path to the JDK used by Gradle. When unset, Gradle uses `JAVA_HOME` if `use_system_java_home` is enabled and `JAVA_HOME` is set; otherwise it falls back to Serena's bundled JRE. |
 | `use_system_java_home` | `false` | Use the system's `JAVA_HOME` environment variable for JDTLS itself and, when `gradle_java_home` is unset, Gradle import. Enable this if your project requires a specific JDK vendor or version for Gradle's JDK checks. |
+| `runtimes` | `[]` | Extra JRE/JDK entries registered with JDT-LS via `java.configuration.runtimes`. Use this when a project's source/target level exceeds the JDK JDT-LS itself runs on (currently JDK 21 in default vscode-java VSIX mode). Each entry is a mapping with required `name` (e.g. `JavaSE-25`, matching the `JavaSE-NN` container the build tool requests) and `path` (JDK/JRE home directory; must exist), plus optional `default`, `sources`, and `javadoc` (passed through to JDT-LS). Entries extend rather than replace the bundled `JavaSE-21` runtime; an entry that reuses the `JavaSE-21` name overrides the bundled one. Changing this setting invalidates the JDTLS workspace hash so a fresh import is performed. |
 | `gradle_version` | `8.14.2` | (vscode-java mode only) Override the Gradle distribution version Serena downloads by default. |
 | `vscode_java_version` | `1.54.0-923` | (vscode-java mode only) Override the bundled `vscode-java` runtime bundle version Serena downloads by default. |
 | `intellicode_version` | `1.2.30` | (vscode-java mode only) Override the IntelliCode VSIX version Serena downloads by default. |
@@ -716,6 +717,10 @@ Notes:
 - In upstream-jdtls mode the `gradle_version`, `vscode_java_version`, `intellicode_version`,
   `intellicode_xmx`, `intellicode_xms` settings are silently ignored — they only apply to the
   vscode-java VSIX mode.
+- Without `runtimes`, JDT-LS only knows about the bundled `JavaSE-21` JRE. Projects that request a newer
+  container (e.g. `sourceCompatibility = JavaVersion.VERSION_25`) then fail to resolve JDK types such as
+  `java.lang.Object`. Register the matching installed JDK via `runtimes` instead of symlinking over Serena's
+  bundled JRE directory.
 
 Example: upstream-jdtls mode (offline / corporate network):
 
@@ -734,6 +739,19 @@ ls_specific_settings:
   java:
     gradle_wrapper_enabled: true
     use_system_java_home: true
+```
+
+Example: register an additional JDK for a project targeting a newer Java version:
+
+```yaml
+ls_specific_settings:
+  java:
+    runtimes:
+      - name: JavaSE-21
+        path: /usr/lib/jvm/java-21-openjdk
+      - name: JavaSE-25
+        path: /home/user/Java/jdk25
+        default: true
 ```
 
 #### Kotlin
--- a/src/solidlsp/language_servers/eclipse_jdtls.py
+++ b/src/solidlsp/language_servers/eclipse_jdtls.py
@@ -15,7 +15,7 @@ import subprocess
 import threading
 from pathlib import Path, PurePath
 from time import sleep
-from typing import cast
+from typing import Any, cast
 
 from overrides import override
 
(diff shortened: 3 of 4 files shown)


opcode81 (changes requested):
Please document the new runtime configuration in 050_configuration.md

arimu1:
Documented `ls_specific_settings.java.runtimes` in `docs/02-usage/050_configuration.md` as requested:

- settings table entry (shape, merge/override behavior, workspace-hash invalidation)
- note explaining why the setting is needed vs. the old symlink workaround
- YAML example registering additional JDKs

Thanks for the review!

opcode81:
Thanks! This is a great addition for Java users.

PR #1719, opened 2026-07-19
Fix: reflect externally-made file edits in symbolic queries (workspace/didChangeWatchedFiles)

### LINKED ISSUE #1718: workspace/didChangeWatchedFiles is defined but never sent — warm LS serves stale symbol data after external edits
`workspace/didChangeWatchedFiles` is defined but never sent — a warm language server serves stale symbol data after external file edits

## Summary

Serena's LSP client layer implements the `workspace/didChangeWatchedFiles`
notification (`solidlsp/lsp_protocol_handler/lsp_requests.py`,
`LspNotification.did_change_watched_files`), and every bundled language server
declares the corresponding client capability
(`didChangeWatchedFiles: {"dynamicRegistration": true}` — see e.g.
`solidlsp/language_servers/pyright_server.py`). However, the notification is
never actually sent: a repo-wide search for `did_change_watched_files(` /
`didChangeWatchedFiles` finds only the method definition and the capability
declarations — zero call sites.

The effect: once a language server has been started for a project and stays
resident (e.g. across an MCP session), Serena has no mechanism to tell it that
files changed on disk *outside of Serena's own read/write tools*
(`create_text_file`, `replace_symbol_body`, etc.). Symbolic queries such as
`find_referencing_symbols` / `find_symbol` then answer from the language
server's stale in-memory snapshot instead of the current file contents.

This is invisible in the common single-agent, single-editor workflow, where
nearly all edits go through Serena's own tools (which correctly call
`did_open_text_document` / `did_change_text_document` / `did_close_text_document`
around each edit). It becomes a real correctness problem whenever the files
change through a channel Serena doesn't control while a language server instance
stays warm — another editor/IDE, a second agent, a `git`
checkout/merge/rebase, a build step, or any multi-agent setup where several
tools operate on the same working tree while one keeps an LSP session resident
for performance.

## Minimal repro

Reproduces with any project and any bundled language server (tested with
Python/pyright); the mechanism is language-server-agnostic since the bug is
"notification never sent".

1. Start a Serena session against a real project so its language server starts
   and stays resident.
2. Pick a symbol referenced from more than one file and confirm
   `find_referencing_symbols` returns the expected caller set (this warms the
   index).
3. **Outside Serena** — not through any Serena tool — edit one of the
   referencing files on disk (e.g. `sed -i`, another editor), adding a new call
   to the symbol or deleting an existing one.
4. Re-run `find_referencing_symbols` for the same symbol in the same
   still-running session (no restart).

**Expected:** the new call site appears (or the deleted one disappears) —
Serena reflects the current file on disk.

**Actual:** the results still reflect the pre-edit contents; the external edit
is invisible until the language server (or the whole session) is restarted. On
a large (~1700-file) Python project, only a small minority of externally-edited
files were picked up by the running session.

## Root cause

`workspace/didChangeWatchedFiles` is the standard LSP mechanism by which a
client tells a language server about file changes it observed on disk
(create/change/delete). Serena defines this notification but never calls it, and
has no file-watching mechanism (no `watchdog`/`watchfiles` dependency, no
polling loop, no `client/registerCapability` handler for a server-initiated
dynamic watcher) to detect and report the changes. Edits made through Serena's
own tools are notified inline via didOpen/didChange/didClose, which is why only
*externally* made edits fall through the gap.

This is a silent-correctness issue, not a crash: the tool call succeeds and
returns a plausible-looking but outdated answer.

## Suggested fix direction

A lightweight, self-contained fix that reuses primitives Serena already has: a
per-symbolic-query freshness check that walks `Project.gather_source_files()`,
diffs mtimes against the previous poll, and fires the existing
`did_change_watched_files` notification (plus a short `open_file()` cycle for
brand-new files, which pyright needs to fold a never-seen file into its
cross-file reference graph). Hooked into `Tool.apply_ex` gated on
`is_symbolic()`, this covers both the stdio-MCP and HTTP query-server paths with
no new dependency and no background thread.

I have a working, tested patch for this (config-gated, with the concurrency lock
and a full end-to-end test against the pyright backend). Opening this issue
first to confirm the maintainers agree with the direction before submitting the
PR — a persistent filesystem watcher would be the heavier alternative and could
be a follow-up if the per-query walk cost ever becomes a problem at scale.

legacynik:
Fix: reflect externally-made file edits in symbolic queries (`workspace/didChangeWatchedFiles`)

Fixes #1718.

## Problem

`workspace/didChangeWatchedFiles` is defined
(`solidlsp/lsp_protocol_handler/lsp_requests.py`) and declared as a client
capability by every bundled language server, but it is never sent — there are
zero call sites. A resident language server therefore never learns about files
changed on disk outside of Serena's own tools (another editor, a second agent, a
`git` checkout, a build step), so symbolic queries such as
`find_referencing_symbols` and `find_symbol` silently return stale results until
a restart. Edits made through Serena's own tools are unaffected (they notify the
LS inline via didOpen/didChange/didClose); only external edits fall through.

See the linked issue for the full write-up and repro.

## Fix

A per-symbolic-query freshness check that reuses primitives already in the tree,
adds no dependency and no background thread:

- **`Project.poll_and_notify_file_changes()`** — walks
  `gather_source_files()` (so the tracked set is exactly what Serena itself
  reads/edits, honoring `.gitignore` / `ignored_paths` / language masks),
  `stat()`s each file, and diffs mtimes against the previous poll. For
  Created/Changed/Deleted files it sends `did_change_watched_files` to every
  language server the project manages. For **Created** files it additionally
  runs a short `open_file()` (didOpen/didClose) cycle — empirically pyright does
  not fold a brand-new file into its cross-file reference graph on the
  notification alone; Changed and Deleted do not need this. The first call only
  establishes the baseline (it must not report a project's entire pre-existing
  file set as newly created).

- **`Tool._maybe_refresh_project_freshness()`**, called in `Tool.apply_ex`
  before the tool runs, gated on `is_symbolic()`. This is the single choke point
  both delivery paths pass through — the stdio-MCP path and the HTTP
  query-server path (`ProjectServer` also calls `tool.apply_ex`). Non-symbolic
  tools skip it: file tools read disk directly, and editing tools notify inline.
  Poll failures are swallowed so a poll can never break a tool call.

- **`SerenaConfig.detect_external_file_changes`** (default `True`) — toggles the
  behavior. Enabled by default because a warm LS otherwise silently serves stale
  data; can be disabled on very large repositories where the per-query
  filesystem walk is not worth the staleness protection.

### Concurrency

The read-diff-swap of the shared last-seen-mtimes state is done under a
`threading.Lock`; the filesystem walk and the LSP calls stay outside the lock so
concurrent callers (`ProjectServer` runs Werkzeug with `threaded=True`) don't
serialize on I/O. A load test on the prototype this is derived from reproduced a
real `dict` mutation race and a torn-read spurious-event storm without the lock.

### Cost / limitations

- `O(tracked source files)` per symbolic query — one `os.walk` + one `os.stat`
  per file. Well within the warm-query budget on a ~1700-file project; on a
  multi-thousand-file monorepo it starts to dominate query latency, which is
  what the config flag (and, as a possible follow-up, a persistent
  filesystem watcher) is for.
- Detects changes only at query time by design — that is exactly what makes it
  cheap and dependency-free.

## Testing

`test/serena/test_external_file_freshness.py`:

- **End-to-end against the real pyright backend** (`test_external_edits_become_visible_after_poll`):
  starts a warm project language server, then — adversarially — asserts a new
  external caller is **invisible before** the poll (reproducing the bug) and
  **visible after** it. Exercises all three change types in one warm session:
  Created (new caller file, via the `open_file()` path), Changed (appended
  caller), Deleted (removed caller file).
- **Baseline-only first poll** (`test_first_poll_is_baseline_only`): a freshly
  loaded project reports zero changes.
- **Wiring/guard unit tests** (no language server): the `apply_ex` hook polls
  only for symbolic tools, is skipped when disabled by config or when there is
  no active project, and swallows poll errors.

`ruff check`, `ruff format --check`, and `ty check` are clean on the changed
files. The existing Python symbolic-tool suites
(`test/serena/test_serena_agent.py` symbolic cases,
`test/solidlsp/python/test_symbol_retrieval.py`, `test/serena/test_symbol.py`,
`test/serena/test_hooks.py`) pass with the change.

## Notes for reviewers

- Whether `poll_and_notify_file_changes` should live on `Project` (it already
  owns `gather_source_files` and the ignore machinery) or in a small standalone
  module is a style call — happy to relocate.
- The default of `detect_external_file_changes=True` treats this as a
  correctness fix for everyone; if you'd prefer it opt-in initially, flipping the
  default is a one-line change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@
 Status of the `main` branch. Changes prior to the next official version change will appear here.
 
 * General:
+  - Fix: a warm language server served stale symbol results for files edited outside of Serena's own
+    tools (another editor, a second agent, a `git` checkout, a build step). The
+    `workspace/didChangeWatchedFiles` notification was defined but never sent, so such edits were
+    invisible to symbolic queries (`find_referencing_symbols`, `find_symbol`, etc.) until a restart.
+    Before each symbolic tool call, Serena now checks its tracked source files for external changes and
+    notifies the language server (new `Project.poll_and_notify_file_changes`). Controlled by the new
+    `detect_external_file_changes` config option (enabled by default; disable on very large repositories
+    where the per-query filesystem walk is not worth it).
   - Fix: `FileUtils.read_file`'s `charset_normalizer` fallback (used when a file cannot be decoded with
     the project's configured `encoding`) decoded the raw bytes directly and therefore skipped the
     universal-newline translation that the primary read path applies. CR characters from disk thus
--- a/src/serena/config/serena_config.py
+++ b/src/serena/config/serena_config.py
@@ -874,6 +874,15 @@ class SerenaConfig(SharedConfig, ModeSelectionDefinitionWithBaseModes):
     """
     tool_timeout: float = DEFAULT_TOOL_TIMEOUT
 
+    detect_external_file_changes: bool = True
+    """Whether, before each symbolic tool call, Serena should check the project's source files for
+    changes made outside of its own tools (another editor, a second agent, a git checkout, a build
+    step) and notify the language server so symbolic queries reflect the current files on disk.
+    Enabled by default because a warm language server otherwise silently serves stale symbol data
+    for such edits. The check performs one directory walk plus an ``os.stat`` per tracked source
+    file per symbolic call; set this to False on very large repositories where that cost outweighs
+    the staleness risk (e.g. when all edits go through Serena anyway)."""
+
     token_count_estimator: str = RegisteredTokenCountEstimator.CHAR_COUNT.name
     """Only relevant if `record_tool_usage` is True; the name of the token count estimator to use for tool usage statistics.
     See the `RegisteredTokenCountEstimator` enum for available options.
--- a/src/serena/project.py
+++ b/src/serena/project.py
@@ -21,6 +21,11 @@ from serena.util.file_system import GitignoreParser, match_path, scan_directory
 from serena.util.text_utils import MatchedConsecutiveLines, search_files
 from solidlsp import SolidLanguageServer
 from solidlsp.ls_config import Language
+from solidlsp.lsp_protocol_handler.lsp_types import (
+    DidChangeWatchedFilesParams,
+    FileChangeType,
+    FileEvent,
+)
 
 if TYPE_CHECKING:
     from serena.agent import SerenaAgent
@@ -59,6 +64,13 @@ class Project(ToStringMixin):
         self._language_server_manager_init_error: Exception | None = None
         self.is_newly_created = is_newly_created
         self._agent: Optional["SerenaAgent"] = None
+
+        # State for detecting externally-made file changes (see poll_and_notify_file_changes).
+        # Maps relative source-file path -> mtime last observed by the poll. None until the first
+        # poll, which only establishes the baseline (it must not report the project's entire
+        # pre-existing file set as newly created).
+        self._freshness_last_seen_mtimes: dict[str, float] | None = None
+        self._freshness_lock = threading.Lock()
 
         # create .gitignore file in the project's Serena data folder if not yet present
         serena_data_gitignore_path = os.path.join(self._serena_data_folder, ".gitignore")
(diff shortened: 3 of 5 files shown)


ebarkhordar (commented):
Read through this closely. The didChangeWatchedFiles gap from #1718 is real, and polling gather_source_files before symbolic calls is a clean way to close it without a watcher thread. One question about the Changed path.

In poll_and_notify_file_changes you force a parse+bind for Created files with the open/close cycle, and the comment explains why: a didChangeWatchedFiles(Created) notification alone was not enough for pyright to fold the file into its cross-file reference graph. Since _maybe_refresh_project_freshness runs synchronously in apply_ex right before the symbolic tool runs, the same reasoning seems to apply to Changed files. The notification is fire-and-forget, so if pyright applies a Changed watched-file event on the same background pass that made the Created notification insufficient, the symbolic query that immediately follows could still be answered from the pre-edit index. In-order message processing would not close this, since that is exactly what was not enough for the Created case.

So: is there something about the Changed path that makes the re-index synchronous (first post-edit query guaranteed fresh), or would Changed also benefit from the same forcing open/close cycle, or a small barrier, that Created gets? If Changed is safe without it, a one-line note on why would help the next reader.

Not blocking, just trying to understand the timing guarantee.

opcode81:
@legacynik thank you for you contribution. I took your implementation as a basis but needed to make some changes.
Superseded by #1726 (you are credited as a co-author)

PR #1647, opened 2026-07-03
Add Grok Build support (context, setup, native hooks)

xormania:
Adds dedicated support for using Serena with xAI's Grok Build CLI.

## Changes

- New `grok` context (single-project, optimized for Grok Build; shaped like the existing `claude-code` context — except `structured_tool_output`, which is left at the auto default as in the `codex`/`codebuddy` contexts, since `claude-code`'s explicit `false` is a Claude-Code-specific bug workaround)
- `serena setup grok` command to register Serena as a user-scoped MCP server (with CLI compatibility detection)
- Grok-native lifecycle hooks (`serena-hooks --client=grok`), including PreToolUse allow/deny output in Grok's `{decision, reason}` format
- Two small hardenings in the PreToolUse remind hook's client-agnostic payload parsing that apply to **all** hook clients: non-string `command` values no longer raise, and `target_file`/`targetFile` file-path keys are recognized (CHANGELOG entry added under Hooks)
- Documentation (`docs/02-usage/030_clients.md`), unit tests for setup/hooks/context, and `scripts/live_test_grok.py` — a zero-inference, state-preserving live smoke test against a real `grok` installation (documented in `CONTRIBUTING.md`)

This lets agents using the `grok` CLI get proper Serena integration with `--project-from-cwd` and correct context behavior.

Note: the `other-langs (macos-latest)` failure is unrelated to this PR — it is a pre-existing break on `main` (shader-sense 1.4.0 broke the HLSL server build from source), fixed in #1652. I will refresh this branch once that lands.

## Checklist

- [x] This PR follows the guidelines in `CONTRIBUTING.md` regarding the scope of PRs.
- [x] For changes that add features or fix problems, I have added an entry to `CHANGELOG.md`, which concisely describes the change.

PR #1720, opened 2026-07-19
Add Grok Build support (context, setup, native hooks)

xormania:
Adds dedicated support for using Serena with xAI's Grok Build CLI. Supersedes #1647 (same change, rebased onto current main after 77 commits of drift, including the v1.6.0 release).

## Changes

- New `grok` context (single-project, optimized for Grok Build; shaped like the existing `claude-code` context — except `structured_tool_output`, which is left at the auto default as in the `codex`/`codebuddy` contexts, since `claude-code`'s explicit `false` is a Claude-Code-specific bug workaround)
- `serena setup grok` command to register Serena as a user-scoped MCP server (with CLI compatibility detection)
- Grok-native lifecycle hooks (`serena-hooks --client=grok`), including PreToolUse allow/deny output in Grok's `{decision, reason}` format
- Two small hardenings in the PreToolUse remind hook's client-agnostic payload parsing that apply to **all** hook clients: non-string `command` values no longer raise, and `target_file`/`targetFile` file-path keys are recognized (CHANGELOG entry under Hooks)
- Documentation (`docs/02-usage/030_clients.md`), unit tests for setup/hooks/context, and `scripts/live_test_grok.py` — a zero-inference, state-preserving live smoke test against a real `grok` installation (documented in `CONTRIBUTING.md`)

This lets agents using the `grok` CLI get proper Serena integration with `--project-from-cwd` and correct context behavior.

## Rebase notes (77 commits, verified against current interfaces)

- CHANGELOG entries re-homed into `# Unreleased (main)` — the in-flight v1.6.0 release cut had left them anchored inside the released section; the new `CLI`/`Hooks` subsections mirror v1.6.0's ordering (own commit, for reviewability).
- Conformance audited against everything that moved in the window: client-setup handler pattern, hook API, context yml schema (incl. the `tool_names` prompt mechanism — predates this branch's base), `SerenaConfig.init`, docs section structure. No interface drift; no prompt-factory regeneration needed.
- Verified locally: `poe format` (no changes), `poe lint`, `poe type-check` all clean; grok test battery 118/118; full `test/serena` suite 725 passed (5 failures are missing local toolchains — pwsh/dotnet — for unrelated C#/PowerShell LS tests that run on CI).

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,8 @@ Status of the `main` branch. Changes prior to the next official version change w
 
 * Hooks:
   - Add `serena-hooks --client=grok`, including Grok-native PreToolUse allow/deny output.
+  - PreToolUse remind hook: coerce non-string shell command values instead of failing, and recognize
+    `target_file`/`targetFile` file-path keys (shared payload parsing, applies to all hook clients).
 
 * Dependencies:
   - Bump mcp from 1.27.0 to 1.28.1
@@ -189,8 +191,6 @@ Status of the `main` branch. Changes prior to the next official version change w
   - Fix: Wait for the subprocess that opens the browser window, preventing zombie processes #1488 
 
 * Hooks:
-  - PreToolUse remind hook: coerce non-string shell command values instead of failing, and recognize
-    `target_file`/`targetFile` file-path keys (shared payload parsing, applies to all hook clients).
   - Handle tool_input passed as string gracefully instead of failing (Copilot CLI sends strings).
 
 * Memories:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,12 @@ Status of the `main` branch. Changes prior to the next official version change w
     `SymbolBody.get_text`. That one well-defined case is now corrected to end at the actual last
     line; any other out-of-range end position now raises `InvalidTextLocationError` instead,
     rather than guessing at a body that could be wrong #1498
+  - Add a `grok` context optimized for xAI's Grok Build CLI.
+  - Add `scripts/live_test_grok.py`, a zero-inference live smoke test verifying the Grok integration
+    (setup, hooks, MCP handshake) against a real `grok` CLI installation (see `CONTRIBUTING.md`).
+
+* CLI:
+  - Add `serena setup grok` to register Serena as a Grok MCP server.
 
 * Language Servers:
   - Java (JDT-LS): add `runtimes` to `ls_specific_settings.java`, a list of extra JRE/JDK entries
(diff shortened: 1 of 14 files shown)


xormania:
**Live behavior evidence — grok 4.5 running this branch (fresh session, `--project-from-cwd`)**

I ran a fixed T0–T8 task battery under a self-reporting protocol (the agent records every call, its exact params, and any friction after each task, before starting the next). Headline: **10/10 serena tool calls succeeded first try, zero failures, zero retries.**

Examples:

- **Navigation**: `find_symbol` with `name_path_pattern="ClientSetupHandlerGrok"`, `include_body=true` → exact hit on the first attempt. `find_referencing_symbols` on `client_setup_handlers` → `cli.py` (init/setup/Choice) plus the test file, all correctly attributed.
- **Symbolic edit**: `find_symbol` (with body) → `replace_symbol_body` reusing the returned name_path `GrokLiveTest/_backup_config_private` → re-fetched the symbol to verify → reverted via native git. Zero missteps; the retrieve-before-replace contract was followed unprompted.
- **Division of labor** (the agent's own reasoning, verbatim): "native grep for text-in-files (Serena has no search_for_pattern/find_file); Serena find_symbol for definition site — symbolic lookup is the right tool when you want the defining symbol, not every textual mention." The six excluded file/shell tools never appeared anywhere in the session.

This is what settled the "context might not be optimal for grok" question for me: the exclusion set maps cleanly onto grok's native toolset in live use, not just on paper — the agent split symbolic work (serena) vs text/fs work (native) correctly without being told to.

xormania:
**Protocol-level evidence — mechanized runs against real grok installs**

- **Hooks**: 12 PASS / 0 FAIL against grok 0.2.101 — native `{decision, reason}` deny/allow shapes, the 120s post-deny remind window, per-session state cleanup.
- **Full live smoke** (`scripts/live_test_grok.py`) against grok 0.2.103: **22 PASS / 3 INFO / 1 WARN / 0 FAIL** — setup incl. idempotent re-run and the negative path, `grok mcp doctor` connectable, and wire-level confirmation of both context deltas: 23 vs 21 tools on the MCP handshake (the single-project exclusions), and `outputSchema`/`structuredContent` served on grok (auto default) vs plain text on claude-code (its explicit `false` workaround).
- The 1 WARN is grok's `/hooks-trust` gating for project-level hooks, which the docs section in this PR already covers.

@MischaPanch that should cover the asks from #1647 — happy to run a specific scenario if you want more.

opcode81:
Thanks a lot for your contribution.
This was merged with squashed commit 9debf7742ab20c40f4075f12a6aa71a3c76852e6.
Since you didn't grant maintainers the right to edit your PR, I had to bypass GitHub.

PR #1728, opened 2026-07-21
Retry LSP ContentModified (-32801) errors instead of failing hard

### LINKED ISSUE #1724: rust-analyzer `ContentModified` (-32801) surfaces as a hard error (no client-side retry) — flaky `test_find_symbol[rust_add_function]` on windows-latest
Preconditions:

Issue details:

**Signature** (identical in 6 occurrences): `test/serena/test_serena_agent.py::TestSerenaAgent::test_find_symbol[rust_add_function]` → `textDocument/hover` (the `include_info=True` enrichment in `request_info_for_symbol_batch`) → `LSPError: content modified (-32801)` → hard `SolidLSPException`. On `main`: five failures between 07-15 and 07-17, zero in the 18 `native (windows-latest)` jobs that ran to completion before that (back to 07-03; all counts exclude cancelled jobs; as of 07-19 ~19:00 UTC):

- https://github.com/oraios/serena/actions/runs/29433865109/job/87415421314 (07-15)
- https://github.com/oraios/serena/actions/runs/29502472168/job/87634643370 (07-16)
- https://github.com/oraios/serena/actions/runs/29551544495/job/87794922043 (07-17 — the run shows "cancelled" because a later push cancelled its remaining jobs; this job had already completed as failure)
- https://github.com/oraios/serena/actions/runs/29577099522/job/87873946692 (07-17)
- https://github.com/oraios/serena/actions/runs/29585582976/job/87901640184 (07-17)

plus #1716's run: https://github.com/oraios/serena/actions/runs/29690002527/job/88203593103 (07-19; that run's re-run attempt only re-executed `jvm (windows-latest)`, so the native job executed once). The signature has zero non-Windows occurrences. (For anyone auditing red runs in the same window: a `native (macos-latest)` ccls flake — empty document symbols — and one `jvm (windows-latest)` failure also appear there; different signatures, not covered here.)

**Onset correlates with a runner-image bump.** All six failures ran runner image `windows-2025-vs2026` version `20260707.563`. The last green windows-native job before onset (07-15 15:23 UTC) ran `20260624.560`; the first failure (07-15 16:47 UTC) is on the new image. Post-onset green jobs also run `20260707.563` — the new image widens a race rather than failing deterministically. Since rust-analyzer is provisioned via `rustup which rust-analyzer` (`RustAnalyzer.DependencyProvider`), an image-side toolchain/rust-analyzer bump at exactly the onset is the likely trigger; the client-side gap below is the longstanding susceptibility it exposed.

**Mechanism**: `-32801` is LSP `ContentModified`. When a state write cancels an in-flight salsa computation, rust-analyzer turns the cancelled handler into a `content modified` error response (`content_modified_error()` in `crates/rust-analyzer/src/handlers/dispatch.rs`) unless the method is in its hard-coded internal-retry set — which does not include hover — so recovery is entirely the client's job. On the Serena side, `send_request` converts any error result into `SolidLSPException` (`LanguageServerInterface`, `ls_process.py`), and nothing outside `pyrefly_server.py` handles `-32801`/`-32800`. A transient cancellation therefore becomes a failed test — and, in production, a failed tool call (the language-server-terminated restart path does not cover this case).

**In-tree precedent / possible directions**: pyrefly's `_install_mutation_retry` already handles the same phenomenon locally — it re-issues requests failing with `RequestCancelled`/`ContentModified` (up to 5 attempts, 0.2s apart), per its docstring for pyrefly's background re-indexing. A rust-analyzer-scoped equivalent restricted to `ContentModified` would close this (`-32800` is client-initiated cancellation per the spec, and a *generic* request-path retry would also re-issue resolve requests that are `ContentModified` by design per LSP 3.17's implementation considerations). Alternatively — or additionally — the `include_info` enrichment could degrade gracefully instead of failing the whole `find_symbol` call.

arimu1:
Closes https://github.com/oraios/serena/issues/1724

## Problem

`-32801` is the LSP `ContentModified` error. Per the spec, it means the server discarded a stale, in-flight computation because the workspace changed underneath it — not that the request itself is invalid — so a well-behaved client is expected to retry, *provided it told the server it would*.

`LanguageServerInterface.send_request` (`src/solidlsp/ls_process.py`) converted any error response, including `ContentModified`, straight into a hard `SolidLSPException`, with no retry, for every language server.

This surfaced as the flaky `test_find_symbol[rust_add_function]` failure on `windows-latest` described in the issue: rust-analyzer returns `ContentModified` for hover requests cancelled by its own background re-indexing (`content_modified_error()` in `dispatch.rs`; hover isn't in rust-analyzer's internal retry set), and nothing on Serena's side retried it, so a transient cancellation became a hard test failure (and, in production, a failed tool call). Concretely: `FindSymbolTool.apply(..., include_info=True)` → `LanguageServerSymbolRetriever._request_info` → `request_hover` → `textDocument/hover`.

## Fix

`send_request` retries a response carrying LSP error code `ContentModified` (-32801), but **only for methods a server has declared it wants retried**, via the standard LSP client capability `general.staleRequestSupport.retryOnContentModified` in its `InitializeParams`. That declaration is the client's contract with the server about exactly which requests it will reissue on cancellation (LSP spec, `StaleRequestSupportClientCapabilities`); blindly retrying every method, including non-idempotent ones like `workspace/executeCommand`, would contradict it.

Mechanism:
- `LanguageServerInterface.set_content_modified_retry_methods(methods)` (new) registers the allowlist on the interface. Empty by default — no declaration means no retries, for any server.
- `SolidLanguageServer._create_initialize_params()` now calls this right after building each server's `InitializeParams`, reading the methods back out of `capabilities.general.staleRequestSupport.retryOnContentModified`. This is a single hook shared by all ~69 language server implementations (they all call `_create_initialize_params()`), so no per-server wiring is needed beyond declaring the capability.
- `send_request` checks the method against that allowlist before retrying; everything else (including undeclared methods on servers that *do* declare something) is surfaced immediately, as before this PR.
- rust-analyzer (`rust_analyzer.py`) now adds `textDocument/hover` to its existing `retryOnContentModified` declaration (previously just the three `semanticTokens/*` methods), since that's the method the issue's repro actually needs.
- No other server declares any methods in `retryOnContentModified` today (`haskell_language_server.py` explicitly declares `[]`; jdtls/kotlin/sourcekit/omnisharp already declared the three `semanticTokens/*` methods pre-existing this PR), so the retry mechanism is inert for them, same as before this PR.

`RequestCancelled` (-32800) is deliberately *not* retried here: per the spec it means the client itself cancelled the request, so retrying it generically wouldn't make sense.

**pyrefly interaction (double-retry avoided):** `pyrefly_server.py` wraps `send_request` with its own pre-existing 5×/0.2s retry for `RequestCancelled`/`ContentModified` (`_install_mutation_retry`), unrelated to LSP capability negotiation. An earlier version of this PR added an *unconditional* base-class retry, which nested inside pyrefly's wrapper (5×5=25 round-trips worst case). With the allowlist, this resolves itself: pyrefly never declares `general.staleRequestSupport`, so the base-class retry is a no-op for every pyrefly request, and pyrefly's own wrapper remains the sole handler, exactly as before this PR touched anything. `pyrefly_server.py` is intentionally left unmodified — verified by reproducing its exact wrapping pattern against a scripted `LanguageServerInterface` double: 5 total dispatches (pyrefly's own `max_retries`), not 5×3.

**Latency:** `_CONTENT_MODIFIED_MAX_ATTEMPTS` reduced from 5 to 3 (2 retries), and now scoped to allowlisted methods only. Worst case for the one method that's actually retried by default (rust-analyzer's `textDocument/hover`) is 2 extra attempts: **`2 * request_timeout + 0.4s`** added latency versus a single failing attempt (down from `4 * request_timeout + 0.8s`).

`send_request` is restructured around a `_send_request_once` helper: the first attempt always runs, then (only if the method is allowlisted) a bounded retry loop runs, and the final `raise` on exhaustion is reached through ordinary control flow rather than a trailing `raise AssertionError("unreachable")`.

## Testing

`test/solidlsp/test_content_modified_retry.py`: a lightweight `LanguageServerInterface` test double that answers requests synchronously from a scripted list of results, with no real language server process. Covers:
- `ContentModified` is retried until a later attempt succeeds (method registered)
- retries are bounded — gives up and raises after the max attempt count
- other LSP errors (e.g. `RequestFailed`) are not retried
- `RequestCancelled` (-32800) is not retried
- `ContentModified` is **not** retried for a method that wasn't registered (also covers `workspace/executeCommand` specifically, as a non-idempotent example)
- `ContentModified` is **not** retried when nothing was ever registered (the pyrefly-shaped default)

`time.sleep` is monkeypatched in this suite so it runs in ~0.01s instead of ~1.2s of real sleeping.

```
$ uv run pytest test/solidlsp/test_content_modified_retry.py -v
test/solidlsp/test_content_modified_retry.py::test_content_modified_is_retried_until_success PASSED
test/solidlsp/test_content_modified_retry.py::test_content_modified_gives_up_after_max_attempts PASSED
test/solidlsp/test_content_modified_retry.py::test_other_lsp_errors_are_not_retried PASSED
test/solidlsp/test_content_modified_retry.py::test_request_cancelled_is_not_retried PASSED
test/solidlsp/test_content_modified_retry.py::test_content_modified_is_not_retried_for_undeclared_method PASSED
test/solidlsp/test_content_modified_retry.py::test_content_modified_retry_methods_default_to_empty PASSED
6 passed in 0.01s
```

Also ran the full non-language-gated `test/solidlsp` unit suite, which includes real Python-language-server integration tests (`test_ls_common.py`, `test_rename_didopen.py`) that exercise `_create_initialize_params()` end-to-end against a running language server — all green, no regressions:
```
$ uv run pytest test/solidlsp/test_content_modified_retry.py test/solidlsp/test_typescript_timeout_policy.py \
    test/solidlsp/test_ls_common.py test/solidlsp/test_symbol_body.py test/solidlsp/test_rename_didopen.py \
    test/solidlsp/test_lsp_protocol_handler_server.py test/solidlsp/test_ls_config.py \
    test/solidlsp/test_is_ignored_path_missing.py test/solidlsp/test_text_utils.py
76 passed in 9.16s
```

Lint / type-check:
```
$ uv run ruff check src/solidlsp/ls_process.py src/solidlsp/ls.py src/solidlsp/language_servers/rust_analyzer.py test/solidlsp/test_content_modified_retry.py
All checks passed!
$ uv run ty check src/serena src/solidlsp
All checks passed!
$ uv run ty check test --exclude test/resources
All checks passed!
```

I don't have a Rust toolchain available locally, so I wasn't able to reproduce/re-run the original `test_find_symbol[rust_add_function]` rust-analyzer test directly, but the retry-methods unit tests exercise the exact mechanism that closes the gap the issue describes, including that it's now scoped to the specific method (`textDocument/hover`) the issue needs.

Also updated the `CHANGELOG.md` entry under "Language Servers" to match.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,13 @@ Status of the `main` branch. Changes prior to the next official version change w
   - PHP/Intelephense: expose `file_filter` via `ls_specific_settings["php"]`, so that additional
     extensions containing PHP sources (e.g. Drupal's `.module` / `.install` / `.inc` / `.theme`)
     become visible to the symbol tools and are indexed by the language server #1710
+  - Fix: an LSP `ContentModified` (-32801) response was surfaced as a hard `SolidLSPException`
+    instead of being retried. Per the LSP spec, `ContentModified` means the server discarded a
+    stale, in-flight computation because the workspace changed underneath it, not that the request
+    itself is invalid; well-behaved clients are expected to retry. `LanguageServerInterface.send_request`
+    now retries such requests a bounded number of times before giving up, for all language servers
+    (this was flaking `test_find_symbol[rust_add_function]` on windows-latest, where rust-analyzer
+    returns `ContentModified` for hover requests cancelled by concurrent indexing). #1724
 
 * JetBrains:
   - Allow external files from dependencies (specified via references like "<ext:FileUtil.class|472e0a13>") to be
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,12 +52,10 @@ Status of the `main` branch. Changes prior to the next official version change w
     extensions containing PHP sources (e.g. Drupal's `.module` / `.install` / `.inc` / `.theme`)
     become visible to the symbol tools and are indexed by the language server #1710
   - Fix: an LSP `ContentModified` (-32801) response was surfaced as a hard `SolidLSPException`
-    instead of being retried. Per the LSP spec, `ContentModified` means the server discarded a
-    stale, in-flight computation because the workspace changed underneath it, not that the request
-    itself is invalid; well-behaved clients are expected to retry. `LanguageServerInterface.send_request`
-    now retries such requests a bounded number of times before giving up, for all language servers
-    (this was flaking `test_find_symbol[rust_add_function]` on windows-latest, where rust-analyzer
-    returns `ContentModified` for hover requests cancelled by concurrent indexing). #1724
+    instead of being retried, as the spec expects for requests a client declared it will reissue.
+    `send_request` now retries such responses for methods declared via a server's
+    `retryOnContentModified` capability; rust-analyzer declares `textDocument/hover`, fixing the
+    flaky `test_find_symbol[rust_add_function]` on windows-latest. #1724
 
 * JetBrains:
   - Allow external files from dependencies (specified via references like "<ext:FileUtil.class|472e0a13>") to be
--- a/src/solidlsp/language_servers/rust_analyzer.py
+++ b/src/solidlsp/language_servers/rust_analyzer.py
@@ -413,6 +413,10 @@ class RustAnalyzer(SolidLanguageServer):
                             "textDocument/semanticTokens/full",
                             "textDocument/semanticTokens/range",
                             "textDocument/semanticTokens/full/delta",
+                            # rust-analyzer's internal cancellation tracking does not cover hover,
+                            # so hover requests it cancels while indexing come back ContentModified;
+                            # we retry them ourselves (see LanguageServerInterface.send_request). #1724
+                            "textDocument/hover",
                         ],
                     },
                     "regularExpressions": {"engine": "ECMAScript", "version": "ES2020"},
--- a/src/solidlsp/ls.py
+++ b/src/solidlsp/ls.py
@@ -3203,4 +3203,17 @@ class SolidLanguageServer(ABC):
         """
         Create the InitializeParams object to send to the language server during initialization.
         """
-        return self._create_initialize_params_builder().with_base_options(self._create_base_initialize_params()).build()
+        params = self._create_initialize_params_builder().with_base_options(self._create_base_initialize_params()).build()
+        self._register_content_modified_retry_methods(params)
+        return params
+
+    def _register_content_modified_retry_methods(self, params: InitializeParams) -> None:
+        """
+        Reads back `general.staleRequestSupport.retryOnContentModified` from the InitializeParams
+        we are about to send and registers it with the underlying `LanguageServerInterface`, so
+        that `send_request` only retries `ContentModified` (-32801) responses for methods we have
+        actually told the server we will retry (see `LanguageServerInterface.send_request`).
+        """
+        general_capabilities = cast(dict, params.get("capabilities", {})).get("general", {})
+        retry_methods = general_capabilities.get("staleRequestSupport", {}).get("retryOnContentModified", [])
+        self.server.set_content_modified_retry_methods(retry_methods)
--- a/src/solidlsp/ls_process.py
+++ b/src/solidlsp/ls_process.py
@@ -18,7 +18,7 @@ from solidlsp.ls_config import Language
 from solidlsp.ls_exceptions import SolidLSPException
 from solidlsp.ls_request import LanguageServerRequest
 from solidlsp.lsp_protocol_handler.lsp_requests import LspNotification
-from solidlsp.lsp_protocol_handler.lsp_types import ErrorCodes
+from solidlsp.lsp_protocol_handler.lsp_types import ErrorCodes, LSPErrorCodes
 from solidlsp.lsp_protocol_handler.server import (
     ENCODING,
     LSPError,
@@ -35,6 +35,13 @@ from solidlsp.lsp_protocol_handler.server import (
 from solidlsp.util import subprocess_util
 
 log = logging.getLogger(__name__)
+
+# Per the LSP spec, `ContentModified` (-32801) means the server discarded a stale, in-flight
+# computation because the workspace changed underneath it, not that the request itself is
+# invalid. A well-behaved client is expected to retry such requests rather than treat them as
+# a hard failure, so we do so here, generically, for all requests to any language server.
+_CONTENT_MODIFIED_MAX_ATTEMPTS = 5
+_CONTENT_MODIFIED_RETRY_DELAY = 0.2
 
 
 class LanguageServerTerminatedException(Exception):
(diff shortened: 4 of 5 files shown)


arimu1:
Thanks for the detailed review — pushed `8fa67596` addressing all of it.

**Chose capability-gated retry (option a).** Added `LanguageServerInterface.set_content_modified_retry_methods()`; `SolidLanguageServer._create_initialize_params()` now reads back each server's own `general.staleRequestSupport.retryOnContentModified` list right after building it and registers it with the interface. `send_request` only retries `ContentModified` for a method that's actually in that list, so the mechanism is a no-op wherever a server declares nothing (most servers today) and never silently contradicts a declared capability. rust-analyzer now declares `textDocument/hover` alongside its existing three `semanticTokens/*` methods — confirmed that's the method #1724's repro (`test_find_symbol[rust_add_function]` with `include_info=True`) actually needs, via `LanguageServerSymbolRetriever._request_info` → `request_hover`.

**#3 (non-idempotent requests):** resolved by the same allowlist — no server declares `workspace/executeCommand`, so it's never retried. Added a regression test (`test_content_modified_is_not_retried_for_undeclared_method`) covering exactly this.

**#1 (pyrefly double-retry):** left `pyrefly_server.py` untouched rather than narrowing its retry tuple as literally suggested. Reasoning: pyrefly never declares `general.staleRequestSupport` at all, so under the allowlist the base-class retry is now a complete no-op for every pyrefly request — pyrefly's own `_install_mutation_retry` wrapper remains the sole handler for both `RequestCancelled` and `ContentModified`, exactly as it was before this PR touched anything. Narrowing pyrefly's tuple to drop `ContentModified` would have removed retry coverage for it entirely once the base layer stopped covering it too — a regression, not a fix. Verified empirically by reproducing pyrefly's exact wrapping pattern against a scripted double: 5 dispatches total (pyrefly's own `max_retries`), not 5×3.

**#4 (latency):** reduced `_CONTENT_MODIFIED_MAX_ATTEMPTS` from 5 to 3, and it's now scoped to allowlisted methods only. Worst case for a `textDocument/hover` request against rust-analyzer is 2 extra attempts: `2 * request_timeout + 0.4s` added latency, down from `4 * request_timeout + 0.8s`.

**#5:** restructured `send_request` around a `_send_request_once` helper — the first attempt happens unconditionally, retries (if the method is allowlisted) happen in a bounded loop, and the final `raise` is reached by ordinary control flow instead of a trailing `raise AssertionError("unreachable")` after the loop. Retry tests now monkeypatch `time.sleep` (suite dropped from ~1.2s of real sleep to ~0.01s) and gained 3 new cases: gating denies retry for an undeclared method (`workspace/executeCommand`), gating denies retry when nothing was ever registered, and the existing success/exhaustion cases were updated to register `textDocument/hover` explicitly.

Also trimmed the CHANGELOG entry from 7 lines to 4, closer to the neighboring `gopls` entry.

Real output, this session:
```
$ uv run pytest test/solidlsp/test_content_modified_retry.py -v
6 passed in 0.01s
$ uv run ruff check src/solidlsp/ls_process.py src/solidlsp/ls.py src/solidlsp/language_servers/rust_analyzer.py test/solidlsp/test_content_modified_retry.py
All checks passed!
$ uv run ty check src/serena src/solidlsp
All checks passed!
$ uv run ty check test --exclude test/resources
All checks passed!
```
Also re-ran the full non-language-gated `test/solidlsp` unit suite (76 tests, includes real Python-language-server integration tests exercising `_create_initialize_params`) — all green.

Re: the `catch-all (macos-latest)` failure on `test_dynamically_activated_mode_is_provided_once_per_session[python]` — not touching it; it's a pre-existing macOS flake (red on `main` at the base commit too, green on ubuntu/windows, and the failing assertion doesn't involve any LSP request).

opcode81:
Thanks a lot for your contribution!

PR #1734, opened 2026-07-22
fix(util): ensure bare * and ? in non-** glob patterns do not match /

### LINKED ISSUE #1732: glob_match(): bare `*`/`?` match across `/`, contradicting its own docstring
- [x] I have consulted the user guide and verified this isn't a config/workflow question.
- [x] I have looked for similar issues and discussions, including closed ones (nearest is #1690, a different function in the same file, already fixed).

- [x] Meaningful title/description above.
- [x] Reproduction steps below.
- [x] Setup details below.

`glob_match()` in `src/serena/util/text_utils.py` documents:

```
- * matches any number of characters except /
- ? matches a single character except /
```

but for any pattern that doesn't contain `**`, the implementation is just:

```python
else:
    # Simple pattern without **, use fnmatch directly
    return fnmatch.fnmatch(path, pattern)
```

`fnmatch.translate()` has no path-separator concept: `*` compiles to `.*` and `?` to `.`, both `/`-crossing. Only the `**` branch gets slash-aware handling (via three separate `fnmatch.translate` fallbacks); the plain `*`/`?` case silently does not honor the contract the docstring promises.

Repro (pure function, no LSP/project needed):

```python
>>> from serena.util.text_utils import glob_match
>>> glob_match('*.py', 'foo.py')
True
>>> glob_match('*.py', 'src/foo.py')
True   # docstring says this should be False
>>> glob_match('a?b.py', 'a/b.py')
True   # docstring says this should be False
```

Confirmed at the call sites that actually expose this to users: `GlobMatcher` (backs `search_for_pattern`'s `paths_include_glob`/`paths_exclude_glob`) and `ReplaceInFilesTool._collect_files` (`replace_in_files`'s same two parameters) both call `glob_match` directly with no extra path-separator handling. So a user who asks for `paths_include_glob="*.py"` expecting "top-level .py files only" (standard gitignore/shell/ripgrep semantics, matching the docstring) silently also gets every `.py` file in every subdirectory. For `paths_exclude_glob` it's the opposite failure: files the user expected to keep get silently dropped too.

This looks like a regression, not an untested corner: before commit 35f2aaa5 ("Fixes in glob matching, extended tests"), path filtering used `pathspec.PathSpec`/`GitWildMatchPattern` (genuinely slash-aware for bare `*`). That commit replaced it with the hand-rolled `fnmatch`-based version above and carried the docstring forward unchanged, but the new non-`**` branch doesn't honor it. `git log -G'def glob_match'` shows only that one commit ever touching the function, and the existing `TestGlobMatch` suite never exercises a bare `*`/`?` pattern against a path containing `/`.

**Setup:** verified against HEAD `9debf774` (current main), pure-Python repro above, no project/LSP client involved. This is a self-contained bug in `text_utils.py`, not something project- or client-specific.

Happy to send a PR (restore path-aware translation for the non-`**` branch, e.g. hand-translate `*`→`[^/]*`/`?`→`[^/]`, or reintroduce `pathspec`, plus tests for `*`/`?` against nested paths) if that's a direction you'd want, let me know.

arimu1:
Fixes #1732.

### Scope of Changes
- Updated `glob_match` in `src/serena/util/text_utils.py` so that for glob patterns without `**`, bare `*` translates to `[^/]*` and `?` translates to `[^/]`, preventing wildcards from matching path separators `/`.
- Added unit test cases in `test/serena/test_text_utils.py` verifying that `*.py`, `a?b.py`, `src/*.py`, and `*agent.py` do not match paths across `/`.
- Updated `CHANGELOG.md`.

### Testing
- Ran pytest on `test/serena/test_text_utils.py` (90 tests passed).
- Verified `ruff check` and `ruff format` pass cleanly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Status of the `main` branch. Changes prior to the next official version change w
   - Add Grok Build support (context `grok`, setup CLI, hooks)
   - The `languages` key in project configurations was changed to `language_servers` to better reflect
     the actual semantics (configurations are automatically migrated)
-  - Fix: `glob_match` bare `*` and `?` in non-`**` patterns matched across `/`, contradicting its documented behavior #1732
+  - Fix: glob matching bare `*` and `?` in non-`**` patterns matched across `/`, contradicting documented behaviour #1732
 
 * Language Servers: 
   - Allow language server priorities to be configured in `serena_config.yml` (for auto-detection during 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Status of the `main` branch. Changes prior to the next official version change w
   - Add Grok Build support (context `grok`, setup CLI, hooks)
   - The `languages` key in project configurations was changed to `language_servers` to better reflect
     the actual semantics (configurations are automatically migrated)
+  - Fix: `glob_match` bare `*` and `?` in non-`**` patterns matched across `/`, contradicting its documented behavior #1732
 
 * Language Servers: 
   - Allow language server priorities to be configured in `serena_config.yml` (for auto-detection during 
--- a/src/serena/config/serena_config.py
+++ b/src/serena/config/serena_config.py
@@ -33,7 +33,7 @@ from serena.constants import (
     SERENA_MANAGED_DIR_NAME,
 )
 from serena.util.inspection import compute_language_server_support_composition
-from serena.util.text_utils import glob_match
+from serena.util.text_utils import GlobMatcher
 from serena.util.yaml import YamlCommentNormalisation, load_yaml, normalise_yaml_comments, save_yaml, transfer_yaml_comments
 from solidlsp.ls_config import LanguageServerId
 
@@ -1437,7 +1437,7 @@ class SerenaConfig(SharedConfig, ModeSelectionDefinitionWithBaseModes):
         """
         project_root_str = str(project_root)
         for pattern in self.trusted_project_path_patterns:
-            if glob_match(pattern, project_root_str):
+            if GlobMatcher(pattern).matches(project_root_str):
                 return True
         return False
 
--- a/src/serena/tools/file_tools.py
+++ b/src/serena/tools/file_tools.py
@@ -16,10 +16,9 @@ from serena.tools import SUCCESS_RESULT, EditedFileContext, EditingToolWithDiagn
 from serena.util.file_system import scan_directory
 from serena.util.text_utils import (
     ContentReplacer,
+    GlobMatcher,
     MultiFileContentReplacer,
     ReplacementOccurrence,
-    expand_braces,
-    glob_match,
 )
 from solidlsp.ls_utils import TextUtils
 
@@ -350,13 +349,13 @@ class ReplaceInFilesTool(EditingToolWithDiagnostics):
                 is_ignored_file=self.project.is_ignored_path,
                 relative_to=self.get_project_root(),
             )
-        include_patterns = expand_braces(paths_include_glob.strip()) if paths_include_glob.strip() else None
-        exclude_patterns = expand_braces(paths_exclude_glob.strip()) if paths_exclude_glob.strip() else None
+        include_glob_matcher = GlobMatcher(paths_include_glob.strip()) if paths_include_glob.strip() else None
+        exclude_glob_matcher = GlobMatcher(paths_exclude_glob.strip()) if paths_exclude_glob.strip() else None
         files: list[tuple[str, str]] = []
         for path in sorted(rel_paths):
-            if include_patterns and not any(glob_match(p, path) for p in include_patterns):
+            if include_glob_matcher and not include_glob_matcher.matches(path):
                 continue
-            if exclude_patterns and any(glob_match(p, path) for p in exclude_patterns):
+            if exclude_glob_matcher and exclude_glob_matcher.matches(path):
                 continue
             try:
                 files.append((path, self.project.read_file(path)))
--- a/src/serena/util/text_utils.py
+++ b/src/serena/util/text_utils.py
@@ -185,83 +185,6 @@ def search_text(
     return matches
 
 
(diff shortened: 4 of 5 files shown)


arimu1:
CI note: the `catch-all (ubuntu-latest)` failure is `test_serena_agent.py::TestPromptProvision::test_dynamically_activated_mode_is_provided_once_per_session[python]` failing with a `ToolCallError: TimeoutError` — an unrelated flaky timeout in the python language-server integration test (the run took 19 min). This change only touches `test_text_utils`, which passes on all three platforms; `catch-all` is fully green on macOS and Windows. Happy to rebase or re-trigger if a re-run helps.

arimu1:
Rebased onto latest `upstream/main` to resolve the merge conflict (only conflict was in `CHANGELOG.md`, resolved by keeping both entries). Glob-matching unit tests (`test/serena/test_text_utils.py`, 95 tests) and `ruff check` / `ty check` all pass locally on the rebased branch.

AmirF194 (commented):
The *//? slash-boundary fix itself looks solid and the test coverage for those cases (src/*.py, a?b.py, etc.) is thorough.

While testing edge cases against `_translate_glob_no_double_star`, I found one regression: a reversed character-class range that the old `fnmatch.fnmatch` path tolerated now raises instead of returning `False`.

Observed on PR head `52cf8552a023f10ae56e543cb6d4c989dd956892` (clean `python:3.12-slim` container, `uv sync --extra dev`):

    >>> import fnmatch
    >>> fnmatch.fnmatch('z', '[z-a]')
    False
    >>> from serena.util.text_utils import glob_match
    >>> glob_match('[z-a]', 'z')
    re.error: bad character range z-a at position 5

`fnmatch.translate` repairs an invalid range like this instead of emitting invalid regex syntax (its "Remove empty ranges, invalid in RE" step). The new hand-rolled translator in `text_utils.py` passes the bracket contents straight into a Python character class, so anything `re` rejects propagates as an exception from `glob_match` instead of a plain non-match.

This is reachable through public tool arguments: `SearchForPatternTool` and `ReplaceInFilesTool`'s `paths_include_glob`/`paths_exclude_glob` (file_tools.py:357-359) call `glob_match` on each pattern directly. A reversed range is an easy typo to hit in practice, e.g. `[a-Z]` (uppercase sorts before lowercase in ASCII, so this looks fine at a glance but is reversed) or `[9-0]`. Previously that pattern just matched nothing; now the tool call errors out instead.

Might be worth catching `re.error` around the `re.match` call in `glob_match`'s simple-pattern branch (or validating/repairing the range in `_translate_glob_no_double_star`) and falling back to no-match, to keep parity with the old graceful behavior for malformed patterns. Not blocking the slash-boundary fix, just flagging before it lands.

opcode81:
@AmirF194 thanks for the analysis.
While it is strictly a regression, I don't think it is practically relevant. We don't explicitly promise the more lenient behaviour, and therefore I think it is acceptable.

wisc17:
I think it is not acceptable.
Using **replace_in_files** is highly bugged and dangerous
Here is what claude said:

```
The probe was conclusive by construction: the needle (TranslateModule) exists only in docs/**/*.md (verified: 0 hits in src/),
the glob was src/app/**/*.component.ts — so a correctly-scoped implementation must report 0 replacements.
Instead it replaced 80 occurrences across 20  markdown files.
```
replacing 80 occurrences across 20 files is by far NOT acceptable
please fix

P.S. may be more related to #1732 
my case is on Windows and I don't think this PR covers it — my pattern was src/app/**/*.component.ts (the ** branch, unchanged here), and the paths Serena reports use backslashes, which the new [^/]* translation doesn't exclude. The filter seems to fail open entirely (.md files replaced under a *.component.ts glob). Happy to file it as a separate issue with the full repro if that's preferred.

opcode81:
> I think it is not acceptable.

@wisc17 I think you must have misread my statement.

opcode81:
I have added two additional commits to this PR:
* Removal of obsolete additional glob translation for `search_text`
* Unification: always use regex translation, generalising the translation to include "**" patterns

opcode81:
@wisc17 please clarify what your error case is.
The expression `"src/app/**/*.component.ts"` does not incorrectly match `"docs/foo/bar.md"` in any implementation, as far as I can tell (neither `main`, nor versions on this PR's branch).
Please provide the exact pattern you used and at least one file's path it incorrectly matched.

wisc17:
@opcode81 You're right on both counts. apologies for the noise.

First, I did misread your earlier comment: "acceptable" was about the reversed-range regression @AmirF194 raised, not replace_in_files. My mistake.                                                                                                                             

Second, I re-tested against the current version and the scoping is working correctly.
The fault was entirely on my side. My original probe passed the glob as a "relative_paths" argument, but the tool has no such parameter (it's paths_include_glob, with relative_path for a single file/dir).
```json
{
  "relative_paths": ["src/app/**/*.component.ts"],
  "mode": "literal",
  "needle": "TranslateModule",
  "repl": "TranslateModule_SCOPETEST"
}
```
The unknown argument was silently ignored, so the call ran with no path restriction at all, across the whole project, which is exactly why it touched all 20 docs/**/*.md files. It was never a matching problem.

Re-running it correctly (dry run, current version) confirms scoping is fine:

- paths_include_glob="src/app/**/*.component.ts" on a needle that exists only under docs/ → Found 0 occurrence(s) — the docs/*.md files are correctly excluded, backslash paths and all.
- no glob → the docs\...\*.md occurrence shows up, as it should.

So please disregard my "please fix"

The only thing I'd gently flag: an unknown parameter is silently ignored rather than rejected, so a client that passes a hallucinated arg like "relative_paths" gets a silent project-wide replace instead of an error. Your dry_run / expected_count guards already cover this well in practice. it only bit me because it was skipped. Thanks for your patience, and for the quick follow-up.

opcode81:
Added some additional commits to consolidate.

opcode81:
@arimu1 thanks for providing a good starting point.

PR #1738, opened 2026-07-22
fix(hooks): emit valid Codex Stop-hook JSON from cleanup (#1533)

### LINKED ISSUE #1533: Codex Stop hook: `serena-hooks cleanup --client=codex` emits invalid Stop-hook stdout

arimu1:
Closes #1533.

## Problem

When `serena-hooks cleanup --client=codex` is configured as a Codex `Stop`
hook, Codex reports:

```
Stop hook (failed)
```

Two causes:
- `SessionEndCleanupHook.execute()` only ran `shutil.rmtree` and printed
  nothing, so stdout was empty — but Codex requires a Stop hook to print
  valid Stop-hook JSON.
- `Hook.__init__` raised `ValueError` when the payload had no `session_id`,
  so a Stop payload without one produced a Python traceback.

## Fix

- `Hook.__init__` gains a `require_session_id` flag (default `True`, so every
  existing hook is unchanged) and tolerates empty stdin.
- `SessionEndCleanupHook` opts out of the requirement — cleanup is a
  no-op-safe operation, so a missing session id simply means there is
  nothing to remove — and always emits `{"continue": true}` for the Codex
  client. Other clients stay silent, preserving current behavior.

## Tests

Added unit tests (missing session id / empty stdin no longer raise) and CLI
tests asserting `cleanup --client=codex` prints `{"continue": true}` both
with and without a session id. `test_hooks.py` green (90/90); `ruff` and
`ty` type-check clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,9 @@ Status of the `main` branch. Changes prior to the next official version change w
   - Fix: `serena-hooks cleanup --client=codex` used as a Codex `Stop` hook now always prints valid
     Stop-hook JSON (`{"continue": true}`) and no longer crashes when the payload has no session id or
     stdin is empty, so Codex stops reporting `hook returned invalid stop hook JSON output` (#1533).
+  - Fix: `serena-hooks activate` (`SessionStartActivateProjectHook`) no longer requires a session id;
+    it only emits a static reminder message and never touches session state, so a client omitting
+    `session_id` on `SessionStart` no longer crashes it the same way cleanup used to (#1533 review).
 
 
 # v1.6.1 (2026-07-21)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,9 @@ Status of the `main` branch. Changes prior to the next official version change w
   - PreToolUse remind hook: coerce non-string shell command values instead of failing, and recognize
     `target_file`/`targetFile` file-path keys (shared payload parsing, applies to all hook clients).
   - Fix hook input parsing for clients that emit raw control characters in JSON string values #1743.
+  - Fix: `serena-hooks cleanup --client=codex` used as a Codex `Stop` hook now always prints valid
+    Stop-hook JSON (`{"continue": true}`) and no longer crashes when the payload has no session id or
+    stdin is empty, so Codex stops reporting `hook returned invalid stop hook JSON output` (#1533).
 
 
 # v1.6.1 (2026-07-21)
--- a/src/serena/hooks.py
+++ b/src/serena/hooks.py
@@ -534,6 +534,12 @@ class PreToolUseRemindAboutSymbolicToolsHook(PreToolUseHook):
 
 
 class SessionStartActivateProjectHook(Hook):
+    def __init__(self, client: HookClient):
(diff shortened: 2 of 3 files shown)


AmirF194 (commented):
The require_session_id opt-in/opt-out split is clean and the new tests cover both the missing-session-id and empty-stdin cases.

One thing to check: SessionStartActivateProjectHook (src/serena/hooks.py:531) doesn't override __init__, so it still inherits require_session_id=True from the base Hook. Docker-verified on this PR's head (dd9eb68):

    printf '%s' '{"hookEventName":"SessionStart"}' | serena-hooks activate --client=codex

raises the same uncaught ValueError: Session ID is required in the hook input data that #1533 reports for cleanup. I don't have evidence Codex ever omits session_id on its SessionStart event, only that #1533's own repro shows it can omit one on Stop, so this might not happen in practice. But if it ever does, activate would crash the same way cleanup used to before this fix. Might be worth deciding whether activate should opt out the same way, or whether SessionStart is guaranteed to always carry a session id.

arimu1:
Good catch — confirmed. `SessionStartActivateProjectHook` didn't override `__init__`, so it inherited `require_session_id=True` from the base `Hook`, even though `execute()` never touches `session_persistence_dir` (it only emits a static reminder message). Pushed a fix (904b071e) that opts it out of the session-id requirement the same way `SessionEndCleanupHook` does, plus tests for missing session_id, empty stdin, and the CLI `activate` path.

On the failing `jvm (windows-latest)` job: the failure (`test_safe_delete_symbol_succeeds_when_no_references[java_model_user]`) is a Java LSP symbol-deletion assertion, unrelated to the hooks.py changes here. I checked recent main CI history and `jvm (windows-latest)` has failed there before too (for a different reason, a Clojure artifact-resolution error), so this looks like a pre-existing flake rather than something introduced by this PR.

MischaPanch:
On one hand I would be happy to merge this to not have a non-functioning codex hook. On the other hand, just giving up and not cleaning up the resources we create does not seem like an appropriate solution, and this PR would cement this as possibility in our code. 

This really is an issue with the codex client, it should absolutely emit info about which session is ending.

Maybe one of you has an idea on how to solve this instead of just disabling the error/warning? I know that codex has been crippling people's SSDs due to excessive logging to disk, and compared to that leaving some tiny files on disk seems harmless, but I still wouldn't want Serena to go down this path

PR #1740, opened 2026-07-22
Security hardening: make shell tool opt-in, confine its cwd, bump CVE-affected deps

jcogilvie:
## What this does

Security hardening in two closely related areas. Opening as a **draft** to get maintainer feedback on the approach and scope before finalizing.

### 1. Scope `ExecuteShellCommandTool`

- Marked `ToolMarkerOptional`, so it is **no longer in the default tool set** and must be opted in via `included_optional_tools`. It is re-enabled by default only in the `agent` and `desktop-app` contexts, which intend to provide shell access. Every IDE/CLI coding context stays shell-free — as before, but now off-by-default rather than relying on each context's explicit `excluded_tools` entry.
- Its `cwd` argument is now confined to the project root via a new `resolve_cwd_within_project()` helper. Containment is component-aware (`Path.is_relative_to` on `realpath`-resolved paths), so sibling paths like `<root>-evil` and in-project symlinks pointing outside are rejected.

**Honest scoping note:** the `cwd` confinement is a best-effort ergonomic guard, **not** a security boundary. Under `subprocess.Popen(shell=True)` the command string itself can reference any absolute path or `cd` elsewhere. The meaningful control is that the tool is now disabled by default. I've written it up this way rather than overstating it.

### 2. Bump dependencies flagged with HIGH CVEs

- `mcp` 1.27.0 → 1.28.1 (session auth bypass, missing origin validation, cross-client task access) — direct dependency and the most relevant, since Serena is an MCP server.
- `pillow` → 12.3.0 (transitive via `pystray`).
- `click` → 8.3.3 (transitive). The specific CVE is in `click.edit()`, which Serena never calls — it uses `click` only for CLI arg parsing — so the vulnerable path is not reachable; bumped as defense-in-depth.

Existing `test_mcp.py` (64 tests exercising `make_mcp_tool` for every tool) passes on `mcp` 1.28.1, confirming API compatibility.

## Notes for reviewers

- Happy to split this into two PRs (deps vs. tool scoping) if you'd prefer.
- The `execute_shell_command` entries under `excluded_tools` in the IDE contexts are now technically redundant (the tool isn't in the default set), but I left them as harmless defense-in-depth. The entries in the `planning` and `onboarding` **modes** are intentionally kept — they are load-bearing when those modes compose over the `agent`/`desktop-app` contexts (a new regression test locks this in).
- New tests live in `test/serena/tools/test_cmd_tools.py`.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ Status of the `main` branch. Changes prior to the next official version change w
   - During project creation, language composition percentages are now computed relative to the total number 
     of recognised source files instead of all files, i.e. unrecognised files are ignored in the percentage 
     computation.
+  - Security: bump `mcp` 1.27.0 -> 1.28.1 (3 HIGH CVEs) and pin transitive `pillow==12.3.0` (6 HIGH CVEs)
+    and `click==8.3.3`.
 
 * CLI:
   - Fix `--project-from-cwd` hijacking git worktrees nested under a Serena project. `find_project_root`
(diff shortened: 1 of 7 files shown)


opcode81:
A shell tool is an essential part of agentic coding; making it formally optional is purely cosmetic, and trying to confine the cwd or otherwise trying to make it "safe" is a hopeless endeavour.
If security is a concern, sandboxing is the only solution; see our [security documentation](https://oraios.github.io/serena/02-usage/070_security.html).

The dependency version changes are fine; `mcp` has already been upgraded on main.

opcode81:
Thanks for your interest in contributing to Serena, but I will close this because of the reasons stated in my earlier comment.

PR #1742, opened 2026-07-22
Fix project CLI language server aliases

TyceHerrman:
## Summary

- Explicitly bind the `--ls` and `--language` aliases to the existing `language` callback parameter for `project create` and `project index`.
- Add regression coverage for the aliases on both commands.
- Document the fix in the changelog.

## Root cause

PR #1739 added `--ls` before `--language` in both Click option declarations. Click consequently inferred `ls` as the callback keyword, while both callbacks still accepted `language`, causing every invocation to fail with `unexpected keyword argument 'ls'`.

## Impact

Both aliases now reach the existing callback parameter, restoring `project create` and `project index` without changing their user-facing options.

## Validation

- `python -m ruff format --check src/serena/cli.py test/serena/test_cli_project_commands.py`
- `python -m ruff check src/serena/cli.py test/serena/test_cli_project_commands.py`
- `python -m ty check src/serena/cli.py test/serena/test_cli_project_commands.py`
- Focused pytest regression coverage: 3 passed

## Checklist

### DIFF
--- a/src/serena/cli.py
+++ b/src/serena/cli.py
@@ -733,9 +733,10 @@ class ProjectCommands(AutoRegisteringGroup):
     @click.option(
         "--ls",
         "--language",
+        "language",
         type=str,
         multiple=True,
-        help="Programming language(s); inferred if not specified. Can be passed multiple times.",
+        help="Language server(s); inferred if not specified. Can be passed multiple times.",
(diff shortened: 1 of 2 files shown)


opcode81:
Thanks.

PR #1746, opened 2026-07-23
Fix: execute_shell_command has no timeout, wedging the task executor on a hung command

AmirF194:
This addresses Finding #2 of #1251 ("Shell execution with no timeout"). The other five findings in that issue are broader design questions (memory path traversal, cross-project memory, directory activation, prompt injection surface, system prompt exposure); this PR is scoped to the shell timeout only and does not close the issue.

Root cause: `execute_shell_command` (src/serena/util/shell.py) calls `process.communicate()` with no timeout, so a hung or long-running command blocks forever. The impact is bigger than the single tool call: `TaskExecutor._process_task_queue` (src/serena/task_executor.py) runs on one dispatcher thread that waits on the current task (`task.wait_until_done(timeout=task.timeout)`, and `ExecuteShellCommandTool`'s call site never passes a `timeout`, so it waits indefinitely) before it will pick up the next queued task. `Tool.apply_ex` separately waits on the same future with `serena_config.tool_timeout` (240s default) and turns an expiry into an error message for the caller, which makes it look like the situation is recovered, but the dispatcher thread underneath is still blocked on the original hung subprocess and never starts any task queued after it.

I confirmed this with a standalone script against the real `TaskExecutor`/`execute_shell_command` classes (no mocks): issuing a `sleep 30` shell task with no timeout, then a trivial second task right after. The caller's own 5s wait on the first task times out as expected, but the second task is still not done 3s later, the dispatcher never even started it.

Fix: `execute_shell_command` takes an optional `timeout` parameter (default `None`, preserving current behavior for its other callers in `client_setup.py`). On expiry it terminates the whole process tree via the existing `terminate_process_tree_with_kill_fallback` helper (already used in `agent.py` and `ls_process.py`) rather than just the top-level shell process, since `shell=True` means the actual long-running command is a child of the shell. `TimeoutExpired` then propagates, which `Tool.apply_ex`'s existing exception handling already turns into a clean error message for the caller. `ExecuteShellCommandTool` now passes `timeout=self.agent.serena_config.tool_timeout`, so a shell command is bounded by the same value the tool call was already conceptually bounded by.

Verified in a clean Docker container (python:3.12-slim, `uv sync --extra dev --locked`):
- New regression test (`test/serena/util/test_shell.py`) fails on main (`TypeError: execute_shell_command() got an unexpected keyword argument 'timeout'`) and passes on this branch, including a case where the hung process is a child of a `sh -c` pipeline (confirms the whole tree is killed, not just the top pid).
- The same standalone dispatcher-wedge script above, re-run against the fixed code with `timeout=2`, shows the second queued task now completes shortly after the first fails, instead of never starting.
- `uv run poe lint`, `uv run poe type-check`, and `codespell` on the changed files are all clean.
- Full `uv run poe test` catch-all matrix: 1007 passed, 32 failed, 97 errors, all failures/errors are in scss/solidity/typescript/vue language-server tests that need Node-based LSP toolchains not installed in this minimal image; none touch `serena/util/shell.py`, `serena/tools/cmd_tools.py`, or `serena/task_executor.py`. The directly relevant suites (`test/serena/util/test_shell.py`, `test/serena/config/test_client_setup.py`, `test/serena/test_cli_setup.py`, `test/serena/config/test_context_mode.py`, all other `execute_shell_command` callers) pass in full.

Not verified: behavior under the real MCP/agent stack end to end (the dispatcher-wedge proof above uses `TaskExecutor` and `execute_shell_command` directly, not a live agent session), and the full pytest matrix on the Node-based language servers (environment limitation, not exercised by this change).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ Status of the `main` branch. Changes prior to the next official version change w
   - The `languages` key in project configurations was changed to `language_servers` to better reflect
     the actual semantics (configurations are automatically migrated)
   - Fix: glob matching bare `*` and `?` in non-`**` patterns matched across `/`, contradicting documented behaviour #1732
+  - Fix: `execute_shell_command` had no timeout, so a hung command blocked indefinitely. Since the
+    agent's task executor runs on a single dispatcher thread that waits for the current task before
+    picking up the next, this also stalled every subsequent tool call, not just the shell command
+    itself. `ExecuteShellCommandTool` now bounds shell commands with the existing `tool_timeout`
+    config value and terminates the process tree (not just the top-level shell) on expiry #1251
 
 * Language Servers: 
   - Allow language server priorities to be configured in `serena_config.yml` (for auto-detection during 
--- a/src/serena/tools/cmd_tools.py
+++ b/src/serena/tools/cmd_tools.py
@@ -47,6 +47,6 @@ class ExecuteShellCommandTool(Tool, ToolMarkerCanEdit):
                         f"Specified a relative working directory ({cwd}), but the resulting path is not a directory: {_cwd}"
                     )
 
-        result = execute_shell_command(command, cwd=_cwd, capture_stderr=capture_stderr)
+        result = execute_shell_command(command, cwd=_cwd, capture_stderr=capture_stderr, timeout=self.agent.serena_config.tool_timeout)
         result = result.model_dump_json()
         return self._limit_length(result, max_answer_chars)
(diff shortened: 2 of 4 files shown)


opcode81:
While adding a timeout to the shell execution is reasonable, I found that a more general fix was needed to unblock the dispatcher thread for all tool executions that exceed the timeout. It was not appropriate to use a timeout only for the resolution of the task's future in `apply_ex` while having no timeout for the task execution itself.
This is addressed in #1751.

AmirF194:
Makes sense, a per-tool timeout on the future without bounding the dispatcher's own wait doesn't fix the general case. Closing this in favor of #1751.

PR #1747, opened 2026-07-23
fix(fsharp): correct selectionRange for module declarations (#925)

### LINKED ISSUE #925: F# LS returns info of general module instead of the specific module asked for on hover
In our testsuite we have

```
module Calculator
```

When requesting info (through hover) we get the docstring of the builtin module instead of hover info about the Calculator module. As a result our test fails (we expect the symbol name to be present in the output of request hover):

```
assert 'Calculator' in 'module\n**Description**\n\n\nUsed to associate a name with a group of related types, values, and functions, to logically separate it from other code.'
```

@dsyme maybe a bug in the F# language server?

AmirF194:
## Problem

FsAutoComplete (Ionide's F# LSP) reports `selectionRange` for a `module <Name>` declaration
pointing at the `module` keyword itself, not at `<Name>`. `LanguageServerSymbol.line`/`.column`
(src/serena/symbol.py:320-333) build hover/reference lookups directly from
`selectionRange.start`, so hovering a module returns the LSP server's info for the generic
`module` keyword instead of the module's own docs, as reported in #925 and confirmed in-thread
by dsyme (F# tooling) and MischaPanch.

## Fix

`FSharpLanguageServer.request_document_symbols` now overrides the base implementation to
correct `selectionRange` for `Module` symbols, matching the declaration line with a regex to
recover the real identifier column (handles `module Foo`, `module Foo =`, `module rec Foo`, and
dotted names). This mirrors the fix already shipped for fortls in `fortran_language_server.py`,
which solves the identical selectionRange-correction problem for Fortran. Only `Module` symbols
are touched; other kinds already select the identifier correctly, per MischaPanch's comment on
the issue.

## Verification

F# language server tests are unconditionally disabled in this repo (`test/conftest.py`,
category 1: "F# language server is currently unreliable"), so a live-FsAutoComplete regression
test cannot run here or on CI. I added a synthetic-payload unit test,
`test/solidlsp/fsharp/test_fsharp_module_selection_range.py`, that instantiates
`FSharpLanguageServer` without running `__init__` (no process, no dotnet), the same technique
already used in `test_typescript_timeout_policy.py`. It covers a top-level module declaration, a
nested module with `=`, a recursive `module rec` declaration, a selectionRange that is already
correct (must be left alone), a non-Module symbol (must be left alone), and one test wiring
`request_document_symbols` end-to-end with a mocked `super()` call to confirm nested symbols get
fixed too.

Ran the new tests plus `ruff check`, `ruff format --check`, `ty check` (both core and test
configs), and `codespell` against the changed files, all clean. Confirmed the new test fails on
current main (the fix method does not exist, and the wiring test's assertion fails since
selectionRange stays unfixed) and passes on this branch.

Not verified: behavior against a live FsAutoComplete instance, since the F# suite cannot run in
this environment. The regex assumes FsAutoComplete always points selectionRange at column 0 of
the `module` line; I confirmed this only against a synthetic payload matching the issue's
description, not against a real FsAutoComplete process.

Fixes #925

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@ Status of the `main` branch. Changes prior to the next official version change w
     and the request used to map diagnostics onto owning symbols just threw. `AnsibleLanguageServer` now
     overrides that request to return `None` directly, so diagnostics fall back to being grouped under
     the file-level path as already documented #1758
+  - Fix: F#'s `module <Name>` declarations reported a `selectionRange` pointing at the `module`
+    keyword instead of at `<Name>`, so looking up hover/references from a module symbol's position
+    returned the keyword's own docs instead of the module's #925
 
 * JetBrains:
   - `jet_brains_find_symbol`: Disallow wildcard-only search, delegating to overview tool if request is for file
--- a/src/solidlsp/language_servers/fsharp_language_server.py
+++ b/src/solidlsp/language_servers/fsharp_language_server.py
@@ -4,6 +4,7 @@ Provides F# specific instantiation of the LanguageServer class.
 
 import logging
 import os
+import re
 import shutil
 import subprocess
 import threading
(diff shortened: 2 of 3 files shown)


opcode81:
Thanks!

AmirF194:
Thanks for reviewing and merging!

PR #1749, opened 2026-07-23
TypeScript: make the tsserver progress-start grace configurable, raise the default

### LINKED ISSUE #1586: tsserver loads files as inferred projects, not the real tsconfig project
find_referencing_symbols returns 0 for TypeScript while find_symbol/definition work 

Preconditions:

`find_referencing_symbols` / `request_references` returns **0 references** for TypeScript symbols that have many real cross-file callers, while `find_symbol`, `get_symbols_overview`, and `request_definition` all work correctly on the same running server. Python references (pyright) work fine in the same project.

I traced the LSP wire protocol: the references request is well-formed and points at the correct identifier position, tsserver does ~2s of real work, then returns `count=0`. This appears to be the unresolved TS tail of #478. Diagnosis: **tsserver is treating each `didOpen`'d file as a standalone inferred project and never loads the package's `tsconfig.json`**, so find-references has no project graph to search.
Issue details:

## Setup
- Serena: **1.5.3** (latest on PyPI)
- MCP client: Claude Code
- typescript-language-server: **5.1.3**; typescript/tsserver: **5.9.3**
- Node: **v22.20.0**; OS: Linux x86_64 (kernel 6.1)
- Language backend: `typescript` (also reproduced with `typescript_vts`)
- Project: TS monorepo, **no root `tsconfig.json`** — each package owns its own. The target package's `tsconfig.json` is plain `include`-based (`"include": ["src/**/*.ts", ...]`), **no `composite`**, with one project `references` entry to a sibling package. ~21k files total.
## Reproduction

Target: an exported `const fromNumber` in `<pkg>/src/util/mathCore.ts`, with 76 in-package caller files, all under `<pkg>/src/`, imported via plain relative paths (`import { fromNumber } from '../util/mathCore'`).

1. Create a `SolidLanguageServer` (mirroring `ls_manager.create_language_server`) with `repository_root_path` = the package root (so its tsconfig is THE project) and `additional_workspace_folders=[]`.
2. `request_document_symbols(file)` → locate the symbol (works; returns it at line 42, char 13).
3. `request_references(file, line=42, column=13)` → returns `[]`.

## Wire trace (the key evidence)

Request is correct; tsserver returns count=0 after real work:

```
solidlsp -> ls: {"jsonrpc":"2.0","method":"textDocument/references","id":4,
  "params":{"textDocument":{"uri":"file://.../<pkg>/src/util/mathCore.ts"},
  "position":{"line":42,"character":13},"context":{"includeDeclaration":false}}}
solidlsp.ls perf: request_references path=src/util/mathCore.ts elapsed_ms=2029 count=0
```

On the same server: `request_definition` returns a hit, and `request_references` for a **Python** symbol returns correct cross-file callers — so the client plumbing is fine; this is specific to the TS server's find-references.

## What I tried (all still return 0)

- `additional_workspace_folders` set to the package roots
- Package opened **as the LSP root** (its tsconfig primary)
- Pre-opening **all 75 caller files** via `request_document_symbols` + a 30s wait
- `typescript_vts` (vtsls) backend
- `serena project index`
- Raising `INDEXING_PROGRESS_TIMEOUT`
- Confirmed neither the symbol nor its callers sit under an ignored dir name (`node_modules`/`dist`/`build`/`coverage`, re: #1523)

## Suspected cause + questions

tsserver only returns references for files in a **loaded project**. When solidlsp `didOpen`s a `.ts` file it seems to land in an **inferred (single-file) project** instead of tsserver loading the owning `tsconfig.json`, so find-references has nothing to search. Definition still works because it follows imports forward.

1. Does solidlsp signal tsserver to **load the tsconfig project** for an opened file (e.g. `_typescript.configurePlugin`, `openExternalProject`, or relying on upward `tsconfig.json` auto-detection)? On this layout it appears to stay inferred.
2. Is find-references known to require `composite`/project-references or a specific `initializationOptions`? Ours is a plain `include`-based config.
3. Your `test_typescript_cross_package` fixture passes — how does it differ structurally (composite? root tsconfig?) from a no-root-tsconfig, `include`-based package?

AmirF194:
## Root cause

`find_referencing_symbols`/`request_references` can return incomplete results on large TypeScript projects. `_wait_for_indexing_start_or_completion` (`src/solidlsp/language_servers/typescript_language_server.py`) waits only `_INDEXING_START_GRACE_S` (hardcoded 2.0s) for tsserver to *start* emitting `$/progress` after files are opened; if nothing has appeared by then, it assumes no indexing is needed and proceeds immediately. On a large project, tsserver's own project-graph resolution can take longer than that before it even creates the first progress token, so the first cross-file reference query can race a project that is still loading and get back an incomplete result.

Every caller of `_wait_for_indexing_start_or_completion` (`_wait_for_cross_file_references_if_needed` in the base class, and the Svelte companion's `_ensure_svelte_files_indexed_on_ts_server`) always calls `expect_indexing()` immediately beforehand, so the "no progress observed" branch is reached specifically when indexing is already expected. From the client's side, "nothing to do" and "not started yet" look identical for as long as tsserver stays quiet, so this cannot be closed by picking a smarter fixed constant; it can only be made adjustable per project.

## Fix

Add `indexing_start_grace` to `ls_specific_settings["typescript"]`, following the exact pattern already used for `indexing_timeout` and `server_ready_timeout` (a class default plus a `_get_indexing_start_grace()` accessor reading `self._custom_settings`), and use it at the one call site behind `find_referencing_symbols`/`request_references` (`_wait_for_cross_file_references_if_needed`). The default rises from 2.0s to 5.0s. This does not eliminate the race (there is no positive "no indexing needed" signal from tsserver to wait for instead), it makes the false-positive window configurable and, at the default, somewhat wider.

## Why this is likely @xormania's #1586

The issue's own wire trace shows `elapsed_ms=2029 count=0` for the failing `references` request; that is within 30ms of the old 2.0s grace, and @MischaPanch separately confirmed a matching intermittent CI flake (`test_find_symbol_references[typescript_helper_refs]`) on the repo's own, much smaller, TS fixture. Both line up with the grace elapsing before tsserver reported progress, not with project size specifically. This also explains why raising `indexing_timeout` alone (already tried, per the issue) did not help: that setting only bounds the *drain* wait once progress has started; the *start* grace was a separate, hardcoded, unexposed constant until now.

## What is and is not verified

Verified in Docker on HEAD `5e8662b2`:
- `test/solidlsp/test_typescript_timeout_policy.py` (3 new tests, plus all 16 existing ones): all fail against unpatched `_wait_for_indexing_start_or_completion` semantics before this change (`elapsed=2.0098s`/`2.0106s`, i.e. the exact old grace, confirmed on two separate runs against HEAD), all pass after.
- Full existing TypeScript and Svelte suites against real, network-installed `typescript-language-server`/`svelte-language-server` processes (`test/solidlsp/typescript/`, `test/solidlsp/svelte/`): 31 + 16 passed, including the cross-package and cross-file reference tests.
- `ruff check`, `ruff format --check`, `codespell` on the changed files: clean. `ty check src/serena src/solidlsp`: same 39 pre-existing diagnostics as on unpatched HEAD, none in the changed file.

Not verified: the reporter's own ~21k-file monorepo scale (the issue itself carries a "could not reproduce" label at that scale, and this PR does not add a reproduction at that size). The `tsconfig.json` `references`-to-a-sibling-package variable mentioned as an alternative candidate in the issue thread was not ruled out separately; if the incomplete results persist after this change at that scale, that variable is the next thing to check.

Fixes #1586

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,15 @@ Status of the `main` branch. Changes prior to the next official version change w
 * JetBrains:
   - `jet_brains_find_symbol`: Disallow wildcard-only search, delegating to overview tool if request is for file
 
+* Language Servers:
+  - `typescript`: Fix: on large projects, the first `find_referencing_symbols`/`request_references` call
+    could silently race tsserver's project load and return incomplete results, because the fixed 2s
+    grace for tsserver to *start* reporting `$/progress` (distinct from the separate, already
+    configurable `indexing_timeout` used to wait for it to *drain*) was hardcoded and not large enough
+    for projects where the initial project-graph resolution itself takes longer than that. The grace
+    is now `indexing_start_grace` (default 5.0s), configurable the same way as `indexing_timeout` and
+    `server_ready_timeout` #1586
+
 * Hooks:
   - Add `serena-hooks --client=grok`, including Grok-native PreToolUse allow/deny output.
   - PreToolUse remind hook: coerce non-string shell command values instead of failing, and recognize
--- a/docs/02-usage/050_configuration.md
+++ b/docs/02-usage/050_configuration.md
@@ -1118,8 +1118,9 @@ Supported settings:
 | `typescript_version` | `5.9.3` | Override the bundled `typescript` npm package version Serena installs when `ls_path` is not set. |
 | `typescript_language_server_version` | `5.1.3` | Override the bundled `typescript-language-server` npm package version Serena installs when `ls_path` is not set. |
 | `npm_registry` | `null` | Override the npm registry Serena uses for the managed install. |
-| `indexing_timeout` | `30.0` | Timeout in seconds for waiting on tsserver's `$/progress` project-indexing signal (both at startup and before the first cross-file reference query). If indexing does not complete within this window, Serena logs a warning and proceeds anyway. Increase it for very large projects. |
+| `indexing_timeout` | `30.0` | Timeout in seconds for waiting on tsserver's `$/progress` project-indexing signal to *drain* once it has started (both at startup and before the first cross-file reference query). If indexing does not complete within this window, Serena logs a warning and proceeds anyway. Increase it for very large projects. |
 | `server_ready_timeout` | `10.0` | Timeout in seconds for waiting on the server-ready signal after initialization. If the signal does not arrive within this window, Serena logs a message and proceeds anyway. |
+| `indexing_start_grace` | `5.0` | Timeout in seconds to wait for tsserver to *start* reporting `$/progress` before the first cross-file reference query. tsserver must resolve the project graph before it can emit the first progress token, and that can take longer than the default on a very large project; if it takes longer than this window, Serena assumes no indexing was needed and may return incomplete cross-file references. Raising `indexing_timeout` alone does not help here, since this grace elapses first. Increase this for very large projects if `find_referencing_symbols`/`request_references` returns incomplete results shortly after project load. |
 
 #### Svelte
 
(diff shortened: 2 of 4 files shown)


opcode81:
Thanks a lot for your contribution!

AmirF194:
Glad it helps! Happy to look into more indexing timing edge cases if you run into any.

PR #1750, opened 2026-07-23
fix(ls): don't orphan the language server process when Serena is killed

### LINKED ISSUE #1490: KotlinLanguageServer processes orphaned across MCP sessions: 90 instances / 21 GB leaked
## Summary

On Linux, every Serena MCP session that activates a Kotlin project spawns a `KotlinLanguageServer` JVM child but does not terminate it on shutdown. The children get reparented to PID 1 and persist indefinitely. After accumulating across many Claude Code sessions, I found **90 orphaned `KotlinLanguageServer/jre/bin/java` processes consuming ~21 GB total RSS** (~240 MB each). Available memory was down to 7 GB of 62 GB before cleanup.

This appears to be the KLS-itself variant of the same family of bugs as #1488 (Python zombie), #1081 (Gradle daemons spawned by KLS), and the closed #1281 (WebView2 25-child leak).

## Environment

- `serena-agent` v1.3.0 (installed via `uv tool install`)
- OS: Arch Linux, kernel 7.0.3-arch1-2
- MCP client: Claude Code (10+ parallel sessions over several days)
- Launched as: `serena start-mcp-server --context claude-code --project-from-cwd`
- KLS path: `~/.serena/language_servers/static/KotlinLanguageServer/kotlin_language_server/jre/bin/java`

## Reproduction

```
$ ps -eo pid,ppid,rss,args | grep "KotlinLanguageServer/kotlin_language_server/jre/bin/java" | grep -v grep | wc -l
90

$ ps -eo pid,ppid,rss,args | grep "KotlinLanguageServer/kotlin_language_server/jre/bin/java" | grep -v grep | awk '{print \$2}' | sort -u
1

$ ps -eo pid,rss,args | grep "KotlinLanguageServer" | grep -v grep | awk '{s+=\$2} END {printf \"%.1f GB total\n\", s/1024/1024}'
21.1 GB total
```

All 90 KLS processes have `PPID=1`, confirming the original Serena parents have exited but the children kept running.

## Likely cause

`KotlinLanguageServer` (the LSP wrapper class) does not have its own subprocess registered for SIGTERM-on-shutdown in the MCP server's exit path, OR the subprocess is registered but not killed when the MCP server is itself killed ungracefully (SIGKILL from the client, OS shutdown, etc.). The KLS Java process happily keeps running with no stdio peer, since nothing closes its stdin or sends it the LSP \`shutdown\`/\`exit\` notifications.

Related but distinct fixes:
- #1082 (merged): KLS now kills GradleDaemons it spawned. But the bug here is one level up — Serena doesn't kill the KLS itself.
- #567: TypeScript LS cleanup fix — likely the right model to follow.
- #1488: Python LS zombie — same shape (Serena doesn't reap its LSP children) but cosmetic; this one is load-bearing because the child stays alive and holds memory.

## Workaround

\`\`\`
pkill -f \"KotlinLanguageServer/kotlin_language_server/jre/bin/java\"
\`\`\`

Reclaimed ~20 GB of RAM + freed ~4 GB of swap immediately.

## Suggested fix

When Serena's MCP server shuts down (clean exit, SIGTERM, atexit), iterate over registered language-server child PIDs and SIGTERM them with a short grace period before SIGKILL. Additionally on startup, check for orphaned KLS processes (PPID=1, matching this user's \`~/.serena/language_servers/static/KotlinLanguageServer/...\` JRE path) and reap them — same heuristic #1082 used for matching \"daemons spawned by *this* KLS instance\".

Happy to test a patch.

AmirF194:
## Root cause

Every language server is started with `start_new_session=True` (`ls_config.py`'s
`start_independent_lsp_process`, on by default): its own process group, detached from Serena's,
so `subprocess_util.terminate_process_tree_with_kill_fallback` can walk and terminate the whole
tree on a clean shutdown. That cleanup code only runs if Serena's own process is alive to run it.
If Serena is killed outright (SIGKILL, OOM, a crash), nothing runs it, and because the child is in
its own session, no signal delivered to Serena's process group reaches it either. #1490 reports the
result for Kotlin's JVM process: 90 orphaned instances, 21 GB. Nowhere in the codebase is a
kernel-level parent-death mechanism set, which is the standard fix for a child that must die when
its parent dies for any reason.

A prior discussion on the issue (`#1082`) is scoped to Gradle daemons Kotlin's language server
spawns internally during indexing; that is a different, narrower gap (those daemons aren't
children of the KLS process at all) and doesn't touch this one.

## Fix

On Linux, `_start` now passes `preexec_fn=set_pdeathsig_on_parent_exit`
(`subprocess_util.py`), which calls `prctl(PR_SET_PDEATHSIG, SIGTERM)` in the child before exec, so
the kernel signals it if its parent dies.

That alone is not sufficient: the process is launched with `shell=True`, so the immediate child is
`/bin/sh -c <cmd>`, and `sh` forks the actual language server as its own child rather than
exec-ing into it. `PR_SET_PDEATHSIG` only protects the process that set it, so on its own this
kills the shell and leaves the real language server behind exactly as before. Confirmed with an
isolated repro of the identical spawn shape (`shell=True`, `start_new_session=True`) before
settling on this fix: with `preexec_fn` alone the shell dies but its child survives; adding the
`exec` prefix is what makes the child die too.
`convert_shell_cmd` now takes a `use_exec` flag that prefixes the command with the shell's `exec`
builtin, so `sh` replaces its own process image with the language server instead of forking it. The
PDEATHSIG registration survives `execve` (same PID), so it now reaches the process that matters.

Both changes are gated to `platform.system() == "Linux"` and to
`start_independent_lsp_process`. `PR_SET_PDEATHSIG` has no equivalent on macOS or Windows; those
platforms keep the current behavior, unchanged by this PR.

## Verification

Added `test/solidlsp/test_pdeathsig.py`: it drives a real `StdioLanguageServer` in a child
process (standing in for Serena), SIGKILLs that child, and asserts the language server process it
started does not survive. Confirmed the test fails on current `main` (the process survives) and
passes with this fix, in a clean `python:3.11-slim` Docker container.

Also ran, in the same container: `poe lint`, `poe type-check`, and the full `test_python_basic.py`
suite against a real `basedpyright` server launched through `uvx` (exercising the modified spawn
path end to end, not just the synthetic repro), all passing.

Not verified: the actual JVM/Kotlin process from the original report, or behavior when the
language server is invoked through a wrapper like `uvx` that may itself fork a further child
(`uv tool run`'s own subprocess handling is outside this fix's scope; the mechanism here only
guarantees the immediate program named in `cmd` inherits the signal).

## Checklist

Fixes #1490

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ Status of the `main` branch. Changes prior to the next official version change w
   - The `languages` key in project configurations was changed to `language_servers` to better reflect
     the actual semantics (configurations are automatically migrated)
   - Fix: glob matching bare `*` and `?` in non-`**` patterns matched across `/`, contradicting documented behaviour #1732
+  - Fix: on Linux, a language server process spawned in its own session (the default) is no longer
+    orphaned when Serena is killed without a chance to shut down cleanly (e.g. SIGKILL, OOM) #1490
 
 * Language Servers: 
   - Allow language server priorities to be configured in `serena_config.yml` (for auto-detection during 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,6 @@ Status of the `main` branch. Changes prior to the next official version change w
   - The `languages` key in project configurations was changed to `language_servers` to better reflect
     the actual semantics (configurations are automatically migrated)
   - Fix: glob matching bare `*` and `?` in non-`**` patterns matched across `/`, contradicting documented behaviour #1732
-  - Fix: on Linux, a language server process spawned in its own session (the default) is no longer
-    orphaned when Serena is killed without a chance to shut down cleanly (e.g. SIGKILL, OOM) #1490
-  - Language servers and their dependency providers now go through the `subprocess_run` helper instead of
-    calling `subprocess.run` directly, so all such subprocesses get `stdin=DEVNULL` and can no longer
-    interfere with the stdio MCP connection (as happened for Julia, #1577) #1748
 
 * Language Servers: 
   - Allow language server priorities to be configured in `serena_config.yml` (for auto-detection during 
@@ -32,6 +27,11 @@ Status of the `main` branch. Changes prior to the next official version change w
     for projects where the initial project-graph resolution itself takes longer than that. The grace
     is now `indexing_start_grace` (default 5.0s), configurable the same way as `indexing_timeout` and
     `server_ready_timeout` #1586
+  - Fix: On Linux, a language server process spawned in its own session (the default) is no longer
+    orphaned when Serena is killed without a chance to shut down cleanly (e.g. SIGKILL, OOM) #1490
+  - Language servers and their dependency providers now go through the `subprocess_run` helper instead of
+    calling `subprocess.run` directly (e.g. for installation processes), so all such subprocesses get 
+    `stdin=DEVNULL` and can no longer interfere with the stdio MCP connection #1748
 
 * Hooks:
   - Add `serena-hooks --client=grok`, including Grok-native PreToolUse allow/deny output.
--- a/src/solidlsp/ls_process.py
+++ b/src/solidlsp/ls_process.py
@@ -513,13 +513,7 @@ class StdioLanguageServer(LanguageServerInterface):
     def _start(self) -> None:
         log.info("Starting language server process via command: %s", self.process_launch_info.cmd)
 
-        # An independent session (its own process group, detached from ours) is what lets a
-        # SIGKILLed Serena leave the language server behind as an orphan: our own graceful
-        # shutdown code (subprocess_util.terminate_process_tree_with_kill_fallback) never runs in
-        # that case, and being in its own session means no signal to our process group reaches it
-        # either. LanguageServerSubprocessLauncher closes that gap on Linux at spawn time via
-        # PR_SET_PDEATHSIG when start_new_session is set.
-        process = subprocess_util.LanguageServerSubprocessLauncher.instance().launch(
+        process = subprocess_util.LanguageServerSubprocessLauncher.get_instance().launch(
             self.process_launch_info, start_new_session=self.start_independent_lsp_process
         )
         self.process = process
--- a/src/solidlsp/ls_process.py
+++ b/src/solidlsp/ls_process.py
@@ -1,8 +1,6 @@
 import asyncio
 import json
 import logging
-import os
-import platform
 import socket
 import subprocess
 import threading
@@ -11,7 +9,7 @@ from abc import ABC, abstractmethod
 from collections.abc import Callable, Iterable
 from dataclasses import dataclass
 from queue import Empty, Queue
-from typing import IO, Any, AnyStr, cast
+from typing import IO, Any, AnyStr
 
 from sensai.util.string import ToStringMixin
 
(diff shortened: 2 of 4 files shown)


opcode81:
Since this appears to be primarily a workaround for `start_independent_lsp_process` being enabled, we could investigate whether the flag could be disabled.

AmirF194:
Disabling `start_independent_lsp_process` wouldn't close this gap on its own: a killed process doesn't propagate SIGKILL to its children just by sharing a session, so the graceful-shutdown path (which walks the process tree) still can't run once Serena itself is gone. The flag would need its own investigation for whatever it's protecting against elsewhere, so I kept this fix scoped to the SIGKILL/crash case in #1490 rather than touching it.

Separately: the four failing Linux jobs on this PR are a real regression from this change, not flakiness. `PR_SET_PDEATHSIG` ties its registration to the specific calling *thread*, not the process, and `LanguageServerManager.from_languages` starts each server on a short-lived `StartLSThread` that returns right after `Popen()`. That thread's own exit was firing the death signal within milliseconds of a normal startup, killing the language server on every run, independent of any SIGKILL scenario, which is exactly the "stdout read process terminated unexpectedly" failures in the Rust jobs above. Fixed by routing the `Popen()` call through one dedicated daemon thread that only ever dies alongside the whole process (`popen_preserving_pdeathsig` in `subprocess_util.py`), plus a regression test (`test_language_server_process_survives_a_short_lived_calling_thread`) that fails without it and passes with it. Verified both directions in Docker against this branch's prior commit. Pushed.

opcode81:
The fix is functionally sound and well-documented, but poorly factorised: it only works when four fragments are combined correctly — the `exec` prefix in `convert_shell_cmd`, the `preexec_fn` registration, routing `Popen()` through `popen_preserving_pdeathsig`, and the Linux/own-session condition at the call site. Nothing enforces their co-occurrence; `_start()` assembles the ritual by hand, guarded only by comments.

Please instead reify the concern "launching a subprocess whose lifecycle is tied to the parent process (on Linux)" as a dedicated component, e.g. `LanguageServerSubprocessLauncher` with a single public method `launch(process_launch_info: ProcessLaunchInfo, start_new_session: bool) -> Popen[bytes]`. Everything else — the libc/prctl preexec_fn, the `exec` prefixing, the platform condition, and the dedicated spawn thread — becomes internal to it.

Two specific points:

1. **This should be a singleton.** The component is home to the persistent worker thread that must perform all pdeathsig-registering `fork()` calls (the registration is tied to the forking OS *thread*, which must therefore live as long as the process). One process-wide thread means one process-wide launcher; with `start_new_session` passed per call, it carries no configuration state, so this is unproblematic. The thread should be created lazily, as it's only needed for own-session launches on Linux; the `_SpawnerHolder` + module-level lock arrangement then disappears into the class.

2. **Revert `convert_shell_cmd` to its original signature.** The `use_exec` flag grafts a lifecycle concern onto a pure command-formatting utility whose other callers have no use for it — and for which it is even invalid on Windows. The `exec` prefixing belongs inside the launcher, applied to the utility's output, where it can never be requested without the preexec_fn and dedicated-thread machinery it depends on.

AmirF194:
Both points are fair, pushed in ce685c89.

1. Reified the four fragments into `LanguageServerSubprocessLauncher`, a lazily-constructed singleton in `subprocess_util.py` with one public method:

`launch(process_launch_info: ProcessLaunchInfo, start_new_session: bool) -> Popen[bytes]`

The libc/prctl preexec_fn, the `exec` prefixing, the Linux/own-session gate, and the dedicated spawner thread are all internal now. `_start()` just calls `LanguageServerSubprocessLauncher.instance().launch(self.process_launch_info, start_new_session=self.start_independent_lsp_process)`. The spawner thread is created lazily on first own-session launch on Linux, not at import or construction time, so `_SpawnerHolder` and the module-level lock are gone, folded into the class.

2. Reverted `convert_shell_cmd` to `convert_shell_cmd(cmd: str | list[str]) -> str`, no `use_exec`. Checked its other two callers (`serena/agent.py`, `solidlsp/language_servers/common.py`) and neither ever passed it. The `exec` prefix is now applied only inside the launcher, to the plain output of `convert_shell_cmd`.

Adapted `test/solidlsp/test_pdeathsig.py`: the two existing end-to-end tests (short-lived calling thread, SIGKILLed driver) needed no changes since they drive `_start()`/`start()` and now go through the launcher transparently. Added three narrower tests: `convert_shell_cmd` has no `use_exec` param, the launcher is a singleton, and its spawner thread stays uncreated for a non-own-session launch.

Verified this run in a clean `python:3.11` container via `uv sync --extra dev --locked`: all 5 tests in `test_pdeathsig.py` pass, `poe lint` (ruff format + check) is clean, and `poe type-check` (`ty`) is clean on `src/serena`, `src/solidlsp`, and `test`. I did not re-run the full test suite or the other language-server test batches, only this file and the two lint/type-check tasks.

opcode81:
Thanks. We pushed a follow-up commit ([cbc7f28](https://github.com/oraios/serena/pull/1750/commits/cbc7f286b305cb59a762a0649e750d3db1d24487)) that completes the encapsulation and fixes two issues we found along the way; summary of what changed and why:

1. **Full internalization.** The launcher's collaborators were still loose module-level siblings: `set_pdeathsig_on_parent_exit` was even public, so it could be passed as `preexec_fn` to a plain `Popen()` — recreating the exact thread-affinity bug your second commit fixed. `_PDeathSigSpawner` is now a private inner class, the preexec_fn a private method, and `PR_SET_PDEATHSIG` a private class constant. The invariant "the registration cannot be requested without the machinery it depends on" is now enforced by visibility rather than convention. (Also: `instance()` → `get_instance()`.)

2. **libc resolution was overfitted to glibc.** `ctypes.CDLL("libc.so.6")` ran unconditionally at import time, so on incompatible systems, the import itself raised `OSError`, making solidlsp unimportable. libc is now resolved lazily on launcher construction via `CDLL(None)` — which resolves `prctl` from the process image on any Linux libc — with an `OSError` fallback to `None` that logs a warning and degrades to plain own-session behavior instead of crashing. `use_errno` was dropped as unused.

3. **pdeathsig is now gated on libc availability, not a platform check.** `_load_libc()` returning `None` off-Linux subsumes the platform test, and a libc-less Linux now behaves exactly like a non-Linux platform: no useless `exec` prefix, no no-op preexec_fn (which alone would force `subprocess` off its `posix_spawn` fast path), no spawner thread.

4. **Docs and tests.** Docstrings/comments were tightened to state what each element *is for*, keeping implementation rationale at the site that implements it; the lifecycle commentary in `_start()` was removed entirely, as it is fully the launcher's concern. The three targeted tests (signature/singleton/laziness) were dropped — they asserted construction details rather than behavior. Your two end-to-end regression tests, which cover the actual guarantees, remain unchanged and are what CI relies on.

AmirF194:
Thanks for finishing this, the full internalization and the lazy libc resolution are both real improvements over what we had. Good catch on the unconditional CDLL("libc.so.6") call failing import on non-glibc Linux, that would have been an ugly regression to chase down later. Happy for this to land as is.

PR #1752, opened 2026-07-24
Improve pattern search exclusions and ranked overflow results

ilganeli:
## Why
In large repos with extensive documentation, serena had insufficiently flexible configuration for what to exclude as relevant and was crashing out, blowing up context. 

## Summary

- add `paths_exclude_globs` while preserving the existing singular exclusion parameter; both forms combine additively
- add ordered `paths_priority_globs` for deterministic live-path-first pattern searches
- return a budget-aware ranked excerpt with shown/total match and file counts before falling back to smaller summaries
- keep unprioritized result ordering compatible with existing behavior

## Compatibility and behavior

The existing `paths_exclude_glob` schema and behavior are unchanged. New list exclusions are optional and additive.

Priority ranking is also opt-in. When supplied, matches are ordered by the first matching priority glob, normalized path, and matched line. When omitted, Serena preserves its existing source ordering.

Oversized ranked excerpts contain whole match entries and report how many matches and files were shown versus omitted. The implementation sizes the excerpt linearly and serializes only the full result and final excerpt.

## Verification

- `uv run poe test -q -k 'search_for_pattern or TestSearchText or TestSearchFiles or TestGlobMatch'` — 93 passed, 1 skipped
- `uv run poe lint` — passed
- `uv run poe type-check` — passed
- 20,000-match regression verifies excerpt rendering performs only two JSON serializations

## Checklist

### DIFF
--- a/src/serena/tools/file_tools.py
+++ b/src/serena/tools/file_tools.py
@@ -611,7 +611,12 @@ class SearchForPatternTool(Tool):
             code_files_only=restrict_search_to_code_files,
         )
 
-        # rank prioritized matches deterministically while leaving no-priority searches unchanged
+        # normalize the public path contract before ranking and serialization
+        for match in matches:
+            assert match.source_file_path is not None
+            match.source_file_path = GlobMatcher.normalize_path(match.source_file_path)
+
+        # rank prioritized matches deterministically while preserving source order when no priorities are supplied
         priority_matchers = [GlobMatcher(glob.strip()) for glob in paths_priority_globs or [] if glob.strip()]
         if priority_matchers:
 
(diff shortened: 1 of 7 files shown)


src/serena/tools/file_tools.py
```diff
@@ -580,6 +584,9 @@ def apply(
         :param context_lines_after: number of context lines to include after each match.
         :param paths_include_glob: optional glob (relative to project root, e.g. ``"src/**/*.ts"``) restricting which files are searched.
         :param paths_exclude_glob: optional glob to exclude files; takes precedence over `paths_include_glob`.
+        :param paths_exclude_globs: additional optional globs to exclude files; additive with `paths_exclude_glob`.
+        :param paths_priority_globs: optional ordered globs that rank matching paths before unmatched paths.
+            The first matching glob determines a path's priority.
```
opcode81, line 589: Having both `paths_exclude_glob` and `paths_exclude_globs` is not reasonable.
To generalise the interface, we can use a list-based interface for both inclusions and exclusions.

The priority feature is a new addition, but it's unlikely to be useful in the vast majority of circumstances.
@MischaPanch what are you thoughts regarding the usefulness of the priority feature? Worth keeping?

MischaPanch: Yes, let's generalize the interface to lists.

I don't think the priority feature is worth the significant additional complexity, we have progressive output shortening and agents can just make a follow-up call if the first call did not deliver the right results

PR #1754, opened 2026-07-24
Chore: Replace subprocess.run with subprocess_run util function

### LINKED ISSUE #1748: Chore: Replace `subprocess.run` with `subprocess_run` util function
To systematically avoid the problem which surfaced in #1577 for Julia, language servers (and their dependency providers) must not directly call `subprocess.run`. Instead, they should call the newly added util method `subprocess_run`.

Instructions for adding new language servers were adapted accordingly, but many existing implementations still call `subprocess.run` directly without specifying `stdin=DEVNULL`. (The parameter is hardcoded in the util function.)

This task is about replacing these calls.
Note: Some tests mock `subprocess.run`.

poysa213:
Closes #1748.

Routes `subprocess.run` calls in the language servers and their dependency providers through the `subprocess_run` helper, which forces `stdin=DEVNULL` so a subprocess can't disrupt the stdio MCP connection and crash the server (as happened for Julia, #1577).

Apart from the enforced stdin the conversion is behaviour-preserving: calls that streamed to the terminal keep doing so via `capture_output=False`, and the two helpers that merge stderr into stdout are handled likewise. `cli.py` (interactive editor, needs a real stdin) and the `subprocess_run` helper itself are left untouched.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ Status of the `main` branch. Changes prior to the next official version change w
   - The `languages` key in project configurations was changed to `language_servers` to better reflect
     the actual semantics (configurations are automatically migrated)
   - Fix: glob matching bare `*` and `?` in non-`**` patterns matched across `/`, contradicting documented behaviour #1732
+  - Language servers and their dependency providers now go through the `subprocess_run` helper instead of
+    calling `subprocess.run` directly, so all such subprocesses get `stdin=DEVNULL` and can no longer
+    interfere with the stdio MCP connection (as happened for Julia, #1577) #1748
 
 * Language Servers: 
   - Allow language server priorities to be configured in `serena_config.yml` (for auto-detection during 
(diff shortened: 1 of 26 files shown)


opcode81:
Why did you close this?

poysa213:
> Why did you close this?

copilot made me doubt my PR lol. i will reopen it again

opcode81:
Copilot's comment was correct, but it's nothing to worry about. I used `text=True` in the util method as a default, because it is what you'll typically want. The `subprocess` run default is `False`, so in cases where `text` is not explicitly specified, we need to either
* remove the decoding of the output - if it is processed (as Copilot suggested - preferred)
* explicitly set it to `False`.

Very few calls should be affected by this. Most explicitly set `text=True` anyway - or don't process the output, so it doesn't matter.

poysa213:
@opcode81  Thanks! Went with the preferred fix -- removed the .decode(); it was the only affected call. Pushed, feel free to merge once CI passes.

opcode81:
Thanks for your contribution!

PR #1759, opened 2026-07-25
fix(hooks): accept raw control characters in payloads

### LINKED ISSUE #1743: serena-hooks crashes (JSONDecodeError) on PreToolUse payloads with raw control characters in tool_input (CodeBuddy / freeform tools)
**Observed behavior**

When CodeBuddy (client `codebuddy`) fires the `serena-hooks remind` / `auto-approve` PreToolUse hooks for a freeform tool such as `Edit` with a patch payload, the hook crashes with:

```
  File "$HOME/.local/bin/serena-hooks", line 10, in <module>
    sys.exit(hook_commands())
  ...
  File "$HOME/.local/share/uv/tools/serena-agent/lib/python3.13/site-packages/serena/hooks.py", line 33, in __init__
    input_data = json.loads(raw)
  File ".../python3.13/json/__init__.py", line 352, in loads
    return _default_decoder.decode(s)
json.decoder.JSONDecodeError: Invalid control character at: line 1 column 100 (char 99)
```

This surfaces in the agent as a `⚠ Hook Notification [warning]` on every such tool call.

**Root cause**

In `serena/hooks.py`, `Hook.__init__` parses the hook payload with `json.loads(raw)` (default `strict=True`). CodeBuddy serializes freeform-tool `tool_input` (e.g. an `Edit` patch) as a JSON string value containing **raw newline control characters** rather than escaped `\n`. Such input is invalid JSON, so `json.loads` raises `JSONDecodeError` before any of the field handling logic runs.

This is distinct from (and not fixed by) issue #1583: #1583 only guarded the `.get()` access when `tool_input` is a `str`, but the failure here happens one step earlier, at the JSON parse itself. The #1583 fix is therefore insufficient for the `codebuddy` client.

**Expected behavior**

`serena-hooks` should tolerate control characters in the payload (or otherwise not crash) for any client, e.g. by parsing with `strict=False` and normalizing non-dict `tool_input` to `None`, consistent with the existing handling introduced for #1583.

**Reproduction**

```bash
# raw newline inside tool_input -> crashes (JSONDecodeError)
printf '%s' '{"hook_event_name":"PreToolUse","session_id":"abc","tool_name":"Edit","tool_input":"*** Begin Patch
*** Add File: /tmp/x.txt
+hi
","permission_mode":"default"}' | serena-hooks remind --client=codebuddy

# same payload with escaped newlines -> valid JSON, hook runs fine (no output, no error)
printf '%s' '{"hook_event_name":"PreToolUse","session_id":"abc","tool_name":"Edit","tool_input":"*** Begin Patch\n*** Add File: /tmp/x.txt\n+hi\n","permission_mode":"default"}' | serena-hooks remind --client=codebuddy
```

**Environment**

- serena-agent version: `1.6.1`
- client: CodeBuddy (`--client=codebuddy`)
- hook: `serena-hooks remind --client=codebuddy` (PreToolUse)

**Suggested fix**

```python
# serena/hooks.py, Hook.__init__
input_data = json.loads(raw, strict=False)
```

Optionally also reject/normalize `tool_input` robustly when it is a `str`, mirroring the #1583 guard, so downstream `.get()` access is always safe.

**Related**

- #1583 (same freeform-tool scenario, partial fix at the `.get()` layer only)

Bortlesboat:
## Summary

- accept raw control characters in hook JSON string values while keeping malformed JSON rejected
- cover the reported CodeBuddy freeform `Edit` payload at the `serena-hooks` CLI boundary
- document the fix in the Hooks changelog

## Root cause

`Hook` used Python's default strict JSON decoder, so unescaped newlines inside CodeBuddy's string-valued `tool_input` failed before the existing non-dict input normalization could run.

## User impact

`serena-hooks remind` and `serena-hooks auto-approve` no longer crash on these payloads. Freeform string inputs remain ignored by hook heuristics, matching the existing behavior.

## Testing

- `uv run pytest test/serena/test_hooks.py -q --basetemp <isolated-temp-dir>` — 83 passed
- `uv run poe lint`
- `uv run poe type-check`

Closes #1743.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ Status of the `main` branch. Changes prior to the next official version change w
   - Add `serena-hooks --client=grok`, including Grok-native PreToolUse allow/deny output.
   - PreToolUse remind hook: coerce non-string shell command values instead of failing, and recognize
     `target_file`/`targetFile` file-path keys (shared payload parsing, applies to all hook clients).
+  - Fix hook input parsing for clients that emit raw control characters in JSON string values #1743.
 
 
 # v1.6.1 (2026-07-21)
--- a/src/serena/hooks.py
+++ b/src/serena/hooks.py
@@ -31,7 +31,7 @@ class HookClient(Enum):
 class Hook(ABC):
     def __init__(self, client: HookClient):
         raw = sys.stdin.read()
-        input_data = json.loads(raw)
+        input_data = json.loads(raw, strict=False)
         self._input_data = input_data
         self._client = client
 
--- a/test/serena/test_hooks.py
+++ b/test/serena/test_hooks.py
@@ -992,6 +992,20 @@ class TestHookCli:
         output = json.loads(result.output)
         assert output["hookSpecificOutput"]["permissionDecision"] == "deny"
 
+    def test_remind_command_accepts_raw_control_characters_in_tool_input(self, tmp_path: Path):
+        """CodeBuddy freeform tool input may contain unescaped control characters."""
+        stdin_json = (
+            '{"hook_event_name":"PreToolUse","session_id":"cli-codebuddy-raw-input",'
+            '"tool_name":"Edit","tool_input":"*** Begin Patch\n*** Add File: /tmp/x.txt\n+hi\n",'
+            '"permission_mode":"default"}'
+        )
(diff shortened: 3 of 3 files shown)


opcode81:
Thanks!

PR #1119, opened 2026-03-02
Issue #1055: Add minimal frontmatter parser for memory files

Meh10t:
This PR starts work on issue #1055.

It introduces a minimal frontmatter parser for memory files, supporting arbitrary YAML-like fields.

Currently supported:
- Optional YAML-like frontmatter block at the top of a memory file
- Supports arbitrary frontmatter fields (no restriction to specific keys)
- Returns both the parsed frontmatter and the remaining body content

Example supported format:

---
summary: Some short description
---

This is the memory content...

This is the first step towards supporting opt-in frontmatter tools (MemoryAddFrontmatterTool and MemoryGetFrontmatterTool).

Next steps:
- Integrate frontmatter stripping in memory loading
- Implement optional frontmatter tools
- Extend list_memories behavior when frontmatter tools are enabled

PR #1761, opened 2026-07-25
feat: add optional memory frontmatter tools

TyceHerrman:
## Summary

- port the memory frontmatter work from #1119 onto the current memory-manager architecture
- add optional `memory_get_frontmatter` and `memory_add_frontmatter` tools
- recognize metadata only through explicitly marked, versioned Serena frontmatter
- keep `read_memory` body-only and keep the default `list_memories` response unchanged
- include non-empty metadata in `list_memories` only when the get-frontmatter tool is active
- preserve the exact serialized metadata prefix across body writes, edits, and reference autofixes
- update individual metadata fields without reserializing unrelated lines
- enforce current ignored-memory and read-only rules

## Compatibility and format

Managed frontmatter has the following scalar-only shape:

```yaml
---
serena_frontmatter_version: 1
type: "Serena Memory"
description: "Short summary"
---
```

`serena_frontmatter_version: 1` must be the first field, and `type` must be non-empty. The marker is reserved for Serena and is omitted from metadata tool and listing output. Newly created blocks default to `type: "Serena Memory"`.

Unmarked legacy content remains ordinary memory body content, including valid-looking, empty, and malformed leading `---` blocks. Parsing is independent of active tools, so identical files read consistently in every tool context. Marked blocks are hidden from `read_memory`; their original quotes, whitespace, field order, and newline style are retained during body-only writes. Field updates replace only the selected value or insert one line before the closing delimiter, rendering new values as reversible JSON-style quoted strings.

Duplicate keys, malformed marked blocks, missing or empty `type`, unsupported Serena frontmatter versions, and writes to the reserved marker fail before the file is modified.

## Minimal OKF alignment

While addressing the compatibility feedback, this also takes the opportunity for minimal field-level alignment with Google's [Open Knowledge Format v0.2](https://github.com/GoogleCloudPlatform/knowledge-catalog/blob/main/okf/SPEC.md): Serena uses the useful field names `type` and `description` where applicable.

This does **not** claim OKF bundle conformance. Serena does not use `okf_version`, does not implement reserved `index.md` or `log.md` behavior, and keeps structured OKF fields outside this scalar-only feature.

## Validation

- `171 passed` across the focused frontmatter, memory-manager, and memory-tool suites
- `poe lint`
- `poe type-check`
- `poe doc-build`
- `python -m compileall` on changed Python files
- `git diff --check`

The new-head GitHub Actions runs for commit `85086235` are final: 10 checks passed, 6 failed, 6 were cancelled, and 2 were skipped. Every non-green executable check failed before repository checkout: GitHub either did not acquire a hosted runner after multiple attempts or returned `Service Unavailable` / `Failed to resolve action download info` while resolving actions. No repository code or tests failed; notably, the JVM matrix passed on Linux, Windows, and macOS. Runs: [Tests](https://github.com/oraios/serena/actions/runs/31118182729), [Python Versions](https://github.com/oraios/serena/actions/runs/31118182438), [CodeQL Advanced](https://github.com/oraios/serena/actions/runs/31118182480), [Codespell](https://github.com/oraios/serena/actions/runs/31118182500), and [Docs Build](https://github.com/oraios/serena/actions/runs/31118182510).

The local macOS sandbox limitations remain recorded separately in #1817 and #1818.

Original implementation: #1119 by @Meh10t.

### DIFF
--- a/src/serena/memories/frontmatter.py
+++ b/src/serena/memories/frontmatter.py
@@ -1,0 +1,73 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True)
+class FrontmatterParseResult:
+    frontmatter: dict[str, str]
+    body: str
+
+
+class FrontmatterParser:
+    """Parser and renderer for simple scalar frontmatter fields."""
+
+    DELIMITER = "---"
+
+    @staticmethod
+    def _validate_field(key: str, value: str) -> None:
+        if not key:
+            raise ValueError("Frontmatter key must not be empty")
+        if ":" in key:
+            raise ValueError("Frontmatter key must not contain ':'")
+        if "\n" in key or "\r" in key:
+            raise ValueError("Frontmatter key must be a single line")
+        if "\n" in value or "\r" in value:
+            raise ValueError("Frontmatter value must be a single line")
+
+    @classmethod
+    def parse(cls, content: str) -> FrontmatterParseResult:
+        lines = content.splitlines(keepends=True)
+        if not lines or lines[0].rstrip("\r\n") != cls.DELIMITER:
+            return FrontmatterParseResult(frontmatter={}, body=content)
+
+        closing_index: int | None = None
+        for index, line in enumerate(lines[1:], start=1):
+            if line.rstrip("\r\n") == cls.DELIMITER:
+                closing_index = index
+                break
+
+        if closing_index is None:
+            return FrontmatterParseResult(frontmatter={}, body=content)
+
+        frontmatter: dict[str, str] = {}
+        for line in lines[1:closing_index]:
+            field = line.rstrip("\r\n")
+            if not field:
+                continue
+            if ":" not in field:
+                return FrontmatterParseResult(frontmatter={}, body=content)
+
+            key, value = field.split(":", 1)
+            key = key.strip()
+            if not key:
+                return FrontmatterParseResult(frontmatter={}, body=content)
+
+            value = value.strip()
+            if len(value) >= 2 and value[0] == value[-1] and value[0] in {'"', "'"}:
+                value = value[1:-1]
+            frontmatter[key] = value
+
+        body = "".join(lines[closing_index + 1 :])
+        return FrontmatterParseResult(frontmatter=frontmatter, body=body)
+
+    @classmethod
+    def render(cls, frontmatter: dict[str, str], body: str) -> str:
+        if not frontmatter:
+            return body
+
+        for key, value in frontmatter.items():
+            cls._validate_field(key, value)
+
+        fields = "\n".join(f"{key}: {value}" for key, value in frontmatter.items())
+        return f"{cls.DELIMITER}\n{fields}\n{cls.DELIMITER}\n{body}"
--- a/src/serena/memories/frontmatter.py
+++ b/src/serena/memories/frontmatter.py
@@ -1,23 +1,64 @@
 from __future__ import annotations
 
-from dataclasses import dataclass
+import json
+from dataclasses import dataclass, field
+
+
+@dataclass(frozen=True)
+class _FrontmatterField:
+    key: str
+    line_index: int
+    value_start: int
+    value_end: int
 
 
 @dataclass(frozen=True)
 class FrontmatterParseResult:
     frontmatter: dict[str, str]
     body: str
+    prefix: str | None = None
+    _fields: tuple[_FrontmatterField, ...] = field(default=(), repr=False)
+
+    @property
+    def is_managed(self) -> bool:
+        return self.prefix is not None
+
+    def with_body(self, body: str) -> str:
+        if self.prefix is None:
+            return body
+        return self.prefix + body
 
 
 class FrontmatterParser:
-    """Parser and renderer for simple scalar frontmatter fields."""
+    """Parser and byte-preserving updater for Serena scalar frontmatter."""
 
     DELIMITER = "---"
+    VERSION_KEY = "serena_frontmatter_version"
+    VERSION = "1"
+    TYPE_KEY = "type"
+    DEFAULT_TYPE = "Serena Memory"
+
+    @staticmethod
+    def _line_ending(line: str) -> str:
+        if line.endswith("\r\n"):
+            return "\r\n"
+        if line.endswith("\n"):
+            return "\n"
+        if line.endswith("\r"):
+            return "\r"
+        return ""
+
+    @classmethod
+    def _without_line_ending(cls, line: str) -> str:
+        ending = cls._line_ending(line)
+        return line[: -len(ending)] if ending else line
 
     @staticmethod
     def _validate_field(key: str, value: str) -> None:
         if not key:
             raise ValueError("Frontmatter key must not be empty")
+        if key != key.strip():
+            raise ValueError("Frontmatter key must not have surrounding whitespace")
         if ":" in key:
             raise ValueError("Frontmatter key must not contain ':'")
         if "\n" in key or "\r" in key:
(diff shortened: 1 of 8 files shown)


opcode81:
Thanks, @TyceHerrman, @MischaPanch and I will evaluate whether the spec that is implemented here still fits with what happened within the memory system in the meantime.

src/serena/memories/frontmatter.py
```diff
@@ -0,0 +1,70 @@
...
+
+def parse_frontmatter(content: str) -> tuple[dict[str, str], str]:
+    """Return a frontmatter mapping and the unchanged body."""
+    result = FrontmatterParser.parse(content)
+    return result.frontmatter, result.body
+
+
+def render_frontmatter(frontmatter: dict[str, str], body: str) -> str:
+    """Render scalar frontmatter followed by the unchanged body."""
+    return FrontmatterParser.render(frontmatter, body)
```
opcode81: These functions (used only in tests) are clutter; only the OO interfaces should be used.

AmirF194 (commented):
`load_memory`/`save_memory` (memory_manager.py:223-243) run `FrontmatterParser.parse()` unconditionally on every read, not just when the new opt-in frontmatter tools are active. `ReadMemoryTool.apply` calls `load_memory` directly, so this changes what the *existing* `read_memory` tool returns for any pre-existing memory whose body happens to open with a line-exact `---`, one or more `key: value` lines, and a closing `---` (a plausible shape: a markdown horizontal rule followed by a short status block, e.g. after a heading).

Docker repro against this PR's head (`d2c3b5c`), no mocks, real `MemoryManager`:

```python
original = (
    "---\n"
    "replicas: 3\n"
    "image: myapp:latest\n"
    "---\n"
    "This is the actual body text a user wrote, describing a k8s manifest above.\n"
)
mgr._save_memory_raw("k8s-notes", original, is_tool_context=False)
loaded = mgr.load_memory("k8s-notes")
```

Output:
```
loaded == 'This is the actual body text a user wrote, describing a k8s manifest above.\n'
loaded != original   # True: the replicas/image lines are gone
```

`read_memory` silently drops that block for a memory nobody ever opted into frontmatter for. It happens to survive a load-then-save round trip here because `render()` reconstructs it, but that reconstruction isn't guaranteed to be byte-identical (`render` trims whitespace and never re-adds quotes, so a quoted value like `url: "https://x.com:443"` loses its quotes on the next round trip and can then mis-split at the extra colon).

Given the CHANGELOG frames this as "opt-in," would it make more sense to only parse frontmatter when at least one of the new tools is active, or to require an explicit marker so existing memory bodies that start with `---` aren't reinterpreted?

TyceHerrman:
Thanks for the concrete reproduction. This is addressed in `85086235` by requiring an explicit, persisted Serena marker instead of making parsing depend on which tools are active:

- only a block whose first field is `serena_frontmatter_version: 1` is treated as managed metadata
- unmarked legacy `---` content, including the reproduction shape, remains byte-for-byte body content through read/save cycles
- marked metadata is hidden from `read_memory` and its exact serialized prefix is retained regardless of active tools
- duplicate keys, malformed marked blocks, unsupported versions, and missing/empty `type` fail before writing
- field updates replace only the selected value or insert one line, preserving unrelated quotes, whitespace, ordering, and newline style
- newly written values use reversible JSON-style quoting, while existing simple unquoted scalar values and colon-containing values remain accepted
- adding metadata to a legacy memory wraps the complete original bytes beneath the new marked block

While addressing this feedback, I also took the opportunity for minimal alignment with Google's Open Knowledge Format v0.2 field names: new blocks require `type` (defaulting to `"Serena Memory"`) and use `description` where applicable. This is only field-level alignment, not OKF bundle conformance: Serena does not use `okf_version`, reserve `index.md`/`log.md`, or implement structured OKF fields in this PR.

Focused validation is `171 passed`, plus Ruff formatting/lint, full `ty` checks, the docs build, `compileall`, and `git diff --check`.

TyceHerrman:
the failing ci checks on commit `85086235` seem to be a GitHub Actions hosted-infrastructure incident, not a failure introduced by this PR:

Multiple unrelated repositories reported the same action-resolution and runner-assignment failures during the same UTC window

MischaPanch:
Thanks for the PR. @opcode81 and I still haven't decided whether we want to support this, as it brings serena's memories a bit away from "just standard md working with any memory", and the memory tool proliferation also carries a cost. Just a heads up on why there was no comment from us yet

opcode81:
@TyceHerrman just so we know what motivated your participation: Is this a feature you would find genuinely useful? If so, please describe how you are/plan on using this.

TyceHerrman:
@opcode81 

one use case i'm thinking of is to use Serena memories in Obsidian vault and use the memory frontmatter as machine-readable input for views and AI-agent automations.

something like:

```yaml
---
serena_frontmatter_version: 1
type: "Serena Memory"
description: "Design and compatibility rules for managed memory frontmatter"
project: "serena"
repository: "oraios/serena"
path: "src/serena/memories/memory_manager.py"
symbol: "MemoryManager.add_memory_frontmatter"
obsidian_note: "Projects/Serena/Memory architecture"
---
```

some ideas for what that could enable - with `memory_get_frontmatter`, an agent can obtain this metadata when listing memories. An Obsidian Base or automation could group memories by project or repository, link a memory to the corresponding project note, and expose its current code location. An agent could select memories matching the active project, read the explanation, inspect the referenced code, and update `path` or `symbol` if the implementation changes.

I am also looking at using [CodeGraph MCP server](https://github.com/colbymchenry/codegraph) alongside Serena, which builds a Treesitter/SQLite graph of symbols and resolved relationships. Its default `codegraph_explore` tool accepts a natural-language question or symbol/file names and returns the relevant line-numbered source, connected call paths, and context.

Frontmatter in serena memories could store inputs for both serena and codegraph. As examples, relative_path and serena_name_path for Serena and codegraph_query for CodeGraph. An agent could rerun the saved CodeGraph query then use the stored path and name path to locate the current symbol with Serena. A main agent could also pass these fields to a subagent instead of asking it to extract them from prose.

The OKF schema alignment in pr was deliberately limited for now, but could be reassessed for further alignment later. Reusing names from OKF such as `type` and `description` doesn't really cost anything and does help ensure that serena memories work with tools expecting OKF (if that actually takes off as a standard).

src/serena/memories/memory_manager.py
```diff
@@ -201,24 +202,72 @@ def _check_write_access(self, name: str, is_tool_context: bool) -> None:
...
+            return name, f.read()
 
-    def save_memory(self, name: str, content: str, is_tool_context: bool) -> str:
+    def _save_memory_raw(self, name: str, content: str, is_tool_context: bool) -> str:
         name = self._sanitize_name(name)
         self._check_not_ignored(name)
         self._check_write_access(name, is_tool_context)
         memory_file_path = self.get_memory_file_path(name)
-        with open(memory_file_path, "w", encoding=self._encoding) as f:
+        with open(memory_file_path, "w", encoding=self._encoding, newline="") as f:
```
feiiiiii5, line 219: This is the pre-#1969 form of the write. Current `main` has, at the end of `save_memory`:

```python
write_file_atomic(str(memory_file_path), content, encoding=self._encoding)
```

When rebasing, taking this line instead of that one would undo the atomicity fix for every memory write. `write_file_atomic` already accepts `newline`, so both requirements are satisfiable at once — inside `_save_memory_raw`:

```python
write_file_atomic(str(memory_file_path), content, encoding=self._encoding, newline="")
```

src/serena/memories/memory_manager.py
```diff
@@ -395,12 +444,12 @@ def edit_memory(
...
-        with open(memory_file_path, encoding=self._encoding) as f:
-            original_content = f.read()
+        with open(memory_file_path, encoding=self._encoding, newline="") as f:
+            parsed = self._parse_frontmatter(name, f.read())
         replacer = ContentReplacer(mode=mode, allow_multiple_occurrences=allow_multiple_occurrences, regex_multiline=regex_multiline)
-        updated_content = replacer.replace(original_content, needle, repl)
-        with open(memory_file_path, "w", encoding=self._encoding) as f:
-            f.write(updated_content)
+        updated_body = replacer.replace(parsed.body, needle, repl)
+        with open(memory_file_path, "w", encoding=self._encoding, newline="") as f:
```
feiiiiii5, line 451: Same situation as the other write site: `edit_memory` on current `main` writes with `write_file_atomic(str(memory_file_path), updated_content, encoding=self._encoding)` (`memory_manager.py:404` at `c6fbd1c5`, from #1969), and this hunk replaces it with a plain `open(..., "w")` because the branch predates that fix. Adding `newline=""` here is exactly what `write_file_atomic`'s `newline` parameter is for, so the rebase can keep both.

feiiiiii5 (commented):
Heads-up for the rebase, not a design objection: this branch is based on `2561478a`, which predates **#1969** (`14f1a3ce`, 2026-09-03, "stop losing memory content on an interrupted write"), and that commit rewrote exactly the two statements this PR also rewrites. On current `main` both memory write paths end in `write_file_atomic(...)` (`memory_manager.py:221` for `save_memory`, `:404` for `edit_memory` as of `c6fbd1c5`), while this branch still uses `open(path, "w", newline="")`. Since the PR is already `mergeable=false / dirty`, the conflict will land on these lines, and a resolution that keeps the branch side would silently put back the truncate-before-write that #1969 removed — a partially-written memory file is the failure mode it was fixing.

The good news is that the helper already supports what this PR needs: `write_file_atomic(path, content, *, encoding, newline=None)` passes `newline` through to `open()` (`src/serena/util/file_system.py`), so the byte-preserving behaviour can be kept while staying atomic. Detail in the two inline comments.

PR #1762, opened 2026-07-26
Docs: correct stale references in the language-addition guide

xormania:
Four items in .serena/memories/adding_new_language_support_guide.md are wrong against the current source. Two are references that no longer resolve:

- the subprocess helper lives at solidlsp.util.subprocess_util and is a module-level function, not a method on a class; the guide named solidlsp.utils.subprocess_utils, which does not exist
- pytest markers are declared under [tool.pytest.ini_options].markers in pyproject.toml; there is no pytest.ini in the repository

and two are formatting defects:

- the Documentation section used "### 5" while every other step uses "## Step N:", so it nested under Step 4 in rendered output
- a stray double colon in the "Implementation Pointers" label

Promoting Documentation to a top-level step also completes the Overview list, which stopped at four.

Step 5 gains src/serena/resources/project.template.yml, which the checklist omitted even though a new language server has to appear in its commented list. The instruction uses `uv run` because the script imports solidlsp and fails under an ambient interpreter, and it notes the trailing spaces the script pads each line with, which the committed list does not carry.

## Checklist

### DIFF
--- a/.serena/memories/adding_new_language_support_guide.md
+++ b/.serena/memories/adding_new_language_support_guide.md
@@ -10,6 +10,7 @@ Adding a new language involves:
 2. **Language Registration** - Adding the language to enums and configurations  
 3. **Test Repository** - Creating a minimal test project
 4. **Test Suite** - Writing comprehensive tests
+5. **Documentation** - Updating the user-facing language lists and the changelog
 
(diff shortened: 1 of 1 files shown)


opcode81:
Thanks!

PR #1357, opened 2026-04-16
Add Gleam language server support

Koushik-Salammagari:
Closes #1334

## Summary

Adds support for the [Gleam](https://gleam.run) programming language by integrating its built-in language server (`gleam lsp`). The LSP is bundled with the Gleam compiler — no separate installation is required.

### Files changed

| File | Change |
|------|--------|
| `src/solidlsp/language_servers/gleam_language_server.py` | New `GleamLanguageServer` class |
| `src/solidlsp/ls_config.py` | `Language.GLEAM` enum value, `*.gleam` file matcher, `get_ls_class` registration |
| `test/resources/repos/gleam/test_repo/` | Minimal Gleam project: `gleam.toml` + two source modules |
| `test/solidlsp/gleam/test_gleam_basic.py` | Symbol, within-file reference, cross-file reference, and bare-name tests |
| `pyproject.toml` | `gleam` pytest marker |
| `docs/01-about/020_programming-languages.md` | Gleam entry in language list |
| `README.md` | Gleam added to language list |

### Implementation notes

- `GleamLanguageServer` follows the same pattern as `CrystalLanguageServer` (requires user-installed binary, no auto-download).
- Launch command: `["gleam", "lsp"]` — the LSP is part of the compiler since Gleam v0.21.
- Build artifacts are ignored via `is_ignored_dirname` (`build/`, `_build/`).
- Tests are skipped automatically when `gleam` is not on PATH (`shutil.which("gleam") is None`).

### Test repository

```
test/resources/repos/gleam/test_repo/
├── gleam.toml                 # depends on gleam_stdlib
└── src/
    ├── calculator.gleam       # Calculator type + add/subtract/multiply/format_result/describe
    └── utils.gleam            # format_output + is_non_empty (cross-file reference target)
```

### CI note

A GitHub Actions step for installing the Gleam compiler will be needed to run these tests in CI (similar to how other language servers are set up in `.github/workflows/pytest.yml`). I'm happy to add that in a follow-up or in this PR if preferred — let me know.

PR #1765, opened 2026-07-27
feat: add Gleam language support via the Gleam compiler's bundled `gleam lsp`

### LINKED ISSUE #1334: Add support for Gleam

weiconghe:
## What does this PR do?

Adds **Gleam** language support to solidlsp, using the language server bundled with the Gleam compiler (`gleam lsp`). Closes #1334 (good first issue).

### Implementation

- New `GleamLanguageServer` (`src/solidlsp/language_servers/gleam_language_server.py`) using the `LanguageServerDependencyProviderSinglePath` pattern (same as qmlls): discovers `gleam` on PATH, launches `[gleam, "lsp"]`.
- `LanguageServerId.GLEAM` registered in `ls_config.py` (enum + `*.gleam` matcher + factory), in the main (non-experimental) section alongside QML.
- **Readiness**: `gleam lsp` emits `$/progress` (begin/end) for its initial dependency-resolution phase on first start. The server tracks active progress tokens and waits for all phases to end before declaring ready (bounded: 10 s for the first `begin`, 180 s for all to drain). Once dependencies are resolved, symbol / reference / hover / diagnostics queries are served synchronously — **no per-request compile-wait or retry is needed** (verified directly against gleam 1.17).
- Dependencies are pre-fetched via `gleam deps download` in `_start_server` so the LSP's own download phase is short and there is no first-request race on slow CI runners.
- **`_supports_pull_diagnostics` overridden to return `False`**: `gleam lsp` does not implement `textDocument/diagnostic` (no `diagnosticProvider` capability is advertised). Sending a pull request hangs (the server neither responds nor errors), so we opt out and rely on `textDocument/publishDiagnostics` (push), which the base class stores via `_observe_server_notification`.

### Why not build on #1357?

This PR supersedes two prior attempts that went stale:

- #1352 (jakeefr) — closed by a maintainer ("No reaction, closing")
- #1357 (Koushik-Salammagari) — open but abandoned (last activity 2026-04-21, 3+ months stale; MischaPanch's "you are still on this, right?" went unanswered)

Diagnosing why #1357's CI was flaky surfaced two root causes, both fixed here:

1. **Fixture import syntax was wrong.** `calculator.gleam` used `import test_repo/utils`, which Gleam 1.x rejects with `Unknown module`. The correct form is `import utils` (in-project module paths are relative to `src/`, without the project-name prefix). This is why "reference-finding tests fail" was observed on #1357 — the cross-file reference target was never resolvable, so the cross-file references test could not pass regardless of timing. *(Credit to Koushik-Salammagari for the original fixture structure and test coverage; the import line is the fix.)*

2. **`textDocument/diagnostic` (pull) hung the diagnostics test.** Gleam LSP advertises no `diagnosticProvider`; the base class's default `_supports_pull_diagnostics → True` sent a pull request the server never answered, timing the test out (240 s). Overriding to `False` routes through `publishDiagnostics` (push) instead.

#1357 was also built against a now-outdated `main`: it used `ProcessLaunchInfo` and the old `_get_initialize_params(repository_absolute_path)` signature, both superseded by the `DependencyProvider` + `_create_base_initialize_params` refactors. This PR is built on current `main` with the modern patterns (the same ones requested in the QML review of #1635).

### Notes

- The Gleam compiler is a self-contained Rust binary — **neither Erlang/OTP nor a separate LSP package is required** for symbol / reference / hover / diagnostics queries. (Full project compilation to BEAM bytecode does require Erlang, but the LSP's analysis path does not.)
- `gleam deps download` runs at server start and only needs network access to the Hex package mirror; no Erlang runtime.

## How did you verify your code works?

- **Local (Windows 11, gleam 1.17.0 standalone binary)**: all 6 tests pass in ~16 s:
  - `test_ls_is_running`
  - `test_document_symbols` — asserts `add` / `subtract` / `multiply` / `Calculator` / `demo` are present
  - `test_find_symbol` — `request_full_symbol_tree` (backing serena's `find_symbol` tool) sees symbols from both `calculator.gleam` and `utils.gleam`
  - `test_find_references_within_file` — `add` definition → its call site in `demo` (same file)
  - `test_find_references_across_files` — `format_output` defined in `utils.gleam` → its call sites in `calculator.gleam`
  - `test_file_diagnostics` — `undefined_symbol` in `diagnostics_sample.gleam` is flagged
- **CI (Linux ubuntu-latest, gleam 1.17.0 x86_64-musl)**: `gleam` is installed on the `other-langs` Linux batch (mirrors the qmlls setup); the `gleam` marker is added to `MARKERS_OTHER_LANGS`. Tests skip on Windows/macOS CI and off-CI without the binary via the central conftest guard.
- `ruff check`, `ruff format`, and `ty check` are all clean.

## Checklist


MischaPanch:
Third time's the charm (after #1352  and #1357) :)

Thanks for the contribution, merging!

Issue #839, opened 2025-12-14
Include scalaLSP operation tests in CI


YuMuuu:
**Is your feature request related to a problem? Please describe.**

https://github.com/oraios/serena/pull/528 We decided not to add various parts of this to CI due to compilation and execution time issues.

**Describe the solution you'd like**

Add the scalaLSP test to CI and fix it so that it works properly.

**Describe alternatives you've considered**

sbt projects generally use a strategy of sharing the build cache with artifacts to shorten dependency download and compile times in CI. In the previous PR, the context became too large, so we removed it from CI, but I think that by using the build cache, you can check the operation in CI without any problems.

**Additional context**

YuMuuu:
Since it is a multi-OS solution, I may not have much knowledge about it, but I think it should be possible.

PR #1767, opened 2026-07-27
Scala: give Metals the build roots rather than the repository root

### LINKED ISSUE #1766: Scala: Metals is given the repository root, so only one build of a monorepo is ever served
Preconditions:

- [X] I have made sure it's an actual issue, not a question (use [GitHub Discussions](https://github.com/oraios/serena/discussions) instead).
- [X] I have consulted the [user guide](https://oraios.github.io/serena/02-usage/) and verified that the issue cannot be resolved by adjusting configuration/following recommended workflows.
- [X] I have looked for similar issues and discussions, including closed ones.

Issue details:

- [X] I have provided a meaningful title and description.
- [X] I have explained how the issue arose and, where possible, added instructions on how to reproduce it.
- [X] I have added details on my setup: Serena version, MCP client, OS, the programming language(s), relevant configuration adjustments and project specifics.
- [X] If the issue relates to an application of Serena to an open-source project, I have added the link.

### Summary

In a repository that holds more than one Scala build below its root, `find_referencing_symbols` / `request_references` silently return nothing for every build but one. Serena sends Metals a single workspace folder — the repository root — and Metals serves one build per workspace folder, so the rest are left with no build target and answered by the fallback presentation compiler, which cannot see across files.

### Setup

- Serena `main` @ `9a9d07e83d8c1cba3458992707f440c624446c6d`; also reproduces on released 1.6.1, with `LanguageServerConfig(code_language=Language.SCALA)` in place of the `ls_id=` line below (the #1739 rename)
- Metals 1.6.4 (Serena's bundled default), Scala 3.3.6, sbt 1.11.7 (pinned by the fixture), Coursier 2.1.24
- OpenJDK 25.0.3, Python 3.14.6, Debian 13 (aarch64)
- No MCP client involved — driven directly through `SolidLanguageServer`, see below
- No configuration adjustments

### Reproduction

Save this as `sscce.py` at the root of a Serena checkout and run it against that checkout's environment (`uv sync --all-extras` first if you haven't):

```
uv run python sscce.py /tmp/mono
```

Needs `java`, `sbt` and `cs` on `PATH`. It writes two independent, self-contained sbt builds under `/tmp/mono`, then asks for the references to `Lib.greet` in each, waiting 4 minutes for a response.

```python
"""Serena + Metals in a monorepo: two independent sbt builds below the repository root."""

import os
import shutil
import sys
import time

REPO = os.path.abspath(sys.argv[1])

# --- fixture: <repo>/alpha and <repo>/beta, each a self-contained sbt build -------------
shutil.rmtree(REPO, ignore_errors=True)
for name in ("alpha", "beta"):
    files = {
        "build.sbt": 'ThisBuild / scalaVersion := "3.3.6"\n\nlazy val %s = (project in file("."))\n' % name,
        "project/build.properties": "sbt.version=1.11.7\n",
        f"src/main/scala/{name}/Lib.scala": f"package {name}\n\nobject Lib {{\n  def greet(who: String): String = s\"hello $who\"\n}}\n",
        f"src/main/scala/{name}/App.scala": f'package {name}\n\nobject App {{\n  def main(args: Array[String]): Unit = println(Lib.greet("world"))\n}}\n',
    }
    for rel, content in files.items():
        path = os.path.join(REPO, name, rel)
        os.makedirs(os.path.dirname(path), exist_ok=True)
        with open(path, "w") as f:
            f.write(content)

# --- run Serena's Scala language server over it ----------------------------------------


def show_message_request(params):
    """Answer Metals' "would you like to import the build?" prompt.

    Serena has no handler for window/showMessageRequest, and without one Metals aborts
    initialization ("Unexpected error initializing server ... method
    'window/showMessageRequest' not handled on client") and imports nothing at all. That is a
    separate gap; this handler removes it from the picture so the workspace folders are the
    only variable. There is precedent for such a handler in
    src/solidlsp/language_servers/ansible_language_server.py.
    """
    actions = params.get("actions") or []
    for action in actions:
        if action.get("title") == "Import build":
            return action
    return actions[0] if actions else None


ls.server.on_request("window/showMessageRequest", show_message_request)

    folders = ls._create_initialize_params().get("workspaceFolders", [])
    print("workspaceFolders:", [f["uri"] for f in folders])

    results: dict[str, list] = {}
    deadline = time.time() + 240
    while time.time() < deadline:
        for name in ("alpha", "beta"):
            if not results.get(name):
                # Lib.scala line 3 (0-based) is `  def greet(...)`; column 6 is on `greet`.
                results[name] = ls.request_references(os.path.join(name, "src", "main", "scala", name, "Lib.scala"), 3, 6)
        if all(results.values()):
            break
        time.sleep(10)

    for name, refs in sorted(results.items()):
        print(f"references to {name}.Lib.greet: {[r['relativePath'] for r in refs]}")
```

### Expected

Both builds resolve their own cross-file reference: `alpha/src/main/scala/alpha/App.scala` and `beta/src/main/scala/beta/App.scala`.

### Actual

```
workspaceFolders: ['file:///tmp/mono']
references to alpha.Lib.greet: ['alpha/src/main/scala/alpha/App.scala']
references to beta.Lib.greet: []
```

`beta` gets nothing, and keeps getting nothing for the full four minutes the script waits. `/tmp/mono/.metals/metals.log` says why:

```
INFO  Started: Metals version 1.6.4 in folders '/tmp/mono' for client Serena .
[…]
INFO  Connected to Build server: Bloop v2.0.17
INFO  no build target found for /tmp/mono/beta/src/main/scala/beta/Lib.scala. Using presentation compiler with project's scala-library version: 3.3.6
```

(Excerpt.) `alpha` gets that same line, but only before the build server connects, never after — it is the build that got imported. `beta`'s repeats on every attempt for the full four minutes.

Which of the two wins is not something you can rely on either; it falls out of a `collectFirst` over a directory listing.

### Analysis

Three things line up:

1. `DefaultInitializeParamsBuilder._apply_updates` sets `workspaceFolders` from `LanguageServerConfig.workspace_folders` (`src/solidlsp/initialize_params.py:59-69`), which defaults to `["."]` (`src/solidlsp/ls_config.py:902`) — the repository root. The Scala server does not override this; its `_create_base_initialize_params` sets only `locale`, `initializationOptions` and `capabilities`.
2. Metals builds one `Folder` per entry of the `workspaceFolders` it is given (`MetalsLanguageServer.initialize`, `metals/src/main/scala/scala/meta/metals/MetalsLanguageServer.scala:134-153`, falling back to `rootUri` only when the list is empty) and creates one `ProjectMetalsLspService` per folder (`WorkspaceFolders.initServices`, `.../internal/metals/WorkspaceFolders.scala:28-45`). One folder, one service, one build.
3. Given a workspace that is not itself a build root, Metals does look one level down for a build — but with `collectFirst`, so it stops at the first (`BuildTools.searchForBuildTool`, `.../internal/builds/BuildTools.scala:195-212`). That is the fallback the single-build-in-a-subdirectory case survives on, and it is exactly why the second build cannot.

(Metals references are at `scalameta/metals` @ `987b0cdfa11905abeee575de0968293f9837936e`.)

### On `ls_workspace_folders`

To be straight about the workaround: for a repository that is only Scala, `ls_workspace_folders: [alpha, beta]` does fix the repro above — those folders are passed straight through as `workspaceFolders`, and Metals gives each its own service. So this is not "no configuration can help"; it is that the default is wrong and the escape hatch does not reach far enough.

Where it stops reaching is a polyglot repository, which is the shape that put me here. `ls_workspace_folders` is one project-level list applied identically to every language server (`src/serena/ls_manager.py:49-56`, `src/serena/config/serena_config.py:326`) and it also sets what SolidLSP indexes. If another server needs the repository root as its folder — a TypeScript project with `package.json` at the top, say — then including `.` puts the root back in front of Metals and widens indexing again, and omitting it cuts that server off from its own sources. There is no value that suits both.

Even in the pure-Scala case it asks the user to know that Metals is one-service-per-folder, work out which directories are builds, and keep that list current by hand — for something Serena can see for itself.

Metals' own `metals.customProjectRoot` is no help either: it names a single root, and it lives in Metals' `UserConfiguration` rather than its initialization options, so it would arrive by `workspace/didChangeConfiguration`, which Serena never sends.

### Related

- #1627 / #1631 gave us `ls_workspace_folders`, but for scoping what gets indexed in an umbrella repo. This is a different axis: which directories are *builds*, per language server.
- #1444 fixed the same shape of problem for Elixir by searching subdirectories for `mix.exs`.

### Fix

Incoming PR I have working, modelled on #1444 and hanging off the `InitializeParamsBuilder` mechanism from #1631. With it, the script above prints:

```
workspaceFolders: ['file:///tmp/mono/alpha', 'file:///tmp/mono/beta']
references to alpha.Lib.greet: ['alpha/src/main/scala/alpha/App.scala']
references to beta.Lib.greet: ['beta/src/main/scala/beta/App.scala']
```

---

*Drafted with the help of Claude Code. I reproduced the bug myself, verified the source citations and the log output, and edited the text; questions are mine to answer.*

merlinorg:
Fixes #1766.

## Root cause

Metals serves one build per workspace folder — `MetalsLanguageServer.initialize` makes a `Folder` per entry of the `workspaceFolders` it is given and `WorkspaceFolders.initServices` creates a `ProjectMetalsLspService` for each. Serena sends exactly one: the repository root, from `LanguageServerConfig.workspace_folders` (default `["."]`) via `DefaultInitializeParamsBuilder`, which `ScalaLanguageServer` does not override.

Where the builds live below that root, Metals falls back to its own search, `BuildTools.searchForBuildTool` — one level down, `collectFirst`. That is what keeps a single build in a subdirectory working, and it is exactly why a second one cannot: every build but the first is left with no build target and answered by the fallback presentation compiler, so `find_referencing_symbols`/`request_references` come back empty with nothing in the logs at Serena's level to say why.

`ls_workspace_folders` is a real workaround for a pure-Scala repository, and #1766 says so. It runs out in a polyglot one: it is a single project-level list shared by every language server and it also sets what SolidLSP indexes, so once another server needs the repository root as its folder there is no value that suits both.

## Fix

`ScalaLanguageServer` detects the build roots under the repository root and passes them all as workspace folders, one Metals service per build. Detection follows the markers `BuildTools` probes for — sbt (including a `project/build.properties` naming an `sbt.version` with no `build.sbt`), mill, gradle, maven, scala-cli, Bazel, Deder, and a `.bsp`/`.bloop` holding a connection file. It stops descending at a build root, skips build output and dependency directories, follows symlinks as Metals does, and is bounded by depth.

Where the repository root is itself a build root it returns immediately and nothing changes for existing single-root users.

Two settings under `ls_specific_settings.scala` for where the heuristic guesses wrong: `project_roots` (an explicit list, relative to the repository root) and `project_root_scan_depth` (default 3), both validated in the manner of the existing `on_stale_lock`.

`ls_additional_workspace_folders` is preserved and appended — those folders can lie outside the repository, so detection could never recover them.

This uses the `_create_initialize_params_builder` extension point from #1631 as intended; it is the first override of it. The shape is #1444's `_find_mix_exs` for Elixir, generalised from "first match, one level" to "all of them, bounded depth".

## Behaviour change worth knowing

For a repository whose builds are below its root, Metals now writes `.metals/` and `.bloop/` into each build directory instead of the repository root, and those users get a one-time re-import. The H2 stale-lock check follows, running per build root rather than only at the repository root. Single-root repositories are untouched.

## Testing

33 unit tests in `test/solidlsp/scala/test_scala_build_roots.py` covering detection, the two settings, and the workspace folders that actually reach `initialize`. None of them need Metals, so they run in CI rather than joining the end-to-end suite #839 is about.

End-to-end, against the reproduction in #1766 — before, `beta` returns nothing for the full four minutes; after:

```
workspaceFolders: ['file:///tmp/mono/alpha', 'file:///tmp/mono/beta']
references to alpha.Lib.greet: ['alpha/src/main/scala/alpha/App.scala']
references to beta.Lib.greet: ['beta/src/main/scala/beta/App.scala']
```

`ruff check`, `ruff format --check`, `ty check src/serena src/solidlsp` and `ty check test --exclude test/resources` are clean.

## Checklist

---

*Prepared with the help of Claude Code; I have reviewed and tested the change and will handle review myself.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,8 @@ Status of the `main` branch. Changes prior to the next official version change w
     root as a workspace folder, and its own one-level search takes just the first build it finds — so in
     a monorepo all but one build were served with no build target, silently returning no cross-file
     references. The build roots are now detected and passed as workspace folders, one Metals service per
-    build; `ls_specific_settings.scala.project_roots` and `project_root_scan_depth` override the detection
+    build; `ls_specific_settings.scala.project_roots` and `project_root_scan_depth` override the
+    detection #1766
 
 * Hooks:
   - Add `serena-hooks --client=grok`, including Grok-native PreToolUse allow/deny output.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,11 @@ Status of the `main` branch. Changes prior to the next official version change w
   - Language servers and their dependency providers now go through the `subprocess_run` helper instead of
     calling `subprocess.run` directly (e.g. for installation processes), so all such subprocesses get 
     `stdin=DEVNULL` and can no longer interfere with the stdio MCP connection #1748
+  - `scala`: Fix: in a repository whose builds live below its root, Metals was given only the repository
+    root as a workspace folder, and its own one-level search takes just the first build it finds — so in
+    a monorepo all but one build were served with no build target, silently returning no cross-file
+    references. The build roots are now detected and passed as workspace folders, one Metals service per
+    build; `ls_specific_settings.scala.project_roots` and `project_root_scan_depth` override the detection
 
 * Hooks:
   - Add `serena-hooks --client=grok`, including Grok-native PreToolUse allow/deny output.
--- a/docs/03-special-guides/scala_setup_guide_for_serena.md
+++ b/docs/03-special-guides/scala_setup_guide_for_serena.md
@@ -64,6 +64,21 @@ Notes:
 - Ensure you completed the manual or auto‑import steps so that the build is compiled and indexed; otherwise, code navigation and references may be incomplete until the first successful compile.
 
 ---
+## Monorepos: builds below the repository root
+
+Metals serves one build per workspace folder, so what it needs is the build roots, not the repository root. Serena detects them: if the repository root is not itself a build root (no `build.sbt`, `build.mill`, `pom.xml`, `.bsp/`, …), it searches up to three levels below for directories that are, and passes those to Metals — one Metals service per build.
+
+Override the detection where it guesses wrong:
+
+```yaml
+# ~/.serena/serena_config.yml or .serena/project.yml
+ls_specific_settings:
+  scala:
+    project_roots: ["backend", "tooling/plugin"]  # relative to the repository root
+    project_root_scan_depth: 3                    # only applies when project_roots is unset
+```
+
+---
 ## Running Multiple Metals Instances
 
 Serena can run alongside other Metals instances (e.g., VS Code with Metals extension) on the same project. This is **fully supported** by Metals via H2 AUTO_SERVER mode.
--- a/src/solidlsp/language_servers/scala_language_server.py
+++ b/src/solidlsp/language_servers/scala_language_server.py
@@ -31,9 +31,14 @@ DEFAULT_ON_STALE_LOCK = "auto-clean"
 DEFAULT_LOG_MULTI_INSTANCE_NOTICE = True
 DEFAULT_PROJECT_ROOT_SCAN_DEPTH = 3
 
-# Files whose presence marks a directory as the root of a build Metals can import.
-# Mirrors the per-build-tool probes in Metals' `BuildTools` (scala/meta/internal/builds/BuildTools.scala).
+# Files whose presence marks a directory as the root of a build Metals can import, following the
+# per-build-tool probes in Metals' `BuildTools` (scala/meta/internal/builds/BuildTools.scala) and
+# `BazelBuildTool.workspaceSupportsBsp`. Deliberately partial: the probes that need to read a file's
+# contents (scala-cli's BSP scope) are left out, since missing a build root only returns us to the
+# previous behaviour, whereas a false positive would hide the real builds beneath it.
 BUILD_ROOT_MARKER_FILES = (
+    "MODULE.bazel",
+    "WORKSPACE",
     "build.gradle",
     "build.gradle.kts",
     "build.mill",
@@ -41,17 +46,22 @@ BUILD_ROOT_MARKER_FILES = (
     "build.mill.yaml",
     "build.sbt",
     "build.sc",
+    "deder.pkl",
+    "mill",
+    "mill.bat",
     "pom.xml",
     "project.scala",
     "settings.gradle",
     "settings.gradle.kts",
 )
 
-# Directories whose presence marks a directory as an already-configured Metals project.
-BUILD_ROOT_MARKER_DIRS = (".bloop", ".bsp")
+# Directories which, when they hold a JSON file, mark an already-configured Metals project
+# (`BuildTools.hasJsonFile`). An empty one is a leftover, not a build.
+BUILD_ROOT_MARKER_JSON_DIRS = (".bloop", ".bsp")
 
-# Directories never worth descending into when scanning for build roots: build output, dependencies,
-# and the two directories (`project`, `src`) that belong to a build we would already have recognised.
+# Directories not worth descending into when scanning for build roots: build output, dependencies,
+# and the directories belonging to a build we would have recognised at their parent. They are still
+# probed themselves — only the descent below them is skipped.
 BUILD_ROOT_SCAN_SKIP_DIRS = frozenset({"node_modules", "out", "project", "src", "target", "venv"})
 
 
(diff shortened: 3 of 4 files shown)


MischaPanch:
Merging this since I suspect that it makes sense. I have at one point given some attention to scala support in Serena, but the setup was so painful and impossible to make work in CI, so I gave up and am still in that given-up mindset.

I'm going to trust you that this is indeed an improvement for scala programmers. Since the changes are contained to scala, there's no danger of it influencing performance of other LS.

merlinorg:
> Merging this since I suspect that it makes sense. I have at one point given some attention to scala support in Serena, but the setup was so painful and impossible to make work in CI, so I gave up and am still in that given-up mindset.
> 
> I'm going to trust you that this is indeed an improvement for scala programmers. Since the changes are contained to scala, there's no danger of it influencing performance of other LS.

Thank you. I can confirm we run both these patches locally with success. If you have any need of scala CI assist, please yell, we run much of it.

MischaPanch:
>Thank you. I can confirm we run both these patches locally with success. If you have any need of scala CI assist, please yell, we run much of it.

@merlinorg well, my main issue is that apparently a project needs to be fully compiled in some special, Metals-specific way (which seemed to require arcane incantations and lot of waiting) to give references to symbols, otherwise it gives kind of references to  symbols, but those are not reliable (? I guess, at least - seems like it uses regex), there is no way to know whether Metals is done compiling, whether the way the project was previously compiled is accepted by metals, whether the returned references are complete or not. Metals starts some GUI and writes stuff to journals on disc, which I believe no LS user would want by default.

If you want to have a look at how to get tests for scala running in our github CI without the test taking over 10 mins to compile for a test project and without the test's success depending on whether the compile once ran through or not in the past (this last point in particular was driving me crazy), I'd be happy about that :). At the moment Scala is one of the least stably supported languages in Serena due to all that.

PR #1170, opened 2026-03-15
Add Ansible language server support

arikon:
## Summary

- Add Ansible language server integration (`@ansible/ansible-language-server`) with hover, completion, and definition capabilities
- Add path-based YAML/Ansible routing: `is_ignored_path()` override ensures only ansible-specific `.yml` files go to Ansible LS, while non-ansible YAML files fall through to YAML LS

### Language Server (`src/solidlsp/language_servers/ansible_language_server.py`)
- Full LSP client with `initializationOptions` matching VS Code Ansible extension
- Auto-installs `@ansible/ansible-language-server@1.2.3` via npm; supports `ls_path` override for custom builds
- `is_ignored_path()` classifies files by directory components (`roles/`, `tasks/`, `handlers/`, etc.) and filename patterns (`playbook*.yml`, `site.yml`)

### Tests
- `test_ansible_basic.py` — initialization, hover, completions (3 tests)

PR #1769, opened 2026-07-27
Scala: answer Metals' build-import prompt

### LINKED ISSUE #839: Include scalaLSP operation tests in CI

### LINKED ISSUE #1768: Scala: Metals' build-import prompt is never answered, so a project that was not imported by another editor never gets a build server
Preconditions:

- [X] I have made sure it's an actual issue, not a question (use [GitHub Discussions](https://github.com/oraios/serena/discussions) instead).
- [X] I have consulted the [user guide](https://oraios.github.io/serena/02-usage/) and verified that the issue cannot be resolved by adjusting configuration/following recommended workflows.
- [X] I have looked for similar issues and discussions, including closed ones.

Issue details:

- [X] I have provided a meaningful title and description.
- [X] I have explained how the issue arose and, where possible, added instructions on how to reproduce it.
- [X] I have added details on my setup: Serena version, MCP client, OS, the programming language(s), relevant configuration adjustments and project specifics.
- [X] If the issue relates to an application of Serena to an open-source project, I have added the link.

### Summary

Metals asks, via `window/showMessageRequest`, whether to import a workspace it has not seen before. Serena registers no handler for that request, so it comes back as `MethodNotFound`, Metals logs `Unexpected error initializing server` and imports nothing. The project then has no build server and no build target, and every cross-file query is answered by the fallback presentation compiler, which sees one file at a time — so `find_referencing_symbols` returns nothing, with no error at Serena's level to say why.

This is not a monorepo edge case. It is any Scala project that has not already been imported by something else. The Scala setup guide's Quick Start ("open your project in VS Code, accept Import build…") is a workaround for it, and simply doesn't work for many workflows (multiple independent worktrees, devcontainers).

### Setup

- Serena `main` @ `9a9d07e83d8c1cba3458992707f440c624446c6d`
- Metals 1.6.4 (Serena's bundled default), Scala 3.3.6, sbt 1.11.7 (pinned by the fixture)
- OpenJDK 25.0.3, Python 3.14.6, Debian 13 (aarch64)
- No MCP client involved — driven directly through `SolidLanguageServer`
- No configuration adjustments

### Reproduction

Save as `sscce.py` at the root of a Serena checkout, then (`uv sync --all-extras` first if needed):

```
uv run python sscce.py /tmp/plain
```

Needs `java`, `sbt` and `cs` on `PATH`. It writes a perfectly ordinary single-module sbt project — no `.bloop`, no `.bsp`, nothing pre-imported — and asks for the references to `Lib.greet`, retrying for three minutes. Note that on a working Serena this causes Metals to run `sbt bloopInstall` and fetch Scala 3.3.6, so run it somewhere that is fine.

```python
"""Serena + Metals on a single-root sbt project that has never been imported."""

import os
import shutil
import sys
import time

REPO = os.path.abspath(sys.argv[1])

# --- fixture: an ordinary sbt project at the repository root, no .bloop/.bsp ------------
shutil.rmtree(REPO, ignore_errors=True)
files = {
    "build.sbt": 'ThisBuild / scalaVersion := "3.3.6"\n\nlazy val plain = (project in file("."))\n',
    "project/build.properties": "sbt.version=1.11.7\n",
    "src/main/scala/plain/Lib.scala": 'package plain\n\nobject Lib {\n  def greet(who: String): String = s"hello $who"\n}\n',
    "src/main/scala/plain/App.scala": 'package plain\n\nobject App {\n  def main(args: Array[String]): Unit = println(Lib.greet("world"))\n}\n',
}
for rel, content in files.items():
    path = os.path.join(REPO, rel)
    os.makedirs(os.path.dirname(path), exist_ok=True)
    with open(path, "w") as f:
        f.write(content)

    refs = []
    deadline = time.time() + 180
    while time.time() < deadline:
        # Lib.scala line 3 (0-based) is `  def greet(...)`; column 6 is on `greet`.
        refs = ls.request_references(os.path.join("src", "main", "scala", "plain", "Lib.scala"), 3, 6)
        if refs:
            break
        time.sleep(10)
    print("references to Lib.greet:", [r["relativePath"] for r in refs])
```

### Expected

```
references to Lib.greet: ['src/main/scala/plain/App.scala']
```

### Actual

```
references to Lib.greet: []
```

`/tmp/plain/.metals/metals.log`:

```
INFO  Started: Metals version 1.6.4 in folders '/tmp/plain' for client Serena .
ERROR Unexpected error initializing server: 
org.eclipse.lsp4j.jsonrpc.ResponseErrorException: method 'window/showMessageRequest' not handled on client.
	at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.handleResponse(RemoteEndpoint.java:220)
	at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.consume(RemoteEndpoint.java:204)
	…
INFO  no build target found for /tmp/plain/src/main/scala/plain/Lib.scala. Using presentation compiler with project's scala-library version: 3.3.6
```

(Stack trace trimmed.) That last line then repeats for every attempt, for the full three minutes. There is no build server: no `Connected to Build server`, no `.bloop/` produced.

The question Metals was asking is `Messages.ImportBuild`: *"New sbt workspace detected, would you like to import the build?"*, offering `Import build` / `Not now` / `Don't show again`.

### Analysis

Serena answers only the server-to-client requests a language server has registered a handler for; anything else is failed back as `MethodNotFound` (`src/solidlsp/ls_process.py:433-442`). `ScalaLanguageServer._start_server` registers none, so Metals' prompt is one of them. There is precedent for handling it elsewhere in the codebase — `ansible_language_server.py:308` registers exactly this handler, added in #1170, for exactly this reason ("without this handler the client replies with `MethodNotFound`, which the ansible LS treats as fatal").

Declining the question rather than answering it would not help. There is a switch for it — `disableShowMessageRequest`, read from the `metals.disable-show-message-request` system property (`MetalsServerConfig.scala:147-150`, `binaryOption` at `:193`) — and since Serena launches the Metals JVM it could set it, though not over LSP, as it is not a client capability. But where it is set, `ConfiguredLanguageClient.showMessageRequest` (`:52-64`) answers with the prompt's own default, and for this prompt that is `Not now` (`BloopInstall.scala:190-205`): no import either. Answering is the only route from a fresh checkout to a build server.

Nor is there a Serena-side setting that helps: the workaround is to import the build with another tool first, which is what the setup guide's Quick Start describes. Once `.bloop/` exists Metals auto-connects and never asks, which is presumably why this has gone unreported — the documented workflow steps around it.

(Metals references are at `scalameta/metals` @ `987b0cdfa11905abeee575de0968293f9837936e`, under `metals/src/main/scala/scala/meta/internal/`: `metals/Messages.scala:207-233` for the prompt, `builds/BloopInstall.scala:174-213` for the request and how the answer is compared against `ImportBuild.yes`, `metals/clients/language/ConfiguredLanguageClient.scala:52-64` and `metals/MetalsServerConfig.scala:147-150` for the fallback.)

### Related

- #688 — different bug (that reporter had run `bloopInstall` by hand), but the thread's "the fact that this special compilation is needed makes scala support very annoying" is about this friction. With the prompt answered, the special compilation is no longer the user's job.
- #839 — the end-to-end Scala suite is skipped partly because a project must be compiled before the tools work. This does not close that, but it removes the manual step from it.
- #1766 — the monorepo issue I filed; its reproduction script installs a `showMessageRequest` handler of its own precisely to work around this, and could drop it once this is fixed.
- #303 — the same class of bug (unhandled server request) for the C# server.
- #1170 — added the Ansible handler this would follow.

### Fix

PR incoming. Registers the handler on the Scala server and answer the three prompts that lead to a build server — `Import build`, `Import changes`, `Connect` — dismissing anything else with `null`.

With it, the script above prints the expected line, and the log shows `time: ran 'sbt bloopInstall' in 5.54s` followed by `Connected to Build server: Bloop v2.0.17`.

---

*Drafted with the help of Claude Code. I reproduced the bug myself, verified the source citations and the log output, and edited the text; questions are mine to answer.*

merlinorg:
Fixes #1768, where the reproduction and the source trail are.

In short: Metals asks whether to import a workspace it has not seen before, Serena registered no `window/showMessageRequest` handler, so the request came back `MethodNotFound` and no build was ever imported — leaving every cross-file query to the fallback presentation compiler. It affects any Scala project not already imported by another editor, not just monorepos.

## Fix

Register the handler, following `ansible_language_server.py:308` (added in #1170 for the same reason).

Answer the three prompts that lead to a build server — `Import build`, `Import changes`, `Connect` — and dismiss anything else with `null`. The allow-list is deliberate rather than "pick the first action": a prompt we do not recognise is one whose consequences we cannot judge, and Metals' others include offers to kill a Bloop process (`OldBloopVersionRunning`) and to open a new window (`NewScalaProject`). `Don't show again` is never chosen, since Metals persists that dismissal in the project's own state.

The one prompt on the path this does *not* answer is `ChooseBuildTool` ("Multiple build definitions found. Which would you like to use?"), whose actions are the build tools' executable names. A workspace holding more than one kind of build therefore still goes unimported; naming the build tool for someone felt like a guess of a different order, but say the word and I'll add it. The comment and the setup guide both call it out rather than implying full coverage.

Answering yes lets Metals run the project's build tool (for sbt, `sbt bloopInstall`), which is a real side effect on the user's machine, so `ls_specific_settings.scala.auto_import_build: false` declines instead and restores the previous behaviour minus the protocol error.

## Testing

Nine tests (thirteen cases) in `test/solidlsp/scala/test_scala_show_message_request.py`, built on Metals' own prompt payloads — the affirmative cases, the opt-out and its route from `ls_specific_settings`, `Don't show again`, an unrecognised prompt, the build-tool choice, no actions, and malformed actions. No Metals needed, so they run in CI.

End-to-end on a fresh single-module sbt project with no `.bloop`/`.bsp` — before, `request_references` returns `[]` for three minutes and the log shows the `MethodNotFound` error; after:

```
references to Lib.greet: ['src/main/scala/plain/App.scala']
```

with the log showing `time: ran 'sbt bloopInstall' in 5.54s` then `Connected to Build server: Bloop v2.0.17`.

`ruff check`, `ruff format --check`, `ty check src/serena src/solidlsp` and `ty check test --exclude test/resources` are clean.

## Docs

The Scala setup guide was written around this bug, so its Quick Start is now "start Serena and let it import", with the VS Code and manual `bloopInstall` routes kept below as the ways to import the build yourself. The language-support page no longer says Scala "requires some manual setup".

This does not close #839, but it removes the manual pre-import step that is part of why the end-to-end Scala suite is hard to run.

## Checklist

---

*Prepared with the help of Claude Code; I have reviewed and tested the change and will handle review myself.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,12 @@ Status of the `main` branch. Changes prior to the next official version change w
   - Language servers and their dependency providers now go through the `subprocess_run` helper instead of
     calling `subprocess.run` directly (e.g. for installation processes), so all such subprocesses get 
     `stdin=DEVNULL` and can no longer interfere with the stdio MCP connection #1748
+  - `scala`: Fix: Metals asks via `window/showMessageRequest` whether to import a workspace it has not
+    seen before, and Serena had no handler, so the request failed with `MethodNotFound` and Metals
+    aborted its initialization ("Unexpected error initializing server"). No build was ever imported and
+    every cross-file query fell back to the presentation compiler, unless the project happened to have
+    been imported beforehand by another editor. The prompts that lead to a build server are now answered;
+    anything else is dismissed. `ls_specific_settings.scala.auto_import_build: false` opts out
 
 * Hooks:
   - Add `serena-hooks --client=grok`, including Grok-native PreToolUse allow/deny output.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,11 +34,12 @@ Status of the `main` branch. Changes prior to the next official version change w
     calling `subprocess.run` directly (e.g. for installation processes), so all such subprocesses get 
     `stdin=DEVNULL` and can no longer interfere with the stdio MCP connection #1748
   - `scala`: Fix: Metals asks via `window/showMessageRequest` whether to import a workspace it has not
-    seen before, and Serena had no handler, so the request failed with `MethodNotFound` and Metals
-    aborted its initialization ("Unexpected error initializing server"). No build was ever imported and
-    every cross-file query fell back to the presentation compiler, unless the project happened to have
-    been imported beforehand by another editor. The prompts that lead to a build server are now answered;
-    anything else is dismissed. `ls_specific_settings.scala.auto_import_build: false` opts out
+    seen before, and Serena had no handler, so the request failed with `MethodNotFound` and Metals gave
+    up on the import ("Unexpected error initializing server"). No build server was ever connected and
+    every cross-file query fell back to the presentation compiler, which sees one file at a time, unless
+    the project happened to have been imported beforehand by another editor. The three prompts that lead
+    to a build server are now answered; anything else is dismissed, including the choice between several
+    build definitions in one workspace. `ls_specific_settings.scala.auto_import_build: false` opts out
 
 * Hooks:
   - Add `serena-hooks --client=grok`, including Grok-native PreToolUse allow/deny output.
--- a/docs/01-about/020_programming-languages.md
+++ b/docs/01-about/020_programming-languages.md
@@ -146,7 +146,7 @@ Some languages require additional installations or setup steps, as noted.
 * **Rust**  
   (requires [rustup](https://rustup.rs/) - uses rust-analyzer from your toolchain)
 * **Scala**  
-  (requires some [manual setup](../03-special-guides/scala_setup_guide_for_serena); uses Metals LSP)
+  (uses Metals LSP, which imports the build on first use — see the [setup guide](../03-special-guides/scala_setup_guide_for_serena))
 * **SCSS / Sass / CSS**
   (experimental; requires Node.js + npm; uses [some-sass-language-server](https://github.com/wkillerud/some-sass) to handle
   `.scss`, `.sass`, and `.css`)
--- a/docs/03-special-guides/scala_setup_guide_for_serena.md
+++ b/docs/03-special-guides/scala_setup_guide_for_serena.md
@@ -21,7 +21,7 @@ Start Serena in your project root. Metals asks whether to import a workspace it 
 
 Set `auto_import_build: false` under `ls_specific_settings.scala` to decline instead; you then need to import the build yourself by one of the routes below, or cross-file queries will be served by the fallback presentation compiler and see only one file at a time.
 
-Serena answers only the prompts that lead to a build server (“Import build”, “Import changes”, “Connect”). Anything else Metals asks is dismissed and logged.
+Serena answers three of Metals' prompts — “Import build”, “Import changes”, “Connect”. Anything else Metals asks is dismissed and logged, including “Multiple build definitions found. Which would you like to use?”, so a workspace holding more than one kind of build (say both an sbt and a Maven definition) still needs importing by one of the routes below.
 
 ---
 ## Importing the build yourself (VS Code)
--- a/docs/03-special-guides/scala_setup_guide_for_serena.md
+++ b/docs/03-special-guides/scala_setup_guide_for_serena.md
@@ -15,7 +15,16 @@ Install the following on your system and ensure they are available on `PATH`:
   - Serena uses `cs` if available; if only `coursier` exists, it will attempt to install `cs`. If neither is present, install Coursier first.
 
 ---
-## Quick Start (Recommended: VS Code + Metals auto‑import)
+## Quick Start
+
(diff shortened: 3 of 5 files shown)


opcode81:
Thanks. We haven't tested this, but it looks good.

PR #1775, opened 2026-07-28
ci: attach provenance and SBOM attestations to the published image

kobihikri:
Hi, and thanks for Serena.

`.github/workflows/docker.yml` publishes the image, but the pushed manifest carries no provenance or SBOM attestation. Someone pulling it cannot check that it was built by this workflow, from this repository, at that tag.

Serena is an MCP server that gets pointed at a codebase and given the ability to read and edit it, so the image is trusted with source that people generally do not want leaving their machine. That is the case where being able to confirm the image's origin before wiring it into an assistant config is worth having.

The change is two lines on the build step:

```yaml
          push: ...
          provenance: mode=max
          sbom: true
```

BuildKit attaches both to the image manifest, so they travel with the image. **No permissions change is needed** — nothing has to gain `id-token`.

```
docker buildx imagetools inspect <image>:<tag> --format '{{ json .Provenance }}'
```

Two caveats: `mode=max` records build arguments (`provenance: true` is the smaller option), and attestations add an extra manifest to the index, which the registry handles.

No SLSA level claimed.

Disclosure: I used AI assistance to help spot this and prepare the change, and I read the workflow myself.


MischaPanch:
Thanks! Merging

PR #1777, opened 2026-07-29
Upgrade mcp SDK to 2.0.0 (protocol revision 2026-07-28)

### LINKED ISSUE #1776: Upgrade to mcp 2.0 (MCP protocol revision 2026-07-28)
`mcp` 2.0.0 was released on 2026-07-28, implementing MCP protocol revision `2026-07-28`. Serena currently pins `mcp==1.28.1`, and the upgrade cannot be a plain version bump: the SDK renamed `mcp.server.fastmcp` to `mcp.server.mcpserver`, renamed `FastMCP` to `MCPServer`, and removed some symbols Serena imports. On `mcp==2.0.0` the current code fails at import:

```
src/serena/tools/tools_base.py:11: in <module>
    from mcp.server.fastmcp import Context
E   ModuleNotFoundError: No module named 'mcp.server.fastmcp'
```

Opening this first rather than sending a PR straight away, since CONTRIBUTING.md asks for an issue when a change affects behaviour, and two user-visible behaviours do change. I have a working branch ready and will link it below.

### What the upgrade requires

Mechanical:
- Import path renames in `src/serena/mcp.py`, `src/serena/tools/tools_base.py`, `test/serena/test_mcp.py`.
- `ToolAnnotations` fields are now snake_case (`readOnlyHint` -> `read_only_hint`, `destructiveHint` -> `destructive_hint`).
- `Context` lost its `ServerSessionT` type parameter (3 -> 2); `ServerSessionT` no longer exists. The base `Tool.run()` also no longer accepts an optional context.

Behaviour-affecting, and the reason for this issue:

1. **`FASTMCP_*` environment variables no longer configure the server.** `SerenaMCPFactory.create_mcp_server()` currently does `Settings.model_config = SettingsConfigDict(env_prefix="FASTMCP_")` to stop `.env` files in user projects from overriding MCP settings. In `mcp` 2.0 `Settings` is no longer a pydantic-settings `BaseSettings`; it is a plain pydantic `BaseModel` populated from the `MCPServer` constructor, so it reads neither the environment nor `.env` files.

   Worth flagging explicitly: that assignment does not raise on 2.0, it simply stops doing anything, so the tests stay green either way. The `.env` problem the override existed to solve is now handled by the SDK itself, so my branch deletes the override (which also drops the need for pydantic-settings, previously transitive via mcp 1.x). But losing `FASTMCP_*` support is a real change for anyone relying on it, and whether you want it reintroduced explicitly is your call, not something I wanted to decide in a migration PR.

2. **`create_mcp_server()` no longer accepts `host`/`port`.** The SDK moved them from the `MCPServer` constructor to the transport-specific `run()` call. My branch drops them from the factory signature and has `cli.py` pass them to `server.run()` for non-stdio transports. Only one in-repo caller exists, but the factory is plausibly used externally, so an alternative is keeping the parameters on the factory and storing them until `run()`.

There is a related trap I hit that is worth recording: `from mcp.server.mcpserver import server` must bind the **submodule**, not the package, or the existing `server.configure_logging` override in `mcp.py` silently stops taking effect (the SDK reads that name from `mcp/server/mcpserver/server.py`'s own namespace). Both of Serena's SDK monkey-patches fail invisibly under a naive port.

### Status of the branch

`ty check` and `ruff` clean, `test/serena/test_mcp.py` 64/64, and `test/serena` 746 passed / 14 failed, where all 14 failures are Go/Rust/C#/PowerShell language-server tests failing on this machine for want of those toolchains (`RuntimeError: Go is not installed`) and are unrelated to the upgrade.

Happy to adjust either behavioural decision above, or to split the change, whichever you prefer.

Kayaba-Attribution:
Closes #1776 (opened first per CONTRIBUTING.md, since two user-visible behaviours change).

`mcp` 2.0.0 shipped on 2026-07-28. Serena pins `mcp==1.28.1`, and on 2.0.0 the current code fails at import (`No module named 'mcp.server.fastmcp'`), so this is a rename-and-reshape port rather than a version bump.

## Checklist

## What changed

Mechanical:
- `mcp.server.fastmcp.*` -> `mcp.server.mcpserver.*` in `mcp.py`, `tools_base.py`, `test_mcp.py`; `FastMCP` -> `MCPServer` (kept under the existing `FastMCP` local alias to keep the diff small).
- `ToolAnnotations`: `readOnlyHint`/`destructiveHint` -> `read_only_hint`/`destructive_hint`.
- `Context[ServerSessionT, LifespanContextT, RequestT]` -> `Context[LifespanContextT, RequestT]`; `ServerSessionT` no longer exists. The `| None` default also goes, because the base `Tool.run()` no longer accepts an optional context (no call site passed `None`).

Behaviour-affecting, both described in #1776 and in `CHANGELOG.md`:
- **`FASTMCP_*` env vars no longer configure the server.** The `Settings.model_config = SettingsConfigDict(...)` override is deleted: `Settings` is now a plain pydantic `BaseModel` populated from the constructor, so it reads neither the environment nor `.env` files, and the override had become a no-op that does not raise. The `.env`-bleed problem it existed to prevent is handled by the SDK now. This also removes Serena's need for pydantic-settings (previously transitive via mcp 1.x). If you would rather keep `FASTMCP_*` support, it needs reintroducing explicitly and I am happy to do that instead.
- **`create_mcp_server()` no longer takes `host`/`port`.** The SDK moved them to the transport-specific `run()` call, so `cli.py` passes them to `server.run()` for non-stdio transports. Only one in-repo caller exists; if you consider the factory external API, I can keep the parameters and store them until `run()`.

One subtlety worth a review look: `from mcp.server.mcpserver import server` intentionally binds the **submodule**, not the package. The SDK resolves `configure_logging` from `mcp/server/mcpserver/server.py`'s own namespace, so binding the package instead would leave the existing `server.configure_logging = configure_logging` override in `mcp.py` silently ineffective.

## Verification

- `poe type-check`: clean (both `_ty_core` and `_ty_test`).
- `poe lint`: clean (`ruff format --check`, `ruff check`).
- `pytest test/serena/test_mcp.py`: 64 passed.
- `pytest test/serena`: 746 passed, 14 failed, 8 skipped. All 14 failures are Go/Rust/C#/PowerShell language-server tests that fail on this machine for lack of those toolchains (`RuntimeError: Go is not installed`); none reference mcp. I could not exercise the Go/Rust/C#/PowerShell paths locally, so CI coverage there is worth confirming.


ebarkhordar (commented):
Nice migration, and thanks for putting the two behaviour changes in the CHANGELOG rather than leaving them to be discovered.

I checked the SDK claims in a clean container against `mcp==2.0.0` and they hold: `MCPServer.__init__` no longer accepts `host`/`port`, `run_sse_async` and `run_streamable_http_async` both take them keyword-only, and `run_stdio_async` takes neither, so the `transport != "stdio"` guard in `cli.py` is doing real work rather than being defensive. `Settings` is a plain `BaseModel` with an empty `model_config`, so dropping the `env_prefix` override is right. `ToolAnnotations` keeps `populate_by_name=True`, so `read_only_hint=` and `readOnlyHint=` serialise identically. `test/serena/test_mcp.py` is 64 passed at 1e436e53.

One thing I think the migration misses, `src/serena/tools/tools_base.py:346`:

```python
client_info = cast(Implementation, client_params.clientInfo)
```

2.0 renamed the protocol models to snake_case with camelCase aliases, and `InitializeRequestParams.clientInfo` is one of them. Pydantic aliases apply to validation and serialisation, not to attribute access, so this raises `AttributeError`. It sits inside the `except Exception` at :351, so it does not break a tool call: it logs `Failed to get client info` at INFO on every call instead, and leaves `Tool._last_tool_call_client_str` at `None`, which is what `dashboard.py:632` passes as `current_client`.

Reproduced in a clean container, in the venv `uv sync --frozen` resolves for this branch at 1e436e53, executing line 346 as read from the file against an `InitializeRequestParams` validated from the camelCase payload a client sends on initialize:

```
mcp == 2.0.0 | serena.tools.tools_base imported OK
--- executing: client_info = cast(Implementation, client_params.clientInfo)
RESULT: AttributeError: 'InitializeRequestParams' object has no attribute 'clientInfo'
```

The same script on `mcp==1.28.1`, so it is the upgrade and not my harness:

```
mcp == 1.28.1
RESULT: client_str = 'claude-code 2.1.0'
```

`client_params.client_info` should be all it needs. `session.client_params` is unchanged in 2.0, and `Implementation` keeps `name`, `title` and `version`, so nothing else on those two lines has to move.

On the rest of the tree: I diffed the pydantic field names of every model the two SDK versions share and got 78 camelCase fields removed across 45 models, then searched for all 39 distinct names. `clientInfo` at that line is the only hit, so I do not think a second one is hiding.

I did not drive a real client handshake, so the dashboard consequence is read off the call path rather than observed in the UI.

MischaPanch:
On the surface, the diff looks good, thanks for the contribution! @opcode81 will check this in more depth when he has time

xormania:
Ran this branch (head 1e436e53) through some independent verification while the deeper review is pending — sharing the data.

**Tests**: fresh clone, `uv sync --extra dev` (resolves mcp 2.0.0 from PyPI), then `test/serena/test_mcp.py`: 64/64 pass. The wider `test/serena` suite passes on my machine apart from environment-dependent language-server startups (.NET 10 / pwsh not installed here) — none of those failures touch this PR's seven files.

**Independently reproduced @ebarkhordar's `clientInfo` finding, end-to-end on a live server**: ran the branch's `serena start-mcp-server --transport stdio` — initialize, `tools/list` (29 tools via the existing registration path), and tool calls all work, with exactly one INFO line per call: `Failed to get client info: 'InitializeRequestParams' object has no attribute 'clientInfo'` (raised at tools_base.py:346, swallowed at :351). The one-token `client_info` fix works — verified attribute access against 2.0.0's `Implementation`. I also re-ran a renamed-field sweep independently (my diff of `mcp.types`: ~80 fields across 46 models) and can confirm their conclusion: that line is the only camelCase SDK-model attribute access anywhere in serena src+test.

**On timing**: at the wire level this merge is conservative — mcp 2.0.0 serves every protocol revision 2024-11-05 through 2026-07-28 from one server (`mcp_types/version.py`; handshake negotiation `runner.py:425`), and the stateless 2026-07-28 semantics engage only when a client sends the per-request `_meta` envelope (`runner.py:504-517`; verified empirically against the branch server with 2024-11-05 and 2025-06-18 clients). What current users actually see are the PR's two documented behavior changes plus the dependency swap below.

**One measured improvement worth having on the record**: on 1.28.1, sync tool handlers run on the event loop — during a deliberately slow 6s sync tool, a mid-tool `ping` goes unanswered for 5.50s (three runs: 5.5023/5.5032/5.5035s; minimal harness, both SDK versions). On 2.0.0 the same mid-tool ping answers in 1.2–1.3ms (sync handlers moved to worker threads, `resolve.py:556`). For a server whose tool calls routinely run for seconds, clients that treat silence as death would have been getting false positives during long tools.

One item that may deserve an explicit maintainer look because it is a decision rather than a bug: the lockfile swaps the HTTP/TLS stack (httpx→httpx2, certifi→OS trust store) and adds `opentelemetry-api` (on by default in the SDK; no-op without an exporter). Flagging for visibility, not objecting.

The deeper 2026-07-28 items (per-request client identity via `_meta`, session-handle semantics, the per-connection lifespan comment in `mcp.py`, SSE deprecation messaging) look correctly out of scope for a port PR — happy to help map those separately if useful.

opcode81:
## Blocker: client session identification breaks under v2

`Tool.apply_ex` derives the client session key from the identity of the SDK's session
object (`session_id = "%x" % id(mcp_ctx.session)`) and injects it into session-aware
tools.

Tested against **mcp 2.1.1**: `ctx.session` is a **new object for every request**,
including over **stdio**. Under v1 the instance is stable for the lifetime of the
connection. The mechanism degrades to "every tool call is a new session" — silently, with
no error.

Consequence: the project activation message and all mode prompts are re-emitted on every
call instead of once per session. Affected consumers, both routing into
`SerenaAgent._project_prompt_status`:

- `ActivateProject.apply` → `get_project_activation_message(session_id)`
- `InitialInstructions.apply` → `create_system_prompt(session_id)`

Related: `apply_ex` also reads `mcp_ctx.session.client_params.clientInfo` for
`_last_tool_call_client_str`. `client_params` is the result of the `initialize`
handshake, which no longer exists at protocol revision 2026-07-28.

This is intentional at the protocol level, not an SDK bug:
[SEP-2575](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2575)
removes the handshake,
[SEP-2567](https://modelcontextprotocol.io/seps/2567-sessionless-mcp) removes sessions
and `Mcp-Session-Id` entirely. For modern HTTP clients there is no transport-level way to
tell whether two calls come from the same caller. Whether the per-request `ctx.session`
object we observe **on stdio** is likewise intended is undocumented, and worth checking
before we design around it.

opcode81:
@MischaPanch please see my comment above. We'll have to decide how to handle this...

MischaPanch:
Man, this is absolutely crazy... I wish English had a good word for Verschlimmbesserung

opcode81:
Superseded by #2082

Issue #1771, opened 2026-07-28
Add opt-in lazy language-server initialization


TyceHerrman:
**Problem**

Codex may create multiple stdio MCP runtimes for one project. With Serena 1.6.1, four Serena processes started four `rust-analyzer` instances within 23 seconds; three reached 9.25–9.76 GB before the machine crashed.

Codex owns runtime reuse and cleanup, as discussed in #693 and openai/codex#30408. This request is for an optional Serena-side guard.

PR #227 made language-server initialization asynchronous but still starts it during project activation. True first-use initialization was previously prototyped but not merged:

- https://github.com/oraios/serena/pull/209#issuecomment-2993770930
- https://github.com/dayowe/serena/commit/9816058869eacb0e09ffe40b472195f7782c3317

**Requested solution**

Add an opt-in lazy mode:

- Do not start language servers during project activation.
- Initialize once through the existing serialized queue when the first dependent tool runs.
- Let non-LSP tools run without triggering initialization.
- Handle concurrent first calls, failures, restarts, project switching, and shutdown.
- Keep eager initialization as the default.

The tradeoff is latency on the first dependent call.

opcode81:
This probably won't be necessary if #1556 is addressed, so I would prioritise that.
PR #1557 already exists, but isn't finished.

TyceHerrman:
Great! missed that in reviewing prior work. I'll check it out

opcode81:
Closing as "not planned" for now.

PR #1779, opened 2026-07-29
Rust: reduce rust-analyzer memory and reload churn

### LINKED ISSUE #1556: rust-analyzer: unbounded memory (40+ GB) on large Rust workspaces — hardcoded heavyweight init options

TyceHerrman:
## Summary

continuation of #1557.

- Disable rust-analyzer cache priming
- Disable Cargo autoreload while keeping `checkOnSave` enabled
- Give published Rust diagnostics an eight-second minimum wait while preserving larger configured timeouts
- Trigger the retained check-on-save path when diagnostics are requested, avoiding the empty initial result seen when cache priming is disabled
- Add focused settings/diagnostics tests and a changelog entry

The selected commits from @davidcforbes and @MischaPanch were cherry-picked the test completion, diagnostics stabilization, generic commentary, and current changelog entry are a separate follow-up commit.

## Scope

This does not add a public API, a new configuration option, or lazy language-server initialization. Lazy initialization from #1771 is intentionally out of scope.

## Validation

- `uv sync --extra dev --locked`
- New rust-analyzer settings tests: 5 passed
- Existing Rust diagnostics test: passed three consecutive runs
- `uv run poe test -m rust -q --tb=short`: 39 passed, 1 skipped
- `uv run poe lint`
- `uv run poe type-check`
- `git diff --check`

Controlled regression probe on the previously affected workspace:

- One patched rust-analyzer instance, guarded by a 6 GiB RSS cap
- Document symbols, `TimeProvider` workspace-symbol lookup, and Rust diagnostics passed
- `cargo check -p codex-core --locked` passed twice: 203.45 s and 3.04 s
- Post-build symbol lookup passed
- Peak rust-analyzer RSS: 3,758.2 MiB, below the 6 GiB cap and the recorded 9.25–9.76 GiB incident baseline
- Final two-minute idle window stabilized from a 3,067.4 MiB initial mean to a 4.1 MiB final mean
- No Cargo-phase server-status events or workspace-reload errors were observed

Fixes #1556

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ Status of the `main` branch. Changes prior to the next official version change w
   - `jet_brains_find_symbol`: Disallow wildcard-only search, delegating to overview tool if request is for file
 
 * Language Servers:
+  - Rust: reduce rust-analyzer memory usage and reload churn by disabling cache priming and Cargo autoreload while preserving diagnostics.
   - `typescript`: Fix: on large projects, the first `find_referencing_symbols`/`request_references` call
     could silently race tsserver's project load and return incomplete results, because the fixed 2s
     grace for tsserver to *start* reporting `$/progress` (distinct from the separate, already
--- a/src/solidlsp/language_servers/rust_analyzer.py
+++ b/src/solidlsp/language_servers/rust_analyzer.py
@@ -505,9 +505,19 @@ class RustAnalyzer(SolidLanguageServer):
                 "showUnlinkedFileNotification": True,
                 "showDependenciesExplorer": True,
                 "assist": {"emitMustUse": False, "expressionFillDefault": "todo"},
-                "cachePriming": {"enable": True, "numThreads": 0},
+                # rustback lightweight patch: serena's symbol tools don't need
+                # rust-analyzer's eager whole-workspace+deps cache priming, nor a
+                # re-index on every cargo metadata / target/ change. On large,
+                # actively-built Rust workspaces these drove rust-analyzer to
+                # 40+ GB RSS and re-indexed on every external `cargo build`.
+                # Disabling them keeps symbol navigation working at a fraction of
+                # the memory. NOTE: checkOnSave is intentionally left enabled below
+                # because serena's get_diagnostics_for_file relies on flycheck
+                # (`cargo check`) for Rust diagnostics. (Upstream: expose these as
+                # a config knob / a "lightweight" RA profile.)
+                "cachePriming": {"enable": False, "numThreads": 0},
                 "cargo": {
-                    "autoreload": True,
+                    "autoreload": False,
                     "buildScripts": {
                         "enable": True,
                         "invocationLocation": "workspace",
--- a/src/solidlsp/language_servers/rust_analyzer.py
+++ b/src/solidlsp/language_servers/rust_analyzer.py
@@ -690,7 +690,7 @@ class RustAnalyzer(SolidLanguageServer):
         timeout = super()._get_published_diagnostics_wait_timeout(pull_diagnostics_failed)
         # Rust diagnostics are often published asynchronously after the pull-diagnostics request,
         # so keep a wider fallback wait window across all platforms.
-        return max(timeout, 4.0)
+        return max(timeout, 8.0)
 
     def _start_server(self) -> None:
         """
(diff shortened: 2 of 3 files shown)


opcode81:
The changes as such look good to me. However, as a non-Rust user, it is not clear to me whether the changes could have detrimental effects. Was the eager indexing truly superfluous or are there cases where the new configuration might cause the server to return incomplete or incorrect results?
If we expect no detrimental effects with a reasonable degree of certainty, then I will merge this.
@TyceHerrman @davidcforbes

TyceHerrman:
I re-reviewed rust-analyzer

 Disabling cache priming is safe afaik. rust-analyzer computes the same indexes on demand, with slower initial queries as the tradeoff.

Disabling Cargo autoreload can leave project metadata stale after manifest or workspace changes, so I’d say cargo.autoreload=true should be restored

opcode81:
> I’d say cargo.autoreload=true should be restored

OK, so undoing this change still keeps the main benefits while eliminating potential problems?

TyceHerrman:
yea i think most of the benefit comes from disabling cache priming

opcode81:
Thanks for finishing this, @TyceHerrman!

PR #1786, opened 2026-07-30
ci: capture Windows JVM hang diagnostics

TyceHerrman:
## Summary

- add a per-test watchdog for the Windows JVM pytest batch
- capture all Python threads, the pytest process tree, JDTLS JVM thread dumps, and active JDTLS workspace logs
- fall back from `jcmd Thread.print -l` to `jstack -l` when JVM attachment fails
- capture after 10 minutes and terminate after 12 minutes in normal runs
- repeat the cross-language sequence surrounding the failed Java startup on this diagnostic branch
- shorten branch-only capture/termination to 5/7 minutes, preserve each attempt log, and continue past fast assertion failures until a hang is captured or all repetitions finish
- run 20 cold JDTLS workspaces after a 50-warm-start run completed without a hang
- print any Python/JVM dump into the Actions job log before artifact upload

## Why

The Windows JVM job in [run 30496140622](https://github.com/oraios/serena/actions/runs/30496140622/job/90725317302?pr=1761) stalled in a Java test until the 60-minute job timeout. The existing log does not show whether Python was waiting for IntelliCode or `ServiceReady`, or whether JDTLS was blocked inside Maven/m2e initialization.

The first 50-repetition run did not hang. [Attempt 26 failed quickly](https://github.com/oraios/serena/actions/runs/30588744848/job/91026120690) because Java hover information omitted the expected Javadoc. Its artifact correctly contained only `pytest.log`; the watchdog had not fired.

A second 50-repetition run also completed without a hang. Eight attempts failed with the same missing-Javadoc result. A third run cleared only the isolated JDTLS workspace between attempts; all 20 repetitions passed without a hang or missing-Javadoc failure.

This draft PR remains diagnostic instrumentation. It is intended to capture the evidence needed to distinguish a Serena/SolidLSP wait or event-ordering problem from an upstream JDTLS/m2e deadlock.

## Validation

- `uv run pytest test/test_ci_hang_diagnostics.py -q`
- `uv run poe lint`
- `uv run poe type-check`
- forced-hang smoke test confirming the artifact completes before pytest-timeout exits the process
- workflow YAML parsed successfully; actionlint reported no diagnostics on the changed lines

PR #1788, opened 2026-07-31
ci: exclude JDTLS workspaces from language-server cache

TyceHerrman:
## Summary

- keep caching downloaded language-server installations
- exclude JDTLS project workspaces from the shared GitHub Actions cache
- force each CI job to initialize its own Eclipse/Buildship runtime state

## Why

Serena currently caches all of `~/.serena/language_servers/static`, including
`EclipseJDTLS/workspaces`. Those directories contain mutable, project-specific
Eclipse and Buildship state. Restoring that state into another job can make Java
test behavior depend on an earlier job rather than only on the checked-out
revision and current runner.

This is an independent cache-correctness change. It does not claim that restored
workspace state caused the Windows JDTLS hang; the diagnostics work in #1786 is
still needed to classify that stall.

The cache action supports multiple glob paths and leading `!` exclusions. Its
cache version also incorporates the configured paths, so the new definition
does not match the old broad cache archive.

#1214 intentionally preserves JDTLS workspaces across local Serena restarts. #1576 established stale JDTLS workspace state as a real correctness risk. This change preserves that user behavior while preventing GitHub Actions from restoring project-derived workspace state across revisions that share the same runner checkout path.

## Validation

- workflow YAML parses successfully
- `git diff --check`
- `actionlint` reports only diagnostics already present on `main`

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -706,7 +706,11 @@ jobs:
         id: cache-language-servers
         uses: actions/cache@v3
         with:
-          path: ~/.serena/language_servers/static
+          # JDTLS workspaces contain mutable Eclipse/Buildship runtime state.
+          # Cache the installed servers, but create each JDTLS workspace locally.
+          path: |
+            ~/.serena/language_servers/static
+            !~/.serena/language_servers/static/EclipseJDTLS/workspaces/**
           key: language-servers-${{ runner.os }}-${{ matrix.batch }}-v1
           restore-keys: |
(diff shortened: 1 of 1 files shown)


opcode81:
Is this ready?

TyceHerrman:
Not yet I need to review it further still

TyceHerrman:
@opcode81 this is ready for review now afaik

opcode81:
@TyceHerrman sorry, I had accidentally closed the wrong PR.

opcode81:
Thanks for your contribution.

PR #1791, opened 2026-07-31
Fix: macOS tray manager showed a Dock icon

caost:
## Problem

On macOS, running the dashboard with `web_dashboard_interface: tray_manager` puts a **Python icon in the Dock** (and in the ⌘Tab app switcher), even though the tray manager only owns a status item and has no windows.

## Cause

`SerenaDashboardTrayManager.run()` creates the shared `NSApplication` so that pystray can attach its status item to it:

```python
if sys.platform == "darwin":
    from AppKit import NSApplication

    kwargs["darwin_nsapplication"] = NSApplication.sharedApplication()
```

but never sets an activation policy. The default is `NSApplicationActivationPolicyRegular`, which marks the process as a regular foreground app — hence the Dock icon.

## Fix

Set `NSApplicationActivationPolicyAccessory`, the policy intended for menu-bar-only agents. This is the programmatic equivalent of `LSUIElement` in an app bundle's `Info.plist`, which isn't available here since the tray manager runs as a plain Python process rather than a bundle.

Only the status item remains visible; the tray menu and dashboard behaviour are unchanged.

## Testing

Verified on macOS 26.5.2 (Apple Silicon), Python 3.13:

- Dock icon and app switcher entry are gone; the menu bar icon still appears
- Tray menu lists the registered instances and opening a dashboard from it still works
- Reproduced the old behaviour by reverting the two lines

Also ran `poe format` and `poe type-check` — both clean.

Notably, the tray manager works on macOS in general — the config template currently marks `tray_manager` as tested on Windows only, and `DashboardManager.Mode.from_platform()` carries a `TODO: Switch to TRAY_MANAGER once support is tested` for Darwin. This Dock icon was the one rough edge I hit; happy to follow up on the default-mode question separately if that's useful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@ Status of the `main` branch. Changes prior to the next official version change w
   - Fix: Serena PyPI version check triggered by callback on main thread could delay agent startup #1774
   - Fix: on macOS, the `tray_manager` interface put an icon in the Dock and in the app switcher
     (should only use menu bar icon)
+  - Use `tray_manager` interface as new default on macOS
 
 * Hooks:
   - Add `serena-hooks --client=grok`, including Grok-native PreToolUse allow/deny output.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,8 @@ Status of the `main` branch. Changes prior to the next official version change w
 
 * Dashboard:
   - Fix: Serena PyPI version check triggered by callback on main thread could delay agent startup #1774
+  - Fix: on macOS, the `tray_manager` interface put an icon in the Dock and in the app switcher
+    (should only use menu bar icon)
 
 * Hooks:
   - Add `serena-hooks --client=grok`, including Grok-native PreToolUse allow/deny output.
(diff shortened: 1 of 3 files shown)


opcode81:
Thanks!

I think you may be the first person to have tried this on macOS. Would you say the tray manager would be a good default to use on macOS? So far, the default has been to use only the browser, like on Linux, but if the tray manager works well on macOS, it would be a better default.

CHANGELOG.md
```diff
@@ -21,6 +21,11 @@ Status of the `main` branch. Changes prior to the next official version change w
 * Tools:
   - `find_symbol`: Change tool description to improve tool search results in clients that load tools dynamically
 
+* Dashboard:
+  - Fix: on macOS, the `tray_manager` interface put a Python icon in the Dock and in the app switcher,
+    because the shared `NSApplication` was left at the default (regular) activation policy; it now uses
+    the accessory policy, so only the menu bar icon remains
```
opcode81: The dashboard section already exists below. Please move the entry there.

PR #1149, opened 2026-03-10
feat: add ELP (Erlang Language Platform) as default Erlang language server

zerone0x:
Closes #771

The [erlang_ls](https://github.com/erlang-ls/erlang_ls) project was archived on 2025-08-15. This PR switches the default Erlang language server to [ELP (Erlang Language Platform)](https://github.com/WhatsApp/erlang-language-platform) by WhatsApp, as recommended by the erlang_ls maintainers.

## Changes

- **New**: `ErlangLanguagePlatform` class in `elp_language_server.py` using `elp server` (stdio LSP transport)
- **`Language.ERLANG`** now points to ELP (the new default)
- **`Language.ERLANG_LS`** added as a deprecated/experimental fallback for users who still have `erlang_ls` installed — set `language: erlang_ls` in `project.yml` to use it
- **Docs** updated to reference ELP and link to the installation guide

## Usage

ELP must be installed and `elp` must be in `PATH`. Installation: https://whatsapp.github.io/erlang-language-platform/docs/get-started/

Users who still need the legacy `erlang_ls` can opt in via `project.yml`:
```yaml
language: erlang_ls
```

ELP is actively maintained, supports go-to-definition, find references, call hierarchy and more. It should also resolve the CI hangs that made erlang_ls testing unreliable.

---
🤖 Generated with Claude Code

Issue #771, opened 2025-11-18
Erlang LS has been archived on 2025-08-15, recommends switching to Erlang Language Platform


andreswebs:
The [erlang_ls](https://github.com/erlang-ls/erlang_ls) project used by Serena has been archived on 2025-08-15. They recommend switching to [Erlang Language Platform](https://github.com/whatsapp/erlang-language-platform).

I have:

- [x] Read the readme and verified that the issue cannot be solved by adjusting configuration
- [x] Understood that Serena's dashboard can be disabled through the config
- [x] Understood that by default a client session will start a separate instance of a Serena server. 
- [x] Understood that for multi-agent setups, the SSE mode should be used.
- [x] Verified that non-project files are ignored using either gitignore or the corresponding setting in `.serena/project.yml`
- [x] Have looked for similar issues and discussions, including closed ones
- [x] Made sure it's an actual issue and not a question - those should be opened as discussion instead.

If you have encountered a real and new issue:

- [x] I performed `<uv invocation> serena project health-check`
- [x] I indexed the project as described in the readme
- [x] Added sufficient explanation of my setup: the MCP client, the OS, the programming language, any config adjustments or relevant project specifics
- [x] Explained how the issue arose, added instructions on how to reproduce it (if possible)
- [x] If the issue happens on an open source project, I have added the link
- [x] Wrote a meaningful title and description

MischaPanch:
Yes, it's nice that an alternative exists. We could also never get erlang_ls to run in the CI. We would be happy to accept a PR that switches to the language server provided by ELP.

andreswebs:
Awesome, I'll be happy to contribute. Will take a couple of weeks to get started, but you can assign this to me @MischaPanch

turboslapper:
@andreswebs Hey! Just wondering if you're still working on this? If not, I'd love to take a go.

PR #1763, opened 2026-07-26
Docs: add the language servers missing from the supported-language lists

xormania:
Several backends that are registered and tested were absent from the user-facing lists: six from the README (pascal, qml, rego, systemverilog, terraform, vue), six from the Language Support docs page (matlab, powershell, rego, systemverilog, terraform, toml), and two from the commented language-server list in the project template (python_basedpyright, qml).

The drift is long-standing and runs in both directions: matlab, powershell and toml reached the README but never the docs page; pascal, vue and qml reached the docs page but never the README; terraform, rego and systemverilog reached neither, the oldest of them since 2025-07-06. The two template omissions are the most recent, from #1705 and #1635. The guide's Documentation step does ask for the README and the docs page, so most of this was missed rather than unspecified; it does not mention the project template at all.

The setup notes on the new docs-page entries name the versions Serena actually pins, taken from the corresponding language-server modules: MATLAB extension 1.3.9, PowerShell Editor Services 4.4.0 and PSScriptAnalyzer 1.25.0, Verible v0.0-4051-g9fdb4057, terraform-ls 0.36.5, and Taplo 0.10.0. The PowerShell entry mentions both automatic installations, since enabling that backend also runs Save-Module against the user's configured PowerShell repository.

The new Regal link points at open-policy-agent/regal, the current canonical repository; StyraInc/regal still redirects there, and the adapter and the CI workflow have not been changed.

The template block is refreshed from scripts/print_language_list.py, which is why its column widths change: python_basedpyright is longer than any identifier the previous list contained. The pointer directly below it still named the Language enum, renamed to LanguageServerId in d2cf18dd.

## Checklist

- [x] This PR follows the guidelines in `CONTRIBUTING.md` regarding the scope of PRs.
- [x] For changes that add features or fix problems, I have added an entry to `CHANGELOG.md`, which concisely describes the change.

PR #1794, opened 2026-08-01
fix: make ErlangLanguageServer instantiable (implement _create_base_initialize_params)

ibakard:
## Problem

`ErlangLanguageServer` cannot be instantiated at all. Every Serena tool fails on an Erlang project during project initialisation:

```
The language server manager is not initialized, indicating a problem during project initialisation.
Failed to start 1 language server(s):
erlang: Can't instantiate abstract class ErlangLanguageServer without an implementation
for abstract method '_create_base_initialize_params'
```

Because the failure is at initialisation, it takes down every tool, not just the one called — `get_symbols_overview`, `find_symbol`, `find_referencing_symbols` and the editing tools all fail identically.

## Cause

When server-specific initialize params were moved behind the abstract `_create_base_initialize_params` hook on `SolidLanguageServer` (`ls.py`), the Erlang backend was not converted with the others. It still builds its initialize dict inline inside `_start_server` and passes it straight to `self.server.send.initialize(...)`, so it never implements the abstract method and the class is left abstract.

This is not environment-specific: it reproduces with `erlang_ls` correctly installed and on `PATH`, on current `main`.

## Fix

Convert the backend to the same pattern the other language servers use (compare `ocaml_lsp_server.py`):

- Move the `capabilities` block out of `_start_server` into `_create_base_initialize_params`.
- Drop `processId` / `rootPath` / `rootUri` from the literal — `DefaultInitializeParamsBuilder` supplies those, and the method's docstring explicitly says implementations should not set them.
- Have `_start_server` build its params via `self._create_initialize_params()`.
- Remove the now-unused `pathlib` import.

No behavioural change beyond making the class instantiable; the capabilities advertised to `erlang_ls` are the same ones the inline dict sent. The `# type: ignore[arg-type]` on the `initialize` call is no longer needed, since `_create_initialize_params()` returns a properly typed `InitializeParams`.

## Verification

Against a real rebar3 umbrella project (~12 OTP applications), with `erlang_ls` on `PATH`:

- `ErlangLanguageServer.__abstractmethods__` is now empty — the class instantiates.
- `get_symbols_overview` on a 700-line module returns its functions, records and macros.
- `find_referencing_symbols` on a record returns 26 cross-references with correct locations and snippets.

## Note on erlang_ls vs ELP

`erlang_ls` was archived on 2025-08-15, and #771 tracks moving to ELP; #1149 attempted that and was closed over a branch/CI problem rather than on the merits. This PR takes no position on that migration — it only makes the Erlang backend that currently ships actually usable. If ELP support lands later, this file is likely replaced wholesale, but until then the shipped backend is 100% non-functional and this is a small fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -105,6 +105,7 @@ Status of the `main` branch. Changes prior to the next official version change w
     coverage-report output, but matched by bare dirname and so also hid legitimate source directories named
     `coverage` (e.g. `src/routes/coverage/`) from symbol tools. Generated report dirs are already covered by
     gitignore. Fixes #1523.
+  - Fix: Restore `erlang` support (broken since v1.6.0 due to an implementation error) 
 
 * Tools:
   - Fix: `search_for_pattern` marked one line too many as matched whenever a match ended with a line
--- a/src/solidlsp/language_servers/erlang_language_server.py
+++ b/src/solidlsp/language_servers/erlang_language_server.py
@@ -2,7 +2,6 @@
 
 import logging
 import os
-import pathlib
 import shutil
 import subprocess
 import threading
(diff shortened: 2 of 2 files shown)


AmirF194 (commented):
CONTRIBUTING.md asks for a CHANGELOG.md entry on every PR (under "Submitting Pull Requests"), and every recently merged PR here has one (#1779, #1763, #1759, #1758, #1754). This one doesn't yet; probably belongs under "Language Servers".

The fix itself checks out. `_create_base_initialize_params` matches the pattern `ocaml_lsp_server.py` and the other converted backends use, and `SolidLanguageServer._create_initialize_params()` (ls.py:3193) already supplies processId/rootUri/clientInfo/workspaceFolders through the builder, so dropping them from the literal here is correct rather than a behavior change.

opcode81:
Thanks for your contribution. It's a shame that `ty` did not detect this. IDEA does:
image

PR #1800, opened 2026-08-01
fix(erlang): address functions by `name/arity` without the name path separator

### LINKED ISSUE #1797: Symbol names containing `/` (Erlang `foo/1`) can't be addressed — collides with the name_path separator
## Summary

Symbols whose *name* contains a `/` cannot be addressed by any tool that takes an exact `name_path`, because `/` is the name-path hierarchy separator. Erlang is the clear case: `erlang_ls` names functions `name/arity`, so `prn_to_sv_id/1` is parsed as "symbol `1` nested inside `prn_to_sv_id`" and never matches — even though that is the exact `name_path` Serena itself just printed.

The result is that `find_referencing_symbols`, `replace_symbol_body` and `insert_after_symbol` are unusable on Erlang *functions*. Read-only browsing still works, so the language looks supported until you try to do anything with a function.

## Reproduction

Any Erlang module. Given a function `prn_to_sv_id/1` in `refsrv.erl`:

```
find_symbol(name_path_pattern="prn_to_sv_id", substring_matching=true)
  -> [{"name_path": "prn_to_sv_id/1", "kind": "Function", ...}]      # works

find_symbol(name_path_pattern="prn_to_sv_id/1")
  -> []                                                              # exact name_path it just returned

find_referencing_symbols(name_path="prn_to_sv_id/1")
  -> ValueError: No symbol matching 'prn_to_sv_id/1' found

find_referencing_symbols(name_path="prn_to_sv_id")
  -> ValueError: No symbol matching 'prn_to_sv_id' found
```

For contrast, a symbol with no `/` in its name behaves correctly — `find_referencing_symbols(name_path="state")` on a record returned 26 references with correct locations and snippets. So the language server and the reference machinery are both fine; only the name addressing is broken.

`replace_symbol_body` and `insert_after_symbol` take the same exact `name_path` with no substring option, so they are expected to fail identically. I did not confirm those two by running a destructive edit against real code — flagging that explicitly rather than overstating it.

## Scope

Not Erlang-specific in principle — it affects any language whose symbol names can contain `/`. Erlang is just where it is unavoidable, since `name/arity` is the standard way to identify a function and is what the language server reports.

## Note on reproducing this

The Erlang backend on current `main` cannot start at all (`ErlangLanguageServer` does not implement the abstract `_create_base_initialize_params`), so this limitation is currently masked by an earlier failure. #1794 fixes that; the behaviour above was observed with that fix applied, against a rebar3 umbrella project with `erlang_ls` on `PATH`.

## Possible directions

Deliberately not prescribing an implementation, but the options seem to be:

- An escape for `/` inside a name segment.
- An explicit "treat this string as a single symbol name, not a path" flag on the tools that take `name_path`.
- Accepting the language server's flat name as-is when the pattern matches a symbol exactly, before falling back to splitting on `/`.

Whatever the fix, it would help if `find_symbol` never returned a `name_path` that cannot be fed back into the other tools — that round-trip failure is what makes this confusing in practice.

ibakard:
Fixes #1797.

## The problem

Erlang LS identifies functions, types and parameterised macros as `name/arity` (e.g. `create_user/4`), but `/` separates the components of a Serena name path. `create_user/4` was therefore parsed as "symbol `4` nested inside `create_user`" and could never match — not even the name path Serena itself reported for the symbol. Read-only browsing kept working, so Erlang looked supported until you tried to do anything with a function:

```
find_symbol("create_user", substring_matching=true)  -> [{"name_path": "create_user/4", ...}]   # works
find_symbol("create_user/4")                         -> []                                      # its own output
find_referencing_symbols("create_user/4")            -> ValueError: No symbol matching ...
```

## The fix

As suggested by @opcode81 in the issue: `ErlangLanguageServer` now overrides `_normalize_symbol_name` to substitute the `/`, and — per that method's docstring — also overrides `_document_symbols_cache_fingerprint` so that document-symbol caches written before this change are invalidated rather than served with the old names.

Two choices worth spelling out:

* **`#` rather than `@`.** `#` cannot occur in an unquoted Erlang atom, so it can never collide with a real function, type or macro name. `@` is a legal atom character (`foo@bar` is a valid function name), so it would be ambiguous.
* **The arity is kept, not stripped.** Unlike the parameter lists that Elixir/Dart strip, arity is part of a function's identity in Erlang: `create_order/3` and `create_order/2` are different functions and may coexist in one module. Dropping it would merge distinct functions into overloads of one name.

Symbols that have no arity — records, macros — are unaffected, since their names contain no `/`.

I also added the "must not contain `/`" rule to the `_normalize_symbol_name` docstring in `ls.py`. It is currently tribal knowledge, and it is what the next backend for a language with `name/arity`-style identifiers needs to know.

## Verification

Run against real `erlang_ls` (OTP 28) on `test/resources/repos/erlang/test_repo`:

```
src/models.erl    -> create_user#4, create_order#3, update_user#2, validate_email#1, ...
src/services.erl  -> create_order#2, handle_call#3, start_link#0, state (record, unchanged)
include/records.hrl -> user, order, item, config, RECORDS_HRL   (all unchanged)

find("create_user#4")                        -> [("create_user#4", "src/models.erl")]
find("create_order#3")                       -> [("create_order#3", "src/models.erl")]
find("create_order#2")                       -> [("create_order#2", "src/services.erl")]
find_referencing_symbols("create_user#4")    -> 5 references across 3 files
replace_symbol_body("validate_email#1", ...) -> applied
insert_after_symbol("create_user#4", ...)    -> applied, in the right place
```

(The issue listed `replace_symbol_body`/`insert_after_symbol` as *expected* to fail, since I had not run a destructive edit against real code at the time. I have now run both against a throwaway copy of the Erlang test repo; they work.)

New tests in `test/solidlsp/erlang/test_erlang_symbol_names.py` cover: no symbol name contains the separator; a function's reported name path round-trips; arity distinguishes same-named functions in different modules; `find_referencing_symbols` locates a function; and names without an arity are left alone. With the normalization removed, 4 of the 5 fail — the round-trip one with exactly the error from the issue (`ValueError: No symbol matching 'create_user#4' found`). The fifth is the guard against over-normalizing and passes either way, by design.

`ruff check`, `ruff format --check` and `ty check` are clean.

Full local Erlang suite (`erlang_ls` on PATH, `--timeout=150` since the suite otherwise hangs):

| module | result |
| --- | --- |
| `test_erlang_basic.py` | 3 passed, 1 failed — `test_file_diagnostics` |
| `test_erlang_ignored_dirs.py` | 6 passed, 3 xpassed |
| `test_erlang_symbol_retrieval.py` | 15 passed, 2 xpassed |
| `test_erlang_symbol_names.py` (new) | 5 passed |

`test_file_diagnostics` is the one the CI comment means by "always hangs on ubuntu": it runs until the timeout kills it. I checked it against unmodified `main` on the same machine and it fails there in the same way and at the same duration (151s), so it is pre-existing and untouched by this change. Everything else passes, including `test_bare_symbol_names` — `#` is not among the characters `has_malformed_name` rejects.

CHANGELOG and the language-support docs page are updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,11 @@ Status of the `main` branch. Changes prior to the next official version change w
   - Fix: F#'s `module <Name>` declarations reported a `selectionRange` pointing at the `module`
     keyword instead of at `<Name>`, so looking up hover/references from a module symbol's position
     returned the keyword's own docs instead of the module's #925
+  - Fix: Erlang functions could not be addressed by any tool taking an exact name path, because
+    Erlang LS identifies them as `name/arity` and `/` separates name path components. The arity is
+    now separated by `#` instead (e.g. `create_user#4`), so the reported name path round-trips and
+    `find_referencing_symbols`/`replace_symbol_body`/`insert_after_symbol` work on Erlang
+    functions #1797
 
 * JetBrains:
   - `jet_brains_find_symbol`: Disallow wildcard-only search, delegating to overview tool if request is for file
--- a/docs/01-about/020_programming-languages.md
+++ b/docs/01-about/020_programming-languages.md
@@ -68,7 +68,8 @@ Some languages require additional installations or setup steps, as noted.
 * **Elm**  
   (requires Elm compiler)
 * **Erlang**  
-  (requires installation of beam and [erlang_ls](https://github.com/erlang-ls/erlang_ls); experimental, might be slow or hang)
+  (requires installation of beam and [erlang_ls](https://github.com/erlang-ls/erlang_ls); experimental, might be slow or hang;
+  note that functions are addressed as `name#arity`, e.g. `create_user#4`, because `/` is reserved as the name path separator)
 * **F#**  
   (requires [.NET v8.0+](https://dotnet.microsoft.com/en-us/download/dotnet); uses FsAutoComplete/Ionide, which is auto-installed; for Homebrew .NET on macOS, set DOTNET_ROOT in your environment)
 * **Fortran**   
--- a/src/solidlsp/language_servers/erlang_language_server.py
+++ b/src/solidlsp/language_servers/erlang_language_server.py
@@ -6,16 +6,25 @@ import shutil
 import subprocess
 import threading
 import time
+from collections.abc import Hashable
 
 from overrides import override
 
-from solidlsp.ls import SolidLanguageServer
+from solidlsp.ls import RawDocumentSymbol, SolidLanguageServer
 from solidlsp.ls_config import LanguageServerConfig
 from solidlsp.lsp_protocol_handler.server import ProcessLaunchInfo
 from solidlsp.settings import SolidLSPSettings
 from solidlsp.util.subprocess_util import subprocess_run
 
 log = logging.getLogger(__name__)
+
+ARITY_SEPARATOR = "#"
+"""
+The character that replaces the `/` in the `name/arity` identifiers reported by Erlang LS.
+
+`#` was chosen because it cannot occur in an unquoted Erlang atom, so it can never collide with a
+real function, type or macro name (unlike `@`, which is a legal atom character).
+"""
 
 
 class ErlangLanguageServer(SolidLanguageServer):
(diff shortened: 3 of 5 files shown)


ibakard:
The failing check (`niche (ubuntu-latest)`) does not come from this PR — it is already failing on `main` at exactly the commit this branch is based on (`c7dbe3bd`, this PR's `baseRefOid`):

https://github.com/oraios/serena/actions/runs/30703366827 → job `niche (ubuntu-latest)`

Same test, same exception, same counts:

| | `main` @ `c7dbe3bd` | this PR |
| --- | --- | --- |
| failing test | `test_nix_basic.py::TestNixLanguageServer::test_file_diagnostics[nix]` | same |
| cause | `LanguageServerTerminatedException` on `textDocument/diagnostic` | same |
| result | 1 failed, 53 passed, 1 skipped, **1974** deselected | 1 failed, 53 passed, 1 skipped, **1979** deselected |

The deselected counts differ by exactly 5 — the five Erlang tests added here, deselected out of that batch (`erlang` is in `MARKERS_OTHER_LANGS`, and `other-langs` passes on all three OSes). The only file this PR shares with the nix path is `ls.py`, where the change is six lines inside a docstring.

For what it's worth, the window on `main` is narrow: `niche` was last green at `c74373e2`, and of the commits since, `f4ecb1a5` ("re-raise LS termination from pull-diagnostics fallback", #1770) is the only one touching that path — it makes `request_text_document_diagnostics` propagate `LanguageServerTerminatedException` instead of falling back to published diagnostics, so a `nixd` crash during the diagnostics pull now fails the test where it was previously swallowed. Two runs, identical failure, so it looks deterministic rather than flaky. Happy to open a separate issue if that would help.

One caveat on this PR's *green* checks, in the interest of not overselling them: `test/solidlsp/erlang/` is SKIPPED in CI (the `erlang_ls` install step in `pytest.yml` is commented out), so CI does not exercise this change at all. The local run against real `erlang_ls` in the PR description is the only verification there is.

opcode81:
Thanks for your contribution!

PR #1803, opened 2026-08-02
fix(ls): resync open buffers and sync find_symbol/get_symbols_overview

### LINKED ISSUE #1593: find_symbol can return stale/incorrect information (observed for Clojure/LSP)
```
INFO  2026-06-22 15:51:24,654 [Task-291:FindSymbolTool] serena.tools.tools_base:_log_tool_application:266 - find_symbol: name_path_pattern='make-nested', depth=0, relative_path='common/src/app/common/test_model/comp/nodes.cljc', include_body=True, include_info=False, include_kinds=[], exclude_kinds=[], substring_matching=False, max_matches=-1, max_answer_chars=-1, session_id='73791df82be0'; session_id: 73791df82be0
INFO  2026-06-22 15:51:24,666 [Task-291:FindSymbolTool] serena.tools.tools_base:task:383 - Result: [{"name_path": "make-nested", "kind": "Function", "relative_path": "common/src/app/common/test_model/comp/nodes.cljc", "body_location": {"start_line": 206, "end_line": 215}, "body": ";;        next `make-nested`) will instantiate. Advances to the new OUTER component\n;;        with each `make-nested`.\n;;     :nesting-count — how many `make-nested` layers were applied.\n;;\n;; \"corresponds to\" between layers is resolved by following the :shape-ref chain\n;; (a copy shape refs the main shape it mirrors); within these single-file setups\n;; the chain resolves in the local page objects, so a local walk suffices.\n;; ===========================================================================\n\n(defn- ref-chain-ids"}]
INFO  2026-06-22 15:51:24,666 [Task-291:FindSymbolTool] solidlsp.ls:_save_raw_document_symbols_cache:2900 - Saving updated raw document symbols cache to /home/penpot/penpot/.serena/cache/clojure/raw_document_symbols.pkl
INFO  2026-06-22 15:51:24,694 [Task-291:FindSymbolTool] solidlsp.ls:_save_document_symbols_cache:2970 - Saving updated document symbols cache to /home/penpot/penpot/.serena/cache/clojure/document_symbols.pkl
```

Observe that the body being returned for `make-nested` is not a body at all - but, suspiciously, a comment mentioning the function.
The true location of the function is not line 206 but line 454.

Interestingly, the language server must have been explicitly invoked to process the file in question, as the two subsequent log messages clearly indicate that the symbol cache was updated.
So this might be a language server bug - or a bug in Serena where occurrences in comments are treated as symbol definitions.

AmirF194:
## Root cause

`find_symbol` and `get_symbols_overview` carried a comment claiming that a file system sync is
not required because the symbol finder "opens all relevant source files explicitly in the case
of changes". That is true when a query opens and closes its own buffer, since a fresh
`didOpen` always carries the current disk content. It stops being true once a file's buffer is
already open (`ref_count > 0`) at query time: `SolidLanguageServer._open_in_ls` guards against
re-sending `didOpen` for a buffer that is already open, and neither tool called
`project.ls_sync_file_system_changes()`, unlike every other symbolic tool
(`find_referencing_symbols`, `find_implementations`, `find_declaration`,
`get_diagnostics_for_file`/`_symbol`, `rename_symbol`). A file already open when edited outside
of Serena's own edit tools (another editor, a second agent, a git checkout) then keeps
answering from the language server's stale view, exactly the "make-nested reported at the old
position" symptom in the issue.

Wiring the sync call in exposed a second gap in the sync mechanism itself
(`LanguageServerFileChangeNotifier.poll_and_notify`, added in #1591 20 days ago): it only sends
`workspace/didChangeWatchedFiles`. That notification is not guaranteed to refresh a document
the server already has open. A server is free to treat the client, not the filesystem, as
authoritative for an open document's content once `didOpen` has been sent, and pyright does
exactly that in a Docker repro: a `didChangeWatchedFiles(Changed)` for an open file is a no-op,
the next `documentSymbol` request still returns pre-edit positions.

## Fix

- `find_symbol`, `get_symbols_overview`: call `project.ls_sync_file_system_changes()`, matching
  the other symbolic tools.
- `SolidLanguageServer.resync_open_buffer()` (new): re-reads a file's current content and, if a
  buffer is open for it, pushes a full-content `textDocument/didChange`.
- `poll_and_notify()`: calls `resync_open_buffer()` for every changed path, in addition to the
  existing `didChangeWatchedFiles` notification.

## Verification

- New regression test (`test_find_symbol_tool_reflects_external_change_to_open_buffer`, pyright
  backend, Docker): holds a file's buffer open, edits it externally, and asserts `find_symbol`
  reports the shifted position. Fails on `main` (stale position), passes with this change.
- `test/serena/` python-marked suite (61 tests, incl. all of `test_ls_file_sync.py`,
  `test_serena_agent.py`, `test_symbol.py`, `test_symbol_editing.py`): green, no regressions.
- `ruff check`/`ruff format --check` and `ty check` on the full `src`/`test` tree: clean.
- Not checked: the effect on any language server other than pyright (no toolchain for
  clojure-lsp, the server named in the original report, in this environment), and the
  performance cost of running the sync's directory walk on every `find_symbol` call for very
  large repositories.

Fixes #1593

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,11 @@ Status of the `main` branch. Changes prior to the next official version change w
 * Tools:
   - `find_symbol`, `jet_brains_find_symbol`: Change tool description to improve tool search results in clients that load tools dynamically
   - `get_current_config`: Result now includes language server status #1782
+  - Fix: `find_symbol` and `get_symbols_overview` could report a stale symbol position or miss a
+    new/deleted file when an already-open file was edited outside of Serena's own edit tools; both
+    tools are now included in the file system sync check every other symbolic tool already runs.
+    That check itself did not refresh an already-open buffer, only the file's watch registration,
+    so an open document could stay stale even where the check did run #1593
 
 * Language Servers: 
   - Allow language server priorities to be configured in `serena_config.yml` (for auto-detection during 
--- a/src/serena/ls_manager.py
+++ b/src/serena/ls_manager.py
@@ -347,6 +347,7 @@ class LanguageServerFileChangeNotifier:
         ]
         params: DidChangeWatchedFilesParams = {"changes": changes}
         created_paths = [rel_path for rel_path, change_type in events if change_type == FileChangeType.Created]
+        changed_paths = [rel_path for rel_path, change_type in events if change_type == FileChangeType.Changed]
 
         for ls in self._language_server_manager.iter_language_servers():
             # send the didChangeWatchedFiles notification to the language server
@@ -367,5 +368,13 @@ class LanguageServerFileChangeNotifier:
                         pass
                 except Exception as e:
                     log.error(f"Failed to refresh newly created file {rel_path!r} in language server", exc_info=e)
+
+            # A server may ignore didChangeWatchedFiles for a document it already has open
+            # (observed with pyright), so push fresh content directly to any open buffer too.
+            for rel_path in changed_paths:
+                try:
+                    ls.resync_open_buffer(rel_path)
+                except Exception as e:
+                    log.error(f"Failed to resync open buffer for changed file {rel_path!r}", exc_info=e)
 
         return len(events)
--- a/src/serena/tools/symbol_tools.py
+++ b/src/serena/tools/symbol_tools.py
@@ -54,7 +54,7 @@ class GetSymbolsOverviewTool(Tool, ToolMarkerSymbolicRead):
             Don't adjust unless there is really no other way to get the content required for the task.
         :return: a JSON object containing symbols grouped by kind in a compact format.
         """
-        # Note: file system sync not required (relevant file is opened in the language server explicitly)
+        self.project.ls_sync_file_system_changes()
 
         if depth == -1:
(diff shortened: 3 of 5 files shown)


opcode81:
Thanks, but this is not the right fix for this.

It was a principle in Serena to never leave files open (`open_file` is a context manager and is the only intended way to open files). Since some language servers (Vue, Svelte) hackily subverted this principle, it was clear that we needed to ultimately account for this in the handling of open files, and I had already had this on my TODO list, but since you brought up the topic, we might as well address it now.

The key observation is that the intention of `open_file` is actually to ensure that the language server is made aware of the (current) contents of the file. Therefore, if a file is already open but has changed on disk since being opened, the function must send the update notification instead of doing nothing.

This PR is superseded by #1804, where I have made the respective adjustments (and solved another cache staleness issue).
I kept your test, so thanks for providing it as a basis for TDD.

AmirF194:
Appreciate the writeup, and glad #1804 landed with the more complete fix. Thanks for keeping the test as the TDD basis.

PR #1809, opened 2026-08-03
fix(gitignore): escape directory names before interpolating into pattern position

### LINKED ISSUE #1806: Directory names are interpolated unescaped into gitignore patterns, silently indexing 0 files
## Summary

`GitignoreParser` rescopes patterns from nested `.gitignore` files by prefixing the containing directory's relative path, but does **not** escape glob metacharacters in that path. A directory whose *name* contains `*` therefore produces a pattern that matches far more than it should — in the worst case silently excluding the entire project, so indexing reports 0 files with exit code 0 and no error.

## Root cause

`serena/util/file_system.py`, `_parse_gitignore_content` (v1.6.1, line 272):

```python
adjusted_pattern = os.path.join(rel_dir, "**", line)
```

`rel_dir` is a filesystem name, but it is being interpolated into a *pattern* context. Lines 263 and 269 have the same issue.

## Concrete case

A stray uv/venv directory named `***` at the project root. Every venv ships a `.gitignore` whose entire content is `*`, so:

- `rel_dir` = `***`, `line` = `*`
- adjusted pattern = `***/**/*`

In gitignore/pathspec syntax `***` is just asterisks — a wildcard matching any first segment. So `***/**/*` means **"ignore every path at depth >= 2."** All sources under `src/`, `packages/`, etc. are excluded. Root-level files like `pyproject.toml` survive, which is why nothing errors.

Observed: `Indexing: 0it [00:00, ?it/s]` / `Indexed files per language:` (empty), exit 0. With language auto-detection it instead reports `No source files for supported language servers were found`, which points the user at language configuration — the wrong place entirely.

Verified by running `GitignoreParser` directly against the project: it loads a spec from `***/.gitignore` with adjusted pattern `***/**/*`, and `should_ignore()` returns True for **1009 of 1009** tracked `.py` files.

## Reproduction

```bash
mkdir -p repro/pkg && cd repro && git init -q
printf 'def f():\n    pass\n' > pkg/mod.py
mkdir '***' && printf '*\n' > '***/.gitignore'    # what any venv looks like
uvx --from serena-agent serena project index --language python .
# => Indexed files per language:   (empty)   — expected: python=1
```

Removing the `***` directory, or setting `ignore_all_files_in_gitignore: false`, indexes normally.

## Impact

Silent and misdiagnosing. Serena appears configured and healthy, `project.yml` looks correct, exit code is 0 — but every symbol tool (`find_symbol`, `find_referencing_symbols`) returns nothing because the index is empty. In our case this went unnoticed for some time and read as a language-server problem. Any directory named with glob metacharacters triggers it; stray venvs named with asterisks are the realistic path, but `foo[1]`, `a{b}` and similar would also misbehave.

## Suggested fix

Escape glob metacharacters in `rel_dir` before joining, so the directory name is matched literally. `pathspec` honors backslash-escaped metacharacters — `\*\*\*/` matches only the literal `***` directory (confirmed as a workaround). Something equivalent to escaping `*?[]{}` in `rel_dir` at all three join sites.

A cheaper mitigation, worth having regardless: warn when a loaded gitignore spec excludes 100% of candidate files, rather than reporting an empty index as success.

## Workaround

`ignore_all_files_in_gitignore: false` plus an explicit `ignored_paths` list. Note that `ls_workspace_folders` does **not** bypass it — the rescoped pattern is matched against project-root-relative paths.

## Environment

serena-agent 1.6.1, installed via `uvx`, macOS (arm64), Python 3.12, LSP backend.

AmirF194:
## Root cause

`GitignoreParser._parse_gitignore_content` rescopes patterns from a nested `.gitignore` by
joining the gitignore directory's relative path (`rel_dir`) onto each pattern line (three
join sites, `src/serena/util/file_system.py:263,269,272` at HEAD `29d07d4f`). `rel_dir` comes
from `os.path.relpath` and is a filesystem name, but it lands directly in pattern position,
where `pathspec`'s `GitWildMatchPattern` reads `*`, `?`, `[`, `]`, `!` and `#` as syntax rather
than literal characters.

A directory whose name contains one of those characters (the concrete case in the issue: a
stray venv literally named `***`, whose own `.gitignore` is just `*`) turns a scoped pattern
into a much broader one: `***/**/*` reads as "any first segment, then anything at depth >= 2",
so it ends up excluding nearly the whole project. Indexing then reports 0 files with exit code
0, and nothing points at the actual cause.

## Invariant

A pattern segment built from a filesystem name must match that name literally, never as glob
syntax. The fix escapes each path component of `rel_dir` (backslash before `\ * ? [ ] ! #`,
following `pathspec`'s own escaping convention) at all three join sites before it is folded
into the pattern string, so directory names are always matched literally regardless of their
content.

## Verification

- Added three regression tests in `test/serena/util/test_file_system.py`, one per join site
  (anchored, `**/`-prefixed non-anchored, and the implicit-`**` non-anchored case), plus the
  issue's own `***` scenario. Ran in a clean `python:3.11-slim` container against HEAD
  `29d07d4f`: all three fail on unmodified `main` and pass on this branch.
- `uv run poe test test/serena/util/test_file_system.py`: 27/27 pass (24 existing + 3 new).
- `uv run poe lint` and `uv run poe type-check`: clean. `codespell` on the changed files: clean.
- Not run: the full CI matrix (language-server batches across ubuntu/windows/macos) and
  `test/serena/util/test_exception.py`'s three pre-existing GUI-dependent failures, which are
  unrelated to this change and reproduce identically on unmodified `main` in this container.

Fixes #1806

## Checklist

## Update (second commit)

The initial Windows CI run (before any review) failed 3 new tests, not the 3 unrelated
pre-existing flakes documented above: `test_gitignore_dir_name_with_metachars_anchored_pattern`
failed with a real `AssertionError` (not the platform-limitation `OSError` the other two
metachar tests hit for `*`/`?`, which are illegal in Windows filenames). Root cause: the
escaping added above inserts literal backslash escape characters, and on Windows `os.sep` is
also backslash, so `os.path.join(rel_dir_pattern, line)` mixed path-separator backslashes with
escape backslashes; the pre-existing trailing `.replace(os.sep, "/")` then converted all of
them indiscriminately, turning an escaped `a\\[1\\]` into `a/[1/]`, which no longer matches
the literal directory name.

Fixed by building every pattern with a literal `/` from the start (never `os.sep`/
`os.path.join`), and removing the now-unneeded (and on Windows actively harmful) blanket
separator replace. Verified with `ntpath` (Windows path semantics) on this Linux session:
simulating the old chain on a Windows-style relpath produces `a/[1/]/mod.py` (does not match
`a[1]/mod.py` via `pathspec`); the new construction produces `a\\[1\\]/mod.py` (matches).
The two tests that create a directory literally named `***`/`q?` are marked
`skipif(sys.platform == "win32")` (those characters can't exist in a Windows filename, a test
limitation, not a fix limitation); added a pure-function test for
`_escape_gitignore_path_component` so that escaping logic keeps cross-platform coverage without
touching the filesystem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ Status of the `main` branch. Changes prior to the next official version change w
 
 * General:
   - Fix: Race conditions in ProjectServer when used by multiple clients in parallel   
+  - Fix: `GitignoreParser` interpolated a directory's name unescaped into gitignore pattern position;
+    a directory named with pattern metacharacters (e.g. a stray `***`) could turn a scoped pattern
+    into one matching far more than intended, silently excluding most or all of the project from
+    indexing #1806
+  - Fix: the README, the Language Support docs page and the project template omitted several already-supported language servers
   - Fix: a tool call exceeding the timeout blocked the task executor indefinitely; the executor now
     recovers without user-induced cancellation
   - Add Grok Build support (context `grok`, setup CLI, hooks)
--- a/src/serena/util/file_system.py
+++ b/src/serena/util/file_system.py
@@ -236,15 +236,20 @@ class GitignoreParser:
         """
         patterns = []
 
-        # Get the relative path from repo root to the gitignore directory
-        rel_dir = os.path.relpath(gitignore_dir, self.repo_root)
+        # Get the relative path from repo root to the gitignore directory. Normalize to
+        # forward slashes immediately: os.path.relpath returns native separators, but
+        # gitignore/pathspec patterns are always POSIX-style, and on Windows os.sep is
+        # backslash -- the same character pathspec uses as its escape character. Building
+        # the pattern with any raw os.sep would make a later blanket backslash->slash
+        # normalization indistinguishable from the escape backslashes below.
+        rel_dir = os.path.relpath(gitignore_dir, self.repo_root).replace(os.sep, "/")
         if rel_dir == ".":
             rel_dir = ""
 
         # rel_dir is a filesystem path, but the code below interpolates it into pattern
         # position; escape each of its components so a directory name containing pattern
         # metacharacters (e.g. "***") is matched literally instead of as glob syntax.
-        rel_dir_pattern = os.sep.join(_escape_gitignore_path_component(part) for part in rel_dir.split(os.sep)) if rel_dir else rel_dir
+        rel_dir_pattern = "/".join(_escape_gitignore_path_component(part) for part in rel_dir.split("/")) if rel_dir else rel_dir
 
         for line in content.splitlines():
             # Strip trailing whitespace (but preserve leading whitespace for now)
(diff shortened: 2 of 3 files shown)


MischaPanch:
@opcode81 from my side we can merge this, even though the error it fixes is largely theoretical

AmirF194:
Rebased onto main to clear the conflict, tests still pass (28/28 in test_file_system.py). Ready whenever it's convenient.

opcode81:
Thanks!

AmirF194:
Thanks for merging, and for the review. Glad the rebase came through clean.

PR #1811, opened 2026-08-04
feat(kotlin): update language server to 262.9593.0

TyceHerrman:
## Summary

- Update Serena's managed Kotlin LSP from `261.13587.0` to `262.9593.0`.
- Support JetBrains' platform-specific `.sit`, `.tar.gz`, and `.win.zip` archives, bundled JBR launchers, and Windows ARM64 builds.
- Preserve custom-version compatibility across the legacy archive layout and both modern CDN paths.
- Give the modern IntelliJ server a per-project system path and wait for its explicit startup-ready notification.
- Add dependency-resolution and startup regression tests, and update the documentation and changelog.

JetBrains still labels Kotlin LSP pre-alpha and does not publish a formal stable channel. This uses the newest published release while retaining Serena's existing CI quarantine for the resource-sensitive live Kotlin tests.

## Validation

- `uv run pytest -vv test/solidlsp/kotlin/test_kotlin_dependency_provider.py test/solidlsp/kotlin/test_kotlin_startup.py` (17 passed)
- Same focused suite with `CI=true` (17 passed, confirming the new regression tests are not quarantined)
- `GRADLE_USER_HOME=... uv run pytest -vv test/solidlsp/kotlin/test_kotlin_basic.py -m kotlin` on Python 3.13.14 (4 passed)
- `uv run poe lint`
- `uv run poe type-check`
- `git diff --check`

## Checklist

### DIFF
--- a/src/solidlsp/language_servers/kotlin_language_server.py
+++ b/src/solidlsp/language_servers/kotlin_language_server.py
@@ -90,6 +90,7 @@ KOTLIN_SERVER_ARTIFACT_BY_PLATFORM: dict[str, tuple[str, FileUtils.ArchiveType, 
     "osx-x64": (".sit", "zip", ("kotlin-server-{version}", "bin", "intellij-server")),
     "osx-arm64": ("-aarch64.sit", "zip", ("kotlin-server-{version}", "bin", "intellij-server")),
 }
+KOTLIN_SERVER_LAUNCHER_NAMES = frozenset({"intellij-server", "intellij-server.exe"})
 
 
 @dataclass(frozen=True)
@@ -220,9 +221,11 @@ class KotlinLanguageServer(SolidLanguageServer):
             return kotlin_script
 
         def _create_launch_command(self, core_path: str) -> list[str]:
-            kotlin_lsp_version = self._custom_settings.get("kotlin_lsp_version", DEFAULT_KOTLIN_LSP_VERSION)
             command = [core_path, "--stdio"]
-            if _uses_kotlin_server_packaging(kotlin_lsp_version):
+            # A custom ls_path is independent of Serena's managed version. Select
+            # arguments from the actual launcher so existing kotlin-lsp.sh/.cmd
+            # configurations do not inherit IntelliJ-server-only options.
+            if os.path.basename(core_path).lower() in KOTLIN_SERVER_LAUNCHER_NAMES:
                 command.extend(["--system-path", os.path.join(self._project_cache_dir, "kotlin-lsp-system")])
             return command
 
--- a/src/solidlsp/language_servers/kotlin_language_server.py
+++ b/src/solidlsp/language_servers/kotlin_language_server.py
@@ -25,6 +25,7 @@ from dataclasses import dataclass
 
 from overrides import override
 
+from solidlsp.dependency_provider import DownloadedDependency, DownloadedDependencyHashDatabase
 from solidlsp.ls import (
     LanguageServerDependencyProvider,
     LanguageServerDependencyProviderSinglePath,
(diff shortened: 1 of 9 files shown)


MischaPanch:
@opcode81 is using the new mechanism you introduced for managing LS dependencies appropriate here?

TyceHerrman:
I did some other work here beside just updating the kotlin LS, but i could keep only those changes and then the new LS dependency management could be run after to update kotlin ls

src/solidlsp/language_servers/kotlin_language_server.py
```diff
@@ -52,33 +53,96 @@
...
+# Modern archive filename suffix, FileUtils archive type, and launcher path within the archive.
+KOTLIN_SERVER_ARTIFACT_BY_PLATFORM: dict[str, tuple[str, FileUtils.ArchiveType, tuple[str, ...]]] = {
+    "win-x64": (".win.zip", "zip", ("bin", "intellij-server.exe")),
+    "win-arm64": ("-aarch64.win.zip", "zip", ("bin", "intellij-server.exe")),
+    "linux-x64": (".tar.gz", "gztar", ("kotlin-server-{version}", "bin", "intellij-server")),
+    "linux-arm64": ("-aarch64.tar.gz", "gztar", ("kotlin-server-{version}", "bin", "intellij-server")),
+    "osx-x64": (".sit", "zip", ("kotlin-server-{version}", "bin", "intellij-server")),
+    "osx-arm64": ("-aarch64.sit", "zip", ("kotlin-server-{version}", "bin", "intellij-server")),
+}
+KOTLIN_SERVER_LAUNCHER_NAMES = frozenset({"intellij-server", "intellij-server.exe"})
```
MischaPanch: one name should be enough, we can condition on the OS

src/solidlsp/language_servers/kotlin_language_server.py
```diff
@@ -52,33 +53,96 @@
...
+
+    kotlin_suffix = LEGACY_PLATFORM_KOTLIN_SUFFIX.get(platform_id.value)
+    if kotlin_suffix is None:
+        raise ValueError(f"Unsupported platform for Kotlin LSP {version}: {platform_id.value}")
+
+    return KotlinLSPArtifact(
+        url=f"https://download-cdn.jetbrains.com/kotlin-lsp/{version}/kotlin-lsp-{version}-{kotlin_suffix}.zip",
+        archive_type="zip",
+        launcher_parts=("kotlin-lsp.cmd",) if platform_id.is_windows() else ("kotlin-lsp.sh",),
+        sha256=INITIAL_KOTLIN_LSP_SHA256_BY_SUFFIX.get(kotlin_suffix) if version == INITIAL_KOTLIN_LSP_VERSION else None,
```
MischaPanch: why is there no sha on this path? Generally, the branching should be documented

src/solidlsp/language_servers/kotlin_language_server.py
```diff
@@ -52,33 +53,96 @@
...
+
+
+def _parse_kotlin_lsp_version(version: str) -> tuple[int, ...]:
+    try:
+        return tuple(int(part) for part in version.split("."))
+    except ValueError as exc:
+        raise ValueError(f"Kotlin LSP version must contain only dot-separated integers: {version!r}") from exc
+
+
+def _uses_kotlin_server_packaging(version: str) -> bool:
```
MischaPanch: I know this is just an isolated small thing, but still I'm not a fan of these functions. They could be inlined, they seem to only exist to support tests, and tests should generally not import private functions unless there is a very good reason. I also doubt that the tests add too much value giving the amount of mocking going on there.

Overall, I think the code selecting the download path can be shortened and explained a bit better.

MischaPanch (commented):
I will let @opcode81 review, so I just had a brief look and some comments. This bypasses existing download mechanisms - both old and new - thus increasing code complexity. If possible, such bypassing should be avoided

TyceHerrman:
i didn't think about it in previous comment, but now i updated the Kotlin provider to use DownloadedDependency from #1799. 

Addressed the three review points. Launcher detection is now OS-specific, the pinned initial and current artifacts use the central checksum database and arbitrary custom versions explicitly remain unverified, and the download-path selection is consolidated and documented without the top-level private helpers.

opcode81:
Thanks!

PR #1813, opened 2026-08-04
fix(nix): use published diagnostics for nixd

### LINKED ISSUE #1802: nixd terminates during textDocument/diagnostic, failing niche (ubuntu-latest) on main
`main` is currently red: the `niche (ubuntu-latest)` batch fails because `nixd` terminates during a `textDocument/diagnostic` request.

Latest `main` run: https://github.com/oraios/serena/actions/runs/30703366827 (commit `c7dbe3bd`) → job `niche (ubuntu-latest)`
→ `1 failed, 53 passed, 1 skipped, 1974 deselected, 1 xfailed`

It also fails on any PR branched from that commit — on #1800 it is the only red check out of 20.

## Symptom

`test/solidlsp/nix/test_nix_basic.py::TestNixLanguageServer::test_file_diagnostics[nix]`

```
test/solidlsp/nix/test_nix_basic.py:256: in test_file_diagnostics
    assert_file_diagnostics(
test/solidlsp/util/diagnostics.py:12: in assert_file_diagnostics
    diagnostics = language_server.request_text_document_diagnostics(relative_file_path, min_severity=1)
src/solidlsp/ls.py:961: in request_text_document_diagnostics
    response = self.server.send.text_document_diagnostic(
...
E   solidlsp.ls_exceptions.SolidLSPException: Error processing request textDocument/diagnostic with params:
E   {'textDocument': {'uri': '.../test/resources/repos/nix/test_repo/diagnostics_sample.nix'}}
E   (caused by LanguageServerTerminatedException: Language server stdout read process terminated unexpectedly)

----------------------------- Captured stdout call -----------------------------
ERROR solidlsp.ls_process:_read_ls_process_stderr:646 - Language server stderr reader thread terminated unexpectedly
ERROR solidlsp.ls_process:_read_ls_process_stdout:625 - LanguageServerTerminatedException: Language server stdout read process terminated unexpectedly
ERROR solidlsp.ls_process:_send_payload:663 - Failed to write to stdin: [Errno 32] Broken pipe
```

The process is gone, not merely erroring: both reader threads die and stdin is a broken pipe. All ten preceding `nix` tests in the module pass, so `nixd` survives symbol/reference/definition work and dies on the diagnostics pull specifically.

## When it started

`niche (ubuntu-latest)` was last green on `main` at `c74373e2` (run 30694340061). Only three commits have landed since: `c958aa02`, `f4ecb1a5`, `c7dbe3bd`. Of those, `f4ecb1a5` ("fix(ls): re-raise LS termination from pull-diagnostics fallback", for #1770) is the only one touching this code path — it makes `request_text_document_diagnostics` re-raise `LanguageServerTerminatedException` instead of falling back to published diagnostics.

There are two readings, and **I cannot separate them from CI logs alone**:

1. `nixd` has been dying here for some time and the fallback silently masked it; the #1770 fix correctly unmasked a real crash. The last green run's log shows no LS-termination errors — but that is *not* evidence against this, because pytest only prints captured logs for failing tests and CI runs without `--log-cli-level`, so a swallowed crash in a passing test would leave no trace.
2. `nixd` itself started crashing recently and `f4ecb1a5` is merely coincident. Worth noting that `nixd` is installed unpinned from `nixos-unstable` (`nix-env -f '<nixpkgs>' -iA nixd`), so the binary under test can change between runs independently of anything in this repo.

Either way it looks deterministic rather than flaky: two independent runs, identical failure.

## One further note on the test path

`assert_file_diagnostics` calls `language_server.request_text_document_diagnostics(...)` directly, so there is no `tools_base` layer above it to perform the restart-and-retry that the new `raise` is there to enable ("Termination must propagate so tools_base can restart the LS and retry"). However the `nixd` crash itself is resolved, a propagated termination in a direct-call test is an unconditional failure.

## Possible leads

- The `stderr` reader thread died too, so `nixd`'s own output was not captured. Capturing its stderr and exit status would show *why* it dies on `diagnostics_sample.nix` — that seems like the thing to know first.
- Pinning `nixd` in the workflow would make the niche batch reproducible over time and would settle reading 1 vs 2 for good.

I ran into this while triaging a red check on #1800 (an Erlang change that cannot touch the nix path) and could not find an existing issue for it. Happy to help if any of this is useful.

TyceHerrman:
## Summary

- use published diagnostics for nixd instead of sending the unsupported `textDocument/diagnostic` request
- extend the Nix diagnostics regression test to confirm nixd remains running
- document the fix in the changelog

## Root cause

nixd publishes diagnostics after `textDocument/didOpen`, but terminates when Serena sends `textDocument/diagnostic`. The generic diagnostics fallback previously masked that termination; the fix for #1770 correctly exposed it. This change preserves termination propagation globally and selects the published-diagnostics path only for nixd.

Fixes #1802.

## Validation

- `CI=true .venv/bin/pytest test/solidlsp/nix/test_nix_basic.py -vv` (`12 passed, 1 xfailed`)
- `.venv/bin/ty check src/serena src/solidlsp`
- `.venv/bin/ty check test --exclude test/resources`
- `.venv/bin/ruff check src/solidlsp/language_servers/nixd_ls.py test/solidlsp/nix/test_nix_basic.py`
- `.venv/bin/ruff format --check src/solidlsp/language_servers/nixd_ls.py test/solidlsp/nix/test_nix_basic.py`
- `git diff --check`

## Checklist

- [x] For changes that fix problems, I added a concise entry to `CHANGELOG.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@ Status of the `main` branch. Changes prior to the next official version change w
     project creation) 
   - Add `python_basedpyright` as an alternative Python language server
   - Nix/nixd: support custom `ls_path` launchers and external JSON settings through `config_path` #1737
+  - Fix: Nix/nixd diagnostics now use published diagnostics instead of the unsupported
+    `textDocument/diagnostic` request, which terminated nixd #1802
   - Fix: `get_diagnostics_for_file` crashed with `SolidLSPException` for any Ansible file with at least
     one lint finding, because `ansible-language-server` doesn't implement `textDocument/documentSymbol`
     and the request used to map diagnostics onto owning symbols just threw. `AnsibleLanguageServer` now
--- a/src/solidlsp/language_servers/nixd_ls.py
+++ b/src/solidlsp/language_servers/nixd_ls.py
@@ -378,6 +378,12 @@ class NixLanguageServer(SolidLanguageServer):
         }
         return initialize_params
 
+    @override
+    def _supports_pull_diagnostics(self) -> bool:
+        # nixd publishes diagnostics after textDocument/didOpen but terminates when it receives
+        # textDocument/diagnostic. Force the published-diagnostics path instead.
+        return False
+
     def _start_server(self):
         """Start nixd server process"""
         initialize_params = self._create_initialize_params()
--- a/test/solidlsp/nix/test_nix_basic.py
+++ b/test/solidlsp/nix/test_nix_basic.py
@@ -259,3 +259,4 @@ class TestNixLanguageServer:
             (),
             min_count=1,
         )
+        assert language_server.is_running()


opcode81:
Thanks a lot for this fix.

PR #1815, opened 2026-08-06
Add Nextflow language server support

vagkaratzas:
## Checklist

Feature coded and tested by Claude Opus 5 (on high effort).

Adds support for Nextflow (language server id `nextflow`) using the official
[Nextflow language server](https://github.com/nextflow-io/language-server), the same
one that backs the VS Code extension.

**Runtime**: the `language-server-all.jar` (pinned to v26.04.3) is downloaded
automatically via `DownloadedDependency`, with its hash in
`downloaded_dependency_hashes.json` and an `update_dep_hashes()` hook wired into
`scripts/update_downloaded_dependency_hashes.py`. Requires a Java 17+ runtime,
resolved from `ls_specific_settings.nextflow.java_home` -> `JAVA_HOME` -> PATH.
Uses `LanguageServerDependencyProviderSinglePath`.

**Two non-obvious behaviours of this server**, both handled in `_start_server`:

1. It indexes nothing on `initialized`. It scans the workspace only on
   `workspace/didChangeConfiguration`, and only when the incoming config *differs*
   from its own defaults in `errorReportingMode`, `files.exclude` or
   `pluginRegistryUrl` (see `NextflowLanguageServer.shouldInitialize` upstream).
   Sending its defaults back is a no-op and every symbol request then returns `[]`,
   so the pushed config deliberately sets `errorReportingMode: "errors"`.
   The scan is reported via `$/progress`; startup blocks until it ends.
2. `.config` files are deliberately *not* registered as source files. The server
   does have a `ConfigService` for them, but there is no `ConfigSymbolProvider` —
   they yield zero symbols, so claiming them would only pollute the symbol index.
   (Upstream's own `ConfigService.matchesFile` also excludes `nf-test.config`.)

**Symbol names**: the server prefixes every name with its declaration keyword
(`process GREET`, `workflow SAY_HELLO`, `function foo`). `_normalize_symbol_name`
strips it so symbols are addressable as written in source; the implicit entry
workflow keeps the server's `<entry>` placeholder.

**Tests**: 6 tests in `test/solidlsp/nextflow/`, asserting concrete symbol names
plus within-file and cross-file references against a minimal pipeline in
`test/resources/repos/nextflow/test_repo`. Marker `nextflow` added to
`MARKERS_JVM`, whose CI batch already installs Java 21. Tests skip only when no
`java` is on PATH.

Also touched, as the standard accompaniments: README language list,
`docs/01-about/020_programming-languages.md`, `CHANGELOG.md`,
`project.template.yml` (regenerated with `scripts/print_language_list.py`), and one
line in `src/serena/hooks.py` adding `.nf` to `_CODE_FILE_EXTENSIONS`.


MischaPanch:
Thanks, looks good, merging!

vagkaratzas:
Thank you @MischaPanch!

PR #1820, opened 2026-08-07
fix(project_server): serialize access to the process-wide active project

NUMORDO:
## Problem

`ProjectServer._query_project` enters `agent.active_project_context(project)` on the Flask
worker thread, but `tool.apply_ex(...)` hands the actual work to the agent's single-threaded
`TaskExecutor`. The context manager and the tool execution therefore run on **different
threads**, so two concurrent requests can interleave:

```
T1: enter ctx(projA) -> dispatch tool to executor
T2: enter ctx(projB)                                  # overwrites the process-wide slot
T1:                     tool runs -> sees projB       # wrong project
T2: exit ctx -> restores whatever T1 had saved        # wrong restore
```

`self._app.run(..., threaded=True)` makes this reachable whenever a client issues concurrent
`query_project` calls against different projects — which is exactly the multi-agent scenario
the tool exists for. `_loaded_projects_by_root` also has an unsynchronized
read-modify-write on the same path.

Note this affects only the **ProjectServer** route (symbolic tools). The inline path in
`QueryProjectTool.apply` is unaffected: it calls `tool.apply()` synchronously inside the
context on the same thread.

## Reproduction

Two dummy registered projects, each with a same-named file containing distinct content;
`read_file` issued through `/query_project` from 8 threads, alternating projects.

| | sequential (8) | concurrent (8 workers) |
|---|---|---|
| before | 0 wrong | **10–30 wrong out of 24–48** |
| after | 0 wrong | **0 out of 48** |

Two failure shapes were observed: the *other* project's file content, and
`Error: No active project ...` (when one request's restore landed between another's
enter/exit). Measured against 1.5.3 and the same change applied to `main`.

## Fix

Serialize only the section that holds the active project, and give project loading its own
lock so a cold language-server startup does not block queries against already-loaded
projects. Tool execution was already serialized by the task executor, so this adds no
contention beyond what already existed, and no public API changes.

## Alternative considered

Passing the tool application into the executor as a single task (so that the swap and the
call share one thread) also fixes it, but requires an `apply_ex(..., dispatch=False)`
parameter to avoid re-entering the executor from within an executor task (which deadlocks,
since the queue processor waits for the current task before popping the next). The lock is
smaller and keeps `apply_ex`'s behavior — logging, language-server restart retry, usage
recording, exception conversion — entirely intact.

## Context

Found while hardening a Claude Code hook that prevents parallel sub-agents from clobbering
each other's `activate_project` (the active project is a single process-wide slot shared by
every agent on one MCP connection). `query_project` is the recommended isolated path there,
so its server-backed variant needs to be safe under concurrency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,7 @@
 Status of the `main` branch. Changes prior to the next official version change will appear here.
 
 * General:
-  - Fix: concurrent ProjectServer requests no longer block cached project lookups behind an unrelated project's cold language-server startup
-  - Fix: the README, the Language Support docs page and the project template omitted several already-supported language servers
+  - Fix: Race conditions in ProjectServer when used by multiple clients in parallel   
   - Fix: a tool call exceeding the timeout blocked the task executor indefinitely; the executor now
     recovers without user-induced cancellation
   - Add Grok Build support (context `grok`, setup CLI, hooks)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Status of the `main` branch. Changes prior to the next official version change will appear here.
 
 * General:
+  - Fix: concurrent ProjectServer requests no longer block cached project lookups behind an unrelated project's cold language-server startup
   - Fix: the README, the Language Support docs page and the project template omitted several already-supported language servers
   - Fix: a tool call exceeding the timeout blocked the task executor indefinitely; the executor now
     recovers without user-induced cancellation
--- a/src/serena/project_server.py
+++ b/src/serena/project_server.py
@@ -62,12 +62,7 @@ class ProjectServer:
 
         self._agent = SerenaAgent(serena_config=serena_config)
         self._loaded_projects_by_root: dict[str, "Project"] = {}
-        # Keep these locks for the server's lifetime. Removing one after a failed load could
-        # let a retry initialize concurrently with threads already waiting on the old lock.
         self._project_load_locks_by_root: dict[str, threading.Lock] = {}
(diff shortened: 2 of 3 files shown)


AmirF194 (commented):
Nice repro and a clean fix for the cross-thread active-project race. One thing worth a look before merge: `_loaded_projects_lock` doesn't quite deliver the isolation the PR description claims.

The lock in `_get_project` (project_server.py:98-108) is a single global lock, held for the full duration of a cold `get_project_instance()` + `create_language_server_manager()` call, not scoped per project. So a concurrent request for an already-cached project blocks behind an unrelated project's cold LSP startup, which is the exact case the PR body says this design avoids ("project loading is guarded separately so that a cold language server startup does not block queries against already-loaded projects").

Verified with a minimal harness that imports `ProjectServer` directly from this branch (head `ff56d54`) and drives `_get_project` from two threads with a project registry mocked to make one project's load take 2s: a lookup of a second, already-loaded project started 0.3s later waits ~1.7s for the unrelated cold load to release the lock, instead of returning immediately.

```
cold load total: 2.00s
cached lookup wait (started 0.3s after cold load): 1.70s
BLOCKED: cached-project lookup waited on the unrelated cold load
```

Not a correctness bug (the original race is still fixed), just a scalability gap relative to the stated goal. A per-project lock (e.g. a `dict[str, threading.Lock]` keyed by project root, taken only around that project's own load, with the shared dict lock held just for the membership check/insert) would keep already-loaded lookups from stalling on someone else's cold start. Happy to share the repro script if useful.

opcode81:
Thank you for your contribution!

PR #1823, opened 2026-08-08
Fix: LSP process-group cleanup no longer requires process-table enumeration

### LINKED ISSUE #1818: Avoid process-table enumeration when cleaning up independent LSP process groups
## Summary

Serena already starts LSP subprocesses in independent sessions by default, but shutdown still discovers descendants with `psutil.Process.children(recursive=True)`. On sandboxed macOS this requires `sysctl(KERN_PROC_ALL)`, which can be denied even for language-server processes Serena launched itself.

The same dependency also makes the Angular startup-cleanup regression fail before it exercises the behavior under test.

## Reproduction

On macOS with process-table enumeration denied, run:

```text
test/solidlsp/angular/test_angular_error_cases.py::TestAngularStartupCleanup::test_companions_cleaned_up_on_initialize_failure
```

The first snapshot fails at:

```python
psutil.Process(os.getpid()).children(recursive=True)
```

with:

```text
Operation not permitted (originated from sysctl(KERN_PROC_ALL))
```

Production cleanup has the same dependency in `src/solidlsp/util/subprocess_util.py`: `_signal_process_tree` calls `parent.children(recursive=True)`. If enumeration fails during Angular's partial-startup cleanup, the companion reference is cleared but the full process tree cannot be proven terminated and grandchildren may survive.

## Existing ownership boundary

`ProcessIO` launches the LSP with `start_new_session=True` whenever `LanguageServerConfig.start_independent_lsp_process` is enabled, which is the default. Serena therefore already has a known POSIX process group for these owned subprocesses; it does not need a system-wide child-process query to clean them up.

A bounded probe in the same macOS sandbox started a new-session process plus a grandchild and successfully terminated both with `os.killpg` without process enumeration.

## Proposed behavior

- On POSIX, record the process-group ID at launch for independently started LSP processes. With `start_new_session=True`, the new process is the group leader, so its initial PGID is its PID.
- Store that PGID rather than deriving it after the leader may have exited.
- During cleanup, send `SIGTERM` to the known group, wait for the direct `Popen`, then send `SIGKILL` to the same group on timeout.
- Treat `ProcessLookupError` as an already-clean group; log unexpected permission or signaling failures.
- Never call `killpg` when independent-process mode is disabled, because that could signal Serena's own process group.
- Keep a platform-specific Windows strategy. The current psutil path may remain initially; a Windows Job Object would provide the equivalent explicit ownership boundary.

## Regression tests

- Replace the Angular test's before/after enumeration of every pytest descendant with captured `Popen`/PGID handles for the companions Serena starts.
- Verify group disappearance using `os.killpg(pgid, 0)` and `ProcessLookupError`, without consulting the global process table.
- Add a focused POSIX cleanup test whose helper starts a child and grandchild in one new session, invokes the cleanup utility, and verifies both are gone.
- Cover the `SIGTERM` success path, forced `SIGKILL` fallback, an already-exited group, and `start_independent_lsp_process=False`.
- Retain a separate Windows-specific test/implementation path rather than weakening the POSIX assertions.

## Acceptance criteria

- Owned LSP children and grandchildren are terminated on POSIX without `ps`, `sysctl`, or `psutil` descendant enumeration.
- Angular partial-startup failure does not leak the TypeScript or HTML companion process groups.
- Cleanup cannot signal Serena/pytest's own process group when independent-process mode is disabled.
- The regression passes on macOS and Linux and has an explicit Windows strategy.
- Normal graceful LSP shutdown remains the first step before group termination.

## Related context

- #1816 reports a self-daemonized, intentionally shared Bloop server. This issue is narrower: it concerns descendants that remain in the independent process group Serena created and owns. It does not propose killing reparented or shared daemons by name.
- This was discovered while running the full test matrix for PR #1761, but it is unrelated to that PR's memory-frontmatter changes.

I searched open and closed Serena issues for `KERN_PROC_ALL`, recursive psutil child enumeration, `killpg`, Angular cleanup, and language-server process groups and found no exact match.

AmirF194:
## Root cause

`terminate_process_tree_with_kill_fallback` always discovers a language server's descendants via `psutil.Process(pid).children(recursive=True)`, which requires system-wide process-table enumeration (`sysctl(KERN_PROC_ALL)` on macOS). That can be denied in a sandboxed environment even for processes Serena started and owns, so cleanup silently drops to signaling only the single process object it was given, leaking any children it spawned itself.

## Invariant

`StdioLanguageServer` already launches every independent LSP process with `start_new_session=True` (the default), which makes it the leader of its own POSIX process group with a known PGID equal to its PID at creation. Serena does not need a system-wide query to find processes it already owns by group membership.

## Fix

- `StdioLanguageServer._start` captures `process.pid` as the process group ID immediately after launch, only when the process was started independently and on POSIX (`None` on Windows or when independent mode is disabled).
- `terminate_process_tree_with_kill_fallback` takes an optional `process_group_id`; when given, `_signal_process_group` sends `SIGTERM`, then `SIGKILL` on timeout, straight to `-pgid` via `os.killpg`, never touching psutil. `ProcessLookupError` (group already gone) is treated as clean; other errors are logged, not raised.
- When `process_group_id` is `None` (independent mode disabled), behavior is unchanged: the old `psutil`-based path runs, and `killpg` is never called, so a shared-group process can't accidentally signal Serena itself.

## Verification (Docker, Python 3.11, `uv sync --extra dev --locked`)

- `test/solidlsp/test_process_group_cleanup.py` (new, 8 tests): fails to import on `main` (`_signal_process_group` doesn't exist there) and passes on this branch. Covers group termination of a child+grandchild by PGID alone, the `SIGKILL` fallback, an already-exited group, and that `os.killpg` is never called when `process_group_id` is `None`. Two of the tests simulate the actual production consequence by mocking `psutil.Process` to raise `AccessDenied` (the same exception class regardless of which syscall a platform used to deny it): without a group id the leader's own child is left running; with it, the group is still fully reaped.
- `uv run poe lint`, `uv run poe type-check`, and `codespell` (matching the repo's actual CI scope, i.e. excluding `.venv`) all pass. The pre-existing `test_pdeathsig.py` (#1490 regression) still passes unchanged.
- Not verified: I have no macOS environment, so I did not reproduce the original `sysctl(KERN_PROC_ALL)` / `EPERM` denial itself; the fix and its tests target the same `psutil.AccessDenied` exception class the sandbox raises, not the macOS-specific trigger. I also did not rewrite `test/solidlsp/angular/test_angular_error_cases.py::TestAngularStartupCleanup` to use PGID-based verification instead of its current pytest-descendant `psutil` enumeration; that test still passes as-is (enumeration isn't denied on Linux CI), converting it needs the full Angular/TypeScript toolchain and felt like a separately reviewable change rather than bundling it here.

Fixes #1818

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ Status of the `main` branch. Changes prior to the next official version change w
     a directory named with pattern metacharacters (e.g. a stray `***`) could turn a scoped pattern
     into one matching far more than intended, silently excluding most or all of the project from
     indexing #1806
+  - Fix: cleanup of an independently-started LSP process and its children required enumerating the
+    system process table (`psutil`), which can be denied even for processes Serena owns in a
+    sandboxed environment; cleanup now signals the known process group directly (#1818)
   - Fix: the README, the Language Support docs page and the project template omitted several already-supported language servers
   - Fix: a tool call exceeding the timeout blocked the task executor indefinitely; the executor now
     recovers without user-induced cancellation
--- a/src/solidlsp/ls_process.py
+++ b/src/solidlsp/ls_process.py
@@ -1,6 +1,7 @@
 import asyncio
 import json
 import logging
+import os
 import socket
 import subprocess
 import threading
@@ -510,6 +511,9 @@ class StdioLanguageServer(LanguageServerInterface):
         self.process_launch_info = process_launch_info
         self.process: subprocess.Popen[bytes] | None = None
         self.start_independent_lsp_process = start_independent_lsp_process
+        self._process_group_id: int | None = None
+        """the POSIX process group ID owned by ``self.process``, captured at launch time (see
+        ``_start``); ``None`` on Windows or when the process was not started in its own session"""
 
         self._stdin_lock = threading.Lock()
 
@@ -523,6 +527,8 @@ class StdioLanguageServer(LanguageServerInterface):
             self.process_launch_info, start_new_session=self.start_independent_lsp_process
         )
         self.process = process
+        # start_new_session=True makes the new process its own group leader, so its PGID is its PID.
+        self._process_group_id = process.pid if (self.start_independent_lsp_process and os.name == "posix") else None
 
         # Check if process terminated immediately
         if process.returncode is not None:
@@ -559,10 +565,14 @@ class StdioLanguageServer(LanguageServerInterface):
                 # Ignore errors here, we are proceeding to terminate anyway.
             # terminate the process
             subprocess_util.terminate_process_tree_with_kill_fallback(
-                self.process, terminate_timeout=timeout, process_name=f"LS[{self.ls_id.value}]"
+                self.process,
+                terminate_timeout=timeout,
+                process_name=f"LS[{self.ls_id.value}]",
+                process_group_id=self._process_group_id,
             )
         finally:
             self.process = None
+            self._process_group_id = None
 
     @staticmethod
     def _safely_close_pipe(pipe: IO[AnyStr] | None) -> None:
--- a/src/solidlsp/ls_process.py
+++ b/src/solidlsp/ls_process.py
@@ -1,9 +1,7 @@
 import asyncio
 import json
 import logging
-import os
 import socket
-import subprocess
 import threading
 import time
 from abc import ABC, abstractmethod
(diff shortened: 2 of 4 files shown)


opcode81:
I added a second commit to add an abstraction in which the lifecycle management is encapsulated.

opcode81:
Thanks!

AmirF194:
Thanks for merging, and for adding the lifecycle abstraction on top, that's a cleaner shape than what I had.

PR #1827, opened 2026-08-08
Clean up memories and move system prompt content to critical_info memory

opcode81:


MischaPanch (approved):
Cool! Will the sandboxing of the template still permit function execution during templating? I guess, since we pass the function, it should be fine, right?

PR #1828, opened 2026-08-08
Enclose sub-prompts in tags

opcode81:
This creates a bit more order (and traceability) in prompts that are assembled from sub-prompts, as it makes the scope of each prompt clear.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Status of the `main` branch. Changes prior to the next official version change w
     failed `--project-from-cwd` auto-detection (#1773).
   - ProjectServer: Configure trusted hosts (local hosts only) when listening on localhost
   - SerenaDashboardTrayManager: Configure trusted hosts (local hosts only)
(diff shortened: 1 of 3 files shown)


TyceHerrman:
@opcode81 I see commit `d3a8e9b5` changed the activation guidance for a project without memories from:

> Onboarding has not been performed yet, you should call Serena's `onboarding` tool now to set up project memories.

to:

> Onboarding has not been performed yet. Ask the user whether to perform onboarding via the `onboarding` tool.

Was this intended to introduce explicit consent for first-time onboarding? The current [documentation](https://github.com/oraios/serena/blob/main/docs/02-usage/045_memories.md#onboarding) still says Serena performs onboarding by default and that activation triggers it. Noticed this behavior after updating to 1.7.0.

Also now in several tasks in codex, Serena emitted this instruction but the agent didn’t ask me about onboarding. In one case, it explicitly decided onboarding was unnecessary for a read-only task and continued. In others, it only mentioned onboarding after completing the work. The instruction appears to be treated as optional or dependent on task scope

PR #1831, opened 2026-08-09
SolidLSP: admit Struct, Interface and Constant symbols as reference containers

xormania:
## Problem

`find_referencing_symbols` degrades to file-level attribution for any reference located inside a Go struct body, interface body or `const` group. Each such reference logs

```
WARNING ... solidlsp.ls:request_referencing_symbols:2449 - Could not find containing symbol for <file>:<line>:<col>. Returning file symbol instead
```

and the result loses the "which symbol references this" answer the tool exists to give — on real Go codebases this affects a large share of type references, since struct fields, interface signatures and const groups are where named types are used. This is not a constructed case: it was encountered during ordinary use on the Go core of a public codebase ([xormania/looplaw](https://github.com/xormania/looplaw)), where `serena project health-check` surfaces the warning out of the box and reference results degrade to file level during routine refactoring.

## Cause

`request_containing_symbol` filters candidate containers to Python's container kinds `{Method, Function, Class}`, plus `Variable` via the one-liner-exempt path. The language server is blameless: `hierarchicalDocumentSymbolSupport` is declared and gopls's DocumentSymbol tree is complete — but Go structs arrive as `SymbolKind.Struct`, interfaces as `Interface` and const-group members as `Constant`, so every candidate is rejected and `request_referencing_symbols` falls back to the file symbol.

## Fix

Mirrors the filter's existing structure:

- `Struct` and `Interface` join the multi-line container whitelist, subject to the existing one-liner exclusion, which is unchanged;
- `Constant` joins `Variable` on the one-liner-exempt path (const-group members are one-liners by nature, exactly like the module-level variables that path already serves).

This also admits the same kinds for other language servers (e.g. Rust structs, TypeScript interfaces); see Verification for cross-language results. There is no cache interaction: containment is derived per-request from the document-symbol caches, whose stored content this change does not alter, so no cache version is bumped.

## Verification

New Go fixture `containment_sample.go` with a const-iota group, a struct field and an interface method all referencing a named type; four regression tests assert containing-symbol and referencing-symbol attribution, and all four fail against the pre-change `ls.py` (verified). Full `-m go` suite: 21 passed, 1 skipped. `poe format` and `poe type-check` clean.

Cross-language: the rust (36), cpp (36 + 1 xfail), typescript (12), java (8), al (34), scss (25) and python (80) suites all pass on this branch. Because those suites mostly assert reference paths rather than attribution, the behavior was also measured directly: A/B attribution probes (same corpus, base `main` vs this branch, real language servers) over the fixture repos plus adversarial corpora built around the risky shapes — multi-line constant initializers, struct/interface bodies, enum and `const` groups — for Go, Python, Rust, C++/clangd, TypeScript, Java, Kotlin and SCSS found **zero attribution regressions and zero lost references**. For clangd and Some Sass the change was a no-op on every probed corpus (neither server emitted the added kinds there). Everywhere the kinds are emitted (rust-analyzer statics/structs/traits, JDTLS multi-line `static final` constants, TS consts/interfaces, kotlin-lsp, pyright enum members), every difference is either a previously-dropped reference gaining its lexical container, or attribution tightening from the enclosing function/class to the exact declaration containing the reference — the same treatment `Variable` symbols already receive. Finally, the same probe over the originating codebase's own Go packages: 649 attributions identical, zero shifted, zero lost — and six references that base dropped entirely, including the two that motivated this PR, now attribute to their lexical containers (const-group members and structs).

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,10 @@ Status of the `main` branch. Changes prior to the next official version change w
     `Deno.*` globals, which the plain TypeScript language server does not. Overlaps TypeScript on
     file extensions, so it is not auto-detected and must be selected explicitly; requires the
     `deno` CLI on PATH
+  - Fix: `find_referencing_symbols` reported file-level containers for references located inside Go
+    struct bodies, interface bodies and `const` groups, because containing-symbol detection admitted
+    only Python's container kinds; `Struct`, `Interface` and `Constant` symbols are now admitted as
+    containers
 
 * JetBrains:
   - `jet_brains_find_symbol`: Disallow wildcard-only search, delegating to overview tool if request is for file
--- a/src/solidlsp/ls.py
+++ b/src/solidlsp/ls.py
@@ -2533,8 +2533,9 @@ class SolidLanguageServer(ABC):
     ) -> ls_types.UnifiedSymbolInformation | None:
         """
         Finds the first symbol containing the position for the given file.
-        For Python, container symbols are considered to be those with kinds corresponding to
-        functions, methods, or classes (typically: Function (12), Method (6), Class (5)).
+        Container symbols are those with kinds corresponding to functions, methods, classes, structs
+        and interfaces, as well as variables and constants (which may be one-liners, e.g. members
+        of a Go ``const`` group).
 
         The method operates as follows:
           - Request the document symbols for the file.
@@ -2546,7 +2547,7 @@ class SolidLanguageServer(ABC):
             among those above the given line.
           - If no container candidates are found, return None.
 
-        :param relative_file_path: The relative path to the Python file.
+        :param relative_file_path: The relative path to the file.
         :param line: The 0-indexed line number.
         :param column: The 0-indexed column (also called character). If not passed, the lookup will be based
             only on the line.
@@ -2589,8 +2590,15 @@ class SolidLanguageServer(ABC):
                 location["relativePath"] = relative_file_path
                 location["uri"] = Path(absolute_file_path).as_uri()
 
-        # Allowed container kinds, currently only for Python
-        container_symbol_kinds = {ls_types.SymbolKind.Method, ls_types.SymbolKind.Function, ls_types.SymbolKind.Class}
+        # allowed container kinds; Struct and Interface cover languages (e.g. Go, Rust, C) whose
+        # containers are not reported as classes
+        container_symbol_kinds = {
+            ls_types.SymbolKind.Method,
+            ls_types.SymbolKind.Function,
+            ls_types.SymbolKind.Class,
+            ls_types.SymbolKind.Struct,
+            ls_types.SymbolKind.Interface,
+        }
 
         def is_position_in_range(line: int, range_d: ls_types.Range) -> bool:
             start = range_d["start"]
@@ -2613,7 +2621,10 @@ class SolidLanguageServer(ABC):
             for s in document_symbols.iter_symbols()
             if s["kind"] in container_symbol_kinds and s["location"]["range"]["start"]["line"] != s["location"]["range"]["end"]["line"]
         ]
-        var_containers = [s for s in document_symbols.iter_symbols() if s["kind"] == ls_types.SymbolKind.Variable]
+        # variables and constants are admitted even as one-liners (e.g. members of a Go const group)
+        var_containers = [
+            s for s in document_symbols.iter_symbols() if s["kind"] in {ls_types.SymbolKind.Variable, ls_types.SymbolKind.Constant}
+        ]
         candidate_containers.extend(var_containers)
 
         if not candidate_containers:
--- a/test/resources/repos/go/test_repo/containment_sample.go
+++ b/test/resources/repos/go/test_repo/containment_sample.go
@@ -1,0 +1,21 @@
+package main
+
+type Severity int
+
+const (
+    SevLow Severity = iota
+    SevHigh
+)
+
+type Alert struct {
+    Level Severity
+    Msg   string
+}
+
+func (a Alert) IsHigh() bool {
+    return a.Level == SevHigh
+}
+
+type Notifier interface {
+    Notify(level Severity) error
+}
--- a/test/solidlsp/go/test_go_basic.py
+++ b/test/solidlsp/go/test_go_basic.py
@@ -4,11 +4,13 @@ from pathlib import Path
 import pytest
 
 from serena.symbol import LanguageServerSymbol
+from serena.util.text_utils import find_text_coordinates
 from solidlsp import SolidLanguageServer
 from solidlsp.ls_config import LanguageServerId
+from solidlsp.ls_types import SymbolKind
 from solidlsp.ls_utils import SymbolUtils
 from test.conftest import find_identifier_position, get_repo_path, ls_has_verified_implementation_support
-from test.solidlsp.conftest import format_symbol_for_assert, has_malformed_name, request_all_symbols
+from test.solidlsp.conftest import format_symbol_for_assert, has_malformed_name, read_repo_file, request_all_symbols
 
 
 @pytest.mark.go
@@ -87,6 +89,62 @@ class TestGoLanguageServer:
             body = sym["body"].get_text()
             assert body.startswith(name), f"Expected grouped var {name} body to start with the identifier, got {body[:24]!r}"
             assert not body.startswith("var"), f"Grouped var {name} body must not include the 'var' keyword"
+
+    @pytest.mark.parametrize("language_server", [LanguageServerId.GO], indirect=True)
+    def test_request_containing_symbol_const_group_member(self, language_server: SolidLanguageServer) -> None:
+        """A reference inside a ``const`` group must be attributed to the enclosing constant."""
+        file_path = os.path.join("containment_sample.go")
+        file_content = read_repo_file(language_server, file_path)
+        coords = find_text_coordinates(file_content, r"SevLow (Severity) = iota")
+        assert coords is not None, "Could not find the Severity reference in the const group"
+        containing_symbol = language_server.request_containing_symbol(file_path, coords.line, coords.col)
+        assert containing_symbol is not None, "Expected a containing symbol for a reference inside a const group"
+        assert containing_symbol["name"] == "SevLow"
+        assert containing_symbol["kind"] == SymbolKind.Constant
+
+    @pytest.mark.parametrize("language_server", [LanguageServerId.GO], indirect=True)
+    def test_request_containing_symbol_struct_field(self, language_server: SolidLanguageServer) -> None:
+        """A reference in a struct field declaration must be attributed to the enclosing struct."""
+        file_path = os.path.join("containment_sample.go")
+        file_content = read_repo_file(language_server, file_path)
+        coords = find_text_coordinates(file_content, r"Level (Severity)")
+        assert coords is not None, "Could not find the Severity reference in the struct field"
+        containing_symbol = language_server.request_containing_symbol(file_path, coords.line, coords.col)
+        assert containing_symbol is not None, "Expected a containing symbol for a reference inside a struct body"
+        assert containing_symbol["name"] == "Alert"
+        assert containing_symbol["kind"] == SymbolKind.Struct
+
+    @pytest.mark.parametrize("language_server", [LanguageServerId.GO], indirect=True)
+    def test_request_containing_symbol_interface_method(self, language_server: SolidLanguageServer) -> None:
+        """A reference in an interface method signature must be attributed to the enclosing interface."""
+        file_path = os.path.join("containment_sample.go")
+        file_content = read_repo_file(language_server, file_path)
+        coords = find_text_coordinates(file_content, r"Notify\(level (Severity)\)")
+        assert coords is not None, "Could not find the Severity reference in the interface method"
+        containing_symbol = language_server.request_containing_symbol(file_path, coords.line, coords.col)
+        assert containing_symbol is not None, "Expected a containing symbol for a reference inside an interface body"
+        assert containing_symbol["name"] == "Notifier"
+        assert containing_symbol["kind"] == SymbolKind.Interface
+
+    @pytest.mark.parametrize("language_server", [LanguageServerId.GO], indirect=True)
+    def test_request_referencing_symbols_attributes_containers(self, language_server: SolidLanguageServer) -> None:
+        """References to a type used in a const group, a struct field and an interface method must
+        be attributed to the constant, the struct and the interface respectively, not to the file.
+        """
+        file_path = os.path.join("containment_sample.go")
+        all_symbols, _ = language_server.request_document_symbols(file_path).get_all_symbols_and_roots()
+        severity_symbol = next((sym for sym in all_symbols if sym.get("name") == "Severity"), None)
+        assert severity_symbol is not None, "Could not find the 'Severity' type symbol in containment_sample.go"
+        sel_start = severity_symbol["selectionRange"]["start"]
+        ref_symbols = [
+            ref.symbol for ref in language_server.request_referencing_symbols(file_path, sel_start["line"], sel_start["character"])
+        ]
+        assert ref_symbols, "Expected references to the Severity type"
+        ref_names = {ref["name"] for ref in ref_symbols}
+        assert "SevLow" in ref_names
+        assert "Alert" in ref_names
+        assert "Notifier" in ref_names
+        assert all(ref["kind"] != SymbolKind.File for ref in ref_symbols), f"File-level fallback attribution in {ref_names}"
 
     if ls_has_verified_implementation_support(LanguageServerId.GO):
 


xormania:
Everything needed to rerun the cross-language verification from the description, in case anyone wants to check the "zero regressions" claim — or break it with a nastier corpus.

The probe records the containing-symbol attribution of every reference in a corpus file. Run it from a checkout of `main` and of this branch, diff the two JSONs — the delta is the behavioral change, independent of what any test asserts. The corpora below were written to hit the risky shapes: multi-line const/static initializers with references inside them, struct/interface bodies, enum and `const` groups.

```
.venv/bin/python ab_probe.py out.json <language> <path-to-fixture-repo> <corpus-filename>
```

Across these six corpora, the untouched Go/Python fixture repos, and the Go packages of the real-world codebase where the bug was first hit: zero regressions, zero lost references (on the originating codebase: 649 attributions identical, six previously-dropped references recovered). clangd and some-sass emitted none of the added kinds on these corpora (no-op both sides); everywhere else, every difference is a previously-dropped reference gaining its lexical container, or attribution tightening into the declaration that contains the reference.

<details>
<summary><b>Probe harness — ab_probe.py</b></summary>

```python
"""A/B attribution probe for the container-kind fix.

For each language fixture repo: enumerate document symbols, request referencing symbols at each
symbol's selectionRange, and record the containing-symbol attribution of every reference.
Run identically in a base-main worktree and the fix worktree; diff the JSON outputs.

Usage: python ab_probe.py OUT.json lang1 lang2 ...
"""

import json
import os
import re
import sys
import tempfile
from pathlib import Path

sys.path.insert(0, os.getcwd())

from solidlsp.ls_config import LanguageServerId
from test.conftest import get_repo_path, start_ls_context

EXTS = {
    "go": [".go"],
    "rust": [".rs"],
    "typescript": [".ts", ".tsx"],
    "cpp": [".cpp", ".hpp", ".h", ".cc"],
    "scss": [".scss"],
    "clojure": [".clj", ".cljs", ".cljc"],
    "python": [".py"],
    "java": [".java"],
    "kotlin": [".kt"],
}

LS_CANDIDATES = {
    "go": ["GO"],
    "rust": ["RUST", "RUST_ANALYZER"],
    "typescript": ["TYPESCRIPT", "TYPESCRIPT_LANGUAGE_SERVER", "TS"],
    "cpp": ["CLANGD", "CPP"],
    "scss": ["SCSS", "SOME_SASS", "SOMESASS"],
    "clojure": ["CLOJURE", "CLOJURE_LSP"],
    "python": ["PYRIGHT", "PYTHON", "PYTHON_PYRIGHT"],
    "java": ["JAVA", "ECLIPSE_JDTLS", "JDTLS"],
    "kotlin": ["KOTLIN"],
}

SKIP_DIR_PARTS = {"node_modules", "target", "build", ".git", "__pycache__"}


def resolve_ls_id(lang: str) -> LanguageServerId | None:
    for cand in LS_CANDIDATES[lang]:
        if hasattr(LanguageServerId, cand):
            return getattr(LanguageServerId, cand)
    lang_lower = lang.lower()
    for member in LanguageServerId:
        if member.value.lower() == lang_lower or member.name.lower() == lang_lower:
            return member
    return None


def probe(lang: str, repo_override: str | None = None, file_filter: str | None = None) -> list[dict]:
    ls_id = resolve_ls_id(lang)
    if ls_id is None:
        return [{"lang": lang, "error": f"no LanguageServerId resolved; members={[m.name for m in LanguageServerId][:60]}"}]
    repo = Path(repo_override) if repo_override else Path(get_repo_path(ls_id))
    rows: list[dict] = []
    solidlsp_dir_env = os.environ.get("ABPROBE_SOLIDLSP_DIR")
    solidlsp_dir = Path(solidlsp_dir_env) if solidlsp_dir_env else Path(tempfile.mkdtemp(prefix=f"abprobe-{lang}-"))
    # Optional LS version pin (e.g. reuse an already-installed Kotlin LSP so both A/B runs
    # use the identical server binary and no download is needed).
    ls_specific_settings = None
    kotlin_version = os.environ.get("ABPROBE_KOTLIN_LSP_VERSION")
    if lang == "kotlin" and kotlin_version:
        ls_specific_settings = {ls_id: {"kotlin_lsp_version": kotlin_version}}
    with start_ls_context(
        ls_id,
        repo_path=str(repo) if repo_override else None,
        ls_specific_settings=ls_specific_settings,
        solidlsp_dir=solidlsp_dir,
    ) as ls:
        files = sorted(
            p.relative_to(repo).as_posix()
            for ext in EXTS[lang]
            for p in repo.rglob(f"*{ext}")
            if not (set(p.parts) & SKIP_DIR_PARTS) and not any(part.startswith(".") for part in p.parts)
        )
        for f in files:
            if file_filter and file_filter not in f:
                continue
            try:
                syms, _ = ls.request_document_symbols(f).get_all_symbols_and_roots()
            except Exception as e:
                rows.append({"lang": lang, "file": f, "error": str(e)[:160]})
                continue
            file_lines = (repo / f).read_text().split("\n")
            seen_targets: set[str] = set()
            for s in syms:
                sel = s.get("selectionRange")
                name = s.get("name")
                if not sel or not name or name in seen_targets:
                    continue
                seen_targets.add(name)
                # Query positions: the declaration's selectionRange start, plus every textual
                # occurrence of the name in the file (some LSes — e.g. some-sass — return no
                # references when the request originates at the declaration itself, only from
                # usage sites).
                positions = [(sel["start"]["line"], sel["start"]["character"])]
                pat = re.compile(r"(?<![\w$%.-])" + re.escape(name) + r"(?![\w-])")
                for ln, text in enumerate(file_lines):
                    for m in pat.finditer(text):
                        pos = (ln, m.start() + 1)
                        if pos not in positions:
                            positions.append(pos)
                seen_rows: set[tuple] = set()
                for qline, qcol in positions:
                    try:
                        refs = ls.request_referencing_symbols(f, qline, qcol)
                    except Exception as e:
                        rows.append({"lang": lang, "file": f, "target": name, "error": str(e)[:160]})
                        continue
                    for r in refs:
                        cs = r.symbol
                        loc = cs.get("location", {}) or {}
                        key = (name, loc.get("relativePath"), r.line, r.character, cs.get("name"), int(cs.get("kind", -1)))
                        if key in seen_rows:
                            continue
                        seen_rows.add(key)
                        rows.append(
                            {
                                "lang": lang,
                                "target": f"{f}::{name}",
                                "ref": f"{loc.get('relativePath', '?')}:{r.line}:{r.character}",
                                "container": cs.get("name"),
                                "container_kind": int(cs.get("kind", -1)),
                            }
                        )
    return sorted(rows, key=lambda row: json.dumps(row, sort_keys=True))


def main() -> None:
    # usage: ab_probe2.py OUT.json LANG REPO_PATH [FILE_FILTER]
    out_path, lang, repo_path = sys.argv[1], sys.argv[2], sys.argv[3]
    file_filter = sys.argv[4] if len(sys.argv) > 4 else None
    try:
        rows = probe(lang, repo_override=repo_path, file_filter=file_filter)
    except Exception as e:
        rows = [{"lang": lang, "fatal": f"{type(e).__name__}: {e}"[:300]}]
    Path(out_path).write_text(json.dumps({lang: rows}, indent=1, sort_keys=True))
    print(f"[probe] {lang}: {len(rows)} rows -> {out_path}", file=sys.stderr)


if __name__ == "__main__":
    main()
```

</details>

<details>
<summary><b>Corpus: rust — src/rich_probe.rs (add `mod rich_probe;` to src/main.rs)</b></summary>

```rust
//! Rich attribution probe: adverse container shapes for reference attribution.

/// (d) enum whose variants are referenced from initializers and bodies
#[derive(Clone, Copy, Debug, PartialEq)]
pub enum Shape {
    Circle,
    Square,
    Rect,
}

/// (d) const group
pub const UNIT_RADIUS: f64 = 1.0;
pub const BASE_SIZE: f64 = 2.0;

/// helper type used in fields, signatures and bodies
pub struct Label(pub String);

/// (a) KEY ADVERSE CASE: multi-line static data blob whose initializer
/// references the enum (3x) and both constants (3x total)
pub static TABLE: &[(Shape, f64)] = &[
    (Shape::Circle, UNIT_RADIUS),
    (Shape::Square, BASE_SIZE),
    (Shape::Rect, BASE_SIZE),
];

/// (b) struct with typed fields referencing other types
pub struct Registry {
    pub primary: Shape,
    pub scale: f64,
    pub label: Label,
    pub history: Vec<Shape>,
}

/// (c) trait with method signatures referencing types
pub trait ShapeStore {
    fn get_shape(&self, idx: usize) -> Shape;
    fn put_label(&mut self, label: Label) -> Shape;
}

/// impl block with an inherent method
impl Registry {
    pub fn describe(&self) -> Label {
        Label(format!("{:?}@{}", self.primary, self.scale))
    }
}

/// impl block for the trait
impl ShapeStore for Registry {
    fn get_shape(&self, idx: usize) -> Shape {
        TABLE[idx % TABLE.len()].0
    }

    fn put_label(&mut self, label: Label) -> Shape {
        self.label = label;
        self.primary
    }
}

/// (f) one-liner top-level item referencing a type outside any admitted container
pub type ShapeList = Vec<Shape>;

/// (e) control: a normal function body referencing all of the above;
/// these references must stay attributed to `probe_all` on both sides
pub fn probe_all() -> usize {
    let s = Shape::Circle;
    let mut r = Registry {
        primary: s,
        scale: BASE_SIZE + UNIT_RADIUS,
        label: Label(String::from("probe")),
        history: TABLE.iter().map(|(shape, _)| *shape).collect(),
    };
    let picked = r.get_shape(1);
    r.put_label(Label(String::from("again")));
    let list: ShapeList = r.history.clone();
    let described = r.describe();
    (picked == Shape::Square) as usize + list.len() + described.0.len()
}
```

</details>

<details>
<summary><b>Corpus: cpp — rich_probe.cpp (repo root)</b></summary>

```cpp
// rich_probe.cpp -- reference-attribution probe corpus (standalone, no includes needed)

namespace probe {

// (d) const group
constexpr int kMaxEntries = 4;
constexpr int kMinEntries = 1;
constexpr int kBase = 10;

// (d) enum whose member initializers reference constants (adverse case: refs inside enum body)
enum Flags {
    FlagA = kBase,
    FlagB = kBase + 1,
    FlagC = kBase + 2,
};

// (d) scoped enum
enum class Color {
    ColRed,
    ColGreen,
    ColBlue,
};

// (b) struct with typed fields referencing other types (adverse case: refs inside struct body)
struct Config {
    Color color;
    Flags flags;
    int limit;
    const char* label;
};

// (c) interface-like abstract class; method signatures reference types
class IWidget {
public:
    virtual Config makeConfig(Color color) const = 0;
    virtual ~IWidget() {}
};

// (a) KEY ADVERSE CASE: multi-line static const data structure whose initializer
// references types, enum members and constants (>= 3 references inside initializer)
static const Config kTable[] = {
    {Color::ColRed, FlagA, kMaxEntries, "red"},
    {Color::ColGreen, FlagB, kMinEntries, "green"},
    {Color::ColBlue, FlagC, kMaxEntries + kMinEntries, "blue"},
};

// (f) control: one-liner top-level variable referencing kTable and Config
static const unsigned long kTableSize = sizeof(kTable) / sizeof(Config);

// control: multi-line method inside a struct (innermost-container check)
struct Helper {
    int compute(Color c) const {
        return static_cast<int>(c) + kMaxEntries;
    }
};

// (e) control: normal function body referencing the above (must stay attributed here)
int useAll(const IWidget& widget) {
    Config cfg = widget.makeConfig(Color::ColGreen);
    Helper helper;
    int total = cfg.limit + kMaxEntries + helper.compute(cfg.color);
    for (unsigned long i = 0; i < kTableSize; ++i) {
        if (kTable[i].color == cfg.color) {
            total += kTable[i].limit + static_cast<int>(kTable[i].flags);
        }
    }
    return total;
}

}  // namespace probe
```

</details>

<details>
<summary><b>Corpus: typescript — rich_probe.ts (repo root)</b></summary>

```typescript
// Rich attribution probe: several container shapes referencing shared symbols.

export class ServiceA {
    ping(): string {
        return "A";
    }
}

export class ServiceB {
    ping(): string {
        return "B";
    }
}

// (d) enum
export enum ProbeMode {
    Fast = "fast",
    Slow = "slow",
}

export const DEFAULT_LABEL = "default";

// (c) interface with typed members and a method signature referencing types
export interface ProbeConfig {
    primary: ServiceA;
    secondary: ServiceB;
    mode: ProbeMode;
    describe(target: ServiceA): string;
}

// (a) multi-line const data structure: >=3 references inside the initializer
export const routes = {
    alpha: ServiceA,
    beta: ServiceB,
    mode: ProbeMode.Fast,
    label: DEFAULT_LABEL,
};

// (b) class with typed fields referencing other types
export class Registry {
    primary: ServiceA = new ServiceA();
    secondary: ServiceB = new ServiceB();
    mode: ProbeMode = ProbeMode.Slow;
}

// multi-line const arrow function whose body references the interface
export const describeConfig = (config: ProbeConfig): string => {
    const label = `${DEFAULT_LABEL}:${config.mode}`;
    return label + config.describe(config.primary);
};

// (e) control: plain function body referencing the above
export function runProbe(): string {
    const registry = new Registry();
    const config: ProbeConfig = {
        primary: registry.primary,
        secondary: registry.secondary,
        mode: ProbeMode.Fast,
        describe(target: ServiceA): string {
            return target.ping();
        },
    };
    return describeConfig(config) + routes.label;
}

// (f) top-level reference outside any container
runProbe();
```

</details>

<details>
<summary><b>Corpus: java — RichProbe.java (beside existing sources, same package)</b></summary>

```java
package test_repo;

import java.util.Map;

/**
 * Probe file exercising reference-container attribution across danger shapes:
 * multi-line constant initializers, typed fields, interface signatures, enums,
 * and a normal method body as control.
 */
public class RichProbe {

    // (a) multi-line static constant whose initializer references types and other constants
    static final Map<String, ProbeHandler> TABLE = Map.of(
            ProbeDefaults.ALPHA, new EchoHandler(),
            ProbeDefaults.BETA, new NullHandler(),
            ProbeDefaults.GAMMA, new EchoHandler());

    // (a') multi-line array-initializer constant referencing enum members
    static final ProbeStatus[] ORDER = {
            ProbeStatus.READY,
            ProbeStatus.RUNNING,
            ProbeStatus.DONE,
    };

    // (e) control: normal method body referencing the types, constants and enum above
    public String run(String key) {
        ProbeHandler handler = TABLE.getOrDefault(key, new NullHandler());
        ProbePayload payload = new ProbePayload(key, ProbeStatus.READY, handler);
        if (ProbeDefaults.ALPHA.equals(key)) {
            return handler.handle(payload);
        }
        return ORDER[0].name() + handler.handle(payload);
    }
}

// (c) interface with method signatures referencing a class and an enum
interface ProbeHandler {
    String handle(ProbePayload payload);

    ProbeStatus statusOf(ProbePayload payload);
}

// (b) class with typed fields referencing other types
class ProbePayload {
    final String key;
    final ProbeStatus status;
    final ProbeHandler origin;

    ProbePayload(String key, ProbeStatus status, ProbeHandler origin) {
        this.key = key;
        this.status = status;
        this.origin = origin;
    }
}

// (d) enum
enum ProbeStatus {
    READY,
    RUNNING,
    DONE
}

// (d') const group: class of string constants used inside TABLE's initializer
final class ProbeDefaults {
    static final String ALPHA = "alpha";
    static final String BETA = "beta";
    static final String GAMMA = "gamma";

    private ProbeDefaults() {
    }
}

class EchoHandler implements ProbeHandler {
    @Override
    public String handle(ProbePayload payload) {
        return payload.key + ":" + payload.status;
    }

    @Override
    public ProbeStatus statusOf(ProbePayload payload) {
        return payload.status;
    }
}

class NullHandler implements ProbeHandler {
    @Override
    public String handle(ProbePayload payload) {
        return "";
    }

    @Override
    public ProbeStatus statusOf(ProbePayload payload) {
        return ProbeStatus.DONE;
    }
}
```

</details>

<details>
<summary><b>Corpus: kotlin — RichProbe.kt (beside existing sources)</b></summary>

```kotlin
package test_repo

/**
 * Probe fixture: exercises reference attribution for containers of various kinds.
 */

// (c) interface with method signatures referencing types
interface ProbeRenderer {
    fun render(unit: ProbeUnit): String
    fun label(kind: ProbeKind): String
}

// (b) data class with typed fields referencing other types
data class ProbeUnit(
    val kind: ProbeKind,
    val weight: Int,
)

// (d) enum class
enum class ProbeKind(val code: Int) {
    ALPHA(1),
    BETA(2),
}

// (a) companion object with a MULTI-LINE constant whose initializer references
// types and other constants (>=3 references inside the initializer)
class ProbeRegistry {
    companion object {
        const val DEFAULT_WEIGHT = 3

        val TABLE = listOf(
            ProbeUnit(ProbeKind.ALPHA, DEFAULT_WEIGHT),
            ProbeUnit(ProbeKind.BETA, DEFAULT_WEIGHT + 1),
        )
    }

    fun lookup(kind: ProbeKind): ProbeUnit? = TABLE.firstOrNull { it.kind == kind }
}

// (e) normal function body referencing the above (control)
fun describeProbe(unit: ProbeUnit): String {
    val kindName = unit.kind.name
    return "unit kind=" + kindName + " weight=" + unit.weight
}

// (f) top-level declaration whose initializer references types outside any
// function/class container (control)
val DEFAULT_UNIT = ProbeUnit(ProbeKind.ALPHA, ProbeRegistry.DEFAULT_WEIGHT)

fun probeReport(): String = describeProbe(DEFAULT_UNIT)
```

</details>

<details>
<summary><b>Corpus: scss — _rich_probe.scss (repo root)</b></summary>

```scss
@use "variables" as vars;

// (d) const-group analog: several single-line tokens.
$probe-color-a: #112233;
$probe-color-b: #445566;
$probe-color-c: #778899;
$probe-radius: 6px;

// (a) multi-line map constant whose initializer references other constants
// (>= 3 references inside the initializer).
$probe-theme: (
    primary: $probe-color-a,
    secondary: $probe-color-b,
    accent: $probe-color-c,
    surface: vars.$color-bg,
);

// (b) struct analog: placeholder selector with declarations referencing tokens.
%probe-card-base {
    border-radius: $probe-radius;
    color: $probe-color-a;
}

// (c) interface analog: mixin whose signature defaults reference tokens.
@mixin probe-panel($bg: $probe-color-b, $accent: $probe-color-c) {
    background: $bg;
    border: 1px solid $accent;
}

// (e) control: function body referencing the map above.
@function probe-pick($key) {
    @return map-get($probe-theme, $key);
}

// (e) control: rule using the mixin, placeholder, function and variables.
.probe-widget {
    @extend %probe-card-base;
    @include probe-panel($bg: $probe-color-a);
    outline-color: $probe-color-c;
    background-color: probe-pick(primary);
    padding: vars.$space-md;
}

// (f) control: top-level reference outside any container.
@debug $probe-radius;
```

</details>

opcode81:
Thanks for your contribution.

PR #1832, opened 2026-08-09
Force the Nextflow server's deferred workspace scan before finding references

vagkaratzas:
Follow-up to #1815. On a fresh server, `find_referencing_symbols` on any Nextflow symbol returned an empty list for the first seconds of a session.

### Cause

The Nextflow language server does not scan the workspace when it appears to. `LanguageService.initialize` — which the `"Initializing"` progress notification we currently wait for wraps — only marks the workspace as unscanned and clears the AST cache. The scan itself happens in `LanguageService.update0`, behind a 1s debounce, and only on a round that finds no pending file change; a round that finds one re-defers it via `updateLater`:

```java
if( !scanned ) {
    if( uris.isEmpty() ) { uris = getWorkspaceFiles(); astCache.clear(); this.scanned = true; }
    else { updateLater(); }
}
```

The first `didOpen` of a session therefore pushes the scan out by at least one further round. A references request issued right after it sees an AST cache holding nothing but the file it just opened, and answers with `[]`.

### Fix

Force the scan instead of waiting for it. `completion` is one of the two requests that call `updateNow` before consulting their provider, and `DebouncingExecutor.executeNow` cancels the pending debounce and runs the update synchronously on the request thread, so the response is not sent until that round has finished. Two rounds suffice whatever the workspace size: the first drains the pending file changes — which is what re-defers the scan — and the second finds nothing pending, so it takes the `getWorkspaceFiles` branch and compiles the whole workspace before replying.

Polling the workspace symbol index instead would be unsound: `LanguageService.symbol` neither awaits the update nor holds the monitor that `update` synchronises on. On a generated 1050-file workspace, polling it made the server raise `java.util.ConcurrentModificationException` from inside the scan, non-deterministically (6, 3 and 0 occurrences across three runs).

### Verification

Generated workspace of 1000 modules + 50 workflows, where the target process has exactly 100 references (50 includes + 50 calls), with only the defining file ever opened:

| | references found | time |
|---|---|---|
| without the fix | 0 / 100 | 2.2s |
| with the fix | 100 / 100 | 2.5s |

Also verified on a real 68-file pipeline (3/3 references, cold).

### Why the existing test missed it

`test_find_references_across_files` passed because the `language_server` fixture is module-scoped: earlier tests in the class had already opened `main.nf` and forced the server to compile it, so the assertion held for a reason that does not apply to a real session. Run alone it failed:

```
pytest test/solidlsp/nextflow/test_nextflow_basic.py -k test_find_references_across_files
→ 1 failed
```

The added `test_find_references_across_files_on_a_fresh_server` starts its own server and opens nothing but the defining file. It fails on the pre-fix baseline and passes with the fix; the isolated original test now passes too.

I have also reported the underlying server behaviour upstream at nextflow-io/language-server#175.

### Checks

`poe lint`, `poe type-check` and `pytest test/solidlsp/nextflow/` (7 passed) all clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,12 +46,6 @@ Status of the `main` branch. Changes prior to the next official version change w
   - **Add support for Nextflow** (language server `nextflow`), using the official
     [Nextflow language server](https://github.com/nextflow-io/language-server); the JAR is downloaded
     automatically, a Java 17+ runtime is required
-  - `nextflow`: Fix: `find_referencing_symbols`/`request_references` returned an empty list for any
-    symbol during the first seconds of a session. The server defers its workspace scan behind a 1s
-    debounce and re-defers it whenever a file change is pending, so the very first `didOpen` pushed the
-    scan out past the reference request, which then saw an AST cache holding only the file it had just
-    opened. Cross-file requests now force the scan via the server's own synchronous update path, so they
-    are correct regardless of workspace size.
   - Add `python_basedpyright` as an alternative Python language server
   - Java/JDTLS: stop downloading and loading the unused IntelliCode completion-ranking bundle; the retired
     `intellicode_version`, `intellicode_xmx` and `intellicode_xms` settings remain accepted but are ignored #1821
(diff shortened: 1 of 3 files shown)


MischaPanch:
Thanks! Please remove the entry from the changelog (the support and fix come in the same release, no need for two entries) and shorten the docstring of the new method - the reason for its existence can be described much more briefly. Then this can be merged

vagkaratzas:
Done; just the `native` test is failing now, but I think due to request timeouts?

opcode81:
Thanks!

vagkaratzas:
Thanks @opcode81  :D

PR #1839, opened 2026-08-11
Add Bazel/Starlark language support via starpls

TiberiusNemesis:
Adds Bazel/Starlark support using [starpls](https://github.com/withered-magic/starpls) v0.1.22. Single static binary, auto-downloaded with pinned SHA256s (osx/linux arm64+x64, win-x64), so there's no CI install step. You don't need bazel either: main-workspace `//...` labels resolve on their own. With bazel on PATH (or `ls_specific_settings.starlark.bazel_path`) you also get `@repo//...` resolution.

Matches BUILD, BUILD.bazel, MODULE.bazel, WORKSPACE and friends, plus `.bzl` and `.star`. The extensionless names needed a new exact-filename mode in `FilenameMatcher` -- suffix matching alone would also hit stuff like `PREBUILD`. Unit tests in `test/solidlsp/test_ls_config.py`.

Test repo lives in `test/resources/repos/starlark/test_repo`, with 10 integration tests: document symbols, go-to-definition across files through `load()`, references, diagnostics. ~4s locally on macOS. I point starpls at a nonexistent bazel path in the tests, otherwise its `bazel info` probe makes bazelisk download a whole Bazel release on the runners (60-90s of flake for zero value).

One caveat: find-references in starpls 0.1.22 is single-file only. Cross-file navigation still works via go-to-definition. Noted in the docs page and the class docstring.

`bazel-*` output symlinks get skipped during traversal, since the scanner follows symlinks and those point straight into the output base.

Also touched: README language list, docs page, CHANGELOG, `project.template.yml`, and the symbolic-tools hook now counts extensionless BUILD/WORKSPACE as code files.

## Checklist


MischaPanch:
Hi, sorry for no reaction on our side, this slipped our radar. It is an addition we would merge if it's ok for you to sign the CLA.

PR #1842, opened 2026-08-12
SolidLSP: correct the position returned by insert_text_at_position

xormania:
`insert_text_at_position` returned a wrong end position whenever the insertion **merged with an
  adjacent character into a single newline sequence**. The resulting text was always correct — only
  the reported position was wrong.

  | call *(abbreviated)* | resulting text | returned | correct |
  |:--|:--|:--:|:--:|
  | `insert("\r", 1, 0, "\n")` | `"\r\n"` | `(2, 0)` | `(1, 0)` |
  | `insert("x\ry", 1, 0, "\n")` | `"x\r\ny"` | `(2, 0)` | `(1, 0)` |
  | `insert("a\r", 2, 0, "X")` | `"a\r\nX"` | `(2, 1)` | `(1, 1)` |
  | `insert("\r\n", 0, 1, "Z")` | `"\rZ\n"` | `(0, 2)` | `(1, 1)` |

  <sub>Row 3 reaches the fault through the one-line-past-EOF branch, which prepends a `\n` that merges with the trailing `\r`. Row 4 is the converse — inserting *inside* a `\r\n` splits it, and `(0, 2)` names a column on a line the split had just
  emptied.</sub>

  ---

  ## Why

  `_get_updated_position_from_line_and_column_and_edit` stepped the **inserted text in isolation**,
  then added the resulting delta to the start position.

  9208a3ae centralised line/column computation on `TextStepper` and adopted the LSP newline
  definition (`\n`, `\r\n`, `\r`). This helper was rebased onto `TextStepper` in that commit but kept
  its original shape — it steps one string, and that string is the *insertion*, not the *result*.
  Once a newline can be formed by two adjacent characters, no computation over the inserted text
  alone can be correct.

  ## What changed

  - **Position is derived from the resulting text** via `get_line_col_from_index` — the same function
    that resolves positions everywhere else, so it is correct by construction.
  - **The helper is deleted.** It was the last computation in `TextUtils` not going through the
    centralised path, and had no other callers. The past-EOF `line`/`col` adjustment that existed
    only to feed it goes with it.
  - **The column is clamped to the text length** before the splice and the derivation. Not a new
    tolerance: `text[:i]` with `i > len(text)` already yielded the whole text, so an out-of-range
    column has always appended. The clamp keeps position and splice on the same index — without it,
    deriving from the result would newly *raise* on input that previously worked.

  > [!NOTE]
  > **Out of scope:** LSP columns are UTF-16 code units, and Serena consumes them as Python string
  > indices with no conversion at the boundary — several adapters declare
  > `general.positionEncodings: ["utf-16"]` while nothing in the core converts. That is a real and
  > separate defect. This PR only keeps the returned position consistent with the text the function
  > produces. Happy to open an issue for it.

  ## Tests

  `test/solidlsp/test_text_utils.py`, seven cases:

  | case | pins | fails before |
  |:--|:--|:--|
  | boundary merge | row 1 | yes |
  | mid-document merge | row 2 | yes |
  | past-EOF merge | row 3 | yes |
  | newline split | row 4 | yes |
  | out-of-range column | clamping | yes |
  | column beyond a non-BMP line | clamping | yes |
  | `\n`-terminated insert | unchanged behaviour | no — control |

  Full CI matrix green on this commit in my fork: 13 test jobs across Linux, Windows and macOS, plus
  CodeQL, codespell, docs build and the Python-version matrix.

  <sub>Found by a property test comparing `TextUtils` against an independently written implementation of the LSP newline rules over a newline-dense alphabet — which also caught a regression in the clamp.</sub>

  ## Checklist


opcode81:
Correct me if I'm wrong, but isn't the problem being solved here completely theoretical?
If a file is using "\r" as a line separator, then nobody will append "\n" to a position after such a separator. It's just not going to happen.

opcode81:
I will probably still merge this, as it is a clean solution, but you didn't actually run into this problem, did you?

xormania:
Fair - I was working on adding Fuzz checks - and it appeared

xormania:
I intend to continue with the fuzz check ideas - it's a don't know that you don't know sort of test - can't hurt? I didn't want to get ahead of myself, but I didn't want to PR without some evidence beforehand.

MischaPanch:
It doesn't hurt, like the other PR handling a misbehavior of our search tool when a directory was called `**`. We just live in fun times of agentic adversarial reviews coming up with evermore unrealistic failure modes

xormania:
@MischaPanch - wow, you nailed it. Thank you. Damn agents. The key is getting them to realize it - i have some ideas.

PR #1843, opened 2026-08-12
VB.Net Support

dibble-james:
## Checklist

### DIFF
--- a/src/solidlsp/language_servers/vbnet_language_server.py
+++ b/src/solidlsp/language_servers/vbnet_language_server.py
@@ -6,19 +6,19 @@ import logging
 import os
 import platform
 import threading
-from collections.abc import Iterable
+from collections.abc import Hashable, Iterable
 from pathlib import Path
 from typing import Any, cast
 
 from overrides import override
 
 from serena.util.dotnet import DotNETUtil
-from solidlsp.ls import DocumentSymbols, LanguageServerDependencyProvider, LSPFileBuffer, SolidLanguageServer
+from solidlsp.ls import LanguageServerDependencyProvider, LSPFileBuffer, RawDocumentSymbol, SolidLanguageServer
 from solidlsp.ls_config import LanguageServerConfig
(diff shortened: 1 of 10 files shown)


src/solidlsp/language_servers/vbnet_language_server.py
```diff
@@ -0,0 +1,536 @@
...
+                    self._original_symbol_names[cache_key] = original_name
+
+            if type_info and "detail" not in symbol:
+                symbol["detail"] = type_info  # type: ignore
+
+        return normalized_name
+
+    @staticmethod
+    def _extract_base_name_and_type(roslyn_name: str) -> tuple[str, str]:
+        if " : " in roslyn_name and "(" not in roslyn_name:
```
AmirF194: Ran the extraction logic from this file against a few inputs in a clean `python:3.11-slim` container, head `7532a8a`.

`_extract_base_name_and_type`'s property branch guards on `"(" not in roslyn_name` against the whole raw string, not just the identifier before ` : `. VB.NET tuple types use literal parens (`Dim x As (Integer, String)`), so a tuple-typed property or field like `Position : (Integer, String)` trips that guard and falls into the method branch instead, which corrupts the name:

```
>>> extract_base_name_and_type("Name : String")
('Name', ': String')
>>> extract_base_name_and_type("Add(x As Integer, y As Integer) : Integer")
('Add', '(x As Integer, y As Integer) : Integer')
>>> extract_base_name_and_type("Position : (Integer, String)")
('Position :', '(Integer, String)')
```

`find_symbol` defaults to exact name-path matching, so a lookup for `Position` would return nothing once this fires, and no test here exercises a tuple-typed member. Splitting on ` : ` first and only searching for `(` in the remainder would cover the tuple case without changing the two paths above.

AmirF194: Confirmed against `d9b82e39`, the split-then-check-parens order handles the tuple return case too.

PR #1848, opened 2026-08-12
Fix: surface a mid-indexing tsserver crash instead of returning {}

### LINKED ISSUE #1814: `find_referencing_symbols` silently returns `{}` after tsserver dies of OOM — reported as "cross-file indexing complete", `isError: false`
## Summary

When `tsserver` crashes with a V8 heap OOM (`SIGABRT`) during the first cross-file
reference query on a large TypeScript monorepo, Serena does not surface the crash.
It logs **`TypeScript cross-file indexing complete`**, returns `Result: {}`, and sets
`isError: false`.

Every subsequent `find_referencing_symbols` call then returns an empty result in
~0.8s with no error. From the MCP client's perspective this is indistinguishable
from "this symbol genuinely has no references."

For an LLM agent this is the worst possible failure shape: fast, confident, and
wrong. An agent consuming these results will conclude dead code is safe to delete,
or that a refactor has no downstream callers.

## Environment

| | |
|---|---|
| Serena | 1.6.1 (also reproduced on the 1.28.1 MCP server build) |
| typescript-language-server | 5.1.3 |
| TypeScript | 6.0.3 (workspace), tsserver bundled |
| Python | 3.13 (uv tool install) |
| OS | macOS 26.x, Apple M1 Max, 64 GB RAM |
| Repo | elastic/kibana (~190k TS/TSX files) |
| Context | `ide-assistant` |

Host memory was **not** the constraint: ~95% free, 0 swap, ~60 GB available at
time of crash. This is the default V8 old-space cap (~4 GB), not host pressure.

## Reproduction

1. Activate a project on a very large TS monorepo (Kibana; a single plugin
   subdirectory is enough — `x-pack/solutions/security/plugins/security_solution`).
2. Call `get_symbols_overview` on any file — succeeds, cheap.
3. Call `find_referencing_symbols` for a real exported symbol in that file.
4. Watch the tsserver child process RSS climb past ~4 GB and abort.

Observed process RSS across the crash (tree sampled every 15s):

```
[  6.4s] post-warmup(LSP program)   tree_rss=  658.9 MB   4 node procs
[ 16.4s] idle+10s                   tree_rss= 1903.4 MB   one node proc at 1417 MB
[ 26.5s] idle+20s                   tree_rss= 3352.7 MB   one node proc at 2870 MB
[ 36.5s] idle+30s                   tree_rss= 4647.6 MB   one node proc at 4165 MB
[ 40.7s] after 5 ref queries        tree_rss=  193.5 MB   <-- tsserver gone
```

## Actual log output

```
INFO  Task-3:FindReferencingSymbolsTool ... find_referencing_symbols: name_path='AGENT_BUILDER_LAST_AGENT_ID_STORAGE_KEY', ...
INFO  TypeScript LSP progress [646a4e19...]: started - Initializing JS/TS language features…
INFO  LSP: window/logMessage: {'type': 1, 'message': '[lspserver] [tsclient] [tsserver] Exited. Code: null. Signal: SIGABRT'}
INFO  TypeScript LSP progress [646a4e19...]: ended -
INFO  _wait_for_cross_file_references_if_needed:545 - TypeScript cross-file indexing complete   <-- WRONG
INFO  tools_base:task:410 - Result: {}
INFO  Task-3:FindReferencingSymbolsTool completed in 28.870 seconds
```

macOS crash report for the child confirms V8 heap OOM, not an external kill:

```
exception: {'type': 'EXC_CRASH', 'signal': 'SIGABRT'}
termination: {"code": 6, "namespace": "SIGNAL", "indicator": "Abort trap: 6", "byProc": "node"}
  abort
  node::OOMErrorHandler(char const*, v8::OOMDetails const&)
  v8::internal::V8::FatalProcessOutOfMemory(...)
  v8::internal::Heap::FatalProcessOutOfMemory(char const*)
```

## Root cause

`src/solidlsp/language_servers/typescript_language_server.py` (line numbers below
are from current `main`, verified after the post-v1.6.1 refactor of this file).

Indexing completion is inferred purely from `$/progress` token bookkeeping:

- `progress_handler(kind="begin")` adds the token and clears `_indexing_complete`.
- `progress_handler(kind="end")` discards the token, and when the active set is
  empty it **sets `_indexing_complete`** (L430-436):

```python
elif kind == "end":
    msg = value.get("message", "")
    log.info(f"TypeScript LSP progress [{token}]: ended - {msg}")
    with self._progress_lock:
        self._active_progress_tokens.discard(token)
        if not self._active_progress_tokens:
            self._indexing_complete.set()      # <-- also reached on crash teardown
```

When tsserver aborts, typescript-language-server emits `kind="end"` for the
in-flight token as part of tearing the connection down. That is structurally
identical to a successful completion, so the event is set and
`_wait_for_indexing_start_or_completion` returns `True` (L546-558):

```python
def _wait_for_cross_file_references_if_needed(self) -> None:
    ...
    if self._wait_for_indexing_start_or_completion(timeout=timeout, start_grace=start_grace):
        log.info("TypeScript cross-file indexing complete")   # reached after SIGABRT
```

Note the `else` branch here *does* warn on timeout — but a crash never reaches it,
because a crash looks like completion rather than a timeout.

The crash *is* already observed — it arrives via `window/logMessage` with
`type: 1` (Error) — but `window_log_message` only logs it (L383-384):

```python
def window_log_message(msg: dict) -> None:
    log.info(f"LSP: window/logMessage: {msg}")
```

So the signal is received and discarded, and a `kind="end"` triggered by process
death is treated as success.

This is the same failure class as #1681 (pyright: premature "ready" → silent
empty results), which was fixed for the pyright backend; the TypeScript backend
reaches an equivalent false-green by a different route.

## Suggested fix

Three independent hardening options; (1) alone removes the false-green.

1. **Treat tsserver exit as a failure state.** In `window_log_message`, match
   `[tsserver] Exited` / `type == 1` and set a `_server_crashed` flag. Have
   `_wait_for_cross_file_references_if_needed` (and the symbol-retrieval path)
   raise instead of returning `{}` when that flag is set, so the tool reports
   `isError: true`.
2. **Liveness assertion before returning an empty result.** If a reference query
   returns zero results, verify the language-server child process is still alive
   before reporting the empty set as authoritative. Empty-because-dead and
   empty-because-no-refs must not be the same value.
3. **Distinguish `kind="end"` on teardown from real completion** — e.g. don't set
   `_indexing_complete` when the token is discarded as part of a connection
   error/exit.

## Additional suggestion: raise the default heap

The default Node heap (~4 GB) is not sufficient for cross-file references on
monorepos of this size. With `NODE_OPTIONS=--max-old-space-size=20480` the same
queries succeed and return real results.

Measured steady state on the same host after the bump (60 queries across 12 files):

| Scope | Peak RSS | Plateau | Median latency | Real results |
|---|---|---|---|---|
| Single plugin | 11.5 GB | 7.5 GB | 1.0 s | 44/60 |
| Repo root | 9.4 GB | 8.6 GB | 14.0 s | 57/60 |

Serena could either default `--max-old-space-size` higher for the TS backend, or
warn when a project exceeds some file-count threshold without it set. Right now
the failure is silent and the user has no signal that they need the flag.

## Note on scoping (separate, possibly worth its own issue)

Narrowing the activated project to a subdirectory is the natural advice for
memory, but it **changes correctness**: 13 of 60 identical (symbol, file) targets
returned empty when scoped to the plugin directory and returned real references
at repo root — e.g. `mockDataTableModel` `{}` vs 21125 bytes,
`demoTimelineData` `{}` vs 69542 bytes. All 44 targets that returned results in
both scopes returned *more* references at repo root.

If subdirectory activation is expected to under-report cross-project references,
that's worth documenting explicitly, since the failure mode is again a silent
empty result rather than a warning.

AmirF194:
## Root cause

`_wait_for_indexing_start_or_completion` (and its underlying `wait_for_indexing`) infer
project-indexing completion purely from `$/progress` token bookkeeping: when the active
token set drains, `_indexing_complete` is set and the wait returns `True`. When `tsserver`
aborts (e.g. a V8 heap OOM mid cross-file query), `typescript-language-server` sends a
`$/progress` "end" for the in-flight token as part of tearing its connection down, which
is structurally identical to a normal completion. `_wait_for_cross_file_references_if_needed`
then logs "TypeScript cross-file indexing complete" and `find_referencing_symbols` returns
`{}` with `isError: false`, indistinguishable from a symbol that genuinely has no references.

The crash *is* already observed independently, via `window/logMessage` (`type: 1`/Error,
`"[tsserver] Exited. Code: null. Signal: SIGABRT"`), but the handler only logged it.

## Fix

Wire that already-received signal into a `_crash_message` flag (set from
`window_log_message` via a small `_tsserver_exit_message` classifier), and raise a new
`TypeScriptServerCrashedError` from every path that would otherwise report indexing as
complete: `wait_for_indexing`, and both early-return branches of
`_wait_for_indexing_start_or_completion` (not just its tail call into `wait_for_indexing`,
since a crash observed within the same poll interval as its own `$/progress` begin/end can
take either early-return branch). This covers the Vue and Svelte companion TypeScript
servers too, since both hold a `TypeScriptLanguageServer` instance and inherit
`window_log_message` via `_start_server`.

## Verification

- `test/solidlsp/test_typescript_timeout_policy.py::TestTsserverCrashDetection` (new): drives
  the reporter's exact `$/progress begin` -> crash `window/logMessage` -> `$/progress end`
  sequence and pins that `wait_for_indexing`, both branches of
  `_wait_for_indexing_start_or_completion`, and `_wait_for_cross_file_references_if_needed`
  all now raise `TypeScriptServerCrashedError` instead of returning `True`/`{}`; also pins
  that an ordinary completion (no crash message) is unaffected.
- Full `test_typescript_timeout_policy.py` suite (29 tests, all pre-existing + new): passes
  in a clean `python:3.11-slim` container.
- `ruff format --check`, `ruff check`, `ty check` (pinned `ruff==0.12.5`, matching CI): clean
  on both changed files.
- Not verified: the actual `window/logMessage` handler registered inside `_start_server`
  against a real `tsserver` crash, since that requires a live language-server process. The
  classification logic it delegates to (`_tsserver_exit_message`) is unit-tested directly.

Fixes #1814

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@ Status of the `main` branch. Changes prior to the next official version change w
     indicate it was incomplete. Serena now declares work-done progress support and waits for the work
     Metals reports, bounded by the new `indexing_timeout`, `indexing_start_grace` and
     `indexing_quiet_period` settings
+  - Fix: a `tsserver` crash mid-indexing (e.g. a V8 heap OOM) sent the same `$/progress` "end"
+    event as a normal completion, so `find_referencing_symbols` and other cross-file queries
+    silently returned an empty result instead of surfacing the crash. The crash is now detected
+    independently via the `window/logMessage` notification tsserver already sends, and the
+    affected wait now raises instead of reporting success (#1814)
 
 * Dependencies:
   - Remove the redundant `dotenv` dependency; the `dotenv` module is provided by `python-dotenv`
--- a/src/solidlsp/language_servers/typescript_language_server.py
+++ b/src/solidlsp/language_servers/typescript_language_server.py
@@ -4,6 +4,7 @@ Provides TypeScript specific instantiation of the LanguageServer class. Contains
 
 import logging
 import os
+import re
 import shutil
 import threading
 import time
@@ -15,7 +16,9 @@ from sensai.util.logging import LogTime
 from solidlsp import ls_types
 from solidlsp.ls import LanguageServerDependencyProvider, LanguageServerDependencyProviderSinglePath, SolidLanguageServer
 from solidlsp.ls_config import LanguageServerConfig
+from solidlsp.ls_exceptions import SolidLSPException
 from solidlsp.ls_utils import PlatformId, PlatformUtils
+from solidlsp.lsp_protocol_handler.lsp_types import MessageType
 from solidlsp.settings import SolidLSPSettings
 
 from .common import RuntimeDependency, RuntimeDependencyCollection, build_npm_install_command
@@ -63,6 +66,21 @@ def prefer_non_node_modules_definition(definitions: list[ls_types.Location]) -> 
         if rel_path and "node_modules" not in rel_path:
             return d
     return definitions[0]
+
+
+class TypeScriptServerCrashedError(SolidLSPException):
+    """Raised when tsserver reported its own abnormal exit via window/logMessage.
+
+    typescript-language-server sends a $/progress "end" event for the in-flight
+    token as part of tearing its connection down after tsserver dies, which is
+    otherwise indistinguishable from a normal indexing completion.
+    """
+
+
+# Matches typescript-language-server's window/logMessage notification for an abnormal
+# tsserver exit, e.g. "[lspserver] [tsclient] [tsserver] Exited. Code: null. Signal: SIGABRT".
+# A clean shutdown does not produce this message.
+_TSSERVER_EXITED_PATTERN = re.compile(r"\[tsserver\]\s+Exited\b", re.IGNORECASE)
 
 
 class TypeScriptLanguageServer(SolidLanguageServer):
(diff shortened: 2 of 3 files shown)


AmirF194:
Checking in, no rush. This fixes tsserver OOM silently reporting isError: false with an empty result instead of surfacing the crash, happy to adjust scope if you'd rather handle it differently.

MischaPanch:
Thanks, merging! Our schedules are very full these days, so PRs take longer to review

AmirF194:
Understood, thanks for merging. Appreciate you making time for it.

PR #1849, opened 2026-08-12
Tests: property-based coverage for the TextUtils newline handling

xormania:
[`9208a3ae`](https://github.com/oraios/serena/commit/9208a3ae) established that a newline in this codebase is `\n`, `\r\n` or `\r` — the LSP
definition — and deliberately *not* what `str.splitlines()` considers a line break. That contract
is currently pinned by a single fixture string, `"012\n012\r\n012\r012"`, with about four
assertions per conversion direction.

The contract covers a combination space: three newline sequences × empty lines × an optional
trailing newline × the characters `splitlines` breaks on but the LSP does not. One fixture samples
a few points of it. These four properties cover the space.

## Properties

| # | property | asserts |
|:-:|:--|:--|
| 1 | position → index → position | a valid `(line, col)` survives the round trip unchanged |
| 2 | index → position → index | unchanged — **except** an index pointing *inside* a `\r\n`, which maps to the start of the following line: the edge case `get_line_col_from_index` documents |
| 3 | `split_lines` | with endings, rejoins into the original text; without them, yields the line contents |
| 4 | out-of-range index | is rejected |

Each asserts an exact value over generated input — no mocks, no skips, no conditional assertions.

## What the generator does, and why it matters

- **Newlines are drawn explicitly.** Unstructured random text essentially never produces `\r\n`,
  so a naive generator would never reach the cases the contract is about.
- **Line content includes the characters `str.splitlines()` breaks on and the LSP does not** —
  `\v`, `\f`, the separator controls, NEL, `U+2028`, `U+2029`. This is the distinction [`9208a3ae`](https://github.com/oraios/serena/commit/9208a3ae)
  drew, and it is the part no existing test covers. **A `splitlines`-based reimplementation of
  `TextUtils` passes the current suite and fails these properties** — measured, see the table below.

The generator excludes exactly one combination, whose line structure would otherwise be ambiguous:
an `\r` separator, an empty line and an `\n` separator together form a single `\r\n`.

## Why a dependency rather than more parametrized cases

Because the contract is a *product*, not a list. Three newline sequences, at each of up to six line
boundaries, times empty/non-empty lines, times the `splitlines`-only break characters, times every
index in the result. Enumerating that by hand is what the current fixture attempts and why it
samples so thinly; adding twenty more `@pytest.mark.parametrize` rows samples it slightly less
thinly and hard-codes my guesses about which combinations matter.

The properties instead state the invariant once and let the generator search the space — and
shrink any counterexample to its minimal form, which is what made the #1842 defect legible in
seconds rather than as a wall of failing cases.

Cost: one exactly-pinned development dependency plus `sortedcontainers`, no production surface, and
sub-second runtime.

## What the properties catch that the current tests do not

I applied six regressions a refactor could plausibly introduce, one at a time, and ran both suites
against each:

| regression | existing tests | properties |
|:--|:--:|:--:|
| `split_lines` delegates to `str.splitlines()` | **passes** | fails |
| an index inside `\r\n` maps to the *previous* line | **passes** | fails |
| the out-of-range index check is removed | **passes** | fails |
| only `\n` is treated as a newline (the pre-[`9208a3ae`](https://github.com/oraios/serena/commit/9208a3ae) rule) | fails | fails |
| a bare `\r` is not treated as a newline | fails | fails |
| `get_index_from_line_col` is off by one past line 0 | fails | fails |

Three of the six survive the suite as it stands. The `splitlines` row follows directly from the
fixture: `"012\n012\r\n012\r012"` contains digits and the three LSP newlines and nothing else, so
no assertion over it can distinguish `TextStepper` from `str.splitlines()` — the two agree on that
input. The generator adds precisely the characters on which they disagree.

These are hand-picked plausible regressions rather than exhaustive mutation testing, and none of
them survive both suites — so this is a floor on what the properties add, not a full accounting.

## CI behaviour

Languages get disabled outright here when they flake, so the settings are deliberate:

| setting | effect |
|:--|:--|
| `derandomize=True` | identical inputs every run — a failure is reproducible from the report alone, never "it passed on retry" |
| `deadline=None` | no per-example timing assertion, the main flake source on the slower runners |
| `database=None` | no example database carried between runs |
| `max_examples=200` | four properties in well under a second |

Hypothesis still creates `.hypothesis/` for its Unicode category cache; `.gitignore` already covers
that.

No new module, no new marker, no workflow change — the file sits in the `catch-all` batch, so this
runs on Windows too, where `\r\n` is native. `hypothesis` is pinned exactly, matching the
surrounding development dependencies, and brings one transitive dependency, `sortedcontainers`.

> [!NOTE]
> For provenance: this generator is also what surfaced the position defect in #1842. The properties
> that pin *that* fix are deliberately held back — they would fail against current `main` — so this
> PR stands on its own and the two can merge in either order.

## Two things about the diff

**`uv.lock` is larger than "two packages added".** The package inventory change is exactly
`hypothesis` + `sortedcontainers` with **no version change to anything existing** — but the lock
also drops ~40 redundant environment markers (per-dependency `python_full_version` guards on
`ipython`, `cffi`, `clr-loader` and friends). Each of those sits inside a block that already
carries the equivalent `resolution-markers`, so they were no-ops; a newer `uv` normalises them
away. Behaviour-neutral at `requires-python = ">=3.11, <3.15"`.

**No `CHANGELOG.md` entry.** This adds no feature and fixes no problem — it is tests plus a
development-only dependency, invisible to `serena-agent` users — and upstream test-only commits
([`7392d65b`](https://github.com/oraios/serena/commit/7392d65b), [`4bd60f18`](https://github.com/oraios/serena/commit/4bd60f18), [`3a8393bc`](https://github.com/oraios/serena/commit/3a8393bc)) do not add one.

## Checklist

- [x] For changes that add features or fix problems, I have added an entry to `CHANGELOG.md`, which concisely describes the change. *(Not applicable — see above.)*

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,6 +75,7 @@ dev = [
   # In version 1.0.4 we get a NoneType error related to some config conversion (yml_analytics is None and should be a list)
   "ty==0.0.24",
   "poethepoet==0.36.0",
+  "hypothesis==6.165.3",
   "pytest==8.4.1",
   "pytest-xdist==3.8.0",
   "ruff==0.12.5",
--- a/test/solidlsp/test_text_utils.py
+++ b/test/solidlsp/test_text_utils.py
@@ -1,6 +1,62 @@
+from dataclasses import dataclass
+
 import pytest
+from hypothesis import given, settings
+from hypothesis import strategies as st
 
 from solidlsp.ls_utils import InvalidTextLocationError, TextStepper, TextUtils
+
+# Characters that Python's str.splitlines() treats as line breaks but the LSP does not
+# (vertical tab, form feed, file/group/record separators, NEL, line/paragraph separator).
+# TextStepper must treat them as ordinary line content.
+_NON_LSP_LINE_BREAKS = "\v\f\x1c\x1d\x1e\x85\u2028\u2029"
+
+_LINE_CONTENT = st.text(
+    st.sampled_from(list(_NON_LSP_LINE_BREAKS)) | st.characters(exclude_categories=("Cs",), exclude_characters="\r\n"),
+    max_size=6,
+)
+
+# Deterministic so that any failure is reproducible, and without per-example timing assertions,
+# which would be the primary source of flakiness on the slower CI runners.
+_PROPERTY = settings(derandomize=True, deadline=None, database=None, max_examples=200)
+
+
+@dataclass(frozen=True)
+class _GeneratedText:
+    """A generated text together with the line contents from which it was assembled."""
+
+    text: str
+    """the assembled text, using the newline sequences the LSP defines"""
+    lines: list[str]
+    """the newline-free content of each line; a text ending in a newline has an empty final line"""
+
+
+def _assemble_text(contents: list[str], separators: list[str]) -> _GeneratedText:
+    r"""
+    Joins the given line contents with the given newline sequences, adjusting any separator that
+    would combine with its predecessor: an "\r" separator, an empty line and an "\n" separator
+    together form a single "\r\n" newline, which would render the line structure unknown.
+
+    :param contents: the newline-free content of each line
+    :param separators: the newline sequence following each line but the last
+    :return: the assembled text together with the line contents it is composed of
+    """
+    adjusted: list[str] = []
+    for index, separator in enumerate(separators):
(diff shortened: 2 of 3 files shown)


xormania:
this might not be worth it - but it has me thinking, which can't be bad.

PR #1859, opened 2026-08-13
Scala: wait for Metals' reported work, not a fixed 5 seconds

### LINKED ISSUE #1858: Scala: the first `find_referencing_symbols` of a session returns partial results, because Serena waits a fixed 5s rather than for Metals to finish indexing and compiling
Preconditions:

Issue details:

## Setup

Serena 1.7.1.dev0 (`main` at 5107b11f), Python 3.13.5, Linux aarch64 (devcontainer), Java 25.0.4 (Temurin), Metals 1.6.4 (Serena's default) and 1.6.8, Bloop 2.1.1. Language: Scala. No client — driven through `solidlsp` directly, so this is not MCP-specific. No configuration adjustments beyond those shown.

## What happens

`ScalaLanguageServer` waits a fixed 5 seconds after the first `didOpen` before its first cross-file query, and never waits again:

```python
# src/solidlsp/language_servers/scala_language_server.py:577-579
@override
def _get_wait_time_for_cross_file_referencing(self) -> float:
    return 5
```

On a cold Metals, 5 seconds is long before it is ready. Metals imports the build, indexes it, and then has the build server compile the project — and references are served from SemanticDB, which is written *by that compilation*. A query issued in between returns some of the references rather than none, which is the awkward part: there is no error, no empty result, and nothing to distinguish a partial answer from a complete one.

## Reproduction 1 — Serena's own Scala test fixture

Against `test/resources/repos/scala`, with no `.metals`, `.bloop`, `.bsp` or `target` present and no Bloop daemon running:

```python
import os, time

REPO = os.path.abspath("test/resources/repos/scala")
REL = "src/main/scala/com/example/Utils.scala"          # line 3, col 6 = `def printHello`
    print("first call:", len(ls.request_references(REL, 3, 6)))
    time.sleep(30)
    print("+30s:      ", len(ls.request_references(REL, 3, 6)))
```

Expected: both calls report the one reference to `Utils.printHello`, from `Main.scala`.

Actual:

```
first call: 0
+30s:       1
```

## Reproduction 2 — a real project

Same procedure on a 726-file Scala 3 sbt multi-module build (Metals 1.6.8, cold `.metals`, `.bloop` present), asking for the references to a widely-used object:

```
first call (Serena's own wait only): 124 references  (t+22s)
after 30s:                           242 references  (t+53s)
after 90s:                           242 references
after 150s:                          242 references
```

The first answer is half the truth, and looks exactly like the whole truth.

## Why the wait cannot be a constant

Metals already reports what it is doing, as LSP work-done progress. Driving it with a raw LSP client that declares `window.workDoneProgress` and polls references alongside gives the timeline below (cold start, same project as Reproduction 2, repository path replaced with `<repo>`):

```
 0.0s  didOpen
 1.7s  metals/status  [module] 'importing...'
 2.0s  log            Connected to Build server: Bloop v2.1.1
 2.1s  $/progress     begin  "Importing build"
 2.3s  $/progress     end    "Importing build"
 3.6s  $/progress     begin  "Indexing"                    <- references = 0
19.7s  $/progress     end    "Indexing"
19.7s  metals/status  [metals] '🚀  Indexing complete!'
20.0s  $/progress     begin  "Compiling bootstrap", "Compiling common-test"
21.0s  $/progress     begin  "Compiling automation"
22.0s  $/progress     begin  "Compiling api"
28.8s  $/progress     end    "Compiling bootstrap", "Compiling common-test"
                                                            <- references = 78 at 30.0s
40.0s  $/progress     end    "Compiling api"                (the last one)
46.1s  metals/status  [metals] ''  hide=true
                                                            <- references = 319 at 46.5s, stable after
```

Two things follow. Readiness is not a fixed period — it is however long the project takes to compile. And it is not the point Metals calls "Indexing complete" either: that lands at 19.7s, while the compilations that produce the references run to 40s.

(The absolute counts in that trace are from the raw client, which was given the repository root as its workspace folder and does no filtering, so they are larger than the 242 Serena reports for the same symbol. It is the shape over time that matters, not the totals.)

Serena receives none of this today. `_create_base_initialize_params` declares only `textDocument.documentSymbol`, so no `$/progress` ever arrives — the notifications above appear only once a client asks for them. (`metals/status` does arrive, and is dropped as `Unhandled method`; Serena sends `statusBarProvider: "false"`, which is not one of Metals' four accepted values — `off`, `on`, `log-message`, `show-message` — so it is not treated as "off".)

## Analysis

The shape that fits the timeline is the one TypeScript already uses in this repo — track the progress tokens, and have the first cross-file query wait for them to drain rather than for a wall-clock period ([#1639](https://github.com/oraios/serena/pull/1639), which added `_pre_open_for_cross_file_references` and `_wait_for_cross_file_references_if_needed` for exactly this).

One wrinkle showed up in testing: a plain "wait until no tokens are outstanding" returns too early against Metals, because Metals hands off between its phases rather than overlapping them, so its token set empties for around a second between the import ending and indexing beginning. A query that resumes in that gap is no better off than the 5-second one.

## Fix

A PR is incoming: declare `window.workDoneProgress`, track the tokens, and wait for them (with a bounded timeout, and a short quiet period so the gap above is not mistaken for completion). On the project in Reproduction 2 the first `find_referencing_symbols` then returns all 242 references, taking 32s instead of 22s.

## Disclosure

Prepared with the assistance of Claude Code (Claude Opus). I reproduced both cases myself, checked the source citations and the timeline against my own logs, and edited the text. I will answer follow-ups personally.

merlinorg:
Fixes #1858, where the reproduction, the timeline and the source trail are.

In short: `ScalaLanguageServer` waited a fixed 5s after the first `didOpen` before its first cross-file query. On a cold Metals that lands long before it has imported the build, indexed it, and had the build server compile — and references come from SemanticDB, which that compilation writes. A query in between returns *some* of the references rather than none, with nothing to mark it partial. On a 726-file Scala 3 sbt build the first `find_referencing_symbols` returned 124 of 242.

Metals reports all of this as work-done progress. Serena never asked for it: `_create_base_initialize_params` declared only `textDocument.documentSymbol`, so no `$/progress` ever arrived.

## Fix

Declare `window.workDoneProgress`, track the tokens, and have the first cross-file query wait for them — the same event-based shape as TypeScript's indexing wait (#1639), using the `_pre_open_for_cross_file_references` / `_wait_for_cross_file_references_if_needed` hooks it added. `_get_wait_time_for_cross_file_referencing` goes, since nothing reaches it any more.

The bookkeeping lives in a small `MetalsProgressTracker` rather than on the server object, so it can be tested without Metals.

Readiness is "nothing outstanding for a moment", not "nothing outstanding". Metals hands off between its phases rather than overlapping them, so its token set empties for about a second between the import ending and indexing beginning; a plain drain returns in that gap and is no better off than the 5s. That was not a hypothesis — it is what the first version of this patch did, and the end-to-end run caught it. `indexing_quiet_period` (default 3s) is how long silence has to last to count.

Three settings, all under `ls_specific_settings.scala`, all bounded so a pathological project cannot hang a session:

| Setting | Default | |
|---|---|---|
| `indexing_timeout` | `180` | how long to wait once work is known to be outstanding; on expiry the query proceeds and a warning names what was still running |
| `indexing_start_grace` | `15` | how long to wait for Metals to report anything at all — a server that reports nothing is taken to have nothing to do, so this degrades to a short wait rather than a hang |
| `indexing_quiet_period` | `3` | the silence that counts as finished |

## Trade-off worth your opinion

The first cross-file query of a session now costs as much as the project takes to compile: 32s instead of 22s on the build above. Every later one is unaffected. I think that is the right default — the alternative is a fast wrong answer that reads as a right one — but it is a behaviour change on a shared path and you may want it opt-in, or the timeout lower. Say the word and I will flip it.

## Testing

Sixteen unit tests in `test/solidlsp/scala/test_scala_indexing_progress.py`, on the tracker rather than the server, so they need no Metals and run in CI: the drain, work that starts within the grace, a server that reports nothing, the timeout, `window/workDoneProgress/create` arriving before `begin`, unbalanced and repeated `end`s, the settings and their validation, and — the regression test for the bug above — a phase handover that must not end the wait.

Each guard was checked by breaking it: ending the wait on the first drain fails the handover test; ignoring `workDoneProgress/create` fails the create-before-begin test; setting idle on any `end` fails two.

End-to-end, cold `.metals`, same project and Metals version:

```
before: first find_referencing_symbols -> 124 references at t+22s (242 only from a later call)
after:  first find_referencing_symbols -> 242 references at t+32s
```

And on this repo's own `test/resources/repos/scala` fixture, `0` before and `1` after.

## Not included

No change to how Metals is launched or versioned, and none to the `metals/status` notifications, which do reach Serena and are dropped as `Unhandled method`. Serena sends `statusBarProvider: "false"`, which is not one of Metals' four accepted values (`off`, `on`, `log-message`, `show-message`) — worth tidying, but `$/progress` is the richer signal and it seemed better not to mix the two in one PR.

## Disclosure

We actively use this patched version of Serena on multiple systems.

Prepared with the assistance of Claude Code (Claude Opus). I reproduced the problem and the fix myself, read the diff, and edited the text. I will answer follow-ups personally.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 Status of the `main` branch. Changes prior to the next official version change will appear here.
 
+* Language Servers:
+  - Fix: Scala cross-file queries waited a fixed 5s after the first file was opened, which on a cold
+    Metals is long before its build import, indexing and compilation have finished; the first
+    `find_referencing_symbols` of a session could return a fraction of the references with nothing to
+    indicate it was incomplete. Serena now declares work-done progress support and waits for the work
+    Metals reports, bounded by the new `indexing_timeout`, `indexing_start_grace` and
+    `indexing_quiet_period` settings
+
 # v1.7.0 (2026-08-09)
 
 * General:
--- a/docs/02-usage/050_configuration.md
+++ b/docs/02-usage/050_configuration.md
@@ -1112,6 +1112,13 @@ build below the repository root — it is the build roots, not the repository ro
 given. Serena detects them automatically; `project_roots` overrides that detection where it guesses
 wrong, and `project_root_scan_depth` bounds how far it looks.
 
+Metals reports its build import, indexing and compilation as LSP work-done progress, and Serena waits
+for all of it before the first cross-file query of a session. This matters most for references, which
+Metals serves from SemanticDB — a file the build server only writes once it has compiled the sources,
+well after indexing ends — so a query made too early returns a fraction of the true result with
+nothing to say it is partial. The wait is bounded by `indexing_timeout`, after which the query
+proceeds against whatever Metals has so far and a warning is logged.
+
 Supported settings:
 
 | Setting | Default | Description |
@@ -1123,6 +1130,9 @@ Supported settings:
 | `auto_import_build` | `true` | Answer Metals' build-import prompts affirmatively, which lets it run the project's build tool (e.g. `sbt bloopInstall`). Set to `false` to leave the build un-imported; Metals then has no build server, and every cross-file query is served by the fallback presentation compiler. |
 | `project_roots` | auto-detected | The build roots to serve, as paths relative to the repository root. A path that does not exist is skipped with a warning; if none of them exists, the build roots are detected instead. |
 | `project_root_scan_depth` | `3` | How many directory levels below the repository root the detection searches. Applies whenever the roots are detected — that is, when `project_roots` is unset, or when it names nothing that exists. |
+| `indexing_timeout` | `180` | How long to wait, in seconds, for Metals to finish importing, indexing and compiling before the first cross-file query. On expiry the query proceeds and a warning names what was still outstanding. |
+| `indexing_start_grace` | `15` | How long to wait, in seconds, for Metals to report any work at all. A server that reports none within this window is taken to have nothing to do. |
+| `indexing_quiet_period` | `3` | How long, in seconds, Metals must report nothing for its work to count as finished. Metals hands off between its phases rather than overlapping them, so it reports nothing for a moment in between; a shorter period risks mistaking that gap for completion. |
 
 #### SCSS / Sass / CSS
 
--- a/docs/03-special-guides/scala_setup_guide_for_serena.md
+++ b/docs/03-special-guides/scala_setup_guide_for_serena.md
@@ -88,6 +88,26 @@ ls_specific_settings:
 ```
 
 ---
+## Waiting for Metals to be ready
+
+Metals needs its build imported, its index built and the project compiled before it can answer a
+cross-file question completely — references in particular come from SemanticDB, which the build
+server writes only as it compiles. Serena waits for all of that, tracking the work-done progress
+Metals reports, before the first such query of a session; a query made earlier would return a
+fraction of the true result and look no different from a complete one.
+
+The first `find_referencing_symbols` of a session therefore takes as long as the project takes to
+compile. Subsequent queries do not wait. Where the default bound is wrong for a large build:
+
+```yaml
+ls_specific_settings:
+  scala:
+    indexing_timeout: 180        # seconds to wait before giving up and answering anyway
+    indexing_start_grace: 15     # seconds to wait for Metals to report anything at all
+    indexing_quiet_period: 3     # seconds of silence that count as "finished"
+```
+
+---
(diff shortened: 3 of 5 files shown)


opcode81:
This looks good. Thanks for you contribution!

PR #1863, opened 2026-08-13
fix(dart): handle analyzer status notifications

### LINKED ISSUE #1855: Unhandled method '$/analyzerStatus' warnings
I get a lot of the following warnings in the log and when running `serena project index`:

`solidlsp.ls_process:_notification_handler:474 - Unhandled method '$/analyzerStatus'`

I had Claude investigate and here's what it said:

> It's harmless noise from the Dart language server — not a real problem. 
> 
> **What it is**: $/analyzerStatus is a Dart Analysis Server–specific LSP notification carrying {"isAnalyzing": true|false}. It's how the Dart server tells an editor "I started / finished an analysis pass" so IDEs can show the spinning "Analyzing…" indicator. The Dart server sends it as a fallback when the client doesn't advertise window.workDoneProgress support — which Serena's LSP client (solidlsp) doesn't.
> 
> **Why the $/**: Per the LSP spec, $/-prefixed messages are explicitly implementation-defined, and clients are allowed to ignore any they don't recognize. So Dart is behaving correctly, and so is Serena.
> 
> **Why the warning**: solidlsp dispatches incoming notifications against a table of handlers it registered; anything not in the table hits the fallback branch at ls_process.py:474 and gets logged at WARNING. It's a "I don't have a handler for this" log line, not an error — nothing failed, no symbol lookup was affected. The bursts of 4–7 identical lines within milliseconds are just the analysis server toggling status as it processes files.
> 
> **What to do**: Nothing functionally. If it clutters your log window, bump log_level in ~/.serena/serena_config.yml:89 from 20 to 40 (errors only) — though you'd lose genuine warnings too. A proper fix belongs upstream in Serena (register a no-op handler for $/analyzerStatus, or downgrade unknown $/ notifications to DEBUG since the spec says they're ignorable).

Just relaying this in case it's something you want to tackle.

jiatianzhao:
## Problem

The Dart analysis server emits `$/analyzerStatus` notifications while it analyzes a workspace. Serena's Dart adapter did not register this server-specific notification, so the generic notification dispatcher logged every event as an unhandled-method warning. This creates repeated warning noise even though the notification does not require action from Serena.

## Cause

The Dart adapter already registers no-op handlers for other informational notifications such as `$/progress` and `language/status`, but `$/analyzerStatus` was missing from that adapter-specific list. The global dispatcher therefore treated it as genuinely unknown.

## Fix

Register `$/analyzerStatus` with the Dart adapter's existing no-op handler. Global handling of unknown notifications is unchanged, so unexpected methods from other servers continue to produce warnings.

Add a regression test that starts the Dart adapter against an in-memory LSP interface, sends `$/analyzerStatus` through the production notification dispatcher, and verifies that the unhandled-method warning is not emitted.

## Verification

- Regression test fails on `main` with `Unhandled method '$/analyzerStatus'` and passes with this change.
- `uv run poe test -m dart -q --tb=short`: 22 passed, 2 skipped.
- `uv run poe lint`: passed.
- `uv run poe type-check`: passed.
- `uvx --from codespell codespell CHANGELOG.md src/solidlsp/language_servers/dart_language_server.py test/solidlsp/dart/test_dart_startup.py`: passed.

## Compatibility

No public API, configuration, dependency, or global notification behavior changes.

Fixes #1855

## Disclosure

Prepared with assistance from OpenAI Codex. I reviewed the implementation and test design and ran the verification above.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Status of the `main` branch. Changes prior to the next official version change w
     project list in `serena_config.yml`
 
 * Language Servers:
+  - Fix: Dart's `$/analyzerStatus` notifications were logged as unhandled-method warnings during analysis (#1855)
   - Fix: Scala cross-file queries waited a fixed 5s after the first file was opened, which on a cold
     Metals is long before its build import, indexing and compilation have finished; the first
     `find_referencing_symbols` of a session could return a fraction of the references with nothing to
--- a/src/solidlsp/language_servers/dart_language_server.py
+++ b/src/solidlsp/language_servers/dart_language_server.py
@@ -192,6 +192,7 @@ class DartLanguageServer(SolidLanguageServer):
         self.server.on_notification("window/logMessage", window_log_message)
         self.server.on_request("workspace/executeClientCommand", execute_client_command_handler)
         self.server.on_notification("$/progress", do_nothing)
+        self.server.on_notification("$/analyzerStatus", do_nothing)
         self.server.on_notification("textDocument/publishDiagnostics", do_nothing)
         self.server.on_notification("language/actionableNotification", do_nothing)
         self.server.on_notification("experimental/serverStatus", check_experimental_status)
--- a/test/solidlsp/dart/test_dart_startup.py
+++ b/test/solidlsp/dart/test_dart_startup.py
@@ -1,0 +1,75 @@
+"""Regression test for Dart analyzer-status notification handling."""
+
+import logging
+from pathlib import Path
+from typing import Any
+from unittest.mock import patch
+
+import pytest
+
+from solidlsp.language_servers.dart_language_server import DartLanguageServer
+from solidlsp.ls_config import LanguageServerConfig, LanguageServerId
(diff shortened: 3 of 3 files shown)


opcode81:
Precondition for merging: Test this with an actual project using the Dart language server, verifying that 
* the state prior to the change produces the undesirable log message
* the new state does not.

jiatianzhao:
I verified this with the real Dart language server using Serena's pinned Dart SDK 3.7.1 on macOS arm64.

I ran the same four-file Dart project with:

`serena project index <project> --ls dart --log-level WARNING --timeout 30`

At the base commit (`7067f566`), indexing completed successfully and emitted 18 warnings, including:

```text
WARNING ... solidlsp.ls_process:_notification_handler:474 - Unhandled method '$/analyzerStatus'
Indexed files per language: dart=4
```

At the PR head (`8042aaeb`), indexing completed successfully with the same result and no `$/analyzerStatus` warnings:

```text
Indexed files per language: dart=4
```

opcode81:
Thanks for your contribution!

PR #1866, opened 2026-08-14
fix(jetbrains): serialize IDE launches across concurrent Serena sessions

### LINKED ISSUE #1864: `jetbrains_launch_command` spawns one IDE per MCP session with no serialisation, so N concurrent sessions race a single IDE cold start and N-1 get a modal DirectoryLock$CannotActivateException
Preconditions:

- [x] I have consulted the user guide and verified that the issue cannot be resolved by adjusting configuration/following recommended workflows. (The only configuration lever is to unset `jetbrains_launch_command`, i.e. to turn the feature off — which is what I have done locally.)
- [x] I have looked for similar issues and discussions, including closed ones. Searched `jetbrains_launch_command`, `DirectoryLock`, `already running`, `concurrent sessions`, `multiple projects IDE`, `does not respond`, `IDE autolaunch`. #1539 added the setting, #1543 is the startup assert, #1585 is the `shell=True` injection, #1235 is duplicate Serena instances on one project. Nothing covers this.

---

## Summary

`_init_active_project_language_backend` launches the IDE whenever no plugin server serves the activating project's root, but the launch is unsynchronised and fire-and-forget. JetBrains IDEs are single-instance per config directory, and the socket that makes a second `pycharm <path>` invocation a *handoff* rather than a *boot* does not exist until the first instance has finished starting. Every launch that lands inside that window fails with a modal error dialog.

With one MCP client and one project this is invisible. With several agent sessions opening several repos at once — the normal case for a terminal multiplexer, or for any user who keeps more than one Claude Code / Codex session alive — it fires reliably.

## Environment

- Serena 1.7.0 (uv tool, Python 3.13), `--context=claude-code --project-from-cwd`
- macOS 15.6 (Darwin 25.6.0), Apple silicon
- PyCharm 2026.2 with the Serena plugin installed and licensed
- `language_backend: JetBrains`, `jetbrains_launch_command: /opt/homebrew/bin/pycharm`
- 8 Claude Code sessions in 8 different repositories, restored simultaneously by a terminal multiplexer

## Reproduction

1. Set `language_backend: JetBrains` and `jetbrains_launch_command: <IDE launcher>`.
2. With no IDE running, start N ≥ 2 MCP sessions in N different project directories at the same time.
3. Each session activates its project, finds no matching plugin server, and runs `<launcher> <projectRoot>`.

One instance wins. The others show:

```
com.intellij.platform.ide.bootstrap.DirectoryLock$CannotActivateException:
Process "/Applications/PyCharm.app/Contents/MacOS/pycharm" (50142) is still running and does not respond.
...
Suppressed: java.io.IOException: Cannot lock the config directory /Users/<u>/Library/Application Support/JetBrains/PyCharm2026.2
Caused by: java.nio.file.FileAlreadyExistsException: .../PyCharm2026.2/.lock
Suppressed: java.net.SocketException: No such file or directory
    at java.base/sun.nio.ch.UnixDomainSockets.connect0(Native Method)
```

## What is actually happening

From my own `~/.serena/logs`, two multiplexer restores on the same day:

```
2026-08-13 11:41:56,921  /opt/homebrew/bin/pycharm /Users/ismar/repos/Career
2026-08-13 11:41:56,921  /opt/homebrew/bin/pycharm /Users/ismar/repos/learning-opportunities
2026-08-13 11:41:56,924  /opt/homebrew/bin/pycharm /Users/ismar/repos/bootdotdev/my-ai-agent
2026-08-13 11:41:56,924  /opt/homebrew/bin/pycharm /Users/ismar/repos/detailobsessed/agent-skills
2026-08-13 11:41:56,938  /opt/homebrew/bin/pycharm /Users/ismar/dotfiles
2026-08-13 11:41:56,941  /opt/homebrew/bin/pycharm /Users/ismar/repos/detailobsessed/forks/copier-uv-bleeding
2026-08-13 11:41:56,944  /opt/homebrew/bin/pycharm /Users/ismar/repos/detailobsessed/ismar.ch
2026-08-13 11:41:56,988  /opt/homebrew/bin/pycharm /Users/ismar/repos
```

Eight launches inside 67 ms; a second identical burst inside 143 ms at 11:45:13; 17 launch events that day.

The two IDE-side files that decide the outcome:

- `~/Library/Application Support/JetBrains/PyCharm2026.2/.lock` — written almost immediately by the winning instance (it held `50142`, the PID Serena spawned).
- `~/Library/Caches/JetBrains/PyCharm2026.2/.port` — the unix socket a later invocation connects to in order to say "open this project in you". It appears only once the IDE is up.

Between those two moments the launcher can neither take the lock nor hand off, which is exactly the pair of exceptions in the trace. Once the IDE *is* up, the same command works correctly and opens the extra project — so the feature's design is sound; only its concurrency is not.

## Contributing factors in `agent.py`

```python
except jetbrains_plugin_client.ServerNotFoundError:
    if self.serena_config.jetbrains_launch_command:
        cmd = subprocess_util.convert_shell_cmd([self.serena_config.jetbrains_launch_command, project.project_root])
        p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
        stdout, stderr = p.communicate()
        if p.returncode != 0:
            log.error(f"Failed to launch JetBrains IDE: {stderr.decode('utf-8')}")
```

1. **No cross-process serialisation.** Each MCP server decides independently, so the decision is made N times against the same IDE config directory.
2. **No wait for the plugin server.** `p.communicate()` returns when the *launcher script* exits, not when the IDE and its plugin are serving. Even in the single-session happy path the first `jet_brains_*` tool call after activation therefore raises `ServerNotFoundError`, whose message is "STOP. Do not attempt any other tools or workarounds. Ask the user to open this folder as a project in a JetBrains IDE" — telling the agent to give up on an IDE Serena has just launched and which will be ready in ~20 s.
3. **The failure is invisible to Serena.** The error is a GUI modal; the launcher's exit status does not reflect it, so nothing is logged and the agent has no idea the IDE it asked for is not coming.

## Workaround in place

`jetbrains_launch_command` is now empty on this machine, which disables the feature entirely — `if self.serena_config.jetbrains_launch_command:` is falsy, so nothing spawns. The cost is that `jet_brains_*` tools fail with the "ask the user to open this folder as a project in a JetBrains IDE" message in any repo not already open in the IDE. Happy to run diagnostics or test a change against this setup.

AmirF194:
## Root cause

`_init_active_project_language_backend` (`src/serena/agent.py`) launched the configured
JetBrains IDE with a fire-and-forget `subprocess.Popen` whenever no plugin server matched the
activating project, independently in every Serena session. When several sessions activate
different projects at (nearly) the same time (the normal case for a terminal multiplexer, or
any user with more than one agent session alive), each decides independently that no server is
running and races `jetbrains_launch_command` against the same IDE. Only one instance can hold
the IDE's own config-directory lock; the rest fail with a native `DirectoryLock` modal, which
never surfaces as a subprocess exit code, so Serena has no plugin server and no indication why.

A second, related gap: `p.communicate()` only waits for the *launcher script* to exit, not for
the IDE and its plugin to actually be serving, so even the single-session happy path can hit
`ServerNotFoundError` for the tool call immediately following activation.

## Fix

`launch_coordinator.launch_and_wait_for_plugin_server` (new module) serializes launches on a
file lock keyed by `jetbrains_launch_command`, re-checking for a reachable plugin server both
before and after acquiring the lock, so a session that loses the race waits for the winner's
IDE instead of starting its own. After the launch command exits, it polls for the plugin server
to become reachable (bounded by a timeout) instead of returning immediately.

`_init_active_project_language_backend`'s JetBrains branch now delegates to this coordinator;
its control flow is otherwise unchanged.

## Verification

- New unit tests (`test/serena/test_jetbrains_launch_coordinator.py`, 8 cases) mock the plugin
  client and `subprocess.Popen` to exercise: skip-launch-if-already-served, launch-then-poll,
  wait-timeout-without-raising, launch-failure short-circuits the poll loop, lock-acquire-timeout
  gives up without launching, and a real-threads test with an actual `FileLock` on a temp path
  that reproduces the original race (two concurrent sessions, same launch command) and asserts
  exactly one `Popen` call. All pass (`uv run pytest test/serena/test_jetbrains_launch_coordinator.py -q`).
- Full `test/` suite run before and after the change (excluding language-server-toolchain tests
  this sandbox can't run): identical 47 pre-existing failures on both, all in language-server
  fixtures or GUI code unrelated to this change; no new failures.
- `poe lint`, `poe type-check`, and `codespell` on the changed files all pass.
- **Not verified**: the actual JetBrains `DirectoryLock`/`.port` behavior itself, which is
  macOS/IDE-specific and not reproducible in this Linux sandbox. The fix targets the two
  contributing factors from the issue's own analysis (no cross-process serialization, no wait
  for readiness) and is tested against a mock launcher and mock plugin-server client, not a
  real IDE.

Fixes #1864

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ Status of the `main` branch. Changes prior to the next official version change w
   - Fix: Parallel agents auto-registering projects could overwrite each other's changes to the global
     project list in `serena_config.yml`
 
+* JetBrains:
+  - Fix: Concurrent Serena sessions activating different projects at the same time with
+    `jetbrains_launch_command` set would each independently launch the IDE, racing each other for
+    the IDE's own config-directory lock; JetBrains IDE launches are now serialized per launch
+    command and Serena waits for the plugin server to become reachable before proceeding (#1864)
+
 * Language Servers:
   - Fix: Dart's `$/analyzerStatus` notifications were logged as unhandled-method warnings during analysis (#1855)
   - Fix: Scala cross-file queries waited a fixed 5s after the first file was opened, which on a cold
--- a/src/serena/agent.py
+++ b/src/serena/agent.py
@@ -41,7 +41,7 @@ from serena.config.serena_config import (
     ToolInclusionDefinition,
 )
 from serena.dashboard import SerenaDashboardAPI, SerenaDashboardTrayManager, SerenaDashboardViewer, open_url_in_browser
-from serena.jetbrains import jetbrains_plugin_client
+from serena.jetbrains import launch_coordinator as jetbrains_launch_coordinator
 from serena.ls_manager import LanguageServerManager
 from serena.memories.memory_manager import MemoryManager
 from serena.project import Project
@@ -1322,18 +1322,13 @@ class SerenaAgent:
 
         # for JetBrains mode, search for plugin server and spawn IDE (if not found and launch command provided)
         elif self.get_language_backend().is_jetbrains():
-            try:
-                client = jetbrains_plugin_client.JetBrainsPluginClient.from_project(project, log_warning=False)
+            client = jetbrains_launch_coordinator.find_plugin_server(project)
+            if client is not None:
                 log.info("Found Serena JetBrains Plugin server: %s", client)
-            except jetbrains_plugin_client.ServerNotFoundError:
+            else:
                 log.info("Serena JetBrains Plugin server not found for project %s", project.project_name)
                 if self.serena_config.jetbrains_launch_command:
-                    cmd = subprocess_util.convert_shell_cmd([self.serena_config.jetbrains_launch_command, project.project_root])
-                    log.info("Launching IDE with command: %s", cmd)
-                    p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
-                    stdout, stderr = p.communicate()
-                    if p.returncode != 0:
-                        log.error(f"Failed to launch JetBrains IDE: {stderr.decode('utf-8')}")
+                    jetbrains_launch_coordinator.launch_and_wait_for_plugin_server(project, self.serena_config.jetbrains_launch_command)
 
     def activate_project_from_path_or_name(
         self, project_root_or_name: str, update_active_modes: bool = True, update_active_tools: bool = True
--- a/src/serena/jetbrains/launch_coordinator.py
+++ b/src/serena/jetbrains/launch_coordinator.py
@@ -1,0 +1,119 @@
+"""
+Coordinates the launching of JetBrains IDE instances triggered by Serena's own
+`jetbrains_launch_command`, so that concurrent Serena sessions activating different
+projects at (nearly) the same time do not race each other into the same IDE launch.
+"""
+
+import hashlib
+import logging
+import subprocess
+import tempfile
+import time
+from pathlib import Path
+
+from filelock import FileLock, Timeout
+
+from serena.jetbrains import jetbrains_plugin_client
+from serena.project import Project
+from solidlsp.util import subprocess_util
+
+log = logging.getLogger(__name__)
+
+LOCK_ACQUIRE_TIMEOUT = 90.0
+"""
+How long a session waits for another session's IDE launch (for the same launch command) to
+finish before giving up on launching its own instance.
+"""
+
+PLUGIN_SERVER_WAIT_TIMEOUT = 60.0
+"""How long to poll for the plugin server to become reachable after the launch command exits."""
+
+PLUGIN_SERVER_POLL_INTERVAL = 1.0
+"""The interval at which we poll for the plugin server while waiting for it to come up."""
+
+
+def _lock_path_for_launch_command(launch_command: str) -> Path:
+    """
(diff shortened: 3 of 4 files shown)


AmirF194:
Saw your comment on the issue about not wanting to wait for plugin readiness by design. Worth flagging before this goes further: this PR's fix has two independent parts, serializing concurrent launches on a file lock, and polling for the plugin server after the launch command exits. The polling part is the one your comment pushes back on.

If the serialization alone is the wanted fix, I can drop the poll-for-readiness half and just have losers of the lock race wait for the winner's launch (still no explicit readiness wait, closer to the current single-launcher happy path). Let me know which shape you'd rather see, or if the poll is fine for the specific case here (readiness for a launch this PR itself triggered, not launch-then-idle-wait) versus whatever you were speaking to on the issue.

AmirF194:
Bumping this one, no pressure. Happy to adjust the approach if the concurrent-launch serialization isn't the direction you'd prefer.

opcode81:
Thanks for your contribution. This is a reasonable extension.

AmirF194:
Good, glad serialization alone was the right scope. Didn't want to force a readiness wait you'd already pushed back on.

PR #1870, opened 2026-08-14
Docs: Refined layout, optimized for useability and efficient loading

xormania:
**The whole of it is running here:** https://xormania.github.io/serena/ — the current documentation rebuilt, this branch together with the docs PRs that follow it, so the result can be judged as one site rather than as diffs.

This PR is the piece the rest stand on: the theme. It is deliberately the smallest reviewable step — everything else in that preview is content, and lands separately.

The measurable part: **the landing page drops from 1035.8 KB to 242.5 KB gzipped**, 32 requests either side. The rest of the change is what a number cannot carry — navigation that keeps its state, a layout that holds from a 320px phone to a wide desktop, controls that do what they say — itemised below.

One new directory and three lines of config, all inside `docs/`. `docs/_themes/serena/` **inherits** `sphinx-book-theme` — its templates extend the parent's blocks via `super()`, so asset digests stay the parent's business.

> [!NOTE]
> Deleting `docs/_themes/serena/` and the three `html_theme*` lines in `docs/_config.yml` restores stock behaviour exactly.

The shell is four layers deep — jupyter-book → sphinx-book-theme → pydata-sphinx-theme → Bootstrap 5 — and a few of its defaults work against a reader. None are Serena decisions; none are reachable from `_config.yml`.

## Weight

pydata-sphinx-theme delivers its icons twice — a webfont stylesheet in the head, the SVG-with-JS kit at the end of the body — and either one alone draws every icon on this site. The kit goes; so do the preloads for the two font faces no page draws a glyph from; the one face in use is replaced at the same path by a copy subset to the glyphs actually drawn (the regeneration command is a comment in `layout.html`). Per asset, landing page, gzipped:

| asset | before | after | change |
|:--|--:|--:|--:|
| FontAwesome SVG-with-JS kit (`all.min.js`) | 530.6 KB | — | -530.6 KB |
| `fa-solid-900.woff2` — subset to the 24 glyphs drawn | 152.8 KB | 2.5 KB | -150.3 KB |
| `fa-brands-400.woff2` | 115.1 KB | — | -115.1 KB |
| `fa-regular-400.woff2` | 24.8 KB | — | -24.8 KB |
| prefetch of the next page (`020_programming-languages.html`) | — | 10.1 KB | +10.1 KB |
| this theme's script (`serena-docs.js`) | — | 10.0 KB | +10.0 KB |
| this theme's stylesheet (`serena-docs.css`) | — | 7.6 KB | +7.6 KB |
| everything else (28 requests) | 212.4 KB | 212.2 KB | under 1 KB either way |
| **total (32 requests either side)** | **1035.8 KB** | **242.5 KB** | **-793.3 KB** |

A little of the saving is spent back on the reader: where the browser supports speculation rules, neighbouring pages load ahead of the click, so moving through the docs feels immediate.

Two template overrides, both minimal: `layout.html` filters the duplicate icon delivery out of the parent's own rendered blocks via `super()` rather than reproducing them, so anything the parent adds to those blocks keeps arriving untouched; `empty.html` exists because setting `secondary_sidebar_items` to nothing at all aborts the build with a pydata extension error.

<sub>Both sides measured the same way, at `93ec0431` (upstream main) and `9ca37e83` (this branch): the rev built clean in a throwaway worktree, the landing page loaded in Chromium, and every same-origin resource it requested gzipped on disk and summed — gzipped because that is what a static host sends.</sub>

## Moving between pages

| behaviour | stock | here |
|:--|:--|:--|
| a section row | link and expander at once — the label navigates, the chevron expands, a few pixels apart | an expander; the page it pointed at becomes the section's first child, so both actions remain, one control each |
| expand/collapse | resets on every navigation | persists both directions until the reader changes it |
| previous/next | flows with the content — on a long page, only at the end | pinned to the bottom of the window, in the page block's box so it centres on the page |

## On a wide screen

- The article floats mid-viewport because it is centred twice: by `.bd-container` around the three-column block, then by `.bd-main .bd-content`. Both left-aligned now; the article keeps a 72em measure — set by what the parameter tables need, and uniform across pages so the text edge does not shift between sections — and the page block its 88rem cap.
- Identifiers in table cells do not break mid-name, so columns are sized by their longest name and the description column takes the surplus.
- Both sidebars are `position: sticky` with their own `overflow-y` — three scroll regions. Static on desktop, so the page scrolls as one document.
- The right-hand "Contents" rail returns on demand: a pill at the content's right edge expands into the page's heading tree — hover on a pointer device, tap on touch, Escape everywhere — with the current section marked as the reader moves. It reads whatever structure a page has — section headings, or on an API-reference page the classes and members autodoc renders as signatures — and appears only where there are four or more entries to offer, so short pages stay clean, and collapsed it costs the article nothing.

## On a phone, or zoomed in

| what the reader hits | stock mechanism | here |
|:--|:--|:--|
| the nav panel's last entries sit below the fold, unreachable | sized in `vh`: on a phone that is the viewport *without* the browser's bars | sized in `dvh`, `vh` as fallback |
| the drawer shuts and reopens on every link | no state crosses a page load | stays open across pages, restored before first paint; Escape closes it, and that is remembered |
| one identifier pushes the page sideways | at 320px on `01-about/035_tools`, `jet_brains_find_referencing_symbol` renders 294px wide inside `<code><span class="pre">`; `white-space: pre` makes it unbreakable — 64px past the viewport | a wrap rule takes it to 0 |

The stored state is replayed only at drawer widths: from 992px up, sphinx-book-theme reads that same checkbox as "hide the sidebar", the inverse of its meaning below.

## Chrome

- The colour-scheme button cycles three states in one fixed order everywhere; the parent's order follows the OS setting. Its name says the current state and the next, not "light/dark" in all three.
- The fullscreen button is off: it duplicates the browser's own controls and fails silently where the API is unavailable.
- "Back to top" shows once the page is past the header, not only while scrolling upwards.
- "Download .md" pointed at the page source, which browsers display rather than save; it now carries `download`.
- A link that leaves the site opens in a new tab, with `rel="noopener"`; same-site links stay in this one, and an author's explicit `target` is respected.
- Motion is dropped under `prefers-reduced-motion`; links prefetch on hover where speculation rules exist.

## Verification

| check | result |
|:--|:--|
| `poe doc-build` (sphinx `-W`) | build succeeded, 0 warnings |
| `poe lint` | clean |
| landing page, gzipped — before | 1035.8 KB gzipped across 32 requests for /index at upstream 93ec0431 |
| landing page, gzipped — after | 242.5 KB gzipped across 32 requests for /index at 9ca37e83 |
| browser checks (chromium, webkit, firefox) | 15 skipped 141 passed (28.0s) |
| worst horizontal page overflow | 0px worst horizontal page overflow at 320 / 390 / 1280, over 3 page(s) |
| source changes | 8 files changed, 1468 insertions(+), 1 deletion(-) |

<sub>Generated by re-running each check at `9ca37e83`.</sub>

## Checklist


MischaPanch:
Hi @xormania . Not sure why you thought our docs needs revision. This destroys the content nav on the right that we have
image

The docs as linked above introduce a new subpage showing tool interfaces, which we deliberately don't expose as they are not part of the MCP contract (except for names) and I doubt that our users are all too interested in them. I also find the wider style and larger letters less appealing visually, and it would force us to maintain our own theme.

I think our docs (using the default jupyter book theme) are pretty ok and it's far from mission critical to decrease the page size. My instinct is to close this and to stop further work on docs unless the core team explicitly opens issues on that.

@opcode81 wdyt?

xormania:
@MischaPanch I first discovered Serena a couple months ago, i looked at the docs and decided it would be easier to just go straight to the code (no offense - although it did lead to me contributing). I have a fresh perspective (I'm new) and I feel that this is important to do. 

You should consider it - maybe not all of it - but there's a lot of benefits to be had from a rigorous documentation system. Both users and contributors face a steep learning curve as it stands. I believe Serena can be a contender - and this is something worth doing sooner than later. 

Its a start - and its flexible - it doesn't lock you into anything. Its all upside. And I'd be happy to get your input for improving it. It has a little bling to it, but not misplaced.

ps - the right nav is bad idea - the area is still available for other things, but don't put nav there (its bad design).

xormania:
@MischaPanch - check it out - i got the benefits of the that right nav back into it (just for you - and frankly i like it - good call to press me on it - thank you). Also yanked tools ref. And I got my newbie contrib wishlist added. 

remember that the preview site is multiple PRs - so its not tied to the new layout. I did a lot with the existing bootstrap doc setup to get it there (personally - i'm a tailwind/flowbite turbo kinda guy). 

What do you think?

PR #1872, opened 2026-08-15
fix(nextflow): don't mark workspace scan flushed when both probes fail

### LINKED ISSUE #1871: Nextflow LS: _flush_deferred_workspace_scan marks the scan flushed even when both flush attempts fail
`_flush_deferred_workspace_scan` (`src/solidlsp/language_servers/nextflow_language_server.py`) forces the Nextflow language server's deferred workspace scan by sending two `completion` requests and discarding their results, then unconditionally sets `self._workspace_scan_flushed = True` after the loop, even when both requests raised:

```python
for _ in range(2):
    try:
        self.server.send.completion(params)
    except Exception as e:
        log.debug("Completion request used to flush the Nextflow workspace scan failed: %s", e)
self._workspace_scan_flushed = True
```

`_send_references_request` checks that flag to skip flushing on later calls. If the workspace scan was never actually forced (both `completion` calls failed, for example because the server was still starting up or briefly unresponsive), the flag is set as if it had succeeded, and every later `find_referencing_symbols` call for the rest of the session silently answers from a possibly-incomplete workspace AST cache. There is no exception, no warning-level log, and no retry.

**Reproduction** (no real Nextflow JAR needed, since this is a pure state-tracking bug): instantiate `NextflowLanguageServer` via `object.__new__`, set `_workspace_scan_flushed = False` and a fake `server.send` whose `completion` always raises, then call `_flush_deferred_workspace_scan`. The flag flips to `True` after two failed attempts, and a second call makes zero further `completion` attempts.

- Serena: 1.7.1.dev0 (main @ 93ec0431)
- The bug is in the flag-setting logic itself and does not depend on OS, MCP client, or project language.

Introduced in #1832, which added the flush mechanism.

AmirF194:
`_flush_deferred_workspace_scan` sends two `completion` requests to force the server's debounced workspace scan, then sets `self._workspace_scan_flushed = True` unconditionally after the loop, even when both requests raised. `_send_references_request` uses that flag to skip flushing on later calls, so if the scan was never actually forced, every later `find_referencing_symbols` for the rest of the session silently answers from a possibly-incomplete workspace AST cache, with no retry.

Fix: only set the flag when at least one `completion` call did not raise; a total failure leaves it clear so the next call retries.

Verification:
- `test/solidlsp/test_nextflow_workspace_scan_flush.py` (new): drives `_flush_deferred_workspace_scan` against a fake `server.send` with no real language server process. Fails on `main` (flag flips `True` after two failures) and passes on this branch; a second regression test pins the existing success path.
- `ruff check`/`ruff format --check` and `codespell` clean on the changed files.
- Not verified against a real Nextflow JAR: the repro isolates the flag-setting logic itself, which doesn't depend on why the two `completion` calls would fail on a live server.

Fixes #1871

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ Status of the `main` branch. Changes prior to the next official version change w
     project list in `serena_config.yml`
 
 * Language Servers:
+  - Fix: Nextflow's `_flush_deferred_workspace_scan` marked the workspace scan flushed even when both
+    of its `completion` probes failed, permanently skipping the flush (and silencing retries) for the
+    rest of the session (#1871)
   - Fix: Dart's `$/analyzerStatus` notifications were logged as unhandled-method warnings during analysis (#1855)
   - Fix: Scala cross-file queries waited a fixed 5s after the first file was opened, which on a cold
     Metals is long before its build import, indexing and compilation have finished; the first
--- a/src/solidlsp/language_servers/nextflow_language_server.py
+++ b/src/solidlsp/language_servers/nextflow_language_server.py
@@ -320,12 +320,14 @@ class NextflowLanguageServer(SolidLanguageServer):
             "textDocument": {"uri": self._resolve_file_uri(relative_file_path)},
             "position": {"line": 0, "character": 0},
         }
+        flushed = False
         for _ in range(2):
             try:
                 self.server.send.completion(params)
+                flushed = True
             except Exception as e:
                 log.debug("Completion request used to flush the Nextflow workspace scan failed: %s", e)
-        self._workspace_scan_flushed = True
+        self._workspace_scan_flushed = flushed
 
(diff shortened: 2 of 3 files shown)


AmirF194:
Checking in after a week with no review yet, no rush. Happy to rebase or split it if that helps it fit into your queue.

AmirF194:
#1926 fixes the same Nextflow flush bug: `_workspace_scan_flushed` set unconditionally instead of only after a successful completion. This one has been open since Aug 15 with the same fix shape. Happy to stand down in favor of #1926 if maintainers prefer it, or they can pick whichever.

opcode81:
Thanks for the fix.

AmirF194:
Appreciate you syncing the branch to clear the ruff-format drift, saved me a push.

Issue #1663, opened 2026-07-07
Document parallel CI test strategy and add instructions for testing new languages in CI


opcode81:
The new CI test strategy is undocumented, and we should explain how one should decide which CI task a (new) language should be part of.

xormania:
I think this is a great area for improvement - I want to help, big time. I've been on a "project intake" binge for Serena  (and i'm busy) - but i'm getting more comfortable with the architecture and code base (this is a LOT to absorb). I have some system design experience - but this is your baby - maybe start with something simple (like an a draft scope/deliverable document we can work on?)? If you want to skip simple - I got the goods.

xormania:
# CI test strategy — proposal for oraios/serena#1663

Issue #1663 asks for two things: **document the CI test strategy** and **explain how to decide
which CI batch a new language belongs to**. Part 1 below is a paste-ready issue comment scoped to
exactly that, with performance work mentioned only as an optional follow-up. Part 2 is the full
design note behind it (strategy comparison, risks, rollout, acceptance criteria) — the content
for a separate follow-up issue if the maintainers want the performance track.

---

## Part 1 — paste-ready issue comment

```markdown
Here's a draft scope/deliverable for this issue — a docs page plus a placement rubric. Happy to
turn it into a PR if the shape looks right.

## Proposed docs page ("CI test strategy")

Promote the `pytest.yml` header rationale into a short page under `docs/`, linked from
CONTRIBUTING and the new-language guide, covering:

- The batch model: 5 marker groups (`jvm`, `native`, `other-langs`, `niche`) + `catch-all`
  derived as the negation of all named groups — so every test lands in exactly one batch and the
  partition can never desync. 13 jobs = batches × 3 OSes, `niche` ubuntu-only.
- Why in-process parallelism (pytest-xdist) is deliberately not used: tests share fixtures,
  language-server processes, and on-disk resources.
- The concurrency ceilings that bound the shape: 20 concurrent jobs / 5 concurrent macOS jobs
  (free tier); the matrix currently uses 13 / 4.
- OS-specific languages are handled by the central conftest guard
  (`_determine_disabled_languages()` / `language_tests_enabled()`), not by batch placement —
  swift is macOS-only inside `native`; haskell skips Windows inside `other-langs`.
- Caching: language servers under `~/.serena/language_servers/static`, key
  `language-servers-{os}-{batch}-vN` with a prefix `restore-keys` fallback; caches save only on
[...]

PR #1875, opened 2026-08-16
Docs: add a Contributing section for people working on Serena itself

xormania:
> [!NOTE]
> **Two commits, and the first one is #1873.** Turning the code reference on renders `src/serena`'s docstrings through autodoc for the first time, and `poe doc-build` runs with `-W`, so the corrections have to be in place. They were deliberately split into their own pull request rather than buried here, so that each change has a single subject and can be judged on its own. Once #1873 merges, GitHub drops it from this diff and what remains is one commit of 16 files, none of them source. **Review the second commit.**

Serena is not a library, and this PR agrees with that.

The docs build has carried a fully configured Sphinx-autodoc pipeline since it was introduced — `make_rst`, `sphinx.ext.autodoc`, `viewcode`, `sourcelink`, the `autodoc_*` options — with the generator call sitting behind a hardcoded `enable_module_docs = False`. Turning that on and publishing the result at the top level would read as *"here is Serena's API"*. That reading would be fair, and it would be wrong.

So the generated pages live at the bottom of a **Contributing** section, where nothing has to argue that they are not a published interface — the navigation says it. And the section around them is the part that matters: what a new contributor needs before the code makes sense.

**Live preview:** https://xormania.github.io/serena/06-contributing/000_contributing-intro.html

## The section

| page | what it carries |
|:--|:--|
| **Contributing** | the map; defers to `CONTRIBUTING.md` for policy and says so |
| **Getting Started** | fresh clone to verified change: environment, the four commands that gate every commit, how tests are selected and skipped, what CI runs, which files are generated rather than edited |
| **Serena for Serena** | keep your daily server on current `main`; run a specific branch when testing needs it; use the diagnostics tools to see what an edit did |
| **House Style** | the design, testing and docstring conventions, rendered from the project's own checked-in doctrine |
| **Adding Language Support** | the most common substantial contribution, mapped end to end — the dependency-provider base classes, the checksum flow, who owns which initialization keys |
| **Scripts** | what lives in `scripts/`: the demos that run tools without an agent attached, the generators, the introspection utilities |
| **FAQ** | 35 questions that keep arriving on the tracker, answered as the maintainers answered them; 32 link the thread they came from |
| **The CI Matrix** | five marker batches across three operating systems, why the split exists, and what it means for a pull request |
| **SolidLSP** | the language-server library underneath: its shape, how Serena uses it, and the etiquette of changing the core versus adding an adapter |

The FAQ is not invented: every entry came from reading the tracker, and each answer was checked back against the thread it came from before it was written down.

## The code reference

Generated from the docstrings already in `src/serena`, behind a hand-written front page that walks the package in the order you meet it — the agent, then the tools, then configuration and the project layer — so it reads forwards on a first pass and works as a lookup afterwards. 48 generated pages, 613 documented objects, produced by the existing pipeline with the flag flipped and the output pointed at this section.

> [!NOTE]
> Nothing here is a promise. These are internals, documented for people working on them, in the one place where that is unambiguous.

## Verification

| check | result |
|:--|:--|
| `poe doc-build` (sphinx `-W`) | build succeeded, 0 warnings |
| contributing pages | 9 |
| FAQ entries | 35 |
| code-reference pages | 48 generated + 1 hand-written, covering 613 documented objects |
| change, this PR | 16 files changed, 1340 insertions(+), 60 deletions(-) |
| change, including #1873 until it merges | 28 files, 1403 insertions(+), 89 deletions(-) |

<sub>Measured at `8573dbfe`. The theme is not involved — this builds on the stock `sphinx-book-theme`, and the code reference is deliberately flat so that its pages stay within the stock sidebar's four-level limit.</sub>

## Checklist


xormania:
https://github.com/oraios/serena/issues/1663
I'm trying to atone for my past sins (also)

PR #1896, opened 2026-08-20
fix(mcp): run blocking tool calls off the event loop

### LINKED ISSUE #1890: Streamable-HTTP listener stalls during heavy/cold symbol calls: health probes time out, new-session handshakes slow ~30x
## Description

While a cold or heavy symbol tool call is running, the streamable-HTTP listener becomes temporarily unresponsive: lightweight health probes time out entirely and new-session `initialize` handshakes slow down by more than an order of magnitude. MCP clients that attach at that moment fail to connect and then run their whole session without any Serena tools.

## Environment

- serena-agent 1.7.0 (uv tool install), mcp 1.28.1, uvicorn 0.52.1
- Python 3.13, macOS 15.6 (arm64)
- `serena start-mcp-server --project <large TS workspace (~4200 .ts/.tsx files)> --context=<shared context> --transport=streamable-http --host=127.0.0.1 --port=47321 --enable-web-dashboard=true --log-level=WARNING`
- Language server: bundled TypeScript LSP (static TypeScriptLanguageServer)
- The single shared server instance is used simultaneously by several agent clients (Codex / Claude Code / ZCode sessions start and connect at arbitrary times)

## Measurements

Idle server:

| Probe | Latency |
| --- | --- |
| `GET /mcp` (Accept SSE) | ~90 ms (400/406, listener alive) |
| fresh `initialize` handshake | ~80 ms |

During the first seconds of a heavy `tools/call` (repo-wide `find_symbol`, same server, other requests were NOT blocked for the whole call duration — the stall concentrates at the start of a cold/heavy call):

| Probe | Latency |
| --- | --- |
| `GET /mcp` (Accept SSE) with 2–3 s timeout | **timeout (curl exit 000)** |
| fresh `initialize` handshake | **2.35 s** (vs ~0.08 s idle) |

Typical tool latencies on this workspace:

- Cold `find_symbol` with `relative_path` (first symbol call after long idle): **14.5 s**
- Repo-wide `find_symbol` without `relative_path` (warm cache): **28.6 s**

## Observed consequences

1. Supervisor / keep-alive scripts around the shared server use short health probes (2 s). During the stall window the probe fails, the script assumes the server is dead and starts a second instance, which immediately dies with `OSError: [Errno 48] address already in use`. We observed **14 such doomed duplicate spawns in a single day**, each leaving noise in logs and churn in the pid file.
2. MCP clients connecting at session start during the stall window fail the handshake/attach and then run the **entire session without any Serena tools registered**, because there is no reconnect retry. This is the primary "Serena randomly does not work" symptom we see.

## Steps to reproduce

1. Start serena as a shared streamable-http server on a large TS workspace (see Environment).
2. In one terminal, trigger a heavy call:
   ```
   curl -sS -X POST http://127.0.0.1:47321/mcp \
     -H 'Content-Type: application/json' -H 'Accept: application/json, text/event-stream' \
     -H "mcp-session-id: $SID" \
     -d '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"find_symbol","arguments":{"name_path":"SomeSymbol"}}}'
   ```
3. In a second terminal, immediately loop a lightweight probe:
   ```
   curl --output /dev/null --write-out '%{http_code} %{time_total}\n' --max-time 2 \
     -H 'Accept: application/json, text/event-stream' http://127.0.0.1:47321/mcp
   ```
4. Observe `000`/timeouts and multi-second fresh `initialize` latencies during the start of the heavy call.

## Expected behavior

Long-running tool calls should not starve the HTTP listener: new-session handshakes and lightweight requests should stay responsive (e.g. by running blocking LSP work off the request-handling path, or prioritizing the handshake endpoint).

## Actual behavior

For seconds at a time the listener stops answering probes and slows new handshakes by ~30x, which external supervisors and clients interpret as a dead server.

AmirF194:
## Root cause

`SerenaFastMCPTool` builds every Serena MCP tool with `is_async=False` and a synchronous
`execute_fn` that calls `tool.apply_ex` directly (`src/serena/mcp.py`). The pinned SDK
(`mcp==1.28.1`) invokes that as `fn(**kwargs)` inside the single asyncio event loop that also
serves the streamable-HTTP transport (`func_metadata.py`'s `call_fn_with_arg_validation`: `if
fn_is_async: await fn(...) else: fn(...)`, with no thread offload anywhere in the call chain).

`apply_ex` is blocking: LSP subprocess round-trips and symbol-tree walks. A heavy or cold call
therefore runs synchronously inside that loop, starving every other coroutine on it, including
health-probe handling and new-session `initialize` handshakes, for the call's entire duration.
This matches the issue's repro: `curl --max-time 2` timing out and handshakes going from ~80ms
to multiple seconds.

## Fix

`execute_fn` now runs `tool.apply_ex` in a worker thread via `anyio.to_thread.run_sync`
(already a transitive dependency of the SDK, and the pattern the SDK itself uses in
`resources/types.py`), and `is_async` is set to `True` so the SDK awaits it instead of calling
it inline.

## Verification

- New regression test `test_make_tool_execution_does_not_block_event_loop` races a 0.3s
  blocking mock tool against a 10-step probe coroutine on the same event loop: fails on `main`
  (the raw `fn()` call isn't even awaitable, so `asyncio.gather` raises `TypeError` before the
  probe can run), passes on this branch.
- `test/serena/test_mcp.py` full file: 65 passed (Docker, Python 3.11, `uv sync --extra dev
  --locked`).
- `ruff format --check`, `ruff check`, `ty check src/serena src/solidlsp`, `ty check test
  --exclude test/resources`, and `codespell` on the changed files all pass.
- Not verified: a live Serena server behind streamable-HTTP under the full multi-second
  `find_symbol` workload from the issue; the mechanism above reproduces the same shape
  (synchronous execution stalls the loop) against the real pinned SDK, but I did not measure
  the issue's literal "~30x slower" figure end to end.

Fixes #1890

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Status of the `main` branch. Changes prior to the next official version change w
 * General:
   - Fix: Parallel agents auto-registering projects could overwrite each other's changes to the global
     project list in `serena_config.yml`
+  - Fix: a heavy or cold tool call (LSP subprocess round-trips, symbol-tree walks) ran synchronously on
+    the MCP server's single event loop, so it blocked health probes and other sessions' `initialize`
+    handshakes for its whole duration under the streamable-HTTP transport. Tool execution now runs in
+    a worker thread (#1890)
 
 * Language Servers:
   - Fix: Dart's `$/analyzerStatus` notifications were logged as unhandled-method warnings during analysis (#1855)
--- a/src/serena/mcp.py
+++ b/src/serena/mcp.py
@@ -2,6 +2,7 @@
 The Serena Model Context Protocol (MCP) Server
 """
 
+import functools
 import sys
 from collections.abc import AsyncIterator, Iterator
 from contextlib import asynccontextmanager
@@ -9,6 +10,7 @@ from copy import deepcopy
 from dataclasses import dataclass
 from typing import Any, Literal, cast
 
+import anyio.to_thread
 import docstring_parser
 from mcp.server.fastmcp import server
 from mcp.server.fastmcp.exceptions import ToolError
@@ -60,7 +62,9 @@ class SerenaFastMCPTool(FastMCPTool):
         func_name = tool.get_name()
         func_doc = tool.get_apply_docstring() or ""
         func_arg_metadata = tool.get_apply_fn_metadata(structured_output=structured_output)
-        is_async = False
+        # apply_ex does blocking LSP work; run it on a worker thread so it can't stall the
+        # event loop that also serves the MCP transport (health probes, other sessions).
+        is_async = True
         parameters = func_arg_metadata.arg_model.model_json_schema()
         if openai_tool_compatible:
             parameters = SerenaMCPFactory._sanitize_for_openai_tools(parameters)
@@ -94,9 +98,9 @@ class SerenaFastMCPTool(FastMCPTool):
                 param_desc = f"{param_doc.description.strip().strip('.') + '.'}"
                 properties["description"] = param_desc[0].upper() + param_desc[1:]
 
-        def execute_fn(**kwargs) -> str:
+        async def execute_fn(**kwargs) -> str:
             try:
-                return tool.apply_ex(log_call=True, catch_exceptions=False, **kwargs)
+                return await anyio.to_thread.run_sync(functools.partial(tool.apply_ex, log_call=True, catch_exceptions=False, **kwargs))
             except ToolCallError as e:
                 raise ToolError(e.get_error_message()) from e
 
--- a/test/serena/test_mcp.py
+++ b/test/serena/test_mcp.py
@@ -1,5 +1,8 @@
 """Tests for the mcp.py module in serena."""
 
+import asyncio
+import time
+
 import pytest
 from mcp.server.fastmcp.tools.base import Tool as MCPTool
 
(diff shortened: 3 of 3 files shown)


AmirF194:
No rush, just flagging this is still open in case it got buried. Happy to rebase or narrow the fix further if that would help review.

MischaPanch:
Hi @AmirF194 . Thank you for all the work in the last weeks! We have been quite flooded and partially on vacation, but we will get back to more reviewing and merging soon :)

AmirF194:
No rush at all, thanks for the update. Happy to address anything that comes up in review whenever you get to it.

opcode81:
I've returned from my holiday, so I can now resume work on Serena.

 Thanks for looking into this, @AmirF194. The diagnosis is correct.

However, I am going to close this without merging, because this can be fixed one level further down:

The offloading this PR adds by hand should, arguably, already be happening inside the SDK: `is_async=False` should already have resulted in execution being offloaded to a thread pool. FastAPI has always done exactly this: Its sync path goes through `run_in_threadpool`, yet the MCP SDK does not behave analogously. The SDK is even inconsistent with itself on this point: `FileResource.read` and `DirectoryResource.read` offload via `anyio.to_thread.run_sync`, while tool execution does not get the same treatment.

Fortunately, v2 of the MCP SDK fixes this: Synchronous handlers now run in a worker thread (see [migration guide entry](https://py.sdk.modelcontextprotocol.io/migration/#sync-handler-functions-now-run-on-a-worker-thread)), while `async def` handlers are unchanged. This is the relevant piece of code in v2, which is called by `Tool.run`:
https://github.com/modelcontextprotocol/python-sdk/blob/d2290ca3434731b68ea3e2270bc06a6e6575931b/src/mcp/server/mcpserver/utilities/func_metadata.py#L149-L164

Therefore, we shall fix this by landing the v2 PR #1777.

AmirF194:
Thanks for the detailed explanation, that makes sense, landing this via the SDK v2 migration is clearly the better fix. Glad the diagnosis held up. Let me know if it would help to have another set of eyes on #1777.

PR #1897, opened 2026-08-21
fix: persist remove_project to serena_config.yml

### LINKED ISSUE #1895: remove_project() does not persist: the removed project silently reappears on the next config reload
Preconditions:

- [x] I have consulted the user guide; this is a code-level bug, not a configuration issue.
- [x] I have looked for similar issues and discussions, including closed ones (see below).

Issue details:

## What happens

Removing a registered project (`remove_project`) appears to succeed, but the removal does not
survive a reload of the Serena configuration: the "removed" project silently reappears in
`serena_config.yml` the next time the config is loaded (e.g. on the next agent start, or any
later operation that reloads the config).

## Reproduction

Ran against current `main`, commit `2561478a7072c9e048fdfc0aba8c334218239340`, in a clean
`python:3.11-slim` container, `pip install -e .`. Using the same
`SerenaConfig._determine_config_file_path` monkeypatch pattern the repo's own
`test/serena/config/test_serena_config.py::TestSerenaConfigLoadSave` fixture uses, to point the
config at a temp directory:

```python
from serena.constants import SERENA_MANAGED_DIR_NAME

SerenaConfig._determine_config_file_path = classmethod(lambda cls: str(master_config_path))
# ... write a master config with two registered projects, project1 and project2 ...

config = SerenaConfig.from_config_file(generate_if_missing=False)
config.remove_project("project2")
print([p.project_name for p in config.projects])          # in-memory: as expected

reloaded = SerenaConfig.from_config_file(generate_if_missing=False)
print([p.project_name for p in reloaded.projects])        # reloaded from disk: unexpected
```

Output:

```
loaded projects: ['project1', 'project2']
in-memory after remove_project('project2'): ['project1']
on-disk projects after remove_project + reload: ['project1', 'project2']
BUG CONFIRMED: project2 silently reappears on disk after remove_project()
```

This is a single agent process, calling `remove_project()` exactly once, with no concurrency
involved.

## Expected

`remove_project("project2")` should durably remove `project2` from `serena_config.yml`. Instead
the removal only exists in that one process's in-memory state; any reload (agent restart, or
anything else that reconstructs `SerenaConfig` from the file) brings the "removed" project back.

## Setup

- Serena version: `1.7.1.dev0` (current `main`, commit `2561478a`)
- Not MCP-client- or language-server-specific; reproduced by calling the public Python API
  directly (`SerenaConfig.remove_project`, `SerenaConfig.from_config_file`), no LSP/MCP needed.

I searched open and closed issues for `remove_project` / project removal and didn't find an
existing report of this behavior. Happy to dig into the root cause and send a PR if useful.

BetterAndBetterII:
## Summary

`remove_project` dropped the project from in-memory state, then `_persist_projects` unioned the on-disk list with that state. The removed project was copied back from disk and reappeared on the next `SerenaConfig.from_config_file()` load.

Skip roots this instance loaded and no longer has. Parallel adds from other processes are unchanged.

Fixes #1895.

## Tests

```
uv run pytest test/serena/config/test_serena_config.py::TestSerenaConfigLoadSave -vv
```

### DIFF
--- a/src/serena/config/serena_config.py
+++ b/src/serena/config/serena_config.py
@@ -958,6 +958,8 @@ class SerenaConfig(SharedConfig, ModeSelectionDefinitionWithBaseModes):
     the path to the configuration file to which updates of the configuration shall be saved;
     if None, the configuration is not saved to disk
     """
+    _loaded_project_roots: set[str] = field(default_factory=set)
+    """Project roots present when this instance was loaded from disk."""
 
     # *** static members ***
 
@@ -1095,6 +1097,7 @@ class SerenaConfig(SharedConfig, ModeSelectionDefinitionWithBaseModes):
                 project_config=project_config,
             )
             instance.projects.append(project)
+        instance._loaded_project_roots = {str(p.project_root) for p in instance.projects}
 
         # determine language backend
         language_backend = get_dataclass_default(SerenaConfig, "language_backend")
@@ -1302,10 +1305,14 @@ class SerenaConfig(SharedConfig, ModeSelectionDefinitionWithBaseModes):
         if self.config_file_path is None:
             return
         persisted = SerenaConfig.from_config_file()
+        current_paths = {str(p.project_root) for p in self.projects}
+        removed_paths = self._loaded_project_roots - current_paths
         combined_projects = []
         handled_project_paths = set()
         for p in persisted.projects + self.projects:
             str_path = str(p.project_root)
+            if str_path in removed_paths:
+                continue
             if str_path not in handled_project_paths:
                 combined_projects.append(p)
                 handled_project_paths.add(str_path)
--- a/test/serena/config/test_serena_config.py
+++ b/test/serena/config/test_serena_config.py
@@ -604,6 +604,22 @@ class TestSerenaConfigLoadSave:
 
         assert len(resulting_config.projects) == 4
 
+    def test_remove_project_persists_across_reload(self):
+        """remove_project must drop the project from disk, not only from in-memory state."""
+        p1 = self._make_project_dir("project1", 'project_name: "project1"\nlanguages: ["python"]\n')
+        p2 = self._make_project_dir("project2", 'project_name: "project2"\nlanguages: ["python"]\n')
+        self._write_master_config([p1, p2])
+
+        config = SerenaConfig.from_config_file(generate_if_missing=False)
+        assert sorted(p.project_name for p in config.projects) == ["project1", "project2"]
+
+        config.remove_project("project2")
+        assert [p.project_name for p in config.projects] == ["project1"]
+
+        reloaded = SerenaConfig.from_config_file(generate_if_missing=False)
+        reloaded_names = sorted(p.project_name for p in reloaded.projects)
+        assert reloaded_names == ["project1"], f"remove_project('project2') did not persist; reloaded projects: {reloaded_names}"
+
 
 class TestGetRegisteredProjectWithDanglingProject:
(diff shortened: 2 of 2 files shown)


src/serena/config/serena_config.py
```diff
@@ -1302,10 +1305,14 @@ def _persist_projects(self) -> None:
         if self.config_file_path is None:
             return
         persisted = SerenaConfig.from_config_file()
+        current_paths = {str(p.project_root) for p in self.projects}
+        removed_paths = self._loaded_project_roots - current_paths
```
AmirF194, line 1309: I ran this against `31ff5d1b6` in a clean container to check the concurrent case `_persist_projects`'s own docstring claims to handle ("parallel agent instances may have added or removed projects in the meantime").

`removed_paths` here is computed only from this instance's own `_loaded_project_roots`, so it misses a removal made by a different concurrent instance. Two `SerenaConfig` objects that both loaded before the removal: one calls `remove_project` and persists correctly, but the second one still has the removed project in its own `self.projects` (it never removed it), so its `removed_paths` is empty and its next persist (e.g. from `add_registered_project`, the exact multi-agent path this code is meant to cover) writes the removed project straight back to disk.

```
$ python repro_race.py
a loaded: ['project1', 'project2']
b loaded: ['project1', 'project2']
after A removes project2, disk has: ['project1']
final disk state: ['project1', 'project2', 'project3']
AssertionError: BUG: project2 resurrected on disk despite being removed: ['project1', 'project2', 'project3']
```

Two agents load `{project1, project2}`, one removes `project2` and persists (disk is correctly `[project1]`), the other (unaware, still holding `project2` in memory) registers `project3` and persists, and `project2` reappears.

Tracking removals by diffing this instance's loaded/current sets does not survive a second in-flight instance. A per-project tombstone (or reloading `_loaded_project_roots` from the freshly-read `persisted` config right before diffing) would catch this case too.

src/serena/config/serena_config.py
```diff
@@ -1302,10 +1305,14 @@ def _persist_projects(self) -> None:
...
         combined_projects = []
         handled_project_paths = set()
         for p in persisted.projects + self.projects:
```
AmirF194, line 1312: I ran this against `31ff5d1b667a41c5a0891fb4d8ce6f119cf5b5d6` in a clean container, extending `test_parallel_agents_changing_project_lists` (same file) with a sibling scenario where one instance removes a project while another, already-loaded instance only adds a different one.

`_persist_projects` fixes the case where the same instance calls `remove_project` and later persists again, but `removed_paths` is derived from this instance's own `_loaded_project_roots`, so a second instance that never called `remove_project` computes an empty `removed_paths` and still merges in the removed project from its own stale `self.projects` at line 1312. The project a sibling instance removed and already persisted comes back.

```
config_a = SerenaConfig.from_config_file(...)   # projects: project1, project2
config_b = SerenaConfig.from_config_file(...)   # projects: project1, project2

config_a.remove_project("project2")             # persists: project1
config_b.add_project_from_path(p3)              # persists: project1, project2, project3

reloaded = SerenaConfig.from_config_file(...)
# AssertionError: project2 was removed by config_a but resurrected by config_b's persist;
# resulting projects: ['project1', 'project2', 'project3']
```

A fix along these lines would keep the removal visible to `config_b`: only force-include paths from `self.projects` that are new since load (`current_paths - self._loaded_project_roots`), and take everything else from `persisted.projects` as reloaded from disk, rather than merging in the full `self.projects` unconditionally.

PR #1898, opened 2026-08-21
fix: preserve section separators when editing Markdown symbols

GWSAnyone:
`replace_symbol_body` and `insert_after_symbol` corrupt Markdown documents.

A section's body range reaches the start of the next heading, so it contains the blank line separating the two sections. `replace_body` strips the replacement and writes it into that range, which drops the separator: the next heading ends up appended to the last line of the new body, where it is no longer a heading. `insert_after_symbol` inserts at the line following the body end, but for such a range that line is already the next section's heading line, so the insertion lands inside that section.

Both are single-call corruptions, reproduced on a three-section file.

The fix keeps the whitespace the deleted range ended with in `replace_body`, and makes `insert_after_symbol` insert at the body end position when that position is already at column 0. For languages whose body ranges end at the symbol's last character, neither path changes; the existing Python and TypeScript snapshots are untouched.

Markdown cases were added to `test_replace_body` and `test_insert_and_delete_no_change`; both fail without the fix. The Markdown test repository gained a project configuration, because Markdown is excluded from language auto-detection and these tests go through `Project.load`.

### DIFF
--- a/src/serena/code_editor.py
+++ b/src/serena/code_editor.py
@@ -44,8 +44,14 @@ class CodeEditor(Generic[TSymbol], ABC):
             """
 
         @abstractmethod
-        def delete_text_between_positions(self, start_pos: PositionInFile, end_pos: PositionInFile) -> None:
-            pass
+        def delete_text_between_positions(self, start_pos: PositionInFile, end_pos: PositionInFile) -> str:
+            """
+            Deletes the text between the given positions.
+
+            :param start_pos: the start position (inclusive)
+            :param end_pos: the end position (exclusive)
+            :return: the deleted text
+            """
 
         @abstractmethod
         def insert_text_at_position(self, pos: PositionInFile, text: str) -> None:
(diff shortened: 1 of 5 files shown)


src/serena/code_editor.py
```diff
@@ -177,6 +193,11 @@ def insert_after_symbol(self, name_path: str, relative_file_path: str, body: str
         # `line += 1`, is replaced
         body = body.rstrip("\r\n") + "\n"
 
+        # add the separator which was not repurposed above, because no line break succeeded the
+        # body range
+        if line == pos.line:
```
AmirF194, line 198: I added two tests for editing the LAST heading in a Markdown doc (no following heading, so the body range this PR fixes is bounded by EOF) and ran them against `7da6bca8` in a clean container (`DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=1`, marksman needs it under slim).

`replace_body` handles it fine. `insert_after_symbol` does not: it glues the new heading onto the previous line with no blank line, because `is_neighbouring_definition_separated_by_empty_line()` (symbol.py:256) always returns False for Markdown, so `min_empty_lines` stays 0. The existing test (insert after "Configuration", which has a following heading) does not catch this: the separator before the inserted text there is the file's own pre-existing blank line, never touched by the code. At true EOF there is no such blank line to borrow.

```
test_case = InsertInRelToSymbolTest(LanguageServerId.MARKDOWN, "guide.md",
    "User Guide/Next Steps", "## Wrap-up\n\nThat is everything.\n")
test_case.set_mode("after")
...
AFTER: '...CONTRIBUTING.md)\n## Wrap-up\n\nThat is everything.\n\n'
```

code_editor.py:196-199 adds a trailing separator when `line == pos.line` but nothing symmetric on the leading side. Prepending a blank line in that branch would fix the EOF case but double the (already correct) middle-of-document case, since there the leading blank line already exists in the file. Might need to check whether the insertion point is already preceded by a blank line before deciding to add one.

PR #1899, opened 2026-08-21
fix: abort agent-supplied regexes that backtrack catastrophically

GWSAnyone:
## The problem

A regular expression written by the agent can backtrack catastrophically, and `search_for_pattern` makes it easy to do by accident: `multiline=True` is the default, so `re.DOTALL` is on, `.` matches newlines, and `.*` is therefore not "the rest of the line" but "any span up to any newline in the file". A quantifier applied to a group containing such a span makes the number of ways to match grow exponentially.

What follows is not a slow tool application but a frozen server:

- `re` neither releases the GIL nor runs signal handlers while matching, so the match cannot be interrupted from inside the process;
- tool applications are executed one at a time by the task executor, so every subsequent one queues behind the runaway match;
- `tool_timeout` reports a failure to the client but does not stop the match that is already running;
- `search_files` parallelises with joblib's threading backend, which does not help against the GIL.

Killing the process is the only remedy left, after which the client has to reconnect to the server.

Reproduction (Serena 1.7.1.dev0, a 1546-line source file):

```
search_for_pattern(
    substring_pattern=r"^func .*\n(?:.*\n){0,40}?.*someIdentifier",
    relative_path="internal/chat/chat.go",
)
```

The task starts and never finishes; the process holds a core at 100% indefinitely, and the log shows `Processing 1 files.` with no completion line.

## The fix

Expressions that come from the agent are matched through a new `serena.util.regex_timeout`, which uses the `regex` module — already a dependency of Serena — with its `timeout` argument. `regex` checks the clock while matching and is therefore interruptible.

The affected entry points, all in `serena/util/text_utils.py`:

| function | tool |
| --- | --- |
| `search_text` | `search_for_pattern` |
| `ContentReplacer.replace` | `replace_content` |
| `MultiFileContentReplacer.find_occurrences` | `replace_in_files` |
| `find_text_coordinates` | the `regex` argument of `find_declaration` / `find_implementations` |

On expiry, `RegexTimeoutError` is raised with a message that names the likely cause and the way out (`[^\n]*` instead of `.*`, no quantifier over a newline-spanning group, context parameters instead of encoding the window in the expression), so the agent can correct the expression on its own.

Two details worth pointing out for review:

- In `search_files`, the timeout is deliberately re-raised instead of being recorded as a skipped file: skipped files are only logged at debug level, so the agent would have seen the failure as an absence of matches.
- `ContentReplacer` now compiles the expression once and passes the compiled object to the ambiguity check, which previously re-compiled the same expression for every match.

The limit defaults to 10 seconds and can be changed with `SERENA_REGEX_TIMEOUT_SECONDS`. Well-formed expressions are unaffected: a project-wide search over 117 files takes ~50 ms.

Note on compatibility: `regex` operates in its `re`-compatible mode (VERSION0) by default, so accepted syntax and match semantics are unchanged. It ships no type information, hence the `Any` annotation for the compiled pattern in `TimeLimitedPattern`; its match objects are structurally identical to `re`'s and are annotated as such.

## Testing

`TestRegexTimeout` in `test/serena/test_text_utils.py` covers all four entry points, the configuration of the limit, and that a well-formed expression is unaffected. The pathological pattern used there does not terminate within 8 seconds under plain `re`, and aborts in 0.5 seconds with the change.

`poe format`, `poe type-check` and `test/serena/test_text_utils.py` (114 tests) pass. The remaining failures in the full `test/serena` run on my machine are unrelated to this change: language servers that are not installed here (C#, PowerShell, Kotlin), and `test_memories_manager.py`, which reads the real `~/.serena/memories` unless `SERENA_HOME` is set — with an isolated `SERENA_HOME` all 112 of its tests pass.

### DIFF
--- a/src/serena/util/regex_timeout.py
+++ b/src/serena/util/regex_timeout.py
@@ -1,0 +1,153 @@
+"""
+Time-bounded matching for regular expressions that originate from the agent.
+
+Several tools accept a regular expression written by the agent (`search_for_pattern`,
+`replace_content`, `replace_in_files`, `find_declaration`, `find_implementations`), and such an
+expression can backtrack catastrophically. The standard library's `re` module offers no way out of
+it: it neither releases the GIL nor runs signal handlers while matching. A single pathological
+expression therefore freezes the entire Serena process rather than just the tool application that
+issued it - tools are executed one at a time, so every subsequent tool application waits behind the
+runaway match, and the tool timeout can report a failure to the client but cannot stop a match that
+is already running. Killing the process is the only remedy left.
+
+The `regex` module (already a dependency) checks the clock while matching and can therefore be
+interrupted, so agent-supplied expressions are matched through the helpers below, which abort the
+match and explain the likely cause instead of taking the process down with them.
+"""
+
+import logging
+import os
+import re
+from collections.abc import Callable, Iterator
+from contextlib import contextmanager
+from typing import Any, TypeAlias
+
+import regex
+
+log = logging.getLogger(__name__)
+
+DEFAULT_TIMEOUT_SECONDS = 10.0
+"""
+The time limit for a single matching operation. Well-formed expressions match in milliseconds even
+across large repositories, so the limit only has to be small enough to keep the server responsive
+and large enough not to interfere with legitimate work.
+"""
+
+TIMEOUT_ENV_VAR = "SERENA_REGEX_TIMEOUT_SECONDS"
+
+# The `regex` module ships no type information. Its match objects are structurally identical to
+# those of `re` and are described accordingly; its compiled patterns are not, because they accept
+# the `timeout` argument that `re.Pattern` does not have.
+Match: TypeAlias = re.Match[str]
+
+
+def get_timeout_seconds() -> float:
+    """
+    :return: the time limit for a single matching operation, which is `DEFAULT_TIMEOUT_SECONDS`
+        unless the environment variable `SERENA_REGEX_TIMEOUT_SECONDS` provides a positive number.
+    """
+    raw_value = os.environ.get(TIMEOUT_ENV_VAR, "").strip()
+    if not raw_value:
+        return DEFAULT_TIMEOUT_SECONDS
+    try:
+        timeout_seconds = float(raw_value)
+    except ValueError:
+        log.warning("Ignoring %s=%r: not a number", TIMEOUT_ENV_VAR, raw_value)
+        return DEFAULT_TIMEOUT_SECONDS
+    if timeout_seconds <= 0:
+        log.warning("Ignoring %s=%r: must be positive", TIMEOUT_ENV_VAR, raw_value)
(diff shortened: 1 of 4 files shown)


src/serena/util/regex_timeout.py
```diff
@@ -0,0 +1,153 @@
+"""
+Time-bounded matching for regular expressions that originate from the agent.
+
+Several tools accept a regular expression written by the agent (`search_for_pattern`,
+`replace_content`, `replace_in_files`, `find_declaration`, `find_implementations`), and such an
```
AmirF194, line 5: I checked which tools actually route through this module against `8a967d5`.

The docstring lists `find_declaration` and `find_implementations` as tools that accept an agent-supplied regex. `find_declaration` does: `FindDeclarationTool.apply` (symbol_tools.py:407) takes a `regex` parameter and passes it through `find_text_coordinates`, which this PR fixes. But `find_implementations` does not take a regex at all, on either backend. `FindImplementationsTool.apply` (symbol_tools.py:348) and `JetBrainsFindImplementationsTool.apply` (jetbrains_tools.py:542) both take `name_path`, matched by `NamePathMatcher`/`PatternComponent` (symbol.py:171-177), which is plain `==`/`in` string comparison, never `re`. Neither ever calls into `find_text_coordinates`, `search_text`, or any other function this PR touches.

```
$ grep -n "def apply" -A2 src/serena/tools/symbol_tools.py | sed -n '/FindImplementations/,+4p'
$ grep -n "regex\b" src/serena/tools/symbol_tools.py src/serena/tools/jetbrains_tools.py
symbol_tools.py:407:        regex: str,
jetbrains_tools.py:510:    def apply(self, relative_path: str, regex: str, include_body: bool = False) -> str:
$ sed -n '171,180p' src/serena/symbol.py
        def matches(self, name_path_component: NamePathComponent, substring_matching: bool) -> bool:
            if substring_matching:
                if self.name not in name_path_component.name:
                    return False
            else:
                if self.name != name_path_component.name:
                    return False
```

`find_implementations` was never at risk of catastrophic backtracking and this PR does not touch its code path, so it should come out of the docstring at `src/serena/util/regex_timeout.py:4-5`.

davalillo: Confirmed and fixed in the follow-up #2072 (commit 76bb84a4): find_implementations takes a name_path on both backends (symbol_tools.py:348, jetbrains_tools.py:542) and never routes through this module, so it is out of the docstring now. Thanks for checking the actual routing rather than taking the docstring at face value.

While verifying the routing I also noticed the suite's test pattern is the optimized single-alternative shape - the alternation is what actually triggers the exponential backtracking on the observed corpus (~1550 CRLF lines) - so #2072 adds a test pinning that shape as well.

AmirF194: Confirmed, thanks for checking both backends and pinning the alternation shape too.

PR #1900, opened 2026-08-21
fix: let the remind hook see shell-based searching and reading

GWSAnyone:
## The problem

`serena-hooks remind` exists to notice when an agent drifts away from the symbolic tools and to nudge it back. For the `claude-code` and `codebuddy` clients it counts only the built-in `Grep` and `Read` tools:

```python
def is_grep_call(self) -> bool:
    if self._client in (HookClient.CLAUDE_CODE, HookClient.CODEBUDDY):
        return self._tool_name == "grep" or "search_for_pattern" in self._tool_name
```

A search issued as `Bash(grep -rn ...)` and a read issued as `Bash(cat ...)` or `Bash(sed -n '1,40p' ...)` are not counted at all.

That is the wrong half of the behaviour. An agent still pressing the built-in `Grep` is at least using a tool the hook can see; an agent that has already moved to the shell is precisely the one the reminder is meant for — and it is invisible to the hook. In practice the shell is where the drift goes: a session told that the built-in read is discouraged reaches for `cat` next, not for `find_symbol`.

The machinery to classify these calls is already present and already used for `codex` and `grok`: `_GREP_SHELL_COMMANDS`, `_READ_SHELL_COMMANDS`, `_is_shell_command_call()`, and the payload parsing that extracts a `cmd`/`command` field (Claude Code's `Bash` tool uses `command`, which is already read in `__init__`).

## The change

`is_grep_call` and `is_read_call` classify shell-command calls for `claude-code` / `codebuddy` the same way they already do for `grok`.

`is_read_code_file_call` gets the same treatment for a related reason: it inspected the shell arguments only for `codex` and `grok`, so a shell read from any other client fell through to the `return True` default and `cat README.md` was counted as reading source. The target of a shell read is named in its arguments regardless of which client issued it, so the client gate around that branch is dropped rather than extended.

## Scope

Deliberately unchanged: only the first token of the command is examined, so `cd sub && grep -rn foo` is still not classified. That limitation is pre-existing and shared with `codex`/`grok`; splitting commands on `&&`, `||`, `;` and `|` would be a sensible follow-up, but it changes classification for the existing clients too and belongs in its own change.

## Testing

Three new cases in `test/serena/test_hooks.py` covering shell search, shell read, and code-file classification of the shell read target, in the same table style as the existing `codex`/`grok` tests. `test/serena/test_hooks.py` (86 tests) and `test/serena/config` (99 tests) pass, as do `poe format` and `poe type-check`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ Status of the `main` branch. Changes prior to the next official version change w
 * General:
   - Fix: Parallel agents auto-registering projects could overwrite each other's changes to the global
     project list in `serena_config.yml`
+  - Fix: the `serena-hooks remind` PreToolUse hook was blind to shell-based searching and reading for
+    the `claude-code` and `codebuddy` clients: only the built-in `Grep`/`Read` tools were counted,
+    while `Bash(grep …)`, `Bash(cat …)` and `Bash(sed -n …)` were not counted at all. The hook thus
+    nudged the agent that reaches for the built-in tool and ignored the one that had already drifted
+    to the shell - the very behaviour it exists to catch. Shell commands are now classified for these
+    clients exactly as they already were for `codex` and `grok`
 
 * Language Servers:
   - Fix: Dart's `$/analyzerStatus` notifications were logged as unhandled-method warnings during analysis (#1855)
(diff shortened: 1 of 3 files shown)


Yu-0312:
PR #1928 is intended to supersede this PR rather than merge beside it. It contains the same Claude Code and CodeBuddy Bash classification changes covered here, and additionally implements compound-command parsing and broader shell-command coverage. I have updated #1928 to state this relationship explicitly so the overlapping work is reviewed as one change. I am leaving this PR open for maintainer disposition rather than closing it automatically.

PR #1901, opened 2026-08-21
fix: stop duplicating a comment block that follows a nested mapping

GWSAnyone:
## The symptom

`~/.serena/serena_config.yml` grows. A configuration on my machine had accumulated **thirty copies** of the same comment block — 369 lines holding 225 lines of content — and gained one more on every project registration, since registering a project rewrites the file.

Reproduced from a clean file: one `serena start-mcp-server --project <new path>` takes the block count from 1 to 2, and so on.

## The cause

Two mechanisms that are individually correct, with a gap between them.

ruamel attaches a comment block that follows a nested mapping to the **last entry of that mapping**, at any depth. In a configuration like

```yaml
ls_specific_settings:
  java:
    use_system_java_home: true

# list of paths to ignore across all projects.
# Same syntax as gitignore, so you can use * and **.
ignored_paths: []
```

the block is attached to `ls_specific_settings.java.use_system_java_home`, not to `ignored_paths`. Verified by inspecting the parsed `CommentedMap`: the token sits at `ls_specific_settings.java['use_system_java_home']`, comment slot 2 (post-value).

`normalise_yaml_comments` only ever inspects top-level keys, so it neither moves the block up nor removes it, and `save_yaml` writes it back out verbatim. `SerenaConfig.save` then calls

```python
transfer_yaml_comments(template_yaml, commented_yaml, YamlCommentNormalisation.LEADING, force_update_all=True)
```

which puts the template's text onto `ignored_paths` as its leading comment. The saved file now contains the block twice — and the second copy is again parsed as belonging to the nested mapping next time, so the count grows by one per save.

`remove_nested_comments` exists for exactly this class of stray and states the assumption plainly ("we assume that only top-level keys are supposed to be commented"), but it only handled `CommentedSeq`:

```python
for key in commented_map.keys():
    entry = commented_map[key]
    if isinstance(entry, CommentedSeq):
        ...
```

## The change

`remove_nested_comments` now walks nested mappings and sequences alike, at any depth.

## Trade-off worth stating

A comment the user writes *inside* a nested mapping is now dropped on the next save — exactly as one written inside a nested sequence already was. That follows the assumption the function already declares, and it is what makes the fix small.

The alternative is to **move** such a block up to the following top-level key instead of removing it, which preserves user comments and is more faithful, but is a larger change to the normalisation (finding the next top-level key, deciding leading vs trailing). If you prefer that shape, say so and I will rework it.

## Testing

Two tests in `test/serena/util/test_yaml.py` exercising one full save cycle of the global configuration (load → transfer template comments with `force_update_all` → save):

- the block appears exactly once after the first save and stays at one over three more saves, with the settings themselves unchanged;
- two consecutive saves produce byte-identical files.

Both fail on `main` and pass with the change. `poe format`, `poe type-check`, and `test/serena/util` + `test/serena/config` + `test/serena/test_project_server.py` (151 tests) pass.

Verified end to end as well: with the patch applied to an installed Serena, registering a new project rewrites the config and the block count stays at 1.

### DIFF
--- a/src/serena/util/yaml.py
+++ b/src/serena/util/yaml.py
@@ -136,17 +136,28 @@ def normalise_yaml_comments(commented_map: CommentedMap, comment_normalisation: 
 
     def remove_nested_comments() -> None:
         """
-        Removes nested comments, particularly of sequences, which incorrectly capture comments
-        that are actually intended for top-level keys.
+        Removes nested comments, particularly of sequences and nested mappings, which incorrectly
+        capture comments that are actually intended for top-level keys.
+
+        Nested mappings matter as much as sequences here: ruamel attaches a comment block that
+        follows a nested mapping to that mapping's last entry, at any depth. Such a comment survives
+        the normalisation above (which only ever looks at top-level keys), is written back out on
+        save, and is then joined by the comment that the caller transfers onto the top-level key it
+        was actually meant for — so the block is duplicated once per save.
         """
+
+        def remove(node: Any) -> None:
+            if not isinstance(node, CommentedSeq | CommentedMap):
+                return
+            items = node.ca.items
+            if isinstance(items, dict):
+                for i in list(items.keys()):
+                    items[i] = [None] * 4
+            for value in node.values() if isinstance(node, CommentedMap) else node:
+                remove(value)
+
         for key in commented_map.keys():
-            entry = commented_map[key]
-            if isinstance(entry, CommentedSeq):
-                items = entry.ca.items
-                if isinstance(items, dict):
-                    items_keys = list(items.keys())
-                    for i in items_keys:
-                        items[i] = [None] * 4
+            remove(commented_map[key])
 
     match comment_normalisation:
         case YamlCommentNormalisation.NONE:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ Status of the `main` branch. Changes prior to the next official version change w
 * General:
   - Fix: Parallel agents auto-registering projects could overwrite each other's changes to the global
     project list in `serena_config.yml`
+  - Fix: a comment block following a nested mapping in a configuration file was duplicated on every
+    save, so `serena_config.yml` grew by one copy of the same block per project registration (one
+    config in the field had accumulated thirty copies, 369 lines for 225 lines of content). ruamel
+    attaches such a block to the last entry of the nested mapping, where the comment normalisation —
+    which only inspects top-level keys — left it untouched; it was then written back out alongside
+    the copy transferred onto the top-level key it actually documents. Nested mappings are now
+    treated like nested sequences already were
 
 * Language Servers:
   - Fix: Dart's `$/analyzerStatus` notifications were logged as unhandled-method warnings during analysis (#1855)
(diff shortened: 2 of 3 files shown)


src/serena/util/yaml.py
```diff
@@ -136,17 +136,28 @@ def trailing_to_leading(comment_entry: Any) -> Any:
...
         """
+
+        def remove(node: Any) -> None:
+            if not isinstance(node, CommentedSeq | CommentedMap):
+                return
+            items = node.ca.items
+            if isinstance(items, dict):
+                for i in list(items.keys()):
+                    items[i] = [None] * 4
+            for value in node.values() if isinstance(node, CommentedMap) else node:
```
ebarkhordar, line 156: I ran this against 7cd1323 in a clean python:3.11-slim container. The duplication fix works, but it also deletes a comment a user wrote inside a nested mapping, not only the block that follows one: `remove()` clears `ca.items` for every nested node at any depth. `main` keeps those comments.

Filling in the template's `ls_specific_settings`:

```yaml
ls_specific_settings:
  java:
    use_system_java_home: true
    # bump this when the JDT-LS runs out of heap
    vmargs: "-Xmx4g"
```

One save cycle, same sequence as `_save_like_serena_config_does`:

```
7fcbca7e (main)  inner comment after save = True
7cd1323c (here)  inner comment after save = False
```

The block you are targeting always attaches to the mapping's last entry, while an interior comment attaches to an earlier one, so walking only the last entry separates them: clear `items[keys[-1]]`, then recurse with `remove(node[keys[-1]])`. With that change `test/serena/util/test_yaml.py` still passes 6 of 6, including your new regression test, and the comment above survives.

`ProjectConfig` saves `project.yml` through the same normalisation (`serena_config.py:343`, `:756-766`), so this reaches project files too. I did not run that path.

AmirF194, line 156: I ran this against `7cd1323c15b609dee65afbb278d5b09bae587749` in a clean container (`python:3.12-slim`, `ruamel.yaml` + `sensai-utils==1.5.0`).

`remove()` here recurses into every `CommentedSeq`/`CommentedMap` at any depth and wipes `.ca.items` unconditionally. That is broader than the bug it fixes: it also strips a comment a user wrote on an entry *inside* a nested mapping, not just a block that trails the mapping and leaks onto the next top-level key.

Repro: a user config with a comment on `ls_specific_settings.java.use_system_java_home` (a real shape, since `ls_specific_settings` is a per-language nested mapping), run through one `load -> transfer_yaml_comments -> save` cycle like `SerenaConfig` does:

```
$ docker run --rm --memory=1g -v $(pwd):/repo:ro -w /repo python:3.12-slim \
    sh -c "pip install --quiet 'ruamel.yaml' 'sensai-utils==1.5.0' && python3 /repro.py"

ls_specific_settings:
  java:
    use_system_java_home: true
...
user comment survived: False
```

Same repro against the pre-PR base (`7fcbca7e`) keeps the comment (and reproduces the duplication bug this PR is fixing, so the repro itself is sound):

```
ls_specific_settings:
  java:
    # explanation for why system java home is used here, written by the user
    use_system_java_home: true
```

Before this PR, `remove_nested_comments` only touched `CommentedSeq`, so a comment nested inside a mapping was never in scope. Extending it to `CommentedMap` fixes the reported duplication but now also deletes any comment a user placed on a key inside a nested mapping, silently, on every save.

PR #1882, opened 2026-08-16
Upgrade mcp from 1.28.1 to >=2.0.0 (Python 2.x protocol)

hzblwhp:
## Summary

Upgrade serena to the MCP Python 2.x protocol line (`mcp>=2.0.0`).

The 1.x to 2.0.0 Python migration is small and structurally isomorphic: the `mcp.server.fastmcp` subpackage is removed and replaced by `MCPServer` plus the new `mcp.server.mcpserver` / `mcp.shared` subpackage structure.

## Changes (4 files, +21/-30)

- **pyproject.toml**: `mcp==1.28.1` -> `mcp>=2.0.0`
- **src/serena/mcp.py**: `FastMCP` -> `MCPServer` (`mcp.server.mcpserver`); `ToolError` moved to `mcp.server.mcpserver.exceptions`; `Context` / `Settings` import paths updated; `run()` signature aligned (2.0.0 removed `ServerSessionT`, `Context` takes 2 generics not 3); `host` / `port` moved from constructor to `run()`
- **src/serena/tools/tools_base.py**: import paths for `Context` and `FuncMetadata` / `func_metadata`
- **src/serena/cli.py**: `run()` call updated for transport / host / port

## Verification

- serena stdio server boots under mcp 2.0.0: **29 tools loaded**
- `initialize` returns `serverInfo.name=Serena`, `protocolVersion: 2025-11-25`
- `ping` responds
- `tools/list` - all 29 tools exposed, every `inputSchema` valid (`type=object`)

## Compatibility

No breaking changes to serena's public API or tool surface. Only internal mcp import paths and the `run()` call signature changed.

PR #1908, opened 2026-08-22
fix(mcp): report Serena's version in the initialize handshake

### LINKED ISSUE #1889: initialize reports serverInfo.version of the installed mcp SDK (1.28.1), not Serena's version
## Description

The MCP `initialize` handshake reports a `serverInfo.version` that is the version of the installed `mcp` Python SDK, not Serena's own version.

- `initialize` result: `"serverInfo": {"name": "Serena", "version": "1.28.1", ...}`
- `get_current_config` from the same running server: `Serena version: 1.7.0`
- The installed `mcp` package version is exactly `1.28.1`, and `1.28.1` has never been a Serena release (latest is 1.7.0).

## Cause

In `serena/mcp.py` (`start_mcp_server`), the `FastMCP` instance is created without a `version` argument:

```python
mcp = FastMCP(
    name="Serena",
    lifespan=self.server_lifespan,
    website_url="https://oraios.github.io/serena",
    host=host,
    port=port,
    instructions=instructions,
)
```

so FastMCP falls back to its own/SDK version string for `serverInfo.version`, while `serena/__init__.py` already exports `__version__ = "1.7.0"` (and a `serena_version()` helper that even appends git status).

## Steps to reproduce

1. Install serena-agent 1.7.0 with a shared streamable-http server:
   ```
   serena start-mcp-server --project <some repo> --transport=streamable-http --host=127.0.0.1 --port=47321
   ```
2. Send an initialize request:
   ```
   curl -sS -X POST http://127.0.0.1:47321/mcp \
     -H 'Content-Type: application/json' -H 'Accept: application/json, text/event-stream' \
     -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-03-26","capabilities":{},"clientInfo":{"name":"diag","version":"0"}}}'
   ```
3. Compare the returned `serverInfo.version` with `get_current_config` and `pip show mcp`.

## Expected behavior

`serverInfo.version` is Serena's version (e.g. `1.7.0`), for example by passing `version=__version__` (or `serena_version()`) to `FastMCP(...)`.

## Actual behavior

`serverInfo.version` mirrors the installed `mcp` SDK version (`1.28.1` here), which never matches any Serena release and confuses clients that log, display, or compare server versions.

## Environment

- serena-agent 1.7.0 (uv tool install)
- mcp 1.28.1, uvicorn 0.52.1
- Python 3.13, macOS 15.6 (arm64)
- Transport: streamable-http

ebarkhordar:
## Problem

The MCP `initialize` handshake answers `serverInfo.version` with the version of the installed `mcp` SDK rather than Serena's own. On current `main` a client sees `1.28.1`, which has never been a Serena release.

## Root cause

`mcp.server.lowlevel.Server.create_initialization_options()` computes `server_version = self.version if self.version else pkg_version("mcp")`, so an unset `version` falls back to the SDK's own package version. `FastMCP.__init__` builds that low-level server with `name`, `instructions`, `website_url`, `icons` and `lifespan`, and never passes a version, so `self.version` stays `None` on every FastMCP-based server. `create_mcp_server` supplies `name` and `website_url` but has no way to supply a version: `version` is not a parameter of `FastMCP.__init__` in the pinned `mcp==1.28.1`, and passing it raises `TypeError`.

## Fix

Set the low-level server's version after construction:

```python
mcp._mcp_server.version = __version__
```

Two points I did not want to settle on my own:

- This reaches into `_mcp_server`, a private attribute. It is the only way to supply a version through `FastMCP` at `mcp==1.28.1`, which `pyproject.toml` pins exactly, so it cannot drift under a floating dependency. The cleaner long-term shape is a `version` passthrough in FastMCP upstream. If you would rather wait for that, say so and I will close this.
- I used `__version__` (`1.7.1.dev0`) rather than `serena_version()`, which appends the git short sha and `-dirty`. `get_current_config` already reports the plain version, and `serverInfo.version` is what clients compare against releases. Happy to switch if you prefer the git-suffixed string.

## Verification

- New test in `test/serena/test_mcp.py` fails on `main` with `AssertionError: assert '1.28.1' == '1.7.1.dev0'` and passes here. It builds the server through the real `create_mcp_server` and asserts on `create_initialization_options()`, the function that produces the wire value.
- Live handshake in a clean `python:3.11-slim` container at HEAD `7fcbca7e`, streamable-http on `127.0.0.1:47321`: before, `"version":"1.28.1"`; after, `"version":"1.7.1.dev0"`.
- `poe lint`, `poe type-check` and `codespell` pass; the `catch-all` pytest batch runs 1279 passed / 5 failed on Python 3.11 in that container, and those same 5 (the `show_fatal_exception_safe` GUI tests and two process-group cleanup tests) fail identically on unmodified HEAD there, so they are environmental rather than mine.
- Not checked: client-side behaviour that keys on the old value, and the `mcp` 2.x upgrades in #1777 and #1882, which restructure this construction site.

Fixes #1889

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ Status of the `main` branch. Changes prior to the next official version change w
   - Fix: Parallel agents auto-registering projects could overwrite each other's changes to the global
     project list in `serena_config.yml`
   - Fix: `read_only` restriction in project definition was not applied to base tool set when in single-project context (#1938)
+  - Fix: the MCP `initialize` handshake reported `serverInfo.version` as the version of the installed
+    `mcp` SDK rather than Serena's own, so a client that logs, displays or compares the server version
+    saw a string that never matches a Serena release (#1889)
 
 * Language Servers:
   - Fix: Dart's `$/analyzerStatus` notifications were logged as unhandled-method warnings during analysis (#1855)
--- a/src/serena/mcp.py
+++ b/src/serena/mcp.py
@@ -20,6 +20,7 @@ from mcp.types import ToolAnnotations
 from pydantic_settings import SettingsConfigDict
 from sensai.util import logging
 
+from serena import __version__
 from serena.agent import (
     SerenaAgent,
 )
@@ -390,6 +391,9 @@ class SerenaMCPFactory:
             port=port,
             instructions=instructions,
         )
+        # noinspection PyProtectedMember
+        # FastMCP passes no version to the low-level server, whose serverInfo then falls back to the mcp package's version.
+        mcp._mcp_server.version = __version__
         return mcp
 
     @asynccontextmanager
--- a/test/serena/test_mcp.py
+++ b/test/serena/test_mcp.py
@@ -298,3 +298,32 @@ def test_make_tool_all_tools(tool_class) -> None:
 
     # The description should be a string (either from docstring or default)
     assert isinstance(mcp_tool.description, str)
+
+
+def test_create_mcp_server_reports_serena_version(monkeypatch) -> None:
+    """Test that the initialize handshake reports Serena's version, not the mcp SDK's."""
+    from importlib.metadata import version as installed_version
+
+    from mcp.server.fastmcp.server import Settings
+
+    from serena import __version__
+    from serena.config.serena_config import SerenaConfig
+
+    class AgentlessFactory(SerenaMCPFactory):
+        """Builds the server without a SerenaAgent, so only the FastMCP construction is exercised."""
+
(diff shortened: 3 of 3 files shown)


ebarkhordar:
#1913 opened two days after this one on the same fault, and the functional change is identical: `from serena import __version__` plus `mcp._mcp_server.version = __version__` at the same point in `create_mcp_server`. What differs between them is the comment wording, the changelog entry and the length of the test.

Two PRs for one line only costs review time, so I am glad to close this one if #1913 is the one you would rather take. Merge whichever reads better and I will close the other.

opcode81:
Closing in favour of #1913

PR #1909, opened 2026-08-22
fix(php): honor file_filter in alternatives, fix Phpactor indexing

andrebrait:
## Summary

Make the alternative PHP backends honor their `ls_specific_settings.*.file_filter`, so that project-specific PHP extensions (Drupal's `.module`, `.inc`, `.theme`, …) take part in symbol queries as they already do for Intelephense.

- PHPantom and Phpactor extend Serena's source-file matcher from `file_filter`, so `is_ignored_path` stops hiding those files from the traversal behind `find_symbol`.
- Phpactor additionally receives the extensions as `indexer.include_patterns` / `indexer.supported_extensions`, without which its index — and therefore `find_referencing_symbols` — never sees them.
- PHPantom has no equivalent setting; its class docstring now states that its `.module` references depend on Serena's traversal having opened the file.

Review of the first version asked whether the alternative backends index the added extensions unprompted. They do not, and answering that turned up three further defects in Phpactor's path, all fixed here:

- **The first cross-file query raced Phpactor's own index.** Phpactor answers references from its index and reconciles the open documents with it first, so a query before the initial index existed returned nothing, and its dirty-document tracker raised `Dirty index file path ... cannot be created` when the index directory was still absent. Serena now declares `window.workDoneProgress`, tracks the `Indexing workspace` progress and waits for it before the first cross-file query, bounded by the new `indexing_timeout` / `indexing_start_grace` settings — the same pattern as Metals and tsserver.
- **Widening `file_filter` did not reindex.** Phpactor's indexer skips files older than the index, so adding `.module` to an already indexed project left those files out of it indefinitely. The index now lives in the project's cache directory with the indexed extensions in the directory name, so a changed `file_filter` gets a fresh index.
- **A project path containing `%…%` killed the server.** Phpactor expands `%token%` pairs in path settings and aborts during `initialize` on tokens it does not know, so Serena leaves the new index-path setting alone for such paths. Covered by `test_phpactor.py`.

`test_php_basic.py` already excluded Phpactor on Windows CI and without PHP's openssl extension. Both preconditions moved into `_determine_disabled_language_servers`, which the module docstring names as the single place for availability rules, so every Phpactor test observes them instead of only the parametrized ones.

## Phpactor on Windows

Measured on the Windows CI runners with a temporary diagnostic (traced `$/progress`, `indexing_timeout: 600`): Phpactor announces `Indexing workspace` at t=1.5s and then makes no progress at all — no progress report, no end, and an index directory holding nothing but the `dirty` file after 600s. The reference query itself completes in 10ms against that empty index. The same workspace is indexed in ~9s on macOS (`report 500/537` at t=8.8s, 15,479 index files, references resolved).

So Phpactor's cross-file queries do not work on Windows regardless of how long Serena waits; symbol queries, which never touch the index, are unaffected. That is recorded on `PhpactorServer` with a recommendation to prefer Intelephense or PHPantom there, and it is the reason the Phpactor tests stay gated off Windows CI. The cause lies inside Phpactor (progress starts, its indexing generator never advances) and is not addressed here.

## Verification

- `uv run poe lint`, `uv run poe type-check`
- `pytest test/solidlsp/php -m php` — 43 passed, on a cold and on a warm index
- The two regressions above reproduced before the fix and re-checked after: warming the index php-only and then querying `.module` references (failed, now passes), and starting Phpactor in a path containing `%…%` (server terminated during `initialize`, now passes; the test fails again if the guard is removed)
- CI green on `2816a85` across all six batches on Linux/macOS/Windows; the run for the comment-only `6b0498e` was cancelled and needs a maintainer rerun

### DIFF
--- a/src/solidlsp/language_servers/phpactor.py
+++ b/src/solidlsp/language_servers/phpactor.py
@@ -141,6 +141,15 @@ class PhpactorServer(SolidLanguageServer):
                 [*PHPACTOR_DEFAULT_INDEXED_EXTENSIONS, *(ext.lstrip(".") for ext in self.ls_id.get_source_fn_matcher().file_extensions)]
             )
         )
+
+        # Phpactor's dirty-document tracker appends to `<indexer.index_path>/dirty` whenever a
+        # references request reconciles the open documents with the index, and raises if that
+        # directory does not exist yet -- which is the case until its background index has flushed
+        # for the first time. A `find_referencing_symbols` early in the session therefore failed
+        # outright (observed on Windows CI). Pointing the index at a directory Serena creates up
+        # front removes the race and keeps the index next to the other cached state of the project.
+        index_path = self.cache_dir / "phpactor-index"
+        index_path.mkdir(parents=True, exist_ok=True)
         initialize_params = {
             "capabilities": {
                 "textDocument": {
@@ -166,6 +175,7 @@ class PhpactorServer(SolidLanguageServer):
                 # These keys replace Phpactor's defaults, hence the union with them.
                 "indexer.include_patterns": [f"/**/*.{ext}" for ext in indexed_extensions],
                 "indexer.supported_extensions": indexed_extensions,
+                "indexer.index_path": str(index_path),
             },
         }
         return initialize_params
--- a/src/solidlsp/language_servers/phpactor.py
+++ b/src/solidlsp/language_servers/phpactor.py
@@ -2,11 +2,13 @@
 Provides PHP specific instantiation of the LanguageServer class using Phpactor.
 """
 
+import hashlib
 import logging
 import os
 import re
 import shutil
 import stat
+import threading
 
 from overrides import override
 
@@ -55,7 +57,22 @@ class PhpactorServer(SolidLanguageServer):
         - file_filter: list of additional file extensions (with leading dot) to treat as PHP
           sources, e.g. [".module", ".inc"]; these are also pushed to Phpactor's indexer, so
           references contained in such files are found as well
+        - indexing_timeout: float, seconds to wait for Phpactor's initial workspace index before
+          the first cross-file query (default: 120.0)
+        - indexing_start_grace: float, seconds to wait for Phpactor to start reporting indexing
+          progress at all (default: 5.0)
+
+    Serena keeps Phpactor's index in the project's cache directory, one directory per set of
+    indexed extensions, so the first cross-file query of a project (or the first one after
+    `file_filter` changed) waits for a full index, PHP's bundled stubs included.
     """
+
+    # Phpactor indexes the workspace once at startup and reports it as work-done progress; a
+    # cross-file query before that index exists comes back empty (see
+    # `_wait_for_cross_file_references_if_needed`). The timeouts only bound the pathological case
+    # of a server that never reports.
+    INDEXING_TIMEOUT = 120.0
+    INDEXING_START_GRACE = 5.0
 
     @override
     def is_ignored_dirname(self, dirname: str) -> bool:
@@ -127,6 +144,12 @@ class PhpactorServer(SolidLanguageServer):
         if file_filter:
             self.ls_id.get_source_fn_matcher().add_extensions(*file_filter)
 
+        # tracking the workspace indexing that Phpactor reports as work-done progress
+        self._progress_lock = threading.Lock()
+        self._active_progress_tokens: set[str] = set()
+        self._indexing_started = threading.Event()
+        self._indexing_complete = threading.Event()
+
     def _create_dependency_provider(self) -> LanguageServerDependencyProvider:
         return self.DependencyProvider(self._custom_settings, self._ls_resources_dir)
 
@@ -160,7 +183,12 @@ class PhpactorServer(SolidLanguageServer):
         # for the first time. A `find_referencing_symbols` early in the session therefore failed
         # outright (observed on Windows CI). Pointing the index at a directory Serena creates up
         # front removes the race and keeps the index next to the other cached state of the project.
-        index_path = self.cache_dir / "phpactor-index"
+        #
+        # The indexed extensions are part of the directory name because Phpactor's indexer skips
+        # files that are older than the index: widening `file_filter` for an already indexed
+        # project would otherwise leave the newly included files out of the index indefinitely.
+        index_key = hashlib.sha256(",".join(indexed_extensions).encode()).hexdigest()[:8]
+        index_path = self.cache_dir / f"phpactor-index-{index_key}"
         if re.search(r"%.*%", str(index_path)):
             # Phpactor expands `%token%` pairs in path settings and aborts on unknown tokens, so
             # such a path cannot be passed on; the project keeps Phpactor's default index location.
@@ -183,9 +211,54 @@ class PhpactorServer(SolidLanguageServer):
                     "workspaceFolders": True,
                     "didChangeConfiguration": {"dynamicRegistration": True},
                 },
+                # Phpactor reports its workspace indexing as work-done progress, which is what the
+                # first cross-file query waits on
+                "window": {"workDoneProgress": True},
             },
             "initializationOptions": initialization_options,
         }
+
+    def _on_progress(self, params: dict) -> None:
+        """Track the work-done progress tokens Phpactor reports, the workspace index among them.
+
+        :param params: the `$/progress` notification's `ProgressParams`
+        """
+        token = str(params.get("token", ""))
+        value = params.get("value") or {}
+        kind = value.get("kind")
+        if kind == "begin":
+            with self._progress_lock:
+                self._active_progress_tokens.add(token)
+                self._indexing_complete.clear()
+            self._indexing_started.set()
+            log.info(f"Phpactor progress [{token}] started: {value.get('title')}")
+        elif kind == "end":
+            with self._progress_lock:
+                self._active_progress_tokens.discard(token)
+                if not self._active_progress_tokens:
+                    self._indexing_complete.set()
+            log.info(f"Phpactor progress [{token}] ended: {value.get('message')}")
+
+    @override
+    def _wait_for_cross_file_references_if_needed(self) -> None:
+        if self._has_waited_for_cross_file_references:
+            return
+
+        # Phpactor answers reference queries from its index and reconciles the open documents with
+        # it first, so a query issued before the initial index has been written returns nothing at
+        # all -- and its dirty-document tracker even raises if the index directory is still absent.
+        # Wait for the indexing progress Phpactor reports rather than for a fixed period.
+        timeout = float(self._custom_settings.get("indexing_timeout", self.INDEXING_TIMEOUT))
+        start_grace = float(self._custom_settings.get("indexing_start_grace", self.INDEXING_START_GRACE))
+        if not self._indexing_started.wait(timeout=start_grace):
+            log.warning(f"Phpactor reported no indexing progress within {start_grace:.0f}s; proceeding")
+        elif not self._indexing_complete.wait(timeout=timeout):
+            with self._progress_lock:
+                outstanding = ", ".join(sorted(self._active_progress_tokens)) or "<none>"
+            log.warning(f"Phpactor indexing did not complete within {timeout:.0f}s; proceeding (outstanding tokens: {outstanding})")
+        else:
+            log.info("Phpactor indexing complete")
+        self._has_waited_for_cross_file_references = True
 
     def _start_server(self) -> None:
         """Start Phpactor server process."""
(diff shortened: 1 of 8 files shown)


andrebrait:
> [!NOTE]
> Superseded — this described head `53e3db88` and a transient Maven Central 403 in the Windows JVM job, both long replaced. Current state is in the PR description.

CI status on head `53e3db88`:

- `other-langs` passes on Ubuntu, macOS, and Windows; these jobs run the PHP changes.
- The sole failure is unrelated `jvm (windows-latest)`: test collection aborted because `clojure -Spath` received HTTP 403 from Maven Central for `org.clojure:clojure:1.12.5`. No tests ran in that job.
- The same JVM job passed on macOS and Ubuntu.
- GitHub refused my `rerun --failed` request because fork PR authors lack repository admin rights.

A maintainer rerun of the Windows JVM job should clear the transient external failure.

src/solidlsp/language_servers/phpactor.py
```diff
@@ -115,6 +117,11 @@ def __init__(self, config: LanguageServerConfig, repository_root_path: str, soli
             self._ignored_dirnames.add("vendor")
         log.info(f"Ignoring the following directories for PHP (Phpactor): {', '.join(sorted(self._ignored_dirnames))}")
 
+        # extending Serena's source matcher with project-specific PHP extensions
+        file_filter = self._custom_settings.get("file_filter")
+        if file_filter:
+            self.ls_id.get_source_fn_matcher().add_extensions(*file_filter)
```
ebarkhordar, line 151: `file_filter` extends Serena's matcher, which `is_ignored_path` consults, so the traversal behind `find_symbol` picks up `.module`. What I could not tell from the diff is whether Phpactor's own index needs telling too.

For Intelephense you push the extensions to the server (`intelephense.py:204`, `intelephense.files.associations`) so it indexes the same files Serena treats as sources. `phpactor.py` and `phpantom.py` send no equivalent, and the new reference test covers PHPantom only, though `test_php_basic.py::test_find_references_across_files` shows Phpactor does resolve cross-file references on `.php`.

So does Phpactor index `.module` unprompted? If not, `find_referencing_symbols` would miss references in those files while `find_symbol` works. I have not run Phpactor myself.

andrebrait: Good catch — Phpactor does *not* index `.module` unprompted, and PHPantom does not either. Measured on `53e3db8` with a clean Phpactor index cache, requesting references for `helperFunction` (defined in `helper.php`, called from `drupal_module.module:21`) **before** any traversal, so the `.module` file is never opened:

| backend | references found |
|---|---|
| Phpactor | `index.php:13`, `index.php:16` — `drupal_module.module` missing |
| PHPantom | `index.php:13`, `index.php:16` — `drupal_module.module` missing |

`phpactor config:dump` shows why: `indexer.include_patterns = ["/**/*.php", "/**/*.phar"]` and `indexer.supported_extensions = ["php", "phar"]`. So exactly the split you predicted: `find_symbol` worked, `find_referencing_symbols` did not.

Fixed for Phpactor in `84ff164` by pushing the matcher extensions through `initializationOptions` (which Phpactor treats as configuration, same channel as the existing `language_server_*.enabled` keys):

```python
"indexer.include_patterns": [f"/**/*.{ext}" for ext in indexed_extensions],
"indexer.supported_extensions": indexed_extensions,
```

Both keys replace Phpactor's defaults rather than extending them, hence the union with `("php", "phar")`. Same run after the fix: `drupal_module.module:21`, `index.php:13`, `index.php:16`.

PHPantom cannot be fixed the same way: its configuration schema exposes `indexing.strategy` (`composer` / `self` / `full` / `none`) as the only indexing knob and no extension or glob setting at all, and its scan is `.php`-only. Its `.module` references therefore materialize only after Serena's traversal has opened the file — which is what the existing test does, and why it passed. That is now stated in the class docstring instead of being implied by test ordering.

Tests follow the actual guarantees:

- `test_phpactor_file_filter_makes_module_references_visible` — references asserted with no prior traversal, so only the index can answer; this fails on the parent commit.
- `test_phpantom_file_filter_makes_module_references_visible` — references after traversal, with the reason spelled out.

`uv run poe format`, `uv run poe type-check`, and `pytest test/solidlsp/php -m php` (42 passed) are green locally.

One unrelated observation while measuring: Phpactor's on-disk index accumulates duplicate reference entries across runs (`index.php` appears 2×, then 4×, then 6× for the same query with an unchanged workspace). Reproducible on the parent commit without any of these changes, so it is not from this PR, but it does mean repeated Serena sessions on the same project can hand back duplicated `find_referencing_symbols` hits. Worth a separate issue if you want me to file one.

andrebrait: Follow-up: making the Phpactor reference test index-only (no traversal first) exposed two further defects on the Windows runner, both fixed in `2816a85`.

**1. The first cross-file query raced Phpactor's own index.** Phpactor answers references from its index and reconciles the open documents with it first, so a query issued before the initial index has been written returns nothing — and its dirty-document tracker even raises `Dirty index file path ... cannot be created` when the index directory does not exist yet (`DirtyFileListProvider::markDirty` does a bare `fopen(..., 'a')`; the directory is created by the first `FileRepository::flush`). Serena now declares `window.workDoneProgress`, tracks the `Indexing workspace` progress Phpactor reports, and waits for it before the first cross-file query, bounded by the new `indexing_timeout` (120s) and `indexing_start_grace` (5s) settings — the same pattern as Metals and tsserver. Observed locally:

```
Phpactor progress [...] started: Indexing workspace
Phpactor progress [...] ended: Done indexing (5.98s, 69/1,611 mb), watching with pattern matching find (BSD/GNU)
Phpactor indexing complete
```

**2. Widening `file_filter` did not reindex.** Phpactor's indexer skips files older than the index, so on a project it had already indexed, adding `.module` left those files out of the index indefinitely — `find_symbol` saw them, `find_referencing_symbols` never would. Reproducible before the fix:

```
$ pytest test/solidlsp/php/test_php_basic.py -m php      # warms the php-only index
29 passed
$ pytest ...::test_phpactor_file_filter_makes_module_references_visible
1 failed          # .module missing from the index
```

Serena now keeps the index in the project's cache directory with the indexed extensions in the directory name (`phpactor-index-<hash>`), so a changed `file_filter` gets a fresh index. Same scenario after the fix: `29 passed` then `1 passed`. The trade-off is that a new extension set pays for a full index, PHP's bundled stubs included (~6s for the test repo here); it is documented on the class.

That relocation needed one guard: Phpactor expands `%token%` pairs in path settings and aborts on tokens it does not know, so a project path such as `pct%weird%dir` killed the server during `initialize`. Serena leaves the setting alone for such paths, and `test_phpactor.py::test_references_in_project_path_with_percent_signs` covers it (it fails without the guard).

**Test gating.** `test_php_basic.py` already excluded Phpactor on Windows CI and without PHP's openssl extension; my new tests did not. Both preconditions moved into `_determine_disabled_language_servers`, which the module docstring already names as the single place for this, so every Phpactor test observes them.

CI on `2816a85`: all three `other-langs` batches and all three `catch-all` batches pass, jvm/native/niche pass.

Answering the original question directly: neither alternative backend indexes `.module` unprompted. Phpactor can be told to, and now is. PHPantom cannot — no extension or glob setting exists — so its `.module` references depend on Serena's traversal having opened the file, which is now stated on the class rather than implied by test ordering.

ebarkhordar: That is the answer, thanks. The Phpactor half is what I expected from `indexer.supported_extensions`; the PHPantom half is not something I could have predicted from the diff, and stating it on the class beats leaving it implied by the order the test happens to run in.

The index-only test is the part that makes it checkable later. Asserting references before any traversal is what separates "the index knows about `.module`" from "something opened the file first".

andrebrait: Correction to my previous comment: I said I could not tell "slow" from "never reports progress" on Windows. That was wrong on both counts — the evidence was already in the run, and the rest was measurable.

**From the timestamps of the failing run.** Both Phpactor reference tests took ~123s, i.e. the 120s `indexing_timeout` plus startup:

| boundary | timestamp | elapsed |
|---|---|---|
| `..._symbols_visible[phpantom]` PASSED | 10:26:15.4 | |
| `test_phpactor_file_filter_makes_module_references_visible` FAILED | 10:28:18.5 | 123.1s |
| `test_file_diagnostics[php_phpantom]` PASSED | 10:28:24.1 | |
| `test_references_in_project_path_with_percent_signs` FAILED | 10:30:26.7 | 122.7s |

Had progress never arrived, the 5s start-grace branch would have returned in ~5s. So the wait engaged, Phpactor reported `begin`, and no `end` came within 120s.

**Then I measured it directly.** A temporary diagnostic commit (`indexing_timeout: 600`, traced `$/progress`, dumped the index directory; pushed, read, force-pushed away — sorry for the one red run) reports from `other-langs (windows-latest)`:

```json
{"platform": "Windows-10-10.0.26100-SP0",
 "query_seconds": 600.1,
 "references": [],
 "serena_cache": [{"path": "...\\.serena\\cache\\php\\phpactor-index-b99427e2",
                   "exists": true, "files": 1, "top_level": ["dirty"]}],
 "timeline": [{"t": 1.48, "kind": "begin", "text": "Indexing workspace"},
              {"t": 601.55, "kind": "begin", "text": "Finding references"},
              {"t": 601.56, "kind": "end", "text": "Found 0 reference(s)"}]}
```

Same diagnostic on macOS for comparison: `report` at t=8.8s (`500/537`), `end` at ~9s, index 15,479 files, references `["drupal_module.module", "index.php", "index.php"]`.

So Phpactor's indexer on Windows is not slow — it makes **no progress at all**: no `report`, no `end`, and 0 of 537 files written after 10 minutes. The query itself is fine (`Finding references` completes in 10ms); it just has an empty index to read. Two by-products worth noting:

- The pre-created index directory does work as intended: the only file in it is `dirty`, i.e. `markDirty` succeeded where the unmanaged default path raised `Dirty index file path ... cannot be created`.
- Windows is therefore not a timeout-tuning question. Cross-file queries there return nothing whatever Serena waits for; symbol queries, which never touch the index, are unaffected — which is why `..._symbols_visible[phpactor]` passes on Windows.

`6b0498e` records exactly that at the two places a reader would look — the `PhpactorServer` docstring (with a recommendation to prefer Intelephense or PHPantom on Windows) and the test gate — instead of my earlier "too slow" hand-wave. I have not diagnosed it inside Phpactor; the shape (progress starts, generator never advances) suggests its Amp event loop starving behind the stdio reader on Windows, which is worth an upstream issue but not this PR.

andrebrait: Housekeeping on CI state, since the checks on this PR look worse than they are.

- `2816a85` is the last commit with a complete run: all six test batches green across Linux/macOS/Windows (`other-langs`, `catch-all`, `jvm`, `native`, `niche`) plus lint, type-check and the `uv sync` matrix.
- The run for `6b0498e` was cancelled 15 minutes in — every job reports `cancelled`, none `failed`, including the `uv sync` matrix that had passed minutes earlier on `2816a85`. My fault for the churn: five pushes in an hour on a free-tier matrix, one of them a deliberately failing diagnostic. I cannot re-run it myself (`Must have admin rights to Repository`), so it needs a maintainer rerun whenever convenient.
- `6b0498e` over `2816a85` is comments only — a docstring paragraph in `PhpactorServer` and a reworded comment at the test gate, 10 added lines, no executable change:

```
 src/solidlsp/language_servers/phpactor.py | 6 ++++++
 test/conftest.py                          | 7 ++++---
```

Locally on `6b0498e`: `uv run poe lint`, `uv run poe type-check`, and `pytest test/solidlsp/php -m php` (43 passed) are all clean.

No further pushes from me unless review asks for changes.

andrebrait:
Scope grew after review, so the description is rewritten to match; the evidence lives in the review thread on `phpactor.py` and is summarized here for anyone not expanding it.

@ebarkhordar's question — do the alternative backends index `.module` unprompted — is answered: neither does. Measured on the parent commit with a clean index, requesting references for `helperFunction` (defined in `helper.php`, called from `drupal_module.module:21`) before any traversal, so the `.module` file is never opened:

| backend | references |
|---|---|
| Phpactor | `index.php:13`, `index.php:16` — `.module` missing |
| PHPantom | `index.php:13`, `index.php:16` — `.module` missing |

`phpactor config:dump` shows why: `indexer.include_patterns = ["/**/*.php", "/**/*.phar"]`. Phpactor can be told otherwise and now is; PHPantom cannot — its configuration schema exposes `indexing.strategy` as the only indexing knob and no extension or glob setting — so its `.module` references materialize once Serena's traversal opens the file, which is now stated on the class rather than implied by test ordering.

Chasing that turned up three more defects in Phpactor's path (cold-index race, `file_filter` widening not reindexing, `%…%` project paths killing the server) plus one platform finding: on Windows Phpactor's indexer announces `Indexing workspace` and then makes zero progress — empty index after 600s, where the same workspace takes ~9s on macOS. Details, measurements and reproductions are in the description and the review thread.

Current test surface: 3 unit tests on the matcher defaults, 1 Intelephense integration test, and 4 alternative-backend integration tests, of which the Phpactor reference test is index-only (no traversal first) and fails on the parent commit.

PR #1910, opened 2026-08-23
fix(solidity): isolate Hardhat state on macOS

### LINKED ISSUE #1817: Solidity diagnostics fail when Hardhat global state is not writable on macOS
## Summary

Serena's managed `@nomicfoundation/solidity-language-server` can start and index a Solidity project but return no diagnostics when its bundled Hardhat runtime cannot create macOS-global state under `~/Library`.

This affects restricted or sandboxed environments even when the project, Serena data directory, language-server installation, Node.js, and Forge are all usable.

## Reproduction

Environment:

- macOS
- Serena current PR head during testing
- `@nomicfoundation/solidity-language-server` 0.8.4
- managed Forge 1.5.1
- fixture compiler version 0.8.20
- writes allowed in the checkout and `/private/tmp`, but not in `~/Library/Preferences`

Run either integration test:

```text
test/solidlsp/solidity/test_solidity_diagnostics.py::TestSolidityDiagnostics::test_file_diagnostics
test/solidlsp/solidity/test_solidity_diagnostics.py::TestSolidityDiagnostics::test_file_diagnostics_via_validation_completion
```

The language server starts, indexes the fixture, and reports validation completion, but the status is:

```text
validationRun: false
reason: EPERM: operation not permitted, mkdir
'/Users/<user>/Library/Preferences/hardhat-nodejs'
```

Both tests then receive zero diagnostics.

## Root cause

The language server bundles Hardhat and `env-paths`. On Darwin, that `env-paths` implementation constructs `Library/Application Support`, `Library/Preferences`, and `Library/Caches` paths from `os.homedir()` and ignores `XDG_CONFIG_HOME`, `XDG_DATA_HOME`, and `XDG_CACHE_HOME`.

Serena's Solidity dependency provider currently adds only the managed binary directory to `PATH`, so there is no Solidity-specific state isolation.

A probe with all three XDG variables pointing into `/private/tmp` still attempted the same `~/Library/Preferences/hardhat-nodejs` write. A process-local Node preload that redirected `os.homedir()` to a writable temporary root removed the `EPERM`; validation then advanced to a separate `HH502` compiler-download failure.

## Proposed behavior

- Give only the managed Solidity language-server process a Serena-owned state root that works without writing to the user's macOS-global preferences or changing the parent process environment.
- Do not change or repurpose `HOME`.
- Prefer an upstream Nomic/Hardhat option for an explicit state root if one becomes available.
- If current pinned dependencies provide no such option, consider a narrowly scoped Darwin compatibility preload for this child process, with an explicit Serena environment variable and a package-owned preload module.
- Preserve any existing `NODE_OPTIONS` when adding the preload.
- Keep the writable state root task-local/ephemeral where persistent access is unavailable.
- Make the diagnostics failure contextual instead of silently presenting a failed validation run as zero diagnostics.

The compiler download/cache is a second concern: integration tests should either seed a pinned compiler into the isolated cache or otherwise arrange the pinned compiler hermetically. Redirecting the state directory alone should not turn the test into an uncontrolled live-download dependency.

## Acceptance criteria

- The two Solidity diagnostics tests can validate a fixture when `~/Library/Preferences` and `~/Library/Caches` are not writable.
- No write outside the configured Serena/project/temporary roots is attempted.
- The implementation does not set `HOME` and does not rely on XDG variables working on Darwin.
- Unsupported or failed validation produces a contextual failure rather than a false empty-diagnostics result.
- Tests cover preservation of existing child-process environment values and the isolated compiler-cache path.
- Linux and Windows behavior remains unchanged unless the same explicit state-root mechanism is supported there.

## Related context

This was discovered while running the full test matrix for PR #1761, but it is unrelated to that PR's memory-frontmatter changes.

I searched open and closed Serena issues for `hardhat-nodejs`, the macOS Preferences path, `HH502`, and Solidity sandbox failures and found no exact match.

Yu-0312:
Fixes #1817

The managed Solidity language server now uses a child-process-only Node preload on macOS so Hardhat resolves its state under a Serena-managed writable directory instead of ~/Library.

Existing NODE_OPTIONS are preserved, HOME is not changed, and Linux/Windows launch environments remain unchanged. The new solidity_state_dir setting allows users to choose another writable state root when needed.

Tests:

- uv run pytest test/solidlsp/solidity/test_solidity_dependency_provider.py test/solidlsp/solidity/test_solidity_basic.py -q
- uv run poe lint
- uv run poe type-check
- uv build --wheel; verified the preload resource is packaged


src/solidlsp/language_servers/solidity_language_server.py
```diff
@@ -174,7 +183,25 @@ def create_launch_command_env(self) -> dict[str, str]:
...
             solidity_ls_dir = self._resolve_solidity_ls_dir(solidity_language_server_version, forge_version)
             managed_bin_dir = os.path.join(solidity_ls_dir, "node_modules", ".bin")
-            return {"PATH": managed_bin_dir + os.pathsep + os.environ.get("PATH", "")}
+            launch_env = {"PATH": managed_bin_dir + os.pathsep + os.environ.get("PATH", "")}
+
+            if platform.system() == "Darwin":
+                state_dir = self._get_solidity_state_dir()
+                os.makedirs(state_dir, exist_ok=True)
+                launch_env["SERENA_SOLIDITY_STATE_DIR"] = state_dir
+                preload_arg = f"--require {shlex.quote(self._HOMEDIR_PRELOAD)}"
```
AmirF194: I ran this against head 4d194b46f74670fb19038eb1f32e01a5ba1f9ce7 in a clean node:20 container.

The `shlex.quote` here produces POSIX single-quote escaping, but Node's `NODE_OPTIONS` parser does not understand single quotes. Any `solidity_state_dir` (or Serena install path) containing a space breaks the child process launch instead of isolating it.

```
$ export NODE_OPTIONS="--require '/tmp/solidity state/preload.cjs'"
$ node -e "console.log(1)"
node:internal/modules/cjs/loader:1210
  throw err;
Error: Cannot find module ''/tmp/solidity'
Require stack:
- internal/preload
```

Double quotes work instead, on the same Node version:

```
$ export NODE_OPTIONS='--require "/tmp/solidity state/preload.cjs"'
$ node -e "console.log(1)"
PRELOAD LOADED OK
1
```

The new test only asserts the string value of `NODE_OPTIONS`, it never actually runs node against it, so CI is green while the isolated launch itself fails whenever the path has a space. This is not a corner case here: the docs' own example writable macOS location, under `~/Library/Application Support/...`, has a space in it, and so does any macOS username with a space in it.

Swapping `shlex.quote(...)` for a wrap in double quotes (escaping embedded `"` and `\`) fixes it.

AmirF194: Confirmed, the double quote fix and the new test with a real node subprocess cover this. Thanks for the fast turnaround.

PR #1911, opened 2026-08-23
fix(erlang): switch to maintained ELP server

### LINKED ISSUE #771: Erlang LS has been archived on 2025-08-15, recommends switching to Erlang Language Platform

Yu-0312:
Fixes #771

Erlang LS was archived in 2025, so Serena now uses the maintained Erlang Language Platform (ELP).

This change downloads the pinned official ELP 2026-08-10 release through `DownloadedDependency` with the central URL-keyed SHA-256 hash database (`downloaded_dependency_hashes.json`, per the current dependency-management standard for new implementations), always uses the managed installation rather than preferring an `elp` executable from PATH, supports Linux/macOS/Windows release assets, launches `elp server` over stdio, preserves `ls_path` overrides, and updates the Erlang tests and documentation. Dead code left over from the old Erlang-LS readiness polling (`server_ready` event and readiness-signal scanning) is removed.

Document-symbol cache entries written by the archived erlang_ls are invalidated via `cache_version_raw_document_symbols=2`, so unchanged `.erl` files are re-indexed by ELP instead of being served stale Erlang-LS-era symbols.

Validation:
- uv run poe lint
- uv run poe type-check
- uv run pytest test/solidlsp/erlang -q
- Recomputed SHA-256 of all five pinned release assets and matched them against the hash database
- ELP managed-download and stdio LSP handshake smoke tests
- uv build --wheel

Addresses review feedback from @opcode81 (use `DownloadedDependency`, prefer the managed installation, invalidate the symbol cache) and @AmirF194 (dead `server_ready` machinery).


src/solidlsp/language_servers/erlang_language_server.py
```diff
@@ -125,105 +212,65 @@ def _create_base_initialize_params(self) -> dict:
...
+        log.info("Sending initialize request to ELP")
         init_response = self.server.send.initialize(self._create_initialize_params())
-
-        # Verify server capabilities
         if "capabilities" in init_response:
-            log.info(f"Erlang LS capabilities: {list(init_response['capabilities'].keys())}")
+            log.info("ELP capabilities: %s", list(init_response["capabilities"].keys()))
 
         self.server.notify.initialized({})
+        self.server_ready.set()
```
AmirF194: I checked out this PR at 2598732 and downloaded all five `ELP_SHA256_BY_ASSET` release assets from the pinned `2026-08-10` GitHub release directly; every hash matches what ships in the tarball, and `elp server` is a real subcommand of the extracted binary (`elp --help` lists it as "Run lsp server"), so the dependency-provider and launch-command changes check out.

One thing left over from the old Erlang LS polling design, at `src/solidlsp/language_servers/erlang_language_server.py:220-247` and `:264-265`: `window_log_message` and `check_server_ready` still scan every notification for readiness strings and call `self.server_ready.set()`, but `_start_server` now calls `self.server_ready.set()` unconditionally right after `self.server.notify.initialized({})` (line 265), before the fixed settling sleep, and nothing downstream ever calls `.wait()` or checks `.is_set()` on that event again. So the two handlers, the `readiness_signals` list, and the "ELP readiness signal detected" logging in `window_log_message` never influence control flow now; they just run against every message for the life of the connection with no observable effect.

```
$ grep -n "server_ready" src/solidlsp/language_servers/erlang_language_server.py
139:        self.server_ready = threading.Event()
236:                self.server_ready.set()
241:        def check_server_ready(params: dict) -> None:
247:                    self.server_ready.set()
265:        self.server_ready.set()
```

If the fixed settling period is the intended replacement for the old wait-then-branch logic, `server_ready`, `window_log_message`'s signal list and `check_server_ready` can go. If the intent was to keep waiting for an actual readiness signal before the settling sleep, line 265's unconditional `.set()` is the one to remove instead.

src/solidlsp/language_servers/erlang_language_server.py
```diff
@@ -7,56 +7,142 @@
...
+}
+
+ELP_SHA256_BY_ASSET: dict[str, str] = {
+    "elp-linux-x86_64-unknown-linux-gnu-otp-27.3.tar.gz": "c0e672a8381b5ea787e94a872847567b10d9b4d0053ca1148f4b236f61af3c63",
+    "elp-linux-aarch64-unknown-linux-gnu-otp-27.3.tar.gz": "0af71bd62e95998b7e57edd2083141b60edab51d0a2da988e6e71abb88cd3f34",
+    "elp-macos-x86_64-apple-darwin-otp-27.3.tar.gz": "d7c739a6b23ba7bfc0fc8481619ec55e41257d08cb7aa8b16499fb4c4f5e38e2",
+    "elp-macos-aarch64-apple-darwin-otp-27.3.tar.gz": "7317a9934edc411e94392d5cc720f2e0c18112e7959060678aaef4d6dacdd9f2",
+    "elp-windows-x86_64-pc-windows-msvc-otp-27.3.tar.gz": "6c4ed9ab76c9cbb2f6ae7b3a8dc06121a4ac6eda94d13bf7a638d3cb111b7f4e",
+}
+
```
opcode81: New implementations should not manage hashes themselves. Please follow the instructions in the memory and use `DownloadedDependency` accordingly.

src/solidlsp/language_servers/erlang_language_server.py
```diff
@@ -7,56 +7,142 @@
...
+        """Resolves ELP from PATH or downloads the pinned official release asset."""
+
+        _SUPPORTED_PLATFORM_IDS: ClassVar[frozenset[PlatformId]] = frozenset(ELP_ASSET_PLATFORM_BY_ID)
+
+        def _get_or_install_core_dependency(self) -> str:
+            """Return an ELP executable, preferring a user-managed executable on PATH."""
+            elp_in_path = shutil.which("elp")
+            if elp_in_path:
+                log.info("Found ELP in PATH: %s", elp_in_path)
+                return elp_in_path
```
opcode81: For language server dependencies Serena can manage, we always prefer the managed installation to an existing installation on the user's system, because this guarantees reproducible results. Unless the user sets `ls_path` (in which case `_get_or_install_core_dependency` won't be called), the managed installation should be used, i.e. this block should be removed.

opcode81 (changes requested):

Yu-0312:
Merged `upstream/main` and confirmed the earlier review feedback is addressed in the current head:

1. **`DownloadedDependency` + central hash DB** (opcode81): ELP is managed via `DownloadedDependency` / `DownloadedDependencyHashDatabase`; hashes live in `downloaded_dependency_hashes.json`. The module no longer manages hashes itself.
2. **No PATH preference** (opcode81): `_get_or_install_core_dependency` always installs the managed binary; the only override is an explicit `ls_path`.
3. **Dead readiness path** (AmirF194): `server_ready`, `readiness_signals`, and `check_server_ready` are gone. Startup uses the initialize response plus a short settling sleep; `window/logMessage` only logs.

Ready for re-review.

src/solidlsp/language_servers/erlang_language_server.py
Yu-0312: Addressed in the current head: ELP is managed via `DownloadedDependency` / `DownloadedDependencyHashDatabase`; hashes live in `src/solidlsp/resources/downloaded_dependency_hashes.json` (see `ErlangLanguageServer.DependencyProvider._create_dep_elp` and `update_dep_hashes`). `scripts/update_downloaded_dependency_hashes.py` is hooked up.

src/solidlsp/language_servers/erlang_language_server.py
Yu-0312: Addressed: `_get_or_install_core_dependency` always installs the managed binary; the only override is an explicit `ls_path` (which bypasses the provider). The PATH-preference block has been removed.

src/solidlsp/language_servers/erlang_language_server.py
Yu-0312: Addressed: `server_ready`, `readiness_signals`, and `check_server_ready` are gone. Startup uses the initialize response plus a short settling sleep; `window/logMessage` only logs.

Yu-0312:
All review feedback from @opcode81 and @AmirF194 is addressed in the current head (replies added to the respective threads). Ready for re-review.

The previous `niche (ubuntu-latest)` CI failure was an opam sandbox/infra flake during "Install OCaml and opam", unrelated to this PR's Erlang/ELP changes; checks were retriggered.

opcode81:
Thanks for addressing the comments. Only one last question. Even though we have fully replaced the LS implementation, the symbol cache was not invalidated. If the new language server produces better data, we could consider invalidating any cached data from the old server (by incrementing the cache version when calling the super constructor; see docstring). So would this be helpful here, or is it fine to keep the old cached data?

Yu-0312:
Yes, I think we should invalidate here.

My reasoning is the asymmetry rather than any measured difference between the two servers: invalidating costs Erlang users one automatic re-index, while keeping the old data risks silently inconsistent behaviour that neither the logs nor an error message would explain.

Concretely, the upsides of invalidating:

Existing users actually get the benefit of the migration. Cache entries are keyed by content hash and survive upgrades, so without a bump every unchanged .erl file keeps being served Erlang-LS-era symbols indefinitely — for existing projects the switch to ELP would effectively be a no-op.
It avoids mixed-provenance data within a single project. Files touched after the upgrade would come from ELP while untouched ones still come from Erlang LS. Since symbol ranges drive the editing tools, any disagreement between the two servers — e.g. whether a function's range spans all clauses — would make replace_symbol_body behave differently from file to file, silently, with nothing pointing at the cache as the cause.
No user action is required. The alternative is a release note asking Erlang users to delete their cache directory by hand, which is easy to miss and awkward to explain.
It removes a variable from future debugging. Any future report about Erlang symbols can then be assumed to reflect ELP's output, instead of first having to establish when the user's cache was built.

One implementation detail I'd like to confirm with you: bumping cache_version_raw_document_symbols alone doesn't appear to be sufficient. _raw_document_symbols_cache_version() does incorporate _ls_specific_raw_document_symbols_cache_version, but _document_symbols_cache_version() does not — and request_document_symbols returns from the high-level cache before it ever consults the raw cache, so the Erlang-LS-derived high-level entries would survive the bump. I'd therefore also bump Erlang's _document_symbols_cache_fingerprint from 1 to 2. (The same gap seems to affect Vue, which passes cache_version_raw_document_symbols=2 without a fingerprint override, but that's a framework-level change I'd rather leave to your judgement than fold into this PR.)

If you agree, I'll make both changes.

opcode81:
> One implementation detail I'd like to confirm with you: bumping cache_version_raw_document_symbols alone doesn't appear to be sufficient. _raw_document_symbols_cache_version() does incorporate _ls_specific_raw_document_symbols_cache_version, but _document_symbols_cache_version() does not.

Thanks for pointing this out; this is a bug. I have fixed this on `main` with commit a5fd4d685f537c3c5a094edc0b39c8b19a344669, so it will be enough for you to increment the version.


In order for CLA acceptance to work, please squash your commits and assign to an author to the commit that corresponds to your actual GitHub user. Then click the badge in CLAassistant's comment to accept the CLA with your GitHub user.

Yu-0312:
Good point — bumping is the right call here rather than keeping the old entries.

The raw document-symbol cache is keyed per `language_id`, so the `erlang` cache directory written by the archived `erlang_ls` is reused as-is by ELP for any file whose content hash is unchanged. Since ELP reports a different symbol shape, those entries are stale rather than merely suboptimal.

Pushed `cache_version_raw_document_symbols=2` in `ErlangLanguageServer.__init__`, following the precedent in `VueLanguageServer`. Lint and type-check pass.

Yu-0312:
@opcode81 The last open review point is addressed in the current head:

- `cache_version_raw_document_symbols=2` is set in `ErlangLanguageServer.__init__` (see a3dea99b), so ELP will not reuse document-symbol cache entries written by the archived erlang_ls.
- After a5fd4d6 on main, bumping the raw cache version is enough to invalidate the high-level cache as well.
- ELP is managed via `DownloadedDependency` / central hash DB; the PATH-preference block is gone.
- CLA is green on all committers.

Keeping the existing commit history (no squash) since CLA already passes. Ready for re-review.

Yu-0312:
Rebased onto current `main` (branch was conflicting).

Per review guidance:
- Cache invalidation: kept `cache_version_raw_document_symbols=2` on `ErlangLanguageServer` (sufficient after your `main` fix a5fd4d68 that propagates LS-specific versions into the high-level cache).
- ELP is managed via `DownloadedDependency` + central hash DB; PATH preference for `elp` is dropped.
- CHANGELOG kept to a single Erlang/ELP entry after the rebase.

CLA: author is `maxwang <wang.yuchi.312@gmail.com>` (GitHub account Yu-0312). Please re-check `license/cla` after this push; happy to squash further if CLA assistant still flags commits.

Local checks: `ruff` + erlang unit tests → 7 passed, 14 skipped (live-server tests).

Yu-0312:
Per @opcode81 on CLA: history is now a **single squashed commit** authored as `maxwang <wang.yuchi.312@gmail.com>` (GitHub: Yu-0312).

Review status re-stated:
- `DownloadedDependency` + central hash DB; no PATH preference unless `ls_path`
- `cache_version_raw_document_symbols=2` (enough after main `a5fd4d68`)
- Rebased onto current `main`; unit tests: 7 passed / 14 skipped (live LS)

Could you re-review when you have bandwidth, and clear `waiting for OP` if this addresses the open points? Happy to adjust further.

mcgauleybrock508-cmyk:
Yes

PR #1912, opened 2026-08-24
Security/codeql research findings

KevinTran3011:
## Checklist

- [ ] This PR follows the guidelines in `CONTRIBUTING.md` regarding the scope of PRs.
- [ ] For changes that add features or fix problems, I have added an entry to `CHANGELOG.md`, which concisely describes the change.

### DIFF
--- a/src/solidlsp/ls_utils.py
+++ b/src/solidlsp/ls_utils.py
@@ -641,8 +641,25 @@ class FileUtils:
         with tarfile.open(archive_path, tar_mode) as tar_ref:
             for tar_member in tar_ref.getmembers():
                 FileUtils._validate_extraction_path(tar_member.name, target_path)
+                # A safe member-name check does not protect against links: a
+                # symlink/hardlink can redirect a later extraction outside the
+                # destination directory. Reject links and special files.
+                if (
+                    tar_member.issym()
+                    or tar_member.islnk()
+                    or not (tar_member.isdir() or tar_member.isfile())
+                ):
+                    raise SolidLSPException(
+                        f"Unsafe archive member '{tar_member.name}': "
+                        "links and special files are not allowed"
+                    )
 
-            tar_ref.extractall(target_path)
+            try:
+                tar_ref.extractall(target_path, filter="data")
+            except TypeError:
+                # Python < 3.12 has no filter argument. Links were already
+                # rejected above, so the legacy call remains confined.
+                tar_ref.extractall(target_path)
 
 
 class PlatformId(str, Enum):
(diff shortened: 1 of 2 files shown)


KevinTran3011:
Research note: the CodeQL tar-slip findings were CWE-22. Both extraction paths validated member names but passed symlinks/hardlinks/special files to tarfile.extractall(). The patch rejects links and non-file/non-directory members and uses filter="data" where available, with a safe legacy fallback after rejection.

src/solidlsp/ls_utils.py
```diff
@@ -641,8 +641,25 @@ def _extract_tar_archive(archive_path: str, target_path: str, archive_type: str)
         with tarfile.open(archive_path, tar_mode) as tar_ref:
             for tar_member in tar_ref.getmembers():
                 FileUtils._validate_extraction_path(tar_member.name, target_path)
+                # A safe member-name check does not protect against links: a
+                # symlink/hardlink can redirect a later extraction outside the
+                # destination directory. Reject links and special files.
+                if (
+                    tar_member.issym()
+                    or tar_member.islnk()
```
AmirF194, line 649: I checked this against `ea75eb0`, both branches, in a clean `python:3.12-slim` container with the package installed via `pip install -e .`.

The `issym()`/`islnk()` rejection is correct and closes a real hole: on `main` (pre-fix), `_is_safe_tar_member` returns `True` for a `tarfile.TarInfo` symlink member pointing at `/`, and `FileUtils._extract_tar_archive` extracts it without raising, so the archive can drop a symlink under the target directory that resolves outside it, unconstrained by the existing member-name containment check. On this branch, the same member is rejected in both places (`_is_safe_tar_member` returns `False`; `_extract_tar_archive` raises `SolidLSPException`).

Neither file's test suite covers this. `test/solidlsp/pascal/test_pascal_auto_update.py::TestTarfileSafety` only constructs plain-file `TarInfo` members (default `REGTYPE`), and `test/solidlsp/util/test_ls_utils.py` has no `issym`/`islnk` case either, so a later refactor of either check has nothing to catch a regression. This is the test I ran to verify the fix, parametrizable into either suite:

```python
def test_rejects_symlink_member(self) -> None:
    member = tarfile.TarInfo(name="evil")
    member.type = tarfile.SYMTYPE
    member.linkname = "../../etc/passwd"
    assert PascalLanguageServer._is_safe_tar_member(member, "/tmp/target") is False

def test_rejects_hardlink_member(self) -> None:
    member = tarfile.TarInfo(name="evil2")
    member.type = tarfile.LNKTYPE
    member.linkname = "some/other/target"
    assert PascalLanguageServer._is_safe_tar_member(member, "/tmp/target") is False
```

`ls_utils.py:648-654` would need the analogous case against `FileUtils._extract_tar_archive`, asserting it raises `SolidLSPException`.

PR #1914, opened 2026-08-24
feat(lua): add EmmyLua language server option

### LINKED ISSUE #690: Add emmylua as second option for Lua LS

Provide an LSP switching configuration for `lua`, but I'm not sure how the author views the scenario of having multiple LSPs for one language.

For `lua`, the most widely used LSP currently is `luals`, which is also the default LSP downloaded and activated. `luals` is derived from `emmylua`. 
However, we have completely rewritten `emmylua` using `rust`, offering more features (some of which cannot be achieved with luals), and it is the most widely used LSP in IntelliJ IDEA (thanks to the user base of the original emmylua project).

The author proposes a specification to handle situations where multiple LSPs exist for the same language.

Yu-0312:
Fixes #690

Add the experimental Rust-based EmmyLua Analyzer as the opt-in lua_emmylua backend. It uses ls_path as an explicit override and otherwise always performs the managed cross-platform download, verified via `DownloadedDependency` against the central URL-keyed SHA-256 hash database (custom `emmylua_ls_version` overrides download unverified by design). The existing lua backend remains unchanged.

Validation:
- uv run poe lint
- uv run poe type-check
- uv run pytest test/solidlsp/lua/test_emmylua_language_server.py test/solidlsp/test_ls_config.py -q
- Recomputed SHA-256 of all six pinned 0.25.1 release assets via scripts/update_downloaded_dependency_hashes.py; values match the previously pinned constants
- Managed-download smoke test (download, verify, extract, chmod, `emmylua_ls --version`) on macOS arm64


Noizyfish:
https://github.com/patrickchugh/terravision.git
rspnoizy@gmail.com is the New Canonical Email for NOIZYFISH.COM & NOIZY.AI. YOU MUST NOW ORGANIZE AND UPDATE ALL PREVIOUS EMAIL AND MOVE & ORGANIZE AL;L INFORMATION & DATA FROM THE M2ULTRA

Yu-0312:
The latest CI runs fail in `test/serena/test_jetbrains_launch_coordinator.py::TestLaunchAndWaitForPluginServer::test_launches_once_and_polls_until_the_server_is_reachable` (introduced by #1864), not in this PR's changes:

- Run on `catch-all (macos-latest)`: `AssertionError: assert 585023 == 2` on `sleep.call_count`
- Rerun on `catch-all (ubuntu-latest)`: `AssertionError: assert 307259 == 2`

**Update (Sep 5):** this is not specific to this PR — #1990's `catch-all (windows-latest)` failed the same morning with the identical signature (`assert 363603 == 2`), so the test is failing repo-wide across three OSes and multiple authors.

The test patches `serena.jetbrains.launch_coordinator.time.sleep`, which resolves to the shared `time` module, so `time.sleep` is mocked **globally for the whole process** while the coordinator's deadlines (`time.monotonic()`) and filelock's acquire loop stay on the real clock. filelock 3.25.2 busy-spins through `time.sleep` while an acquisition is contended, so any concurrent/lingering thread or lock contention during the mock window inflates `sleep.call_count` into the hundreds of thousands; `from_project` call counting (`calls["n"] == 4`) still passes, confirming the coordinator flow itself ran exactly as intended.

The test passes reliably locally (8 passed in 0.4s, also with the CI-preceding test files in the same process). Main's own CI has been hitting similar shared-runner flakiness this week (ccls timeouts in `native (ubuntu-latest)` on Sep 3 and again on #1990, the OCaml install step in `niche` on Sep 4). Happy to help fix the test upstream (e.g. patch `time.sleep` non-globally or use an injectable clock) if maintainers agree.

Yu-0312:
Pushed two follow-ups:

1. **Cache isolation** — `lua_emmylua` and `lua` share language_id `lua` (and therefore the same document-symbol cache directory). `_raw_document_symbols_cache_fingerprint` now returns the ls_id, so switching backends does not serve symbols produced by the other server.

2. Tests: `test/solidlsp/lua/test_emmylua_language_server.py` 18 passed; ruff clean.

Ready for re-review.

opcode81:
> lua_emmylua and lua share language_id lua (and therefore the same document-symbol cache directory)

Thank you for pointing this out; this is a bug. The cache directory should be unique to the language server and should therefore be based on the language server id's key.
I have fixed this with commit 602837b6e03d7bfa7873438f01b58457eaf75dab on main.

> _raw_document_symbols_cache_fingerprint now returns the ls_id, so switching backends does not serve symbols produced by the other server.

Therefore, please undo this change.

opcode81:
@Yu-0312 why did you close the PR?

Yu-0312:
Sorry, I was reading too quickly and misunderstood your comment, so I closed it by mistake. Totally my mistake. I'll reopen it and follow up on your request.

Yu-0312:
Done — thanks for the proper fix on main.

I removed the local `_raw_document_symbols_cache_fingerprint` override (`76b6a6c`) and merged upstream/main so the branch now picks up `602837b` (cache directory keyed by `ls_id`).

`test/solidlsp/lua/test_emmylua_language_server.py` and `test/solidlsp/test_ls_config.py`: 22 passed.

Yu-0312:
Rebased onto current `main`.

Per @opcode81: the document-symbol cache directory is no longer keyed by `language_id`; `main` already uses `ls_id.get_key()` (602837b). This branch no longer overrides `_raw_document_symbols_cache_fingerprint` / custom fingerprint logic for EmmyLua — that change was dropped as requested. `lua` and `lua_emmylua` therefore share the language id but not the cache directory.

Local checks: `uv run ruff check` on EmmyLua files + `pytest test/solidlsp/lua/test_emmylua_language_server.py` → 18 passed.

Yu-0312:
Per @opcode81: fingerprint override removed; cache isolation comes from main `602837b` (`ls_id.get_key()`). Branch is a **single squashed commit** on current `main`. EmmyLua unit tests: 18 passed. Ready for re-review.

PR #1915, opened 2026-08-24
feat(cli): prefetch language server dependencies

### LINKED ISSUE #664: Run selena in a container without egres

Yu-0312:
Fixes #664

Add download-ls-dependencies to prefetch runtime dependencies for selected language servers or all non-experimental servers with --all. uvx providers now support persistent pre-installation, while custom ls_path and ls_base_cmd settings are respected.

Validation:
- uv run poe lint
- uv run poe type-check
- uv run pytest test/serena/test_cli_download_ls_dependencies.py test/solidlsp/test_dependency_provider.py -q


src/serena/cli.py
```diff
@@ -229,6 +248,71 @@ def setup(client: str) -> None:
...
+        "-a",
+        "all_language_servers",
+        is_flag=True,
+        help="Download dependencies for all non-experimental language servers.",
+    )
+    @click.option(
+        "--include-experimental",
+        is_flag=True,
+        help="Include experimental language servers when used with --all.",
+    )
```
opcode81: Remove these options; they are detrimental.
Users should explicitly install the servers they need only.

src/solidlsp/ls.py
```diff
@@ -616,6 +616,15 @@ def _get_dependency_provider(self) -> LanguageServerDependencyProvider:
             self._dependency_provider = self._create_dependency_provider()
         return self._dependency_provider
 
+    def install_dependencies(self) -> None:
+        """Install all runtime dependencies managed by this language server."""
+        # Providers are normally created lazily when a launch command is needed. The
+        # prefetch CLI calls this method before launching, so create the provider here
+        # for dependency-managed servers. Legacy servers with a fixed launch command
+        # do not have a provider and intentionally require no prefetch work.
+        if self._process_launch_info is None:
```
opcode81: This condition is never true.

Yu-0312:

PR #1916, opened 2026-08-24
feat(hooks): add enforce mode for symbolic tools

### LINKED ISSUE #1470: serena-hook enforce — hard-block variant with directive stop messages and configurable extensions
## Background

I've been running Serena with `serena-hook remind` and found myself wanting a
stricter mode: rather than nudging Claude after N consecutive raw reads/greps,
I wanted every direct code-file access blocked up front, with the exact
Serena replacement call pre-filled so Claude can retry immediately without
guessing the right tool or arguments.

I built this as a standalone hook: **https://github.com/Lenucksi/serena-guard**

## What it does differently from `serena-hook remind`

| | `serena-hook remind` | serena-guard |
|---|---|---|
| Block strength | Soft (nudge after N calls, then allow) | Hard (every call, no exceptions) |
| Stop message | Generic reminder to use symbolic tools | Exact replacement call with params pre-filled |
| Extension list | Hardcoded `frozenset` in source | `extensions.yaml` — edit without touching code |
| LSP awareness | No | Per-extension LSP plugin name in stop message |
| Bash parsing | Shell-command classification | `shlex` parse: grep/find/rg targeting code files |

The directive stop message is the core differentiator. Instead of
"consider using Serena's symbolic tools", Claude receives:

```
⛔ FORBIDDEN — Read on code files. SERENA: full support.  [.ts]
Target: /project/src/auth/token.ts

USE THIS:
  mcp__plugin_serena_serena__read_file(
    relative_path="src/auth/token.ts", start_line=120, end_line=159
  )
  — OR for a named symbol —
  mcp__plugin_serena_serena__find_symbol(name_path="SymbolName", include_body=True)

LSP (typescript-lsp@claude-plugins-official): hover · incomingCalls · outgoingCalls
```

The path, line range, and pattern are taken directly from the intercepted
tool call's parameters, so the replacement call is ready to copy-execute.

## What might be worth folding into the existing hook system

Not proposing a full PR yet — raising this to see if the maintainers find
any of these ideas worth integrating:

1. **Directive stop messages** — the existing deny messages are advisory;
   filling in the concrete Serena call from the blocked attempt's parameters
   seems generally useful, even in the soft-nudge variant.

2. **Configurable extension list** — `_CODE_FILE_EXTENSIONS` in
   `PreToolUseRemindAboutSymbolicToolsHook` (and the analogous set in
   `PreToolUseEnforceSymbolicToolsHook` if that's ever added) is currently a
   hardcoded `frozenset`. A small YAML/TOML config alongside `serena_config.yml`
   would let users add language support without touching source.

3. **LSP plugin awareness** — if the extension list carries metadata about which
   LSP plugin covers a given file type, the stop message can name the specific
   LSP plugin, making it easier to use `hover` / `incomingCalls` etc.

4. **`serena-hook enforce`** — a hard-block command as a first-class option
   alongside `remind`, for setups where the soft nudge isn't strict enough.

Happy to work up a PR for any of these if there's interest. Alternatively,
the standalone repo is there for users who want the hard-block behaviour now.

Yu-0312:
Fixes #1470

Add the opt-in serena-hooks enforce mode. It denies direct code-file reads and grep calls and includes a ready-to-use Serena read_file or search_for_pattern replacement in the denial reason. SERENA_HOOK_CODE_FILE_EXTENSIONS allows the source suffix set to be customized.

The remind hook's deny-threshold behaviour is unchanged. `_iter_shell_path_arguments` (shared with enforce) now skips common search-flag values such as `--type py` / `-g` so the enforce retry suggestion does not mistake a flag value for the search pattern.

Validation:
- uv run poe lint
- uv run poe type-check
- uv run pytest test/serena/test_hooks.py -q


Yu-0312:
Related review-context note: PR #1928 also changes `src/serena/hooks.py` and `test/serena/test_hooks.py`, while this PR adds enforce mode. The features are distinct, but both touch the same hooks implementation and tests, so merge order or conflict resolution may be relevant. This is not a supersedes or dependency claim.

src/serena/hooks.py
```diff
@@ -527,6 +539,66 @@ def _build_non_symbolic_deny(self) -> "PreToolUseHook.OutputData":
...
+                )
+            arguments = [f"relative_path={json.dumps(file_path)}"]
+            if self._tool_input is not None:
+                for key in ("start_line", "end_line", "max_answer_chars"):
+                    value = self._tool_input.get(key)
+                    if value is not None:
+                        arguments.append(f"{key}={json.dumps(value)}")
+            return f"mcp__serena__read_file({', '.join(arguments)})"
+
+        pattern = "<pattern>"
```
AmirF194, line 567: I ran this against 1ab567c9 in a clean python:3.12-slim container.

For a Codex/Grok shell-command grep call, `_replacement_call()` never reads `_command_args_str` (the read branch above it does, via `_iter_shell_path_arguments()`), so the suggested retry drops the actual pattern and path.

Instantiating the hook directly with a Codex `cmd` payload `"rg -n needle src/foo.py"` produces:

```
Retry with: mcp__serena__search_for_pattern(substring_pattern="<pattern>")
```

The same payload with `cat src/foo.py` correctly yields `relative_path="src/foo.py"`, since only the read branch has the shell-argument fallback. `test_codex_omits_additional_context` covers a Codex grep call but only asserts the tool name is present in the reason, not that the pattern/path were recovered, so this doesn't show red.

Fix: give the grep branch the same `_iter_shell_path_arguments()` fallback the read branch has when `_tool_input` carries no `pattern`/`relative_path`.

AmirF194:
Confirmed, this fixes it. The shell-argument fallback on the grep branch matches what the read branch already does, and the new test pins the pattern/path recovery.

Yu-0312:
Pushed a fix for shell pattern extraction in the enforce-hook retry suggestion.

`rg`/`grep` flags such as `--type` / `-g` / `-e` take a separate value token. Previously `rg --type py needle src/foo.py` treated `py` as the search pattern. `_iter_shell_path_arguments` now skips known flag values (both `--type=py` and `--type py` forms).

- New test: `test_codex_grep_skips_flag_values_when_extracting_pattern`
- `test/serena/test_hooks.py`: 104 passed
- ruff check/format clean

Ready for re-review.

Yu-0312:

PR #1918, opened 2026-08-24
fix(process): reap language server descendants

### LINKED ISSUE #1464: Clojure LSP may leave zombies behind
Not sure what is up with that, will investigate

Image

Yu-0312:
Fixes #1464

After signaling a language-server process tree, wait for the discovered descendants before considering cleanup complete. Use one bounded deadline and retain the existing process-group path. Add a regression test covering signal and descendant wait order.

Validation:
- uv run poe lint
- uv run poe type-check
- uv run pytest test/solidlsp/test_process_group_cleanup.py -q

## Follow-up: process-group descendant cleanup

This follow-up addresses the review feedback about descendants surviving when a `start_new_session=True` language-server leader exits immediately after `SIGTERM`.

The cleanup now snapshots descendants unconditionally before signaling. It still uses `os.killpg` for a known process group, preserving the macOS-safe path, but it also waits for the saved descendants. If the leader exits while a descendant remains alive, the shared timeout expires and the existing group `SIGKILL` fallback terminates the remaining processes. Process-table discovery remains best-effort: `psutil.Error` and `OSError` still degrade to an empty snapshot.

Added regression coverage for a process-group leader that exits on `SIGTERM` while its child ignores `SIGTERM`; the test verifies that the child is removed by the fallback. Local validation: `ruff` and `ty` pass, and the process cleanup plus pdeathsig tests pass (`15 passed`).

## Follow-up: reap the leader after the kill budget

The latest review identified one more cleanup edge case. When descendants consumed the full two-second kill-fallback budget, the previous implementation returned before calling `process.wait()` for the language-server leader. The fallback now always calls `process.wait()` with the remaining time clamped to a small positive value (`0.1s`), so an already-killed leader is still reaped and the function does not leave a zombie leader behind.

Added regression coverage for descendants consuming the kill budget and verified the existing process-group and pdeathsig behavior. Local validation: `ruff` and `ty` pass, and the cleanup plus pdeathsig tests pass (`16 passed`).


src/solidlsp/util/subprocess_util.py
```diff
@@ -359,22 +398,27 @@ def terminate_process_tree_with_kill_fallback(
...
         if process_group_id is not None:
             _signal_process_group(process_group_id, terminate=terminate)
         else:
-            _signal_process_tree(process, terminate=terminate)
+            _signal_process_tree(process, terminate=terminate, descendants=None if terminate else descendants)
 
     signal_tree(terminate=True)
     try:
         log.debug(f"Waiting for process {process.pid} to terminate...")
+        descendants_finished = _wait_for_processes(descendants, terminate_timeout)
```
ebarkhordar: I ran this against 0b5eceb5 in a clean python:3.11-slim container. The descendant reaping works. One thing worth weighing before it merges: `terminate_timeout` is now spent twice in sequence, so the terminate phase can block for about double what the caller asked for.

Line 412 gives the descendant wait the full `terminate_timeout`, then line 415 gives `process.wait` the full `terminate_timeout` again. When the descendants exit late but still inside the budget, the leader gets a fresh budget on top of what they used.

I gave a leader and one child that both ignore SIGTERM, with the child exiting on its own at 4.5s and the leader at 9.0s, then called the function with `terminate_timeout=5.0`, the value `src/serena/agent.py:1307` passes (that caller passes no `process_group_id`, so it takes the snapshot path):

```
# PR head, subprocess_util.py sha256 6de4ff15...
RESULT terminate_timeout=5.0  child_exits_at=4.5s  leader_exits_at=9.0s
RESULT terminate phase blocked for 8.72s

# same script, merge-base 7fcbca7e, sha256 dca188dc...
WARNING leader (pid=15) termination timed out, killing process forcefully...
RESULT terminate phase blocked for 5.01s
```

The kill fallback has the same shape: lines 421 and 423 each get their own 2.0s, so that phase can run to 4s.

`_wait_for_processes` already computes one shared deadline internally, so the fix is to hoist that same idea one level up: take `deadline = monotonic() + terminate_timeout` before the descendant wait and pass `deadline - monotonic()` to `process.wait`.

I did not test Windows, and I did not test the `process_group_id` path, which sets `descendants` to the empty list and is therefore unaffected.

src/solidlsp/util/subprocess_util.py
```diff
@@ -359,24 +402,36 @@ def terminate_process_tree_with_kill_fallback(
         also signal us.
     """
     log.debug(f"Terminating process {process.pid}, current status: {process.poll()}")
+    descendants = [] if process_group_id is not None else _get_process_descendants(process)
```
AmirF194: I ran this against `822330d` in a clean Docker container (python:3.12-slim, psutil 7.0.0, oslex 0.1.3, sensai-utils 1.5.0, the exact runtime deps).

`descendants` is hardcoded to `[]` whenever `process_group_id` is set, so the new wait-for-descendants logic never runs on that path, only on the psutil tree-walk path. That path is exactly what `start_new_session=True` uses. I spawned a leader (own session, honors SIGTERM) with a child and grandchild that both ignore SIGTERM, then called `terminate_process_tree_with_kill_fallback(proc, terminate_timeout=2.0, process_group_id=proc.pid)`. The leader exits immediately, so `process.wait()` never times out, so the kill fallback in the except block never fires either. Result: the child and grandchild are still running after the call returns, not reaped, not killed.

```
descendants before terminate: [(1020455, 'python'), (1020456, 'python')]
terminate_process_tree_with_kill_fallback returned after 0.00s
descendants still present after cleanup call: [(1020455, 'sleeping'), (1020456, 'sleeping')]
```

This predates this PR (main has the same gap), but the changelog entry here claims cleanup now waits for discovered descendants, and that is not true for the group-id path, which is the one pdeathsig-launched servers use. `_get_process_descendants` already degrades to `[]` on `psutil.Error`, so calling it unconditionally instead of skipping it for the group case would fix both paths without reintroducing the macOS access-denied problem this PR's docstring describes.

src/solidlsp/util/subprocess_util.py
ebarkhordar: Re-ran at `51ca4950` in a clean container with `psutil==7.0.0` pinned to match pyproject.toml.

The shared deadline works. With the descendant reaped promptly so the descendant wait can succeed, and `terminate_timeout=5.0`, the terminate phase is 5.01s at the merge base, 7.76s at 0b5eceb5, 5.00s at 51ca4950.

The early return at line 431 does leave the leader unreaped though. When descendants outlive the 2s kill budget, line 429 spends all of it, so the function returns before `process.wait` at 435. The kill lands, nothing reaps the leader, and the log says the opposite:

```
# descendant exits at 4.5s and is never reaped, leader ignores SIGTERM, terminate_timeout=5.0
0b5eceb5   leader after return: GONE     returncode -9
51ca4950   leader after return: zombie   returncode None   ERROR "could not be killed within timeout"
```

Both took 7.03s, so it is the return skipping the reap and not timing. That came out of the deadline hoist I suggested, so it is mine rather than yours, and it turns a PR about leaked zombie descendants into one that can leak the zombie leader. Reaping something you already killed is cheap:

```python
exit_code = process.wait(timeout=max(remaining, 0.1))
```

One correction while I am here. My earlier numbers ran on unpinned psutil, which resolved to 7.2.2, with a descendant that was never reaped. In that setup the descendant wait never succeeds, so the old line 415 was never reached and the double spend is not what I measured. The conclusion held, the evidence I gave for it did not.

I did not test Windows or the `process_group_id` path.

Yu-0312:
Thanks for reproducing this so precisely. I was able to confirm the gap: when `process_group_id` is set, the previous code used `descendants = []`. If the group leader exited immediately after `SIGTERM`, the leader's `wait()` returned successfully and the cleanup never waited for or force-killed the still-running child and grandchild.

I changed the cleanup to take a best-effort descendant snapshot unconditionally before signaling. The process-group path still uses `os.killpg`, so it does not reintroduce the macOS process-table enumeration problem, but it now waits for the saved descendants and reaches the group `SIGKILL` fallback when they remain alive.

I also added `test_group_cleanup_kills_descendant_after_leader_exits`, which reproduces the leader-exits/child-ignores-SIGTERM case and verifies that the descendant is cleaned up. The change is in commit `51ca4950`. Local `ruff` and `ty` checks pass, and the cleanup plus pdeathsig tests pass (`15 passed`).

src/solidlsp/util/subprocess_util.py
AmirF194: Confirmed, that matches the gap: descendants was hardcoded to the empty list on the process_group_id path, so the wait-for-descendants logic never ran there. Thanks for the fix.

src/solidlsp/util/subprocess_util.py
ebarkhordar: Confirmed at `a4cff181` in a clean python:3.12-slim container, `psutil==7.0.0` pinned to match pyproject.toml. Same script as before: the descendant exits at 4.5s and is never reaped, the leader ignores SIGTERM, `terminate_timeout=5.0`.

```
51ca4950   leader after return: zombie   returncode None   ERROR "could not be killed within timeout"
a4cff181   leader after return: GONE     returncode -9     no error line
```

Both returned after 7.07s, so it is the reap that changed and not the timing. The false error line is gone with it. I did not test Windows.

Yu-0312:
Thanks for catching this. You are right that the early return after the descendant kill wait can leave the leader unreaped. If descendants consume the full two-second fallback budget, the previous code returned before reaching `process.wait()`, even though the leader had already been killed.

I removed that early return. The fallback now always calls `process.wait()` with `max(remaining, 0.1)`, allowing the already-killed leader to be reaped while preserving the shared deadline for the normal path. I also added `test_kill_fallback_reaps_leader_after_descendant_wait_expires`, which verifies that descendants can consume the fallback budget without skipping the leader reap.

The fix is in commit `a4cff181`. Local `ruff` and `ty` checks pass, and the process cleanup plus pdeathsig tests pass (`16 passed`).

opcode81:
Thanks for addressing this, @Yu-0312, and thanks to @AmirF194 and @ebarkhordar for their detailed reviews and analyses.

I am confident that this is already in a good state. The only thing I am unsure about is the narrowing of except clauses to use explicitly enumerated types of exceptions rather than the catch-all `except:`. We deliberately used the catch-all clause in several places, because these helper functions must really *never* raise an exception interrupting the original callers' control flow. Are you confident that it is impossible for other exceptions to be raised? If not, I would suggest that catch-all clauses be used. (Contrary to IDE and linter recommendations, I do not regard this as an anti-pattern if used deliberately.)

Yu-0312:
Thanks for the careful review, and I’m glad the overall direction looks good.

I agree that the catch-all handling is intentional here. These cleanup helpers are best-effort boundaries whose contract is to avoid interrupting the original caller, even when process discovery, signaling, or waiting encounters an unexpected platform- or psutil-specific failure. For that reason, narrowing every boundary to an enumerated exception list could make cleanup less robust as new OS or library failure modes appear.

I’ll keep the broad `Exception` handling at those cleanup boundaries, while deliberately not catching `BaseException`, so `KeyboardInterrupt` and `SystemExit` still propagate normally. No further functional change is needed from my side.

Issue #1040, opened 2026-02-14
Fix & re-enable flaky tests in CI


opcode81:
Many tests had to be disable to render CI tests reliable.

The following tests were disabled and we should investigate whether we can "fix" something and re-enable them at some point:

* Phpactor tests on Windows:
* TestFSharpLanguageServer::test_find_referencing_symbols_across_files

MischaPanch:
The nix error is an actual bug, it always fails. I briefly investigated at some point but didn't find the solution then, and I never prioritized this. 

For referencing symbols in F# we could maybe just bump the wait time

MischaPanch:
Php-tractor failing on windows is also not flaky but rather an actual bug

opcode81:
> The nix error is an actual bug, it always fails.

If that's the case, then why was it written and enabled?

> Php-tractor failing on windows is also not flaky but rather an actual bug

No, some of them pass on my computer (and also in CI), but *all* of them *can* fail.

> For referencing symbols in F# we could maybe just bump the wait time

Try it if you want. I don't have time for this.

Anyway, this ticket is there to re-investigate the respective issues. With #1041, I am just disabling them.

antigenius0910:
While working on PR #1114 (enabling Rego tests on Windows), we noticed a broader FSharp regression on Windows CI that goes beyond the specific flaky tests tracked here.

**All** FSharp tests are currently failing on Windows with the following error from FsAutoComplete:

```
[FsAutoComplete.Lsp.AdaptiveState] Typecheck failed for ...\Calculator.fs with Check aborted (Aborted).
Errors: [|"The type referenced through 'System.Array' is defined in an assembly that is not referenced.
You must add a reference to assembly 'The system type 'System.Array' was required but no referenced
system DLL contained this type'."|]
```

**Tests now failing** (beyond those already disabled in #1041):
- `TestFSharpLanguageServer::test_find_symbol`
- `TestFSharpLanguageServer::test_get_document_symbols_program`
- `TestFSharpLanguageServer::test_get_document_symbols_calculator`
- `TestFSharpLanguageServer::test_find_referencing_symbols`
- `TestFSharpLanguageServer::test_nested_module_symbols`
- `TestFSharpLanguageServer::test_go_to_definition`
- `TestFSharpLanguageServer::test_completion`
- `TestSerenaAgent::test_find_symbol_references[fsharp-...]`

[...]

PR #1921, opened 2026-08-24
fix(nix): retry transient hover responses

Yu-0312:
Addresses #1040

## Summary

Nixd can return an empty (`null`) hover response while its initial analysis is still completing. This change retries empty hover responses within a bounded interval, while preserving `null` for genuinely unsupported positions.

The integration coverage now uses a dedicated minimal fixture containing `with pkgs; hello`, matching nixd's supported nixpkgs-backed hover-provider form. Keeping this case separate from the larger `default.nix` fixture avoids changing its line layout, which is relied on by existing reference, definition, and snapshot tests. The previous `utils` cross-file reference remained unsupported by nixd and therefore could not demonstrate that the retry fixed the reported failure; keeping it as a non-blocking `xfail` made the PR's evidence weaker than its title and changelog. The test now dynamically locates the supported package selector and asserts that hover information is returned.

The unit coverage also verifies both sides of the retry contract: an empty first response is retried with the identical LSP request and succeeds when the next response is available, while a permanently empty response stops after the configured retry bound.

This addresses the disabled Nix hover coverage tracked by #1040 without claiming that nixd provides documentation for every local or cross-file binding.

## Validation

- `uv run poe lint`
- `uv run poe type-check`
- `uv run pytest test/solidlsp/nix/test_nix_server.py test/solidlsp/nix/test_nix_basic.py -q`

In the local environment, the Nix integration tests are skipped because `nixd` is not installed; the Nix unit tests pass (`19 passed`). The GitHub Nix CI job provides the end-to-end validation with the configured nixd installation.


test/solidlsp/nix/test_nix_basic.py
```diff
@@ -118,14 +118,18 @@ def test_find_references_within_file(self, language_server: SolidLanguageServer)
             # Check if we found the inherit (line 67, 0-indexed: 66)
             assert 66 in ref_lines, f"Should find makeGreeting inherit at line 67, found at lines {[l + 1 for l in ref_lines]}"
 
-    @pytest.mark.xfail(is_ci, reason="Test is flaky")  # TODO: Re-enable if the hover test becomes more stable (#1040)
     @pytest.mark.parametrize("language_server", [LanguageServerId.NIX], indirect=True)
+    @pytest.mark.xfail(
+        strict=False,
+        reason="Current nixd returns no hover for this fixture binding; keep the regression covered until upstream support is available",
```
AmirF194: I read this PR's diff and the discussion on the linked issue, and did not run nixd myself.

`test_hover_information` was moved to a different fixture position (`23, 13`, the `utils` reference) but is still marked `xfail(strict=False)`, with the reason stating nixd currently returns no hover there. That matches opcode81's comment on #1040 ("The nix error is an actual bug, it always fails"), not a startup-timing description, and the CHANGELOG entry and PR title both frame this as fixing a transient first-response case.

```
issue #1040 (opcode81): "The nix error is an actual bug, it always fails."
test/solidlsp/nix/test_nix_basic.py:122-124 (this PR):
    @pytest.mark.xfail(
        strict=False,
        reason="Current nixd returns no hover for this fixture binding; keep the regression covered until upstream support is available",
    )
```

`_request_hover` in `nixd_ls.py` only re-issues the same request after a fixed delay, and the PR's own regression test still cannot assert success at either the old or the new position. Does the retry actually resolve the hover failure reported in #1040, or does it cover a separate, genuinely transient case while the always-empty hover opcode81 described is still open? If the latter, the CHANGELOG line reads as a stronger claim than the diff's own test evidence supports.

Yu-0312:
Thanks for the careful review. You are right that the previous regression test did not demonstrate that the retry fixed the failure reported in #1040: it remained `xfail`, and the `utils` binding returned no hover even after the retry window.

I revised the change to make the distinction explicit. The retry is for transient empty responses while nixd is completing its initial analysis; it does not make unsupported local or cross-file bindings return hover documentation. The integration test now targets a direct nixpkgs-backed selector (`pkgs.bash`) that follows nixd's supported hover-provider path, locates it dynamically in the fixture, and asserts a non-empty hover result. The fixture no longer relies on the unsupported `utils` reference for this regression.

I also strengthened the unit coverage to verify that the identical LSP request is retried after an empty response and that a permanently empty response stops after the configured bound. The CHANGELOG and PR description now use the narrower, evidence-supported wording rather than implying that every hover failure in #1040 is fixed.

Validation completed:

- `uv run poe lint`
- `uv run poe type-check`
- `uv run pytest test/solidlsp/nix/test_nix_server.py test/solidlsp/nix/test_nix_basic.py -q` (`19 passed, 13 skipped locally because nixd is not installed)

The Nix integration path remains enabled and will be exercised by the CI runner's nixd installation.

Yu-0312:
Follow-up after the second CI run: the shared fixture is now unchanged, and all existing Nix tests pass. The remaining failure showed that even the nested `with pkgs; hello` expression in `default.nix` was reported by nixd as `hover/idiom/selector: not an idiom`.

I traced this to the expression's surrounding function/attribute-set context. The hover integration test now uses a separate minimal `hover.nix` fixture containing exactly `with pkgs; hello`, matching nixd's own supported package-hover form. This keeps the existing fixture line layout and snapshots untouched while exercising the nixpkgs-backed hover provider directly. The test still locates `hello` dynamically and asserts a non-empty hover result.

The retry remains intentionally scoped to transient empty responses while nixd is completing its initial analysis; it does not make unsupported local or cross-file bindings return hover documentation. The unit coverage verifies identical-request retry behavior and the bounded stop condition, and the CHANGELOG / PR description use the narrower, evidence-supported wording.

Local validation remains clean:

- `uv run poe lint`
- `uv run poe type-check`
- `uv run pytest test/solidlsp/nix/test_nix_server.py test/solidlsp/nix/test_nix_basic.py -q` (`19 passed, 13 skipped locally because nixd is not installed)

The latest fix adds the dedicated minimal fixture and is ready for another Nix CI run.

test/solidlsp/nix/test_nix_basic.py
AmirF194: Thanks, this addresses it. The narrower CHANGELOG wording and the new hover.nix fixture match what the test now actually demonstrates.

Yu-0312:

PR #1925, opened 2026-08-24
fix(config): persist removed registered projects

### LINKED ISSUE #1895: remove_project() does not persist: the removed project silently reappears on the next config reload

Yu-0312:
Fixes #1895

`remove_project` previously removed a project only from the current SerenaConfig instance. The merge-on-save logic then reloaded the on-disk list and silently re-added the removed project. Track the instance baseline so removals are applied to the latest disk state while concurrent additions and removals from other agents remain preserved.

Added reload and concurrent add/remove regression tests.

Validation:
- `uv run pytest test/serena/config/test_serena_config.py -q`


opcode81:
Thanks for your PR. This fixes a regression introduced by an unreleased change (8a4d116c13bb809ad3b8cc9071d28e8ed9f4f6df), so nothing should be added to the changelog.

AmirF194 (commented):
I checked this against the bug directly: copied this branch's test file onto a clean checkout of main (43ae0211), without the fix, and ran it.

All three new tests fail on main, so the reappearing-project bug is real. The same file passes on this branch's head, and the existing test_parallel_agents_changing_project_lists still passes too, so the concurrent-list merge does not regress.

```
main (43ae0211) + this branch's test file: 3 failed
this branch (308f3a9a): 4 passed, including test_parallel_agents_changing_project_lists
```

Ran both in a clean python:3.11-slim container with the pinned runtime deps from pyproject.toml, PYTHONPATH pointed at each worktree's src.

Yu-0312:
Thanks for verifying this against a clean main checkout and for confirming that the existing concurrent-project-list test still passes. I’m glad the comparison confirms this PR fixes the reappearing-project regression without affecting the concurrent-list behavior. No further changes are needed from my side.

PR #1926, opened 2026-08-24
fix(nextflow): retry failed workspace scan flushes

### LINKED ISSUE #1871: Nextflow LS: _flush_deferred_workspace_scan marks the scan flushed even when both flush attempts fail

Yu-0312:
Fixes #1871

Nextflow marked its deferred workspace scan as flushed even when both synchronous completion requests failed. Subsequent reference queries therefore skipped every retry and could use an incomplete AST cache. Mark the scan flushed only after at least one request succeeds, retain retry behavior after total failure, and log the non-success.

Added deterministic regression tests for all-fail retry and partial success.

Validation:
- `uv run pytest test/solidlsp/nextflow/test_nextflow_scan_flush.py test/solidlsp/nextflow/test_nextflow_basic.py -q`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ Status of the `main` branch. Changes prior to the next official version change w
     project list in `serena_config.yml`
 
 * Language Servers:
+  - Fix: Nextflow workspace scans were marked flushed even when both flush requests failed, preventing
+    later reference queries from retrying (#1871)
   - Fix: Dart's `$/analyzerStatus` notifications were logged as unhandled-method warnings during analysis (#1855)
   - Fix: clojure-lsp was not told that Serena sends `workspace/didChangeWatchedFiles`, so changes made
     outside Serena's own edit tools (a git checkout, another editor, a build step) need not invalidate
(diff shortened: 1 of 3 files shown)


opcode81:
Duplicates #1872

PR #1927, opened 2026-08-24
fix(hooks): emit Codex cleanup JSON response

### LINKED ISSUE #1533: Codex Stop hook: `serena-hooks cleanup --client=codex` emits invalid Stop-hook stdout

Yu-0312:
Fixes #1533

Codex Stop hooks require valid JSON on stdout, but `serena-hooks cleanup --client=codex` previously emitted empty stdout on normal input and a traceback when `session_id` was missing. Codex cleanup now always emits the non-blocking `{"continue":true}` response on stdout; cleanup diagnostics remain on stderr, and other clients retain their existing behavior.

Added CLI regression tests for normal and incomplete Codex hook input and documented the contract.

Validation:
- `uv run pytest test/serena/test_hooks.py -q`
- `uv run poe doc-build`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ Status of the `main` branch. Changes prior to the next official version change w
     independently via the `window/logMessage` notification tsserver already sends, and the
     affected wait now raises instead of reporting success (#1814)
 
+* CLI:
+  - Fix: Codex cleanup hooks now always emit a valid non-blocking Stop-hook JSON response, including
+    when the session input is incomplete (#1533)
+
 * Dependencies:
   - Remove the redundant `dotenv` dependency; the `dotenv` module is provided by `python-dotenv`
 
--- a/docs/02-usage/030_clients.md
+++ b/docs/02-usage/030_clients.md
@@ -370,10 +370,11 @@ Then create `~/.codex/hooks.json` with the following content:
 ```
 
 The `SessionEnd` cleanup hook requires Codex 0.145.0 or newer. Older Codex versions only support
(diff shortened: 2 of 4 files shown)


Yu-0312:
Related review-context note: PR #1928 also changes `src/serena/hooks.py` and `test/serena/test_hooks.py`, while this PR changes Codex cleanup JSON output. The behaviors are independent, but both touch the same hooks implementation and tests, so merge order or conflict resolution may be relevant. This is not a supersedes or dependency claim.

Yu-0312:
Closing as a duplicate: #1540 and #1738 are both older, still-open PRs fixing the same issue (#1533) with the same approach — always emitting valid Stop-hook JSON (`{"continue": true}`) from `serena-hooks cleanup --client=codex` and tolerating a missing `session_id` instead of raising. Keeping three parallel threads for one fix open makes review harder, so I'm withdrawing the newest one.

Happy to help review or test #1738 or #1540 instead.

PR #1919, opened 2026-08-24
feat(hooks): add DSH client

Yu-0312:
Fixes #1869

Add DeepSeek Harness (DSH) as a Claude-compatible serena-hooks client. DSH can use activate, remind, and cleanup with the existing hookSpecificOutput permission payload. Auto-approve is intentionally unavailable: the DSH bridge does not treat Claude Code's `allow` decision as pre-approval, so that hook stays silent for DSH.

Native `Bash` search/read commands are classified the same way as Claude Code's Grep/Read tools, gated on the `Bash` tool name so an unrelated MCP tool that merely has a `command` parameter is not counted (same concern raised in #1928). Shell path arguments are inspected for Bash so `cat README.md` is not treated as a code-file read.

Add DSH documentation and regression coverage.

Validation:
- uv run poe lint
- uv run poe type-check
- uv run pytest test/serena/test_hooks.py -q (101 passed)

Related: #1928 also changes hooks.py and adds a fuller shlex-based parser for Claude Code/CodeBuddy. When both PRs merge, DSH should keep inheriting native Bash classification.

PR #1928, opened 2026-08-24
fix(hooks): classify native shell reads and searches

### LINKED ISSUE #1845: serena-hooks remind --client=claude-code is blind to grep/read on native macOS/Linux builds (Claude Code 2.1.117+)
Preconditions:

- [x] I have consulted the user guide and verified that the issue cannot be resolved by adjusting configuration/following recommended workflows.
- [x] I have looked for similar issues and discussions, including closed ones. (#1394 is the same bug for Codex; #1470 and #1491 are related. Nothing covers Claude Code.)

Issue details:

- [x] Meaningful title and description.
- [x] Explanation of how it arose, with a copy-pasteable reproduction.
- [x] Setup details below.

---

## Summary

Claude Code **2.1.117** removed `Grep` and `Glob` as distinct tools on native macOS/Linux builds, routing them through the `Bash` tool as embedded `bfs`/`ugrep`. `is_grep_call()` matches `HookClient.CLAUDE_CODE` on tool name only and does no shell-command parsing, so on those builds the grep counter can never fire, and every search or file read the model routes through `Bash` is invisible to `serena-hooks remind`.

This is #1394 again for a different client. The shell-parsing machinery that fixed it for Codex already exists and already runs on Claude Code payloads — only the dispatch in `is_grep_call()` / `is_read_call()` excludes it.

## Reproduction

Two runs of four identical payloads, differing only in `tool_name`. `_GREP_USES_THRESHOLD` is 3, so both should deny on the third call.

```sh
run() { sid="$1"; name="$2"; for i in 1 2 3 4; do
  out=$(jq -nc --arg s "$sid" --arg n "$name" \
    '{session_id:$s,hook_event_name:"PreToolUse",tool_name:$n,
      tool_input:{command:"rg SomeSymbol src/"}}' \
    | serena-hooks remind --client=claude-code 2>&1)
  printf '%s' "$out" | grep -q "Too many consecutive" && echo "  call $i: DENY" || echo "  call $i: allowed"
done; rm -rf ~/.serena/hook_data/"$sid"; }

echo "A. tool_name=Bash + rg command   (what Claude Code 2.1.117+ sends)"; run "repro-a-$$" Bash
echo "B. tool_name=Grep                (what it sent before 2.1.117)";    run "repro-b-$$" Grep
```

Actual:

```
A. tool_name=Bash + rg command   (what Claude Code 2.1.117+ sends)
  call 1: allowed
  call 2: allowed
  call 3: allowed
  call 4: allowed
B. tool_name=Grep                (what it sent before 2.1.117)
  call 1: allowed
  call 2: allowed
  call 3: DENY
  call 4: allowed
```

B shows the counter, thresholds, persistence and deny path all working. A is the same search expressed the way Claude Code now expresses it, and it is never counted.

## The upstream change

Claude Code CHANGELOG, 2.1.117:

> Native builds on macOS and Linux: the `Glob` and `Grep` tools are replaced by embedded `bfs` and `ugrep` available through the Bash tool — faster searches without a separate tool round-trip (Windows and npm-installed builds unchanged)

Two follow-ups confirm this is the steady state rather than a transient:

- 2.1.119 — "Fixed Glob and Grep tools disappearing on native macOS/Linux builds when the Bash tool is denied via permissions"
- 2.1.162 — "`--tools`: explicitly listing Grep/Glob now provides the dedicated search tools on native builds with embedded search (previously these names were silently ignored)"

Native install is the default distribution channel on macOS and Linux, so this is the common case, not an edge one.

## The code

`src/serena/hooks.py` @ `main` (identical in 1.7.0):

```python
def is_grep_call(self) -> bool:
    if self._client in (HookClient.CLAUDE_CODE, HookClient.CODEBUDDY):
        return self._tool_name == "grep" or "search_for_pattern" in self._tool_name
    if self._client == HookClient.GROK:
        return self._tool_name == "grep" or (self._is_shell_command_call() and self._command_name in self._GREP_SHELL_COMMANDS)
    if self._client == HookClient.CODEX and self._is_shell_command_call():
        return self._command_name in self._GREP_SHELL_COMMANDS
    # heuristic for other clients
    return "grep" in self._tool_name

def is_read_call(self) -> bool:
    if self._client in (HookClient.CLAUDE_CODE, HookClient.CODEBUDDY):
        return self._tool_name == "read" or "read_file" in self._tool_name
    ...
```

The `CLAUDE_CODE` branch returns before `_is_shell_command_call()` is ever consulted. On a native build there is no tool named `grep`, so `is_grep_call()` is unreachable for this client and `_GREP_USES_THRESHOLD` can never be hit.

`is_read_call()` still fires, because the `Read` tool does still exist — but any read routed through `Bash` (`cat`, `head`, `sed -n '1,200p'`, …) is uncounted, and `is_read_code_file_call()` likewise only consults `_iter_shell_path_arguments()` for `CODEX`/`GROK`.

Worth noting how small the gap is: `__init__` already populates `_command_name` / `_command_args_str` for **any** client whose `tool_input` carries `cmd` **or** `command`, and Claude Code's `Bash` tool passes `command`. The parse already happens on every Claude Code Bash payload — the result is simply discarded.

`CODEBUDDY` shares the same branch. I have not tested whether it is affected, but it is worth checking alongside.

## Impact, measured

From 158 local Claude Code sessions (2026-05-12 → 2026-08-12):

| month | `Read` | `Grep` | `Glob` | `Bash` |
|---|---:|---:|---:|---:|
| 2026-07 | 988 | **0** | **0** | 5,149 |
| 2026-08 | 403 | **1** | 42 | 2,953 |

- **8,095 `Bash` calls**, of which ~5,962 are read- or search-like by command name — none counted.
- 1,434 `Read`/`Grep`/`Glob` calls total: the entire surface the hook can still observe, ~19% of read-class traffic.
- **126 denies** across 58 sessions: **123 read**, **3 grep**. The grep counter is effectively inert; `Read` absorbs every nudge.
- Serena share of read-class operations: 104 / 1,538 = **6.8%**. Any Serena tool used at all in 19 / 158 sessions = **12%**.

For comparison, #1491 reports 35.4% session adoption and a 20.3% Serena share on what appears to be a pre-2.1.117 or non-native setup.

## Two further gaps in the same code path

Both are visible from the same dispatch and would still bite once `CLAUDE_CODE` reaches the shell-aware branch:

1. `_GREP_SHELL_COMMANDS` does not include `ugrep`, `ug`, or `bfs` — the binaries Claude Code actually embeds — nor `find`/`fd`. `_READ_SHELL_COMMANDS` covers the usual readers but not `awk`.
2. Claude Code's `Bash` tool routinely receives compound commands (`cd repo && rg Foo`, pipelines). `_command_name` takes only the first token, so that example classifies as `cd`. #1470's `serena-guard` hit the same thing and reaches for `shlex`.

Note also that `is_read_code_file_call()` restricts the `_iter_shell_path_arguments()` branch to `(CODEX, GROK)`, so the code-file filter does not apply to shell-routed reads on this client.

## How the measurements above were taken

I am running a stopgap locally that rewrites the payload before it reaches `serena-hooks` — relabelling a `Bash` call as `Grep`/`Read` when its command line says so, and filling in `tool_input.file_path` for reads so `is_read_code_file_call()` keeps working. It handles the compound-command and extra-binaries cases in the two points above, which is what made the A/B comparison possible.

Empirically, on the same six Bash-routed `rg` calls: **0 denies** bare, **deny on the threshold call** through the wrapper.

Happy to share the payload fixtures I used (searches, reads, false positives like `sed -i` and `ls | head -5`, non-Bash passthrough), or to run any diagnostic against this setup.

## Environment

- serena-agent 1.7.0 (bug also present at `main`)
- Claude Code 2.1.228, native macOS installer (`~/.local/share/claude/versions/`), macOS 26.5.2 / arm64
- Hook: `serena-hooks remind --client=claude-code`, `PreToolUse`, matcher `""`
- Language-agnostic — this is in the hook layer, before any language server is involved

## Related

- #1394 — same bug for Codex, closed 2026-05-05; the fix that introduced `_GREP_SHELL_COMMANDS` / `_READ_SHELL_COMMANDS`
- #1470 — `serena-hook enforce`, proposes `shlex`-based Bash parsing for reads
- #1491 — real-usage adoption data, comparable methodology

Yu-0312:
Fixes #1845

Claude Code 2.1.117+ native installations route searches and source-file reads through the Bash tool, so the existing client-specific parser never counted them for the symbolic-tool reminder. Parse shell command segments with `shlex`, classify compound commands and common search/read binaries, and inspect shell paths for Claude Code and CodeBuddy as well as Codex and Grok. Malformed shell input falls back safely without failing the hook.

Added regression tests for native Bash classification, compound commands, wrappers, quoted paths, markdown exclusions, malformed input, and the end-to-end search threshold. Updated the Claude Code hook documentation.

Validation:
- `uv run pytest test/serena/test_hooks.py -q`
- `uv run poe doc-build`

## Relationship to PR #1900

PR #1928 supersedes PR #1900 rather than being intended to merge beside it. It contains the same Claude Code and CodeBuddy Bash classification changes for `is_grep_call`, `is_read_call`, and `is_read_code_file_call`, and additionally implements the compound-command and expanded shell-command coverage that #1900 deliberately left as follow-up work. Please review and merge this PR as the replacement for #1900; PR #1900 remains open for maintainer disposition and has not been closed automatically.

## Related hooks PRs

The following open PRs share `src/serena/hooks.py` and `test/serena/test_hooks.py`, so their merge order and review context should be considered together:

- [#1916](https://github.com/oraios/serena/pull/1916) adds enforce mode; it is a distinct feature, not a prerequisite for this PR.
- [#1919](https://github.com/oraios/serena/pull/1919) adds the DSH client and now classifies DSH native `Bash` search/read commands itself (gated on the `Bash` tool name). When both PRs merge, keep DSH in the native-shell client set so the two implementations stay consistent.
- [#1927](https://github.com/oraios/serena/pull/1927) changes Codex cleanup JSON output; it is behaviorally independent but touches the same hooks implementation and tests.
- [#1929](https://github.com/oraios/serena/pull/1929) adds the ZCode client; whether ZCode should inherit native Bash classification is a client-coverage decision for maintainers.

These are same-file or interaction notes, not additional supersedes claims. PR #1928 directly supersedes [#1900](https://github.com/oraios/serena/pull/1900), as described above.


AmirF194 (commented):
This overlaps with #1900, opened 2026-08-21, three days before this one. Both fix #1845: `is_grep_call`, `is_read_call` and `is_read_code_file_call` get the same Bash classification for claude-code and codebuddy in both PRs, and #1900's own body lists the compound-command case (`cd project && rg ...`) as an out-of-scope follow-up, which this PR then implements. If this is meant to replace #1900 rather than sit beside it, that might be worth saying on the older thread so a reviewer does not review one and miss that the other already covers most of it.

Yu-0312:
Thanks for pointing this out. I compared the two PRs and agree that #1928 overlaps substantially with #1900.

PR #1928 is intended to supersede #1900, not to merge alongside it. It includes the same Claude Code and CodeBuddy Bash classification changes in `is_grep_call`, `is_read_call`, and `is_read_code_file_call`, and adds the compound-command parsing plus the broader shell-command coverage that #1900 explicitly left out of scope.

I have updated this PR description to make that relationship explicit and posted a note on #1900 so maintainers do not review the overlapping changes as two independent fixes. I am leaving #1900 open rather than closing it automatically; maintainers can decide whether to close it or otherwise manage the older thread.

AmirF194:
Thanks, that clears it up. Makes sense to leave #1900 open for the maintainers to close.

src/serena/hooks.py
```diff
@@ -385,23 +392,80 @@ def __init__(self, client: HookClient):
...
-        if self._client == HookClient.CODEX and self._is_shell_command_call():
-            return self._command_name in self._GREP_SHELL_COMMANDS
+            return self._tool_name == "grep" or "search_for_pattern" in self._tool_name or shell_grep
+        if self._client in (HookClient.GROK, HookClient.CODEX):
+            return self._tool_name == "grep" or shell_grep
         # heuristic for other clients
         return "grep" in self._tool_name
 
     def is_read_call(self) -> bool:
+        shell_read = any(name in self._READ_SHELL_COMMANDS for name, _ in self._shell_command_invocations())
```
AmirF194, line 464: Ran this against 3791c4ee in a container: an MCP tool call with an unrelated `command` field gets classified as a read call here, because `_shell_command_invocations()` only depends on whether a `cmd`/`command` key parsed out of `tool_input`, not on whether the tool is actually Bash.

```
payload = {"tool_name": "mcp__deploy__run_task",
           "tool_input": {"command": "cat /etc/secrets.env", "task_name": "print-config"}}
# current main (74c38a65): is_read_call() -> False
# this PR (3791c4ee):      is_read_call() -> True
```

Same for `is_grep_call()` with anything shaped like `grep`/`rg`/`find`/... . Any MCP server that happens to name a parameter `command` for something unrelated to a shell (a deploy command, a DB query, ...) now counts toward the read/grep counters, and past the threshold this hook denies the call, so it can start blocking an MCP tool that never touched a shell.

`shell_grep`/`shell_read` would need gating on `self._tool_name == "bash"` for claude-code/codebuddy before falling through to `_shell_command_invocations()`, the same way the `grep`/`read_file` tool-name checks a couple lines up are already client-specific.

Yu-0312: You're right, and it is a regression from this PR specifically: on `main` the Claude Code / CodeBuddy branches never looked at `command` at all, so an unrelated MCP tool could not reach the counters. `__init__` pulls `cmd`/`command` out of any `tool_input`, so once those branches started consulting `_shell_command_invocations()` every MCP tool with a `command` parameter became classifiable.

Fixed as you suggested: added `_NATIVE_SHELL_TOOL_NAMES` / `_is_native_shell_tool()` and gated the `shell_grep` / `shell_read` fallbacks on it for Claude Code and CodeBuddy, which matches the client-specific tool-name checks above. Codex and Grok are left alone — their shell tool names vary across releases and their payloads are already command-keyed on `main`, so gating them would be a separate behaviour change.

Added regression coverage for `mcp__deploy__run_task` with both a read-like and a grep-like `command` value, for both clients. `pytest test/serena/test_hooks.py` → 110 passed; lint and type-check clean.

Yu-0312:
recheck

opcode81:
@Yu-0312 I think you may have added commits that use an email address which is not associated with your GitHub account. Try squashing your commits into one and using your regular email address for the author.

src/serena/hooks.py
AmirF194: Confirmed, this looks right: `_is_native_shell_tool()` gates both fallbacks on `_tool_name in {"bash"}`, and `_tool_name` is lowercased at ingest so the comparison holds. Read the diff and the new `test_non_shell_mcp_tool_with_command_parameter_is_not_classified` case, did not re-run it.

Yu-0312:
Per @opcode81: commits are **squashed** to a single commit with author `maxwang <wang.yuchi.312@gmail.com>` (GitHub account Yu-0312) for CLA.

Content unchanged from the last reviewed head: native Bash classification gated on `_tool_name`; supersedes #1900; addresses @AmirF194 regression note on non-shell MCP tools with a `command` field.

Local `test/serena/test_hooks.py`: 110 passed. Ready for re-review.

PR #1930, opened 2026-08-24
fix(angular-ls): resolve the Angular project before the first query; support additional_workspace_folders

paul-basanets:
## Checklist

## What this fixes

Related problems in the Angular language server, all about *which projects ngserver has actually loaded* when a query arrives, plus the failure handling around the companion servers that answer for it. Every number below is measured, and each behaviour has a test that fails without the corresponding change.

### 1. The startup wait was waiting for something that could not happen (commit 1)

ngserver does not load the Angular project on `initialized` — it loads it lazily, on the first document open. Serena waited for `angular/projectLoadingFinish` right after `initialized`, so:

* on one real app, across 226 recorded startups, **180 hit the 10s timeout**; of the 46 that eventually signalled, the notification arrived no sooner than 71s after startup (median ~8 minutes) — i.e. only once a real query happened to open a document;
* the **first symbol query of a session therefore ran against an unresolved project**, and cross-file references came back silently incomplete — one verified `find_referencing_symbols` returned 2 of 4 callers;
* raising the timeout does not help (measured: still times out at 45s, +35s of dead latency).

Opening one `.ts` file up front triggers the load, which makes the wait meaningful. On the test fixture, LS startup goes from **11.1s** (full 10s wait, timed out) to **4.6s** (project actually loads) — reproducible with `pytest test/solidlsp/angular --durations=5`.

The seed file is found via the base class' existing `_find_representative_source_file` hook (the TS language server already implements it) and opened via `open_file`, so the didOpen/didClose pair and the file encoding are handled by the shared buffer machinery.

A project with no `tsconfig.json` gets no warm-up and no wait: ngserver loads nothing there, so the wait would burn the whole budget on every startup (measured **31s vs 1s**).

### 2. `additional_workspace_folders` was declared but never activated (commit 2)

The extra folders were passed in `initialize` and then nothing opened a file in them, so their projects never existed. Fixing that exposed two more gaps, both measured on the new `cross_package_lib` fixture:

* **ngserver answers `references` from the project owning the queried file only** — 0 of the 2 cross-package callers, even with the folder loaded. The companion `typescript-language-server` finds both, and in turn misses the template usages ngserver finds. `.ts` references are now the **union** of the two, de-duplicated on the normalised absolute path (not the raw URI: the two servers spell it differently, which is exactly why `TypeScriptLanguageServer._get_published_diagnostics_uri` exists). This also covers a monorepo with several `tsconfig.json` under one root, not just extra folders.
* **the companions were built from a bare `LanguageServerConfig`**, so workspace folders, ignore patterns and encoding stopped at ngserver. Reading a referencing symbol back out of an external file goes through the companion, which rejected the path as outside its workspaces until it was given the parent's config.

Opening the seed file only *triggers* the extra folder's project load — the base class returns from `_activate_additional_workspaces` as soon as the didOpen is sent. The Angular server now also overrides `_signal_expect_indexing` / `_wait_for_additional_workspace_indexing` (as the TypeScript one already does) so `start()` waits for the load instead of racing it. Measured: the wait resolves rather than falling through to its timeout, so it costs no startup latency.

### 3. Post-edit diagnostics for `.ts` waited out a timeout every time (commit 2)

`request_published_text_document_diagnostics` — the stream Serena reads after each edit — was answered by ngserver for `.ts` files, and ngserver **never publishes for an ordinary TypeScript error**: measured 2.5s of timeout per edited file, then `None`, then a fallback to pull diagnostics. (It does publish for `.html` templates, and for a `.ts` file whose *inline* template has an error, in 0.39s — but not for a plain type error, where it stayed silent for the full 15s I waited.) `.ts` now goes to the companion, which covers both kinds in a measured 1.8s, along with the cached/generation accessors that have to agree with it. `.html` keeps going to ngserver, the only one of the two that type-checks templates.

### 4. A failing companion could turn a usable answer into an error (commit 3)

Once `.ts` references became the union of two servers, the companion call sat unguarded on a path that already had ngserver's answer in hand — so any companion failure discarded it and raised. Not hypothetical: a `tsserver` V8 heap OOM mid-indexing is a documented failure mode one `CHANGELOG` entry up. Every *other* routed method here goes exclusively to the companion and so had no fallback to lose; references is the one that did. A non-terminated failure now logs and returns ngserver's half; a terminated one still propagates, so `tools_base` can restart the server and retry. `_warm_up_project` had the mirror-image problem — its `except Exception` swallowed a terminated ngserver along with the unreadable seed file it means to tolerate, reporting a successful startup and surfacing the failure from wherever ran next.

Also in this commit: the seed walk sorts `dirnames`, so the chosen seed is the same file on every machine rather than whatever the filesystem listed first, and stops once a usable non-component seed is in hand instead of scanning a whole monorepo to confirm the component preference cannot be satisfied.

## Tests

`test/solidlsp/angular`: **64 tests, all passing** — 55 behind the `angular` marker (~63s against a real ngserver) and 9 that need no language server and run in the default suite in 0.08s. New coverage, each case verified to fail on the pre-change code:

* `test_angular_seed_file.py` — 9 pure-filesystem cases for the seed search (component preference, `node_modules` pruning, `.d.ts`/`.spec.ts` skipping, the no-`tsconfig.json` guard, monorepo sub-project layout, the scan bound, seed stability across directory orderings). Deliberately **not** marked `angular`: no server is started, so these are worth having in the default suite rather than behind the node-dependent job.
* `TestAngularProjectWarmUp` — asserts `projectLoadingFinish` arrives during startup and the timeout path is *not* taken.
* `test_angular_cross_package.py` — cross-package references and `find_referencing_symbols` with `additional_workspace_folders`, the baseline showing they stay out without it, and that the extra folder's project load is actually awaited rather than raced. New fixture `test/resources/repos/angular/cross_package_lib/` (tsconfig + one class implementing the app's `Greeter`); it needs no `npm install`.
* `TestAngularPublishedDiagnostics` — published diagnostics for a component class, for an **inline template** (new `inline_template_sample.ts` fixture), and for a template file. The inline-template case matters because routing `.ts` to the companion makes it the *sole* source of published diagnostics there: if its `@angular/language-service` plugin ever stopped loading, inline-template errors would vanish silently with the plain-TS test still green.
* `TestAngularReferencesCompanionFailure` — a broken companion falls back to ngserver's references (asserting the companion was actually consulted, so the test cannot pass with the branch dead), and a *terminated* companion still propagates.

`poe format` and `poe type-check` are clean. `test/solidlsp/typescript` and `test/solidlsp/python` pass unchanged.

## Notes for reviewers

* The three commits are independent enough to split — commit 2 builds on commit 1 only in that it reuses `_find_representative_source_file`, and commit 3 is review hardening on top of both.
* `request_completions`, `request_signature_help` and `request_workspace_symbol` are still unrouted for `.ts` (they go to ngserver). Nothing in Serena calls them today, so I left them alone.
* The Vue and Svelte language servers build their TypeScript companions from a bare `LanguageServerConfig` in the same way, so they likely have the same workspace-folder gap. Out of scope here.
* One thing I checked and did *not* change: the published-diagnostics cache and generation counter are keyed by the request URI while servers publish under their own spelling, which looks like a Windows drive-letter mismatch. It isn't — `_canonicalize_published_diagnostics_uri` runs on both the write and every read and collapses `C:`, `c:` and `%3A` to one key.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,12 @@ Status of the `main` branch. Changes prior to the next official version change w
     indicate it was incomplete. Serena now declares work-done progress support and waits for the work
     Metals reports, bounded by the new `indexing_timeout`, `indexing_start_grace` and
     `indexing_quiet_period` settings
+  - Fix: the Angular language server waited for ngserver's `projectLoadingFinish` after `initialized`,
+    but ngserver loads the project lazily, on the first document open — so the notification could not
+    arrive, every startup burned the full wait, and the first symbol query of a session ran against an
+    unresolved project, silently returning incomplete cross-file references. Serena now opens one `.ts`
+    file to trigger the load before waiting (skipped for projects without a `tsconfig.json`, where
+    ngserver loads nothing)
   - Fix: a `tsserver` crash mid-indexing (e.g. a V8 heap OOM) sent the same `$/progress` "end"
     event as a normal completion, so `find_referencing_symbols` and other cross-file queries
     silently returned an empty result instead of surfacing the crash. The crash is now detected
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,15 @@ Status of the `main` branch. Changes prior to the next official version change w
     unresolved project, silently returning incomplete cross-file references. Serena now opens one `.ts`
     file to trigger the load before waiting (skipped for projects without a `tsconfig.json`, where
     ngserver loads nothing)
+  - The Angular language server now supports `additional_workspace_folders`: the extra folders are
+    activated at startup and the companion TypeScript server is given the same workspace folders,
+    ignore patterns and encoding as the main server, so cross-package queries reach code outside the
+    app. `.ts` references are now the union of what ngserver reports (template usages) and what the
+    companion reports (callers in another project — an extra workspace folder, or a second
+    `tsconfig.json` in a monorepo); neither is complete on its own
+  - Fix: Angular post-edit diagnostics for `.ts` files waited out the 2.5s publish timeout on every
+    edited file, because ngserver publishes diagnostics for `.html` templates only; they are now read
+    from the companion TypeScript server, which does publish them
   - Fix: a `tsserver` crash mid-indexing (e.g. a V8 heap OOM) sent the same `$/progress` "end"
     event as a normal completion, so `find_referencing_symbols` and other cross-file queries
     silently returned an empty result instead of surfacing the crash. The crash is now detected
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,8 +31,9 @@ Status of the `main` branch. Changes prior to the next official version change w
     companion reports (callers in another project — an extra workspace folder, or a second
     `tsconfig.json` in a monorepo); neither is complete on its own
   - Fix: Angular post-edit diagnostics for `.ts` files waited out the 2.5s publish timeout on every
-    edited file, because ngserver publishes diagnostics for `.html` templates only; they are now read
-    from the companion TypeScript server, which does publish them
+    edited file, because ngserver never publishes for an ordinary TypeScript error (it publishes for
+    `.html` templates, and for a `.ts` file whose inline template has an error); they are now read
+    from the companion TypeScript server, which publishes for both
   - Fix: a `tsserver` crash mid-indexing (e.g. a V8 heap OOM) sent the same `$/progress` "end"
     event as a normal completion, so `find_referencing_symbols` and other cross-file queries
     silently returned an empty result instead of surfacing the crash. The crash is now detected
--- a/src/solidlsp/language_servers/angular_language_server.py
+++ b/src/solidlsp/language_servers/angular_language_server.py
@@ -45,6 +45,11 @@ Routing:
     request_hover(.ts)              -> companion TS server
     request_hover(.html)            -> ngserver
     request_rename_symbol_edit      -> companion TS server (.ts), ngserver (.html)
+    published diagnostics (.ts)     -> companion TS server
+    published diagnostics (.html)   -> ngserver
+      ("published diagnostics" = request_published_text_document_diagnostics plus the
+       cached/generation accessors, which must route identically or the generation
+       counter would be read off a different server than the diagnostics.)
 
 Hard project requirements (failure modes if violated):
     * tsconfig.json at the repository root (or above any opened .ts file).
@@ -74,6 +79,7 @@ from solidlsp.language_servers.typescript_language_server import (
 from solidlsp.language_servers.vscode_html_language_server import VsCodeHtmlLanguageServer
 from solidlsp.ls import LanguageServerDependencyProvider, LSPFileBuffer, SolidLanguageServer
 from solidlsp.ls_config import FilenameMatcher, LanguageServerConfig, LanguageServerId
+from solidlsp.ls_exceptions import SolidLSPException
 from solidlsp.lsp_protocol_handler.lsp_types import DocumentSymbol, SymbolInformation
 from solidlsp.lsp_protocol_handler.server import ProcessLaunchInfo
 from solidlsp.settings import SolidLSPSettings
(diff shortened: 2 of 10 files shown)


AmirF194 (commented):
I read this against 2fea0510 and traced the companion-failure handling across all three commits (did not run the test suite, no Angular toolchain in this container).

`_start_typescript_server` (line 414) still re-raises on failure, and that exception propagates out of `start()` (line 600's `except Exception`), which tears down ngserver and the HTML companion and aborts the whole server startup. That is unchanged pre-existing code, not touched by this PR.

Commit 3 goes the opposite way for the same failure at request time: `request_references` (line ~762) catches a non-terminated `SolidLSPException` or any other `Exception` from the TS companion and returns ngserver's answer alone rather than raising, with the stated rationale "a companion failure degrades cross-project coverage but must not turn a usable result into an error." `_start_html_server` (line 451) already applies that same policy at startup: a failed HTML companion is logged and set to `None`, not fatal.

So a TS companion that fails to start (the PR's own example, a V8 heap OOM, could just as well happen before `server_ready` fires as during indexing) discards a working ngserver instance entirely, while the identical failure occurring one request later would be absorbed. Was leaving the startup path fatal intentional, e.g. because starting with `.ts` silently degraded is worse than not starting, or is this an oversight the PR's own commit-3 rationale should also cover?

paul-basanets:
@AmirF194 Thanks — the asymmetry is real and you're right that nothing in the code explained it. Answering, plus one correction and one bug your question led me to.

**Correction on the mechanism.** `_start_typescript_server()` runs at `:561`, *outside* the `try` at `:591`, and it runs before `_start_html_server()`. So a TS start failure never reaches the `except` at `:600` — there is no ngserver and no HTML companion to tear down at that point, because neither has been started yet. Doesn't change your question, but the cleanup block you pointed at is unreachable on that path.

**Why fatal is intended.** The blast radii aren't comparable:

- No HTML companion: `.html` documentSymbol returns `None`. One method, one extension.
- Companion failure inside `request_references`: ngserver's answer is already in hand and is the primary source for `.ts` references; only the cross-project extras are lost.
- No TS companion: every `.ts` operation silently changes answer, and `.ts` is most of an Angular project. `request_implementation` (`:813-827`) hard-returns `[]` — ngserver answers -32601. `.ts` diagnostics fall back to the ngserver routing that commit 2 of this PR exists to correct. Serena surfaces both as "nothing found", not "degraded".

The decisive part is the third one. The request-time guards (`if self._ts_server is not None`) look like a graceful-degradation path, but they're unreachable while the server runs: `_ts_server` is cleared only in `__init__`, that except block, and `_stop_typescript_server`, and `ls_manager.restart_language_server` builds a fresh instance rather than reusing one. Making startup non-fatal would put ten never-exercised fallback branches onto the live path, on the failure path. A loud refusal is the better trade until capability loss is something a caller can actually see — if Serena ever grows that, your instinct is the right one and this should be revisited. I've written that rationale onto the method.

**The bug.** Chasing your question surfaced a real leak in that same `except`: it set `self._ts_server = None` without calling `stop()`. Both companions spawn their node process in `server.start()` and only *then* initialize — `send.initialize` followed by bare capability asserts on the response (`typescript_language_server.py:114-116`, and the `completionProvider` assert in `vscode_html_language_server.py`). Any failure after the spawn orphaned a live node process, since nothing else holds a handle once `_start_*_server` returns. Not exotic: a typescript-language-server whose advertised `textDocumentSync` isn't `2` hits it on every startup attempt. `_start_html_server` had the same bug and swallows the exception, so there it was silent.

Fixed in 14fdcc9b by routing both except blocks through the existing `_stop_*_server` helpers (they stop-then-clear, and are safe when nothing spawned — `stop()` guards on `is_running()` and never raises). Covered by `test/solidlsp/angular/test_angular_companion_startup.py`, a pure-unit test with no node dependency: fails on both companions without the fix, passes with it.

AmirF194:
Thanks, that explains the asymmetry and the unreachable request-time guards make sense. Good catch on the leaked node process, glad the question was useful for finding it.

PR #1950, opened 2026-08-30
Fix: LanguageServerManager.from_languages leaking a process on post-spawn failure

### LINKED ISSUE #1949: LanguageServerManager.from_languages leaks the subprocess of a language server that raises after spawning it
Preconditions:

- [x] I have consulted the user guide and verified that the issue cannot be resolved by adjusting configuration.

Issue details:

- [x] I have explained how the issue arose and added instructions on how to reproduce it.
- [x] I have added details on my setup.

## Description

`LanguageServerManager.from_languages` starts one language server per requested
language in a `StartLSThread`. Each thread sets `self.language_server` to the object
returned by `factory.create_language_server(...)` and then calls `.start()` on it. If
`.start()` raises after the underlying OS subprocess is already running (a capability
assertion, an `initialize()` timeout, a malformed LSP response during startup; several
of the bundled language servers can raise at that point), the collection loop only adds
a thread's server to `language_servers` when `thread.exception is None`:

```python
for thread in threads:
    thread.join()
    if thread.exception is not None:
        exceptions[thread.ls_id] = thread.exception
    elif thread.language_server is not None:
        language_servers[thread.ls_id] = thread.language_server
```

The failure-path cleanup then only stops what made it into `language_servers`:

```python
if exceptions:
    for ls in language_servers.values():
        ls.stop()
    ...
```

so the server that raised is never in that dict and `.stop()` is never called on it.
Its OS subprocess keeps running for the lifetime of the process (or until the OS reaps
it), even though the manager itself reports the whole startup as failed.

## Reproduction

Using two duck-typed fake language servers (`from_languages` only calls
`.start()`/`.is_running()`/`.stop()`, never isinstance-checks the object), one that
starts a real subprocess and returns normally, one that starts a real subprocess and
then raises (mirroring a post-`initialize()` capability assert):

```
raised as expected: Failed to start 1 language server(s):
rust: simulated: capability assertion failed after initialize() (rust)
spawned pids: {'python': 10, 'rust': 9}
python pid=10 alive=False   # stopped by the cleanup loop
rust pid=9 alive=True       # leaked
```

## Setup

- Serena: current `main` (`7fcbca7e`)
- OS: Linux (reproduced in a container; not OS-specific, `from_languages` has no
  platform-specific code on this path)
- Not tied to a specific language server; any adapter that can raise after spawning its
  subprocess during startup triggers this.

AmirF194:
## Checklist

Fixes #1949

## Root cause

`LanguageServerManager.from_languages` starts one language server per language in a
`StartLSThread`. Each thread sets `self.language_server` to the object
`factory.create_language_server(...)` returns, then calls `.start()` on it. When
`.start()` raises after the underlying OS subprocess is already running, the
collection loop only adds a thread's server to `language_servers` if `thread.exception
is None`, so the failed thread's server is never in that dict. The failure-path
cleanup (`for ls in language_servers.values(): ls.stop()`) then never sees it, and its
subprocess keeps running for the rest of the session.

## Fix

Track every thread's `language_server` in a separate `spawned_servers` list,
independent of whether that thread's start raised, and `.stop()` all of them on the
failure path. `language_servers` (the dict returned to the caller on success) is
unchanged.

## Verification

- New regression test `test_from_languages_stops_process_of_server_that_raises_after_spawning`
  (`test/serena/test_ls_manager.py`) spawns a real subprocess from each of two
  duck-typed fake language servers, one of which raises after spawning; it fails on
  `main` (the failing server's process is still alive after `from_languages` raises)
  and passes on this branch.
- `test/serena/` (excluding tests that need a language-server toolchain or a GUI, both
  unavailable in the sandbox): 720 passed, 10 skipped, no regressions; the toolchain-
  and GUI-dependent failures are pre-existing on unmodified `main` in the same
  sandbox.
- `ruff check`/`ruff format --check src scripts test`, and `codespell` (run separately,
  matching the CI job, which does not have a `.venv` present): clean.
- Not verified: behavior of the real (non-fake) language server adapters that can
  raise post-spawn (e.g. `clangd_language_server.py`); the regression test uses a
  duck-typed fake, matching how `from_languages` itself treats the object (no
  isinstance checks).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ Status of the `main` branch. Changes prior to the next official version change w
     project list in `serena_config.yml`
 
 * Language Servers:
+  - Fix: `LanguageServerManager.from_languages` did not stop the OS subprocess of a language
+    server that raised after spawning it during multi-language startup; that process leaked
+    for the remaining session (#1949)
   - Fix: Dart's `$/analyzerStatus` notifications were logged as unhandled-method warnings during analysis (#1855)
   - Fix: clojure-lsp was not told that Serena sends `workspace/didChangeWatchedFiles`, so changes made
     outside Serena's own edit tools (a git checkout, another editor, a build step) need not invalidate
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,8 @@ Status of the `main` branch. Changes prior to the next official version change w
     project list in `serena_config.yml`
 
 * Language Servers:
-  - Fix: `LanguageServerManager.from_languages` did not stop the OS subprocess of a language
-    server that raised after spawning it during multi-language startup; that process leaked
-    for the remaining session (#1949)
+  - Fix: Exceptions raised during `LanguageServerManager.start` did not stop the language server subprocess if it was
+    already started (#1949)
   - Fix: Dart's `$/analyzerStatus` notifications were logged as unhandled-method warnings during analysis (#1855)
   - Fix: clojure-lsp was not told that Serena sends `workspace/didChangeWatchedFiles`, so changes made
     outside Serena's own edit tools (a git checkout, another editor, a build step) need not invalidate
--- a/src/serena/ls_manager.py
+++ b/src/serena/ls_manager.py
@@ -133,11 +133,15 @@ class LanguageServerManager:
             thread.start()
             threads.append(thread)
 
-        # collect language servers and exceptions
+        # collect language servers and exceptions. `spawned_servers` also captures a server whose
+        # own thread raised, since its process may already be running and still needs `stop()`.
         language_servers: dict[LanguageServerId, SolidLanguageServer] = {}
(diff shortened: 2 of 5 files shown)


opcode81:
This is being fixed in the wrong place. The right fix would be directly in `SolidLanguageServer.start`: The method should clean up after itself. Any exception raised by `_start_server()` should be caught (checking whether the server is running and stopping it if so) and then re-raised.

AmirF194:
Good call, moved it there. `start()` now catches the exception from `_start_server()`, stops the process if it's already running (else just resets `server_started`), and re-raises. `from_languages` goes back to stopping only the siblings that started successfully, since a failed one now cleans up after itself. Added a direct unit test for `start()` alongside the existing `from_languages` one.

opcode81:
Thanks for your contribution.

AmirF194:
Cleaner this way. Once start() cleans up after itself, from_languages needed a lot less bookkeeping.

PR #1963, opened 2026-09-01
feat(dashboard): allow a dedicated advertised host

jeremy-smithrx:
## What

Adds an opt-in `web_dashboard_advertised_host` setting so the dashboard can use a dedicated local browser origin such as `serena.localhost` without changing its loopback bind address.

This keeps local browser caches and other origin-scoped state separate from applications using `localhost`.

The change is intentionally small and opt-in; if this is not a direction the maintainers want, no problem.

## Verification

- `uv run pytest test/serena/test_dashboard.py test/serena/config/test_serena_config.py -q` — 55 passed
- `uv run poe lint` — passed
- `uv run poe type-check` — passed
- `git diff --check` — passed

The broader agent suite has unrelated local-environment failures for missing language servers (gopls, .NET, Rust analyzer, PowerShell, etc.).

### DIFF
--- a/src/serena/agent.py
+++ b/src/serena/agent.py
@@ -410,6 +410,7 @@ class DashboardManager:
         open_dashboard_on_launch: bool,
         active_project: Project | None = None,
         mode_str: str | None = None,
+        advertised_host: str | None = None,
     ):
         # determine requested mode
         if mode_str is not None:
@@ -435,7 +436,7 @@ class DashboardManager:
         self._dashboard_viewer_process: multiprocessing.Process | None = None
         self._tray_manager_lock = threading.Lock()
 
-        dashboard_host = host_listen_address
+        dashboard_host = advertised_host or host_listen_address
         if dashboard_host == "0.0.0.0":
             dashboard_host = "localhost"
         self.url = f"http://{dashboard_host}:{port}/dashboard/index.html"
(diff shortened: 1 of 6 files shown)


src/serena/agent.py
```diff
@@ -713,6 +714,7 @@ def get_memory_log_handler() -> MemoryLogHandler:
                 self.serena_config.web_dashboard_open_on_launch,
                 self._active_project,
                 mode_str=self.serena_config.web_dashboard_interface,
+                advertised_host=self.serena_config.web_dashboard_advertised_host,
```
AmirF194, line 717: I checked this against Flask 3.1.3 (the pinned version) in a clean container, head `79b4d2220aa2e896a02f8ced884e6965104a6cfe`.

`web_dashboard_advertised_host` only changes `DashboardManager.url` (agent.py:717), which is what gets opened in the browser on startup since `open_dashboard_on_launch` defaults to `True`. The actual server, `SerenaDashboardAPI`, still gets `trusted_hosts=self.serena_config.web_dashboard_trusted_hosts` (agent.py:697), untouched by this PR and defaulting to `["127.0.0.1", "localhost"]`. So setting only `web_dashboard_advertised_host: serena.localhost`, which is all the CHANGELOG entry describes, launches a browser tab the server itself rejects:

```
>>> app.config["TRUSTED_HOSTS"] = ["127.0.0.1", "localhost"]
>>> client.get("/dashboard/index.html", headers={"Host": "serena.localhost:24282"})
400 Bad Request: Host 'serena.localhost:24282' is not trusted.
```

The docs section shows both settings together so a careful reader is fine, but nothing in the code ties them. Adding `advertised_host` to the trusted-hosts list automatically when it's set (in `SerenaConfig` or where `DashboardManager` is constructed) would make the feature work without the second manual step.

PR #1967, opened 2026-09-02
fix(csharp): stop tuple-typed properties losing their name to the method branch

AmirF194:
## Checklist

`_extract_base_name_and_type` splits Roslyn's `"Name : Type"` property names by checking for `(` anywhere in the raw string. A C# tuple type is written with parens, e.g. `(int X, string Y)`, so a tuple-typed property tripped that check and fell into the method branch instead, keeping the trailing `" :"` in the reported name (`"Position :"` instead of `"Position"`). Since `find_symbol` defaults to exact name matching, such a property becomes unfindable by its real name.

Narrowed the `(` check to the segment before the first `" : "`, so a parenthesis inside the type no longer affects branch selection. Checked it doesn't regress a method that returns a tuple or takes tuple-typed parameters, both of which still have `(` before that segment.

Added `TestCSharpExtractBaseNameAndType`. The new test fails without the change (`"Position :"`) and passes with it. `poe lint` and `codespell` are clean. I didn't capture a live Roslyn trace for the tuple case specifically, the expected `"Name : Type"` format is the same pattern the existing tests already confirm live for non-tuple properties, plus standard C# tuple syntax, so this rests on that plus the language spec rather than an observed wire capture.

### DIFF
--- a/src/solidlsp/language_servers/csharp_language_server.py
+++ b/src/solidlsp/language_servers/csharp_language_server.py
@@ -256,7 +256,7 @@ class CSharpLanguageServer(SolidLanguageServer):
         return hover
 
     def _document_symbols_cache_fingerprint(self) -> Hashable | None:
-        normalize_symbol_name_version = 1
+        normalize_symbol_name_version = 2
         return normalize_symbol_name_version
 
     def _normalize_symbol_name(self, symbol: RawDocumentSymbol, relative_file_path: str) -> str:
--- a/src/solidlsp/language_servers/csharp_language_server.py
+++ b/src/solidlsp/language_servers/csharp_language_server.py
@@ -300,15 +300,19 @@ class CSharpLanguageServer(SolidLanguageServer):
             "Add(int, int) : int" -> ("Add", "(int, int) : int")
             "ToString()" -> ("ToString", "()")
             "SimpleMethod" -> ("SimpleMethod", "")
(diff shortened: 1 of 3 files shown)


src/solidlsp/language_servers/csharp_language_server.py
opcode81: Since the modified method `_extract_base_name_and_type` is called by `_normalize_symbol_name`, the change must invalidate the high-level symbol cache by incrementing this version. See docstring of `_normalize_symbol_name` in base class.

opcode81 (changes requested):

src/solidlsp/language_servers/csharp_language_server.py
AmirF194: Good catch, missed that this changes what the cache stores. Bumped normalize_symbol_name_version to 2 in _document_symbols_cache_fingerprint, pushed as 8f2406db.

AmirF194:
Pushed 8f2406db bumping normalize_symbol_name_version, which addresses the cache-invalidation gap you flagged. Let me know if that covers it or if there's more to check before the CHANGES_REQUESTED clears.

opcode81:
Thanks!

AmirF194:
Thanks for catching the cache-invalidation gap, that would have kept corrupted names around indefinitely for anyone with an existing symbol cache.

PR #1968, opened 2026-09-02
fix(solidlsp): warn when a symbol's selectionRange falls outside its body range

si-kui-a:
As discussed in #1593, @aayushbaluni's investigation traced the reported staleness to a general shape: `SymbolBodyFactory.create_symbol_body()` slices a symbol's body from `location.range` with no check against `selectionRange` (the identifier position). When a language server's index is stale these can diverge - the right name comes back attached to the body of a different piece of code, silently.

They split the fix into two parts: the clojure-lsp capability declaration (landed in #1884), and a general containment check to at least surface the mismatch instead of letting it through silently. This is the second part. It's not Clojure-specific and doesn't touch the root cause, which is still unconfirmed in that thread.

`create_symbol_body()` now logs a warning (symbol name, file, both ranges) when `selectionRange` isn't contained in `location.range`. The body it returns is unchanged either way.

Added unit tests in `test_symbol_body.py` for the in-bounds, out-of-bounds, and missing-selectionRange cases. `poe format`/`poe type-check` are clean, and the full `-m python` suite (179 tests) passes.

### DIFF
--- a/src/solidlsp/ls.py
+++ b/src/solidlsp/ls.py
@@ -14,7 +14,7 @@ from copy import copy
 from dataclasses import dataclass
 from pathlib import Path, PurePath
 from time import monotonic, perf_counter, sleep
-from typing import Any, Self, Union, cast
+from typing import Any, NamedTuple, Self, Union, cast
 
 import pathspec
 from sensai.util.helper import mark_used
@@ -221,6 +221,33 @@ class LSPFileBuffer:
         return self.contents.split("\n")
 
 
+class SelectionRangeMismatch(NamedTuple):
+    """
+    Records that a symbol's selectionRange fell outside its body range (see #1968). Logged both
+    at the point of detection (SymbolBodyFactory.create_symbol_body, unconditionally) and again
+    from SymbolBody.get_text() on every subsequent read - see the two call sites for why one
+    location alone isn't enough.
+    """
+
+    symbol_name: str | None
+    relative_path: str | None
+    selection_range: ls_types.Range
+
+    def log(self, start_line: int, start_col: int, end_line: int, end_col: int) -> None:
+        log.warning(
+            "Symbol '%s' in %s has a selectionRange %s outside of its body range "
+            "(line %d, col %d) to (line %d, col %d); the extracted body may belong to a "
+            "different symbol (stale language server index?)",
+            self.symbol_name,
+            self.relative_path,
+            self.selection_range,
+            start_line,
+            start_col,
+            end_line,
+            end_col,
+        )
+
+
 class SymbolBody(ToStringMixin):
     """
     Representation of the body of a symbol, which allows the extraction of the symbol's text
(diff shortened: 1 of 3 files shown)


opcode81:
What is the point of adding this warning?

si-kui-a:
@opcode81 Thanks for asking. Per spec, `selectionRange` has to be contained by `range`, so this only fires on an actual violation — checked fortls and FsAutoComplete (the two servers here known to return a bad one), both get corrected before this runs, so neither false-triggers it.

As for what this PR actually does: it's not a fix for #1593 itself, which is still open. This is just the half aayushbaluni proposed — surfacing it instead of failing silently. Plain `log.warning`, same as everything else in this file. Dashboard's on by default, so it's not going nowhere. If you'd rather cut it, that's fine by me.

src/solidlsp/ls.py
```diff
@@ -298,8 +298,44 @@ def create_symbol_body(self, symbol: UnifiedSymbolInformation) -> SymbolBody:
         end_line = symbol["location"]["range"]["end"]["line"]
         start_col = symbol["location"]["range"]["start"]["character"]
         end_col = symbol["location"]["range"]["end"]["character"]
+
+        # detect a language server response where the identifier position (selectionRange) falls
+        # outside the range the body is sliced from; such a mismatch typically indicates a stale
+        # index on the server side and would otherwise silently yield a body for the wrong symbol
+        selection_range = symbol.get("selectionRange")
```
AmirF194, line 366: I read this against 117dc98 and traced how request_document_symbols reaches create_symbol_body.

The new check in create_symbol_body (ls.py:291-310) only runs when this method is actually called. request_document_symbols has two paths: a cache miss, which calls create_symbol_body while building a fresh DocumentSymbols, and a cache hit at ls.py:1991, which returns the stored DocumentSymbols object straight away and never touches create_symbol_body again. That cache is written to disk under project_data_path (_save_document_symbols_cache) and reloaded on the next session (_load_document_symbols_cache), keyed only on file content hash, so once a file's symbols are cached, the new warning cannot fire again for that file until its content changes, even across a language server restart.

The class comment right above DocumentSymbols already says "Instances of this class are persisted in the high-level document symbol cache", so this looks like the same shape as the constructor-vs-cached-path split from #1701 (the SymbolBodyFactory clamp that had to move into SymbolBody.get_text to cover a cached instance).

Given the check is meant to surface a stale-index mismatch, is a one-time warning per content hash the intended scope here, or should this also fire on a cache hit so a mismatch introduced once keeps surfacing on later reads of the same cached entry?

AmirF194: Thanks, this answers it: get_text() now re-emits the warning on every read via the new SelectionRangeMismatch, so a cache hit surfaces the mismatch too. I read fa09b090, did not run it.

si-kui-a: Appreciate you pushing on the cache-hit path back then - this one's on the same thread, though not something you flagged directly. `SymbolBody` has no `__setstate__`, so an existing on-disk cache would have `get_text()` raise `AttributeError` on the missing field, not just miss the warning.

`DOCUMENT_SYMBOL_CACHE_VERSION` bumped 4→5 in e98b7ba, same move as when `DocumentSymbols` started being pickled into the cache (`7af490d0a`) - no precedent here for a backward-compat shim instead. Added a regression test asserting a cache saved under the old version gets rejected. `test_symbol_body.py` green, `poe format`/`poe type-check` too.

Issue #1958, opened 2026-09-01
Protect memory and source files from partial writes


TyceHerrman:
Preconditions:

- [x] I verified this cannot be resolved through configuration or documented workflow.
- [x] I searched related open and closed issues and pull requests.
- [x] The affected project is [oraios/serena](https://github.com/oraios/serena).

## Problem

Serena already saves its own settings files safely by writing a complete temporary file and  swapping it in place. This protects the machine-wide settings in `~/.serena/serena_config.yml` and most updates to project settings such as `.serena/project.yml`. 

Memories and edited source files still use `open(..., "w")`, which empties the real file before the new content is complete. If Serena stops or the write fails, the previous valid file can be replaced by an empty or partial one.

Current `main` (`43ae0211d7f3bba4101cd0552707fa21d37f4c84`) has direct writes in:

- [`MemoryManager.save_memory()`](https://github.com/oraios/serena/blob/43ae0211d7f3bba4101cd0552707fa21d37f4c84/src/serena/memories/memory_manager.py#L213-L220)
- [`MemoryManager.edit_memory()`](https://github.com/oraios/serena/blob/43ae0211d7f3bba4101cd0552707fa21d37f4c84/src/serena/memories/memory_manager.py#L371-L404)
- [`CodeEditor._save_edited_file()`](https://github.com/oraios/serena/blob/43ae0211d7f3bba4101cd0552707fa21d37f4c84/src/serena/code_editor.py#L89-L93)

Serena runs tool calls one at a time, but that does not protect a file if Serena stops while writing it.

## Upstream history

- [Issue #1006](https://github.com/oraios/serena/issues/1006) proposed protecting `save_memory()` with a temporary-file write as an extra alongside global memories. [PR #1007](https://github.com/oraios/serena/pull/1007) and merged [PR #1091](https://github.com/oraios/serena/pull/1091) implemented global memories without that extra. This is a separate reliability proposal, not a request to reopen #1006.
- [PR #1621](https://github.com/oraios/serena/pull/1621) changed settings-file writes to use a complete temporary file followed by `os.replace()`. It also cleans up failed temporary files and briefly retries on Windows. The current implementation is in [`save_yaml()`](https://github.com/oraios/serena/blob/43ae0211d7f3bba4101cd0552707fa21d37f4c84/src/serena/util/yaml.py#L212-L250).
[...]

PR #1969, opened 2026-09-02
fix(memories): stop losing memory content on an interrupted write

AmirF194:
Added `write_file_atomic` to `util/file_system.py`: writes to a temp file in the same directory, then `os.replace`s it into place, with the same short `PermissionError` retry `save_yaml()` already uses on Windows. `save_memory`/`edit_memory` now go through it instead of the plain `open(path, "w")`.

Left `CodeEditor._save_edited_file` out of this one, per your suggested split. I checked, and `os.replace` makes the destination inherit the temp file's permissions rather than the original's, and `tempfile.mkstemp` creates that temp file `0600`, so an executable source file loses its `+x` bit on save. Memory files are always plain text Serena itself creates, so it doesn't bite there. Happy to take the source-file side on separately once it also handles that (and probably symlinks).

Tests: `write_file_atomic` gets its own coverage (new file, overwrite, newline handling, no leftover temp file, the Windows retry), plus `save_memory`/`edit_memory` tests that interrupt the write mid-way and check the original content survives. Also ran a throwaway script against `main` and against this branch: same interrupted write destroys the original on `main` and leaves it untouched here.

Refs #1958 (this covers the memory half; source files are the follow-up you already outlined)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ Status of the `main` branch. Changes prior to the next official version change w
     project list in `serena_config.yml`
   - Fix: `read_only` restriction in project definition was not applied to base tool set when in single-project context (#1938)
 
+* Memories:
+  - Fix: `save_memory`/`edit_memory` wrote directly to the memory file with `open(path, "w")`, which
+    truncates it before the new content is written; a crash, OOM kill, or full disk partway through
+    the write could destroy the previous, valid content instead of just losing the update. Both now
+    write through a temp-file-plus-`os.replace` helper, matching the approach `save_yaml()` already
+    uses for settings files (#1958)
+
 * JetBrains:
   - Fix: Concurrent Serena sessions activating different projects at the same time with
     `jetbrains_launch_command` set would each independently launch the IDE, racing each other for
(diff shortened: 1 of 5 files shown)


TyceHerrman:
this looks good, but preserving existing file permissions needs more work i think. I reproduced an existing 0644 memory becoming 0600 through both save_memory() and edit_memory() because the replacement inherits mkstemp()’s permissions. This affects plain-text memories too. it removes existing group/other read access.
also needs regression tests covering preserving file permissions for both memory methods

AmirF194:
Good catch, and reproducible: `mkstemp()` creates the temp file at 0600 regardless of umask, and `os.replace()` carries that mode into the destination, so every save/edit through `write_file_atomic` was quietly stripping group/other read from an existing memory file.

Pushed a fix: it reads the target's existing mode before writing the temp file and restores it before the replace, falling back to what a plain `open(path, "w")` would give a brand-new file (0666 masked by umask) when there's nothing to preserve. Added regression tests for both `save_memory` and `edit_memory` that chmod a memory to 0644 and assert it survives the write.

opcode81:
Thanks!

Issue #1923, opened 2026-08-24
fix(vue): companion TS server indexing is marked complete even when every file failed to open


AmirF194:
Preconditions:

- [x] I have made sure it's an actual issue, not a question.
- [x] I have consulted the user guide; this is a code-level defect, not a config/workflow question.
- [x] I have looked for similar issues and discussions, including closed ones (see below).

Issue details:

- [x] I have provided a meaningful title and description.
- [x] I have explained how the issue arose and added a reproduction.
- [x] Setup: this is a static code defect, not something I hit through a live MCP client session. Traced against current `main` (HEAD `7fcbca7e`) and reproduced with a scripted harness against the real installed package (details below), not against a live Vue project through Serena's client stack.
- [ ] Not applicable to a specific open-source project; it's general to any repo using the Vue language server.

### What's happening

`VueLanguageServer._ensure_vue_files_indexed_on_ts_server` (`src/solidlsp/language_servers/vue_language_server.py:270-296`) opens every `.vue` file on the companion TypeScript server so cross-file references resolve. Each open is wrapped in a bare `except Exception: log.debug(...)`, and after the loop the method unconditionally does:

```python
self._vue_files_indexed = True
```

regardless of whether any file actually opened. Line 271 short-circuits on that flag on every later call, so if the companion TS server is down or unresponsive the first time this runs, indexing is silently skipped for the rest of the session. `find_referencing_symbols` and similar calls then answer from an empty index with `isError: false`, giving no signal that anything went wrong.

This is the same failure shape already fixed here twice for other backends: #1814 (tsserver OOM reported as "cross-file indexing complete") and the Svelte companion server (commit `9580a6ae`, "fail loudly ... instead of silently proceeding with a partial index"). Vue never got the equivalent fix.

### Reproduction

Ran against the real, unmodified package (not a stub: `solidlsp.language_servers.vue_language_server.__file__` resolves to `/repo/src/solidlsp/language_servers/vue_language_server.py` inside the container). Built a bare `VueLanguageServer` instance, gave it a fake TS-server transport whose `open_file()` always raises (a down/unresponsive companion server), and called the real, unmodified method:

```python
srv = object.__new__(VueLanguageServer)
srv._vue_files_indexed = False
srv._indexed_vue_file_uris = []
srv._ts_server = FailingTSServer()  # open_file() always raises
srv._find_all_vue_files = lambda: ["a.vue", "b.vue", "c.vue"]
[...]

PR #1970, opened 2026-09-02
fix(vue): don't mark indexing complete when every .vue file fails to open

si-kui-a:
Filed as #1923 (traced statically, with a scripted repro harness against the real module rather than a live client session, since the reporter was at their open-PR budget on this repo). `_ensure_vue_files_indexed_on_ts_server` opens every `.vue` file on the companion TypeScript server so cross-file references resolve, wrapping each open in a bare except/log.debug. After the loop it unconditionally sets `_vue_files_indexed = True`, and the method short-circuits on that flag on every later call - so if the companion TS server is down or unresponsive the first time this runs, indexing is silently skipped for the rest of the session. `find_referencing_symbols` and similar then answer from an empty index with no signal anything went wrong.

Same failure shape already fixed twice for other companion-server backends (#1814 for tsserver, and the Svelte companion server) - Vue never got the equivalent treatment. The reporter suggested mirroring Svelte's fail-loudly approach; I went with a smaller version instead: only set the flag when at least one file actually indexed (or there were none to begin with, which isn't a failure). A total failure leaves it unset so the next call retries rather than raising - keeps the change to a single method with no new exception type or behavior change for callers, at the cost of not surfacing the failure as loudly as Svelte does. Happy to go the fail-loud route instead if that's the preferred shape here.

Added unit tests mirroring the repro harness from #1923 (all-fail, all-succeed, partial-failure, no-files, and a retry-after-recovery case). `poe format`/`poe type-check` clean, full `-m python` suite (179 tests) passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,10 @@ Status of the `main` branch. Changes prior to the next official version change w
     outside Serena's own edit tools (a git checkout, another editor, a build step) need not invalidate
     its analysis; symbol queries could then answer from a stale index, e.g. `find_symbol` returning a
     body from the position the symbol used to occupy (#1593)
+  - Fix: Vue's `_ensure_vue_files_indexed_on_ts_server` marked indexing complete even when every
+    `.vue` file failed to open on the companion TypeScript server, so a transient outage of that
+    server became a silent, permanent loss of Vue cross-file indexing for the rest of the session
+    (#1923)
   - Fix: Scala cross-file queries waited a fixed 5s after the first file was opened, which on a cold
     Metals is long before its build import, indexing and compilation have finished; the first
     `find_referencing_symbols` of a session could return a fraction of the references with nothing to
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,8 +32,11 @@ Status of the `main` branch. Changes prior to the next official version change w
     body from the position the symbol used to occupy (#1593)
   - Fix: Vue's `_ensure_vue_files_indexed_on_ts_server` marked indexing complete even when every
     `.vue` file failed to open on the companion TypeScript server, so a transient outage of that
-    server became a silent, permanent loss of Vue cross-file indexing for the rest of the session
-    (#1923)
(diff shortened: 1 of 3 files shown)


AmirF194 (commented):
I ran this against `86bd012ff31ff622e5eb7b8dce4f906f5a147691` in a clean container.

`_ensure_vue_files_indexed_on_ts_server` (vue_language_server.py:270-303) leaves `_vue_files_indexed` False on total failure so a later call can retry. But its only caller, `_ensure_ls_operational` (line 224), sets `_ls_operational_ready_event` unconditionally right after that call (line 227), and every call to `_ensure_ls_operational`, retries included, short-circuits on that event at line 205 before it ever reaches the vue-index call again. The event only clears on server stop/start. So a single companion-server hiccup still loses Vue indexing for the rest of the process; the fix adds a warning but the retry it describes never runs.

```
$ PYTHONPATH=/repo/src python3 -m pytest /tmp/test_retry_bug.py -v
test_retry_bug.py::test_failed_vue_index_is_never_retried_via_ensure_ls_operational PASSED
```

The test calls `_ensure_ls_operational()` with a failing companion server, then again with a working one, and confirms the second call never re-attempts indexing.

Gating the event on the outcome would make the retry real:

```python
self._ensure_vue_files_indexed_on_ts_server()
if self._vue_files_indexed:
    self._ls_operational_ready_event.set()
```

si-kui-a:
@AmirF194 Appreciate the repro test — made this easy to pin down. You're right about the short-circuit. Held off on applying the fix exactly as suggested since `_ensure_ls_operational` also gates definitions/renames/diagnostics/file-references — a dead companion server would cost every one of those a full re-index attempt too, not just vue calls. Kept the event unconditional, added a separate retry under the same lock, plus a 30s backoff. You know this state machine better than I do — flag it if that tradeoff looks off.

AmirF194:
Makes sense, and covers the other gated paths my suggestion did not touch. I have not run this version, but keeping the event unconditional plus a locked retry and backoff is a reasonable way to avoid a repeat re-index storm on a dead companion server.

PR #1973, opened 2026-09-02
fix(jdtls): fall back to a per-instance workspace on concurrent-session collision

### LINKED ISSUE #1944: Concurrent stdio instances on one project hand the same Eclipse JDTLS -data workspace to N language servers, causing unbounded CPU/RAM and silent index corruption
## Checklist

- [x] read the relevant parts of the documentation and verified that the issue cannot be solved by adjusting configuration
- [x] understood that the Serena Dashboard can be disabled through the config
- [x] understood that, by default, a client session will start a separate instance of a Serena server
- [x] understood that, for multi-agent setups, the Streamable HTTP/SSE mode should be used
- [x] understood that non-project files are ignored using either .gitignore or the corresponding setting in `.serena/project.yml`
- [x] looked for similar issues and discussions, including closed ones (see below)
- [x] made sure it's an actual issue, not a question

## Summary

When two stdio Serena instances activate the **same** project, both spawn an Eclipse JDT language server and hand each one the **identical** `-data` workspace directory. Eclipse JDT workspaces are single-writer by design, so the two servers continuously invalidate each other's index. The result is that neither ever reaches an idle state: they sat at **160% and 128% CPU with 1.4 GB + 2.6 GB RSS** on a single Gradle/Java project, indefinitely, on an otherwise-healthy machine.

Nothing warns the user. There is no lock, no error surfaced through MCP, and no log line indicating the collision — it presents only as the machine becoming slow.

## Relationship to existing issues

I found the prior art and want to be explicit about what is and isn't new:

- **#1235** (*Duplicate Serena instances and unsafe concurrent writes when the same project directory is open in multiple MCP clients*) — closed 2026-03-27 pointing at the documented [single persistent server mode](https://oraios.github.io/serena/02-usage/040_workflow.html#multiple-agents-accessing-a-single-serena-instance). That resolution addresses the *duplication*; this report is about the concrete **unsafe-write mechanism** on the JDTLS side, which remains unguarded regardless of whether the user knows about HTTP mode.
- **#1864** — the same *class* of problem (N sessions racing a single-instance resource) but for the **JetBrains** backend and `DirectoryLock`. This is the **LSP/EclipseJDTLS** equivalent.
- **#937** — `init_language_server_manager` never completing despite JDTLS finishing; possibly a downstream symptom of the same contention.

The argument for treating this separately: stdio-per-session is the **default**, HTTP mode is opt-in, and the failure mode here is silent index corruption plus a multi-GB / multi-core resource leak. A user following the defaults gets no signal that anything is wrong.

## Environment

- Serena **1.2.0** (uv tool), `--context=copilot-cli --project-from-cwd`
- `language_backend: LSP`
- Ubuntu 24.04.4 LTS, 12-core i7-1365U
- Project: Java/Gradle (`build.gradle`, `settings.gradle`), moderate size
- MCP client: GitHub Copilot app — each session spawns its own stdio Serena instance

## Reproduction

1. With `language_backend: LSP`, open **two** concurrent MCP sessions whose cwd is the **same** Java project.
2. Each activates the project and starts an Eclipse JDT language server.
3. Inspect the `-data` argument of both JDTLS processes.

Both receive the same path:

```console
$ tr '\0' '\n' < /proc/625237/cmdline | grep -A1 -x -- -data | tail -1
/home/sp/.serena/language_servers/static/EclipseJDTLS/workspaces/2269b098e07d8bdba1bd239a2fac6177/data_dir

$ tr '\0' '\n' < /proc/626656/cmdline | grep -A1 -x -- -data | tail -1
/home/sp/.serena/language_servers/static/EclipseJDTLS/workspaces/2269b098e07d8bdba1bd239a2fac6177/data_dir
```

The workspace hash is derived from the project path, so *every* concurrent session on a given project collides on one workspace directory.

Neither settles. Sampled ~5 minutes after start, with the project otherwise idle:

```
pid=625237   160.0% CPU   1428 MB
pid=626656   128.2% CPU   2579 MB
```

## Secondary observation: repeated StackOverflowError in the shared workspace

The same workspace's Eclipse log contains **149 `StackOverflowError` occurrences** across **10 rotated `.bak_N.log` files** (Eclipse rotates on restart, so the workspace has been re-initialised at least 10 times). The stack is regex recursion:

```
java.lang.StackOverflowError
	at java.base/java.util.regex.Pattern$Branch.match(Unknown Source)
	at java.base/java.util.regex.Pattern$GroupHead.match(Unknown Source)
	at java.base/java.util.regex.Pattern$LazyLoop.match(Unknown Source)
	at java.base/java.util.regex.Pattern$GroupTail.match(Unknown Source)
	at java.base/java.util.regex.Pattern$BranchConn.match(Unknown Source)
	at java.base/java.util.regex.Pattern$CharProperty.match(Unknown Source)
```

This looks like catastrophic backtracking on some input file rather than a concurrency symptom, so it may well be a **separate** bug — I'm including it because it shares the workspace and may compound the restart churn. Happy to split it into its own issue if you'd prefer.

## Suggested fixes

Roughly in order of how well they'd contain the failure:

1. **Namespace the JDTLS `-data` directory per language-server instance**, not solely per project path — e.g. append a PID/session/instance suffix. This makes concurrent sessions safe by construction, at the cost of some disk and duplicate indexing.
2. **Reuse a single language server per project across instances** (in-process reference counting), which avoids the duplicate indexing cost entirely and is the semantically ideal fix.
3. **At minimum, detect and warn.** Take an advisory lock on the workspace directory; if it's already held, log loudly and surface an MCP-visible warning rather than proceeding into silent corruption. Even just a startup log line naming the workspace path would make this diagnosable — I only found it by inspecting `/proc/<pid>/cmdline`.

Option 3 alone would have turned this from "my machine is mysteriously slow" into a one-line diagnosis.

## Workaround for others hitting this

Either switch to the documented single persistent server (HTTP/SSE) mode, or avoid concurrent MCP sessions rooted in the same project. To recover after it has happened, stop the orphaned servers and delete the corrupted workspace so it rebuilds:

```bash
rm -rf ~/.serena/language_servers/static/EclipseJDTLS/workspaces/<hash>
```

si-kui-a:
Fixes #1944 with the fuller of the three mitigations you ranked there - namespace the workspace per instance, not just detect and warn.

Eclipse's -data workspace is single-writer, but its path is hashed purely from the project path plus a few settings, so two concurrent stdio sessions on the same project hand their JDTLS processes the identical directory - neither settles, both keep invalidating each other's index, with nothing surfaced through MCP and no log line pointing at the collision.

The shared, project-hashed workspace is still tried first (a non-blocking FileLock, same library and pattern as the JetBrains launch-serialization fix in #1864/#1866), so the common case - one session, possibly restarting - keeps reusing its cached index exactly as before this existed; that's the whole reason the workspace hash is stable across restarts in the first place (see `_compute_workspace_hash`'s own docstring). Only when that lock is already held by another live process does it fall back to a directory suffixed with this process's PID - safe as a disambiguator specifically because PIDs are unique among concurrently running processes, which is the only scenario this path is for. The fallback session indexes independently rather than reusing the shared cache; that's the accepted cost of avoiding the corruption, not a bug. If even the fallback somehow collides, it proceeds unlocked on the shared workspace rather than refuse to start.

Split into two static methods so each piece is independently testable without a JDK or a real JDTLS process: `_try_acquire_workspace_lock` (silent, purely mechanical locking) and `_acquire_workspace_dir` (owns the fallback decision and all logging). 5 tests cover the orchestration directly, including the case that would catch a naive always-namespace approach: re-acquiring the shared workspace after releasing it has to return the *same* directory, not the fallback.

`poe format`/`poe type-check` clean. Full `-m python` suite (179) re-run clean. No JDK locally to run `-m java`, but nothing outside `eclipse_jdtls.py`'s own workspace-selection logic is touched and the new tests exercise the actual fallback decision directly.

### DIFF
--- a/src/solidlsp/language_servers/eclipse_jdtls.py
+++ b/src/solidlsp/language_servers/eclipse_jdtls.py
@@ -18,6 +18,7 @@ from pathlib import Path, PurePath
 from time import sleep
 from typing import Any
 
+from filelock import FileLock, Timeout
 from overrides import override
 
 from solidlsp import ls_types
@@ -207,6 +208,22 @@ class EclipseJDTLS(SolidLanguageServer):
         self._service_ready_event = threading.Event()
         self._project_ready_event = threading.Event()
 
+    @override
+    def stop(self, shutdown_timeout: float = 2.0) -> None:
+        # release the workspace lock (if we hold one; see DependencyProvider.create_launch_command)
+        # before stopping the process, so a fast-following relaunch on the same workspace doesn't
+        # spuriously see it as still held
+        assert isinstance(self._dependency_provider, self.DependencyProvider)
+        lock = self._dependency_provider.workspace_lock
+        if lock is not None:
+            try:
+                lock.release()
+            except Exception as e:
+                log.warning(f"Error releasing Eclipse JDTLS workspace lock: {e}")
+            finally:
+                self._dependency_provider.workspace_lock = None
+        super().stop(shutdown_timeout)
+
     def _create_dependency_provider(self) -> LanguageServerDependencyProvider:
         ls_resources_dir = self.ls_resources_dir(self._solidlsp_settings)
         return self.DependencyProvider(self._custom_settings, ls_resources_dir, self._solidlsp_settings, self.repository_root_path)
@@ -271,6 +288,10 @@ class EclipseJDTLS(SolidLanguageServer):
             self._solidlsp_settings = solidlsp_settings
             self._repository_root_path = repository_root_path
             self.runtime_dependency_paths = self._setup_runtime_dependencies(ls_resources_dir, custom_settings)
+            self.workspace_lock: FileLock | None = None
+            """Held for as long as this JDTLS instance is running, once acquired in
+            create_launch_command(); released by EclipseJDTLS.stop(). None if another live
+            instance already held the lock at launch time (see create_launch_command)."""
 
         @classmethod
         def _create_dep_gradle(cls, gradle_version: str | None = None) -> DownloadedDependency:
@@ -723,6 +744,35 @@ class EclipseJDTLS(SolidLanguageServer):
                 ws_hash_input = (repository_root_path + "|" + jdtls_launcher_jar_path + "|" + workspace_settings_json).encode()
             return hashlib.md5(ws_hash_input).hexdigest()
 
+        @staticmethod
+        def _try_acquire_workspace_lock(ws_dir: str) -> FileLock | None:
+            """
+            Attempt to acquire an exclusive, non-blocking lock on a JDTLS workspace directory.
+
+            Eclipse's own `-data` directory is single-writer; two live JDTLS processes sharing one
+            continuously invalidate each other's index with no error surfaced through MCP (GH
+            #1944). The lock is advisory (nothing currently refuses to start without it, matching
+            the diagnostic-only first step described where this is called) and OS-level: it is
+            automatically released if the holding process dies, so a crashed JDTLS never leaves a
+            stale lock behind for the next launch to trip over.
+
+            :param ws_dir: the JDTLS workspace directory to lock.
+            :return: the acquired, held lock, or None if another live process already holds it
+                (a warning is logged in that case).
+            """
+            lock = FileLock(str(PurePath(ws_dir, ".serena-workspace.lock")), timeout=0)
+            try:
+                lock.acquire()
+                return lock
+            except Timeout:
+                log.warning(
+                    f"Another live process already holds the Eclipse JDTLS workspace at {ws_dir} "
+                    "(likely a concurrent Serena session on the same project). Both will keep "
+                    "running and continuously invalidate each other's index; see "
+                    "https://github.com/oraios/serena/issues/1944."
+                )
+                return None
+
         def create_launch_command(self) -> list[str]:
             # ws_dir is the workspace directory for the EclipseJDTLS server.
             project_hash = EclipseJDTLS.DependencyProvider._compute_workspace_hash(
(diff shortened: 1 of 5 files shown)


src/solidlsp/language_servers/eclipse_jdtls.py
```diff
@@ -207,6 +208,22 @@ def __init__(self, config: LanguageServerConfig, repository_root_path: str, soli
         self._service_ready_event = threading.Event()
         self._project_ready_event = threading.Event()
 
+    @override
+    def stop(self, shutdown_timeout: float = 2.0) -> None:
+        # release the workspace lock (if we hold one; see DependencyProvider.create_launch_command)
+        # before stopping the process, so a fast-following relaunch on the same workspace doesn't
+        # spuriously see it as still held
+        assert isinstance(self._dependency_provider, self.DependencyProvider)
+        lock = self._dependency_provider.workspace_lock
```
AmirF194, line 217: I ran this against 60ec5e7 in a clean container, calling `_acquire_workspace_dir` directly (no JDK needed, matching this PR's own `TestAcquireWorkspaceDir`).

`workspace_lock` (set line 291, cleared only at line 224) is acquired in `create_launch_command` (line 838), which runs from `__init__` before `start()` is ever called. Its only release path is the new `stop()`. But `start()`'s exception handler (ls.py:3183-3191) calls `stop()` only when `is_running()` is true; if `_start_server()` raises before the process is actually running (Popen failing to spawn, or `initialize()` failing immediately), `stop()` is skipped and the lock leaks for the process's lifetime. `ls_manager.py:148` documents the assumption this breaks: "SolidLanguageServer.start() cleans up after itself on failure."

```
attempt 1 (construction whose start() later fails pre-spawn): .../workspaces/deadbeef1234
attempt 2 (same process, same project, nothing else running): .../workspaces/deadbeef1234-instance-71
BUG CONFIRMED: second same-process attempt silently fell back to per-instance workspace
```

I only exercised the lock half directly; the `start()` control flow that skips `stop()` is read from ls.py, not independently triggered with a real spawn failure.

Dropping the `is_running()` guard at ls.py:3188 and always calling `self.stop()` on the failure path would close this; `stop()` is documented to never raise.

si-kui-a:
Thanks for actually running this and tracing it through - confirmed the same thing independently reading `ls.py`/`ls_manager.py`: `stop()` is documented to never raise, and `ls_manager.py`'s own comment at its multi-server startup call site explicitly depends on "start() cleans up after itself on failure" holding unconditionally, not just for post-spawn failures. Dropped the `is_running()` gate in 08874cb exactly as you suggested.

The existing `test_ls_start_cleanup.py` actually had a test asserting the *old*, buggy behavior (`stop` NOT called on a pre-spawn failure) - updated it to assert the corrected behavior instead of just adding a new test alongside a contradicting one. Full `solidlsp` + `test_ls_manager.py` suite still green, plus `poe format`/`poe type-check`.

AmirF194:
Thanks for confirming and fixing it in 08874cb, and for catching the test that asserted the old behavior.

PR #1974, opened 2026-09-02
fix(text-utils): clamp overflowing column in get_index_from_line_col

AmirF194:
`TextUtils.get_index_from_line_col` (`src/solidlsp/ls_utils.py`) turns a `(line, col)`
position into a text index by stepping to the start of the target line and adding `col`,
with no check that `col` stays within that line. When a language server reports a column
one past the line's actual length, the extra offset rolls past the line's own newline and
lands inside the following line instead.

`replace_symbol_body` deletes `[body_start, body_end)` before inserting the new body at
`body_start`. On GDScript projects, Godot's `documentSymbol` reports a symbol's body-end
column one past the line's length (confirmed: for `\tprint("old")`, 13 characters long,
it reports column 14). With the overflow, the delete range extends into the newline that
separates the symbol from the next one, so replacing a symbol's body also removes the
blank line after it.

Fix: `get_index_from_line_col` now clamps `col` to the target line's length before adding
it to the line's start index, so an out-of-range column resolves to the end of that line.

## Verification

- Reproduced the position-math root cause directly (no LSP involved):
  `delete_text_between_positions(text, 2, 0, 3, 14)` on a two-function snippet returned a
  deleted range that included the separating blank line's newline; after the fix it stops
  exactly at the line's own end, and the full delete-then-insert sequence
  `replace_symbol_body` performs preserves the blank line.
- Added two regression tests in `test/solidlsp/test_text_utils.py`, both failing on
  unmodified `main` and passing on this branch: one pins the clamp on
  `get_index_from_line_col` directly, the other reproduces the reported symptom end to end.
- `pytest test/solidlsp/test_text_utils.py test/serena/test_file_tools.py`: 32/32 passed.
- `ruff format --check` / `ruff check` / `codespell` on the changed files: clean.
- I have not run this against a live Godot LSP session; the fix and tests target the
  position-math primitive the issue traces the bug to.


opcode81:
This is not fixing an actual issue in Serena; it is a workaround for a bug in the Godot LS. (The nature of the change should ultimately be adequately reflected in the change log.)

It is problematic to assume that clamping an invalid column value to the valid range is definitely the right thing to do and will result in the desired effects. Ultimately, this glosses over an error and, as a result, prevents issues in implementations from being detected.

From the perspective of Serena, the "right" thing to do would be to raise `InvalidTextLocationError` in such cases and for this to be fixed upstream - though this is perhaps not pragmatic. 
It might be worthwhile to investigate why Godot is returning an invalid value and whether this might be a systematic issue (Godot returning 1-based indices instead of 0-based ones or sth like that).
If this is a systematic error in Godot, then the right thing to do in Serena is to fix this one layer deeper - i.e. where the incorrect column index is returned and processed by SolidLSP's abstraction layer, such that when the text processing happens, the coordinates are already fixed.

AmirF194:
Fair, the clamp does gloss over it. What I actually verified: the isolated repro (col=14 against a 13-char line) is one case where Godot's documentSymbol reports the body-end column one past the line's length. I haven't surveyed other symbols or positions to say whether that's systematic (an off-by-one across Godot's LSP) or specific to how it closes a body at end of line, and I couldn't find an existing Godot issue that matches it.

get_index_from_line_col is shared across every LS integration, so even confirming the root cause is Godot-side wouldn't let this layer special-case it, it can only be strict or permissive for everyone that calls it. Given that, and the narrow-clamp precedent already in this function family (cb38f6f8), I'd keep the clamp as the pragmatic default here and open a separate issue to chase the Godot-side cause. Want me to hold this PR until that's settled instead?

opcode81:
In the case of cb38f6f82aac111971e0bc96c542dbcd30eef946, it is a different case entirely, because we are only compensating for different ways of encoding a very specific coordinate - the end of the file.

The issue at hand is actually more subtle.
The semantics of LSP text coordinates are as follows: the start coordinate is inclusive, the end coordinate is *exclusive*, and you acknowledge that Serena adheres to it in your initial PR description:

> `replace_symbol_body` deletes `[body_start, body_end)` before inserting the new body

`body_end` itself not being included in the deletion is essential.
If a symbol body ends at the end of a line, language servers will typically return an index one past the actual maximum index in the line. This is not an error but expected behaviour. The combination of Serena simply adding the column coordinate to the relevant line's start index when computing indices, i.e.

https://github.com/oraios/serena/blob/bcb3fc74ff558b9e1ae017f84b11b755cd561ca2/src/solidlsp/ls_utils.py#L237

and Python's end-exclusive slicing,

https://github.com/oraios/serena/blob/bcb3fc74ff558b9e1ae017f84b11b755cd561ca2/src/solidlsp/ls_utils.py#L285

we obtain exactly the right behaviour.

The error in Godot appears to be that it returns an index that is *two* past the actual maximum 0-based index, and the confusing part about the change in your PR is that the `max_col` you compute is actually not the maximum 0-based column coordinate allowed for the line but typically 1 plus that (depending on the line ending), which is why it still works for language servers that correctly add only 1 to the end column coordinate.
Note that the maximum admissible column coordinate for your example `\tprint("old")` is not 13 but 12.

So the core conclusion doesn't change ("This is a bug in Godot."), but the changes in your PR are at least slightly confusing.

opcode81:
In the case of cb38f6f82aac111971e0bc96c542dbcd30eef946, it is a different case entirely, because we are only compensating for different ways of encoding a very specific coordinate - the end of the file.

The issue at hand is actually more subtle.
The semantics of LSP text coordinates are as follows: the start coordinate is inclusive, the end coordinate is *exclusive*, and you acknowledge that Serena adheres to it in your initial PR description:

> `replace_symbol_body` deletes `[body_start, body_end)` before inserting the new body

`body_end` itself not being included in the deletion is essential.
If a symbol body ends at the end of a line, language servers will typically return an index one past the actual maximum index in the line. This is not an error but expected behaviour. With Serena simply adding the column coordinate to the relevant line's start index when computing indices, i.e.

https://github.com/oraios/serena/blob/bcb3fc74ff558b9e1ae017f84b11b755cd561ca2/src/solidlsp/ls_utils.py#L237

and Python's end-exclusive slicing,

https://github.com/oraios/serena/blob/bcb3fc74ff558b9e1ae017f84b11b755cd561ca2/src/solidlsp/ls_utils.py#L285

we obtain exactly the right behaviour.

The error in Godot appears to be that it returns an index that is *two* past the actual maximum 0-based index.
A slightly confusing part in your PR is that the `max_col` you compute is actually not the maximum 0-based column coordinate allowed for the line but exactly 1 plus that (since `prev_line_end_idx` is exclusive), which is why it still works for language servers that correctly add only 1 to the end column coordinate.
Note that the maximum admissible column coordinate for your example `\tprint("old")` is not 13 but 12.

So the core conclusion doesn't change ("This is a bug in Godot."), but the changes in your PR are at least slightly confusing.

AmirF194:
You're right, and it's exactly the naming ambiguity you called out: `max_col` held one past the last valid character index, not that index itself. Renamed it to `line_end_col` and reworded the comment to say it's the exclusive bound (`prev_line_end_idx`), pushed as 1e38b3b. The clamp value itself doesn't change; both new tests and the full `test_text_utils.py` suite still pass.

Agreed the core conclusion holds: this compensates for Godot's specific off-by-two, not a general LSP coordinate convention, and a deeper fix belongs in SolidLSP's Godot integration if it turns out to be systematic.

opcode81:
I would suggest that we

* analyse whether Godot is indeed consistently off by one (max_col+1 being expected for symbols that end at the end of a line, max_col+2 is off by one), and if so, fix it in the language server adapter (and report it upstream).
* (optionally) keep the the maximum end index computation and use it to raise an exception if the index is out of bounds; clamping is not appropriate.

AmirF194:
Resolving the conflict with #1842 (just merged) surfaced something relevant to your second point. Raising instead of clamping in get_index_from_line_col would break #1842's own test_insert_position_inside_newline_sequence_splits_it: that test calls insert_text_at_position with a column that only makes sense as a raw index into a CRLF sequence, not a real end-of-line column. So the two callers want different semantics. delete_text_between_positions wants clamping (or raising) on a genuinely out-of-range LSP column; insert_text_at_position's internal split case wants the old unclamped pointer arithmetic.

On the off-by-one question, I only have the one reproduced case (col=14 on a 13-char line) and no second Godot range to compare it against, so I can't say whether it's consistently max_col+1.

I'd rather not push a merge resolution that quietly picks a semantics here. Could split get_index_from_line_col's LSP-facing callers from insert_text_at_position's internal one if that's the direction you'd want, or wait for your call.

opcode81:
> insert_text_at_position's internal split case wants the old unclamped pointer arithmetic

This case was never a valid case to begin with. Deleting the test is the right measure to take.

opcode81:
> On the off-by-one question, I only have the one reproduced case (col=14 on a 13-char line) and no second Godot range to compare it against, so I can't say whether it's consistently max_col+1.

As I said, the task is to analyse whether it is consistently true_end_col+2. We won't merge the clamping, because it's wrong.

AmirF194:
Traced it into Godot's own source instead of guessing (godotengine/godot, gdscript_extend_parser.cpp / gdscript_parser.cpp / gdscript_tokenizer.cpp). GodotPosition::to_lsp() does character = column - 1, a clean 1-indexed to 0-indexed conversion, so a token's own end_column already lines up correctly on its own. What looks off is gdscript_parser.cpp:837-838, which closes a node's range from `current` (the next lookahead token) rather than `previous` (the token actually consumed last). When `current` is the synthesized NEWLINE, tokenizer.cpp's newline() sets its end_column to the column reached after the newline character is itself consumed, one column past where a real content token would end. Stack that on the usual one-past-end convention and you get exactly the +2, structurally rather than per-instance.

I have not set up a live Godot LSP session to confirm this end to end, only read the source, so call this likely, not verified. Worth building the repro to check, or do you already have a second Godot case handy?

opcode81:
> Traced it into Godot's own source instead of guessing (godotengine/godot, gdscript_extend_parser.cpp / gdscript_parser.cpp / gdscript_tokenizer.cpp). GodotPosition::to_lsp() does character = column - 1, a clean 1-indexed to 0-indexed conversion, so a token's own end_column already lines up correctly on its own. What looks off is gdscript_parser.cpp:837-838, which closes a node's range from current (the next lookahead token) rather than previous (the token actually consumed last). When current is the synthesized NEWLINE, tokenizer.cpp's newline() sets its end_column to the column reached after the newline character is itself consumed, one column past where a real content token would end. Stack that on the usual one-past-end convention and you get exactly the +2, structurally rather than per-instance.

Thanks for looking into it, @AmirF194.
I suppose this is sufficient to build the proper fix:  
* Override `_build_document_symbols_from_raw_symbols` in the Godot LS, adding post-processing: Iterate over all symbols and modify `symbol["location"]["range"]["end"]["character"]` accordingly. We should add a TODO to gate the workaround behind a Godot version check once the issue is fixed upstream.
* Override `_document_symbols_cache_fingerprint` the cache fingerprint method to account for the change.
* Report the issue upstream.

AmirF194:
Pushed 1f5dc3b: reverted the two get_index_from_line_col commits and moved the fix into GodotLanguageServer._request_document_symbols instead, correcting the measured +1 overshoot on range/selectionRange/location recursively before the raw symbols reach the shared position math. Bumped the raw-symbol cache version per ls.py's own instructions for that override.

Added test/solidlsp/util/test_godot_symbol_fix.py, unit tests against the issue's own line fixture (13-char line, reported end column 14), plus the no-overshoot and larger-overshoot cases left alone. Ran it alongside the full test/solidlsp/test_text_utils.py suite in a clean container: 24 passed.

Left the version-gate TODO in place since I haven't nailed down which Godot version(s) this affects beyond 4.6.3, and haven't filed the upstream Godot issue yet, still on my list.

Merge conflict with #1842 is resolved now that the fix no longer touches ls_utils.py at all.

AmirF194:
The type checker caught two real mistakes in that push: `_fix_range_end` took `dict[str, Any]` but the LSP `Range` TypedDicts it actually receives aren't assignable to that, and I'd overridden a method named `_request_document_symbols`, which doesn't exist on the base class. The real hook is `_request_raw_document_symbols`. Fixed both and re-verified locally with the repo's pinned `ty` (0.0.24) via `uv sync --extra dev --locked`: `ty check src/serena src/solidlsp` and `ty check test` both pass clean, and the 7 new tests still pass. Pushed 3e6c48d.

opcode81:
Thanks for continuing this. Is there a reason why you didn't implement it by overriding `_build_document_symbols_from_raw_symbols` as suggested?
Overrides of the raw method necessitate invalidation of the full LS cache; the high-level cache is much faster to recompute.

Also, it seems that your agent is not following the instructions on encapsulation in the `critical_info` memory. Please check whether your agent is reading this memory.
When using Serena, the agent receives it either
* when in a single-project context, upon calling the `initial_instructions` tool (which agents are instructed to always call in Serena's system prompt)
* when not in a single-project context, upon calling the `activate_project` tool.

AmirF194:
No good reason, that's the right call. Moved it: pushed 9368c7c, the correction now runs inside _build_document_symbols_from_raw_symbols on the converted UnifiedSymbolInformation tree (same location/range/selectionRange/children shape), bumping only _document_symbols_cache_fingerprint. The raw cache stays untouched now.

Also moved _fix_range_end/_fix_symbol_ranges into GodotLanguageServer as staticmethods per critical_info's encapsulation note, since they're only used here. Not sure what specifically prompted the memory-reading question beyond that; happy to fix anything else it points at if you can be more concrete.

Ran the full unit suite, ruff, and ty check clean (24/24 tests pass). Didn't repeat the live Godot editor session from the original repro since this change doesn't touch the correction mechanism itself, only which method wires it in, and the two code paths carry the same dict shape by construction.

AmirF194:
Rebased this onto the SPDX/relicensing merge and re-verified locally, still green apart from the unrelated niche OCaml setup flake on main. Let me know if the _build_document_symbols_from_raw_symbols placement from 09-07 needs another pass.

opcode81:
@AmirF194 thank you very much for this contribution!

AmirF194:
Thanks for pushing back on the clamp, tracing it into Godot's parser instead was the right fix.

PR #1977, opened 2026-09-02
fix(dart): wait for analyzer readiness signal before returning from start()

AmirF194:
`DartLanguageServer._start_server` registers handlers for the two notifications the Dart
analysis server sends to report indexing progress, `$/analyzerStatus` (`isAnalyzing`) and
`experimental/serverStatus` (`quiescent`), but both handlers were no-ops, and `_start_server`
returned right after sending `initialized`, with no wait on either signal.

The only readiness gate left is the generic one-time 2s sleep in the base class
(`_get_wait_time_for_cross_file_referencing`). Every other backend that receives these same
two notifications (pyright, basedpyright, rust-analyzer, clojure-lsp, jedi, omnisharp, clangd)
overrides that with a real wait; Dart is the only one that gets the handler but not the logic
behind it. On a project whose initial analysis takes longer than ~2s, `find_symbol` /
`find_referencing_symbols` issued right after activation can return before the workspace scan
finishes, silently incomplete.

This adds a `threading.Event` gated on either notification (mirroring `pyright_server.py`'s
`analysis_complete` pattern) and waits on it, bounded by 60s, before `_start_server` returns.
A timeout logs a warning and proceeds rather than blocking Serena forever if a future Dart SDK
stops sending these.

Verified against the real Dart analysis server (3.7.1, downloaded fresh), not just the mocked
unit tests below: with logging enabled, `_start_server` sends `initialized` at T+0, and the
real server's own `$/analyzerStatus {isAnalyzing: false}` doesn't arrive until roughly 1.4s
later, during which the old code already considered the server ready. All 20 existing
`test_dart_basic.py` / `test_dart_diagnostics.py` tests still pass against the real server
with this change.

- Added `test_start_waits_for_analysis_readiness_signal` (both notification shapes): starts
  `start()` on a background thread, asserts it is still blocked before either signal is sent,
  then asserts it resumes once one is. Fails on `main` on the actual assertion (start() returns
  immediately); passes with this fix.
- Added `test_start_proceeds_after_readiness_timeout_if_no_signal_arrives` for the timeout
  fallback path.
- The existing `test_analyzer_status_notification_is_handled_after_startup` (from #1863) is
  unchanged in behavior; I only patched its timeout down so it doesn't block for a full 60s
  now that `start()` actually waits.
- `ruff format --check`, `ruff check` (pinned 0.12.5) and `codespell` all pass clean on the
  touched files.

I have not tried this on Windows or macOS, or against a pre-LSP-3.18 Dart SDK old enough to
send only one of the two notification shapes; this fix handles either alone, but I don't have
a build that old to confirm against.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,12 @@ Status of the `main` branch. Changes prior to the next official version change w
   - Fix: Exceptions raised during `LanguageServerManager.start` did not stop the language server subprocess if it was
     already started (#1949)
   - Fix: Dart's `$/analyzerStatus` notifications were logged as unhandled-method warnings during analysis (#1855)
+  - Fix: `DartLanguageServer._start_server` discarded both `$/analyzerStatus` and
+    `experimental/serverStatus`, the two notifications the Dart analysis server sends to report
+    indexing progress, and returned as soon as `initialized` was sent instead of waiting for either
+    one; a request issued right after activation (`find_symbol`, `find_referencing_symbols`) could
+    return before the workspace scan finished. Serena now waits (bounded by 60s) for either signal to
+    report completion, matching the pattern already used for pyright, basedpyright and rust-analyzer
   - Fix: clojure-lsp was not told that Serena sends `workspace/didChangeWatchedFiles`, so changes made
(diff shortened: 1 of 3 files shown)


opcode81:
Thanks!

PR #1978, opened 2026-09-03
fix(typescript): drain in-flight indexing on later cross-file queries

### LINKED ISSUE #1937: TypeScript: find_referencing_symbols returns silently partial results while the project graph is still loading — Serena waits only before the FIRST cross-file query
## Setup

Serena 1.7.0 (`uv tool install -p 3.13 serena-agent`), MCP client Claude Code, macOS 26.5 arm64. Language: TypeScript, `typescript-language-server` 5.1.3, workspace TypeScript 6.0.3. Context `claude-code`, started as `serena start-mcp-server --context claude-code --project-from-cwd`. pnpm monorepo, 1406 tracked `.ts`/`.tsx` files, TypeScript project references configured throughout.

## What happens

`_wait_for_cross_file_references_if_needed` latches after the first cross-file query and never waits again:

```python
# src/solidlsp/language_servers/typescript_language_server.py:546-558
def _wait_for_cross_file_references_if_needed(self) -> None:
    if self._has_waited_for_cross_file_references:
        return
    ...
    self._has_waited_for_cross_file_references = True
```

But the first cross-file query is exactly what makes tsserver start pulling in referenced and referencing projects, and that work continues after it has answered. A `find_referencing_symbols` issued into that window returns **some** of the references — not an error, not an empty result, and nothing that distinguishes a partial answer from a complete one.

This is the same shape as #1858 (Scala, fixed by waiting properly before the first query). The difference is that here the first query is fine and the *next* ones are wrong, so the existing fix does not cover it.

## Reproduction

Six identical calls in one session, `find_referencing_symbols` on `formatDisplay` (exported from a leaf package, 11 real consumer files across two apps and one package):

| # | Fired | Files returned |
|---|---|---|
| 1 | after startup wait | **11** ✅ |
| 2 | +0 ms | **1** ❌ |
| 3 | +0 ms | **1** ❌ |
| 4 | +2 s | **11** ✅ |
| 5 | +0 ms | **11** ✅ |
| 6 | +2 s | **11** ✅ |

Note call 5: an immediate repeat is fine *once the graph is quiet*. The failure window is tied to graph loading, not to call spacing.

Mixed symbols show the same thing, and worse. Five different reference queries back to back, each ~200 ms after the previous response:

| Query | Returned | Correct |
|---|---|---|
| `BottomNav` (first) | 2 | 2 |
| `BottomNav` (other platform twin) | **0** | 2 |
| `formatDisplay` | **1** | 11 |
| `getAdminClient` | **0** | 82 |
| `requireFamilyMemberHono` | **1** | 29 |

Re-running the same sequence with a gap between calls returned all five in full. Every query that touches a file not queried yet in the session re-opens the window, because it triggers more project loading.

## Why this matters more than a slow answer

An agent cannot tell this from "this symbol has no references". #1814 makes the same point about the OOM case, and the reasoning applies unchanged here: fast, confident, and wrong. An agent will conclude a refactor has no downstream callers.

## Not #1814

The tsserver process does not die. The server recovers inside the same session — calls 4, 5 and 6 above are complete after calls 2 and 3 were partial — which a crashed process cannot do. Memory was not near the V8 cap; the repo is ~1.4k TS files, not Kibana-scale.

## Suggestions

1. Wait per cross-file query rather than once per session — drop the `_has_waited_for_cross_file_references` latch, or re-arm it whenever a query opens a file that was not already open.
2. If waiting every time is too costly, surface the uncertainty instead of hiding it: when the last `$/progress` indexing token is still open, mark the result partial in the tool response so the client can re-query.

Happy to test a patch against this repo — the discrepancy is reproducible on demand here.

AmirF194:
`_wait_for_cross_file_references_if_needed` gates the pre-request indexing wait behind
`_has_waited_for_cross_file_references`, a latch set once after the first cross-file query and
never reset. The first query is exactly what makes tsserver start pulling in referenced and
referencing projects, and that indexing continues after it has answered.

A later `find_referencing_symbols`/`find_references` call that opens a file from a project
tsserver has not loaded yet restarts that indexing but skips the wait entirely, because the latch
is already set, so it can return while its own `$/progress` token is still open. The result is
silently partial: the reporter's own six-call sequence on one symbol returns 11 references, then
1, then 1, then 11 again, with nothing distinguishing the partial answers from complete ones.

The latch still covers the first query's start-grace wait unchanged (that grace exists to give
tsserver a window to *begin* reporting progress, and re-running it on every query would add a
multi-second delay to queries that trigger no new indexing at all). A later query now also checks
whether a progress token is already active and, if so, waits for it to drain via the same
`wait_for_indexing` the first query uses.

Scoped to the confirmed repro: it waits for progress that has already started by the time the
check runs. A narrower race, where a later query's own progress notification has not arrived yet
at that instant, is the same shape #1586 fixed for the first query but is not covered here, since
closing it would mean re-running the start-grace wait on every query.

## Verification

- New test, `test_second_call_drains_a_progress_token_already_in_flight`, fails on `main` (returns
  in under a millisecond while the token is still active) and passes on this branch (blocks until
  the token drains).
- Full `test/solidlsp/test_typescript_timeout_policy.py` (the file's own regression suite for this
  policy): 31 passed.
- `coverage run --source=solidlsp.language_servers.typescript_language_server -m pytest
  test/solidlsp/test_typescript_timeout_policy.py`: both new branches and the changed lines are
  covered, apart from the pre-existing `_pre_open_for_cross_file_references` latch check, which
  this PR does not touch and was already untested before it.
- `ruff format --check` and `ruff check` on the changed files, and `codespell`: clean.
- Not run: the full pytest matrix (native/jvm/other-langs across OS) and CodeQL, since those spin
  up real language servers this environment cannot provision; the change is confined to
  `typescript_language_server.py`'s own progress bookkeeping and does not touch any other server.

Fixes #1937

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ Status of the `main` branch. Changes prior to the next official version change w
 
 * Language Servers:
   - Add Fatou support as an alternative Julia language server (`julia_fatou`)
+  - Fix: TypeScript's `_has_waited_for_cross_file_references` latch was set after the first
+    cross-file query and never reset, so a later query that opened a file from a project tsserver
+    had not loaded yet (e.g. a monorepo package) skipped the indexing wait even while that
+    project's own `$/progress` indexing was still in flight (#1937)
   - Fix: Nextflow's `_flush_deferred_workspace_scan` marked the workspace scan flushed even when both
     of its `completion` probes failed, permanently skipping the flush (and silencing retries) for the
     rest of the session (#1871)
--- a/src/solidlsp/language_servers/typescript_language_server.py
+++ b/src/solidlsp/language_servers/typescript_language_server.py
@@ -591,18 +591,29 @@ class TypeScriptLanguageServer(SolidLanguageServer):
 
     @override
     def _wait_for_cross_file_references_if_needed(self) -> None:
-        if self._has_waited_for_cross_file_references:
+        timeout = self._get_indexing_timeout()
+        if not self._has_waited_for_cross_file_references:
+            start_grace = self._get_indexing_start_grace()
+            self._log_cross_file_indexing_wait_outcome(
+                self._wait_for_indexing_start_or_completion(timeout=timeout, start_grace=start_grace), timeout
+            )
+            self._has_waited_for_cross_file_references = True
             return
 
-        timeout = self._get_indexing_timeout()
-        start_grace = self._get_indexing_start_grace()
-        if self._wait_for_indexing_start_or_completion(timeout=timeout, start_grace=start_grace):
+        # The latch above only covers the first query; a later one can still open a file from a
+        # project tsserver has not loaded before, starting a fresh $/progress cycle to drain.
+        with self._progress_lock:
+            indexing_in_progress = bool(self._active_progress_tokens)
+        if indexing_in_progress:
+            self._log_cross_file_indexing_wait_outcome(self.wait_for_indexing(timeout=timeout), timeout)
+
+    def _log_cross_file_indexing_wait_outcome(self, completed: bool, timeout: float) -> None:
+        if completed:
             log.info("TypeScript cross-file indexing complete")
         else:
             log.warning(
                 "TypeScript cross-file indexing did not complete within %.0fs; proceeding (%s)", timeout, self.describe_indexing_state()
             )
-        self._has_waited_for_cross_file_references = True
 
     @override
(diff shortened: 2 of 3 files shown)


AmirF194:
This one's been quiet for a week with no review yet, wanted to flag it back to the top. The fix is scoped to the TypeScript language server's cross-file wait latch, nothing shared, so it shouldn't need much back and forth.

opcode81:
Thanks!

AmirF194:
Thanks for taking a look, glad the wait latch scoping made review straightforward.

Issue #1961, opened 2026-09-01
New files are not picked up by the roslyn language server


sse-virto:
## Description

Files created after the C# language server (`csharp`, Roslyn) has loaded its projects never join the project's compilation. `get_diagnostics_for_file` then reports phantom errors for the whole session:

- Files referencing a type defined in a new file: `CS0246 The type or namespace name 'X' could not be found`, `CS0103`
- The new file itself: `IDE0005 Using directive is unnecessary` on usings that are required
- `dotnet build --no-incremental` on the same project: `Build succeeded`, 0 errors

### Hints

It seems the new file is opened as a Miscellaneous Files document — compiled standalone against a bare reference set — so nothing in it resolves and its usings look unused, while the rest of the project cannot see its types.

## Reproduction

1. Start Serena on a C# project and let the language server finish indexing.
2. Move a type out of an existing file into a new file in the same project (any way of creating a `.cs` file works).
3. Call `get_diagnostics_for_file` on the original file and on the new file.

Observed: `CS0246`/`CS0103` on the original file, `IDE0005` on the new file's required usings.
Expected: the diagnostics `dotnet build` produces — none.

## Analysis

File watching does reach the server: `sync_file_system_changes` logs
`File system polling complete; 3 change events sent to language servers.` Content sync of pre-existing files works.
It seems Roslyn adds documents when it re-evaluates a project, not on a watched-file `Created` event.

## Related

#1593 (`find_symbol` can return stale/incorrect information, observed for Clojure/LSP)

Seems similar issue but unrelated fix.

Preconditions:

- [x ] I have made sure it's an actual issue, not a question (use [GitHub Discussions](https://github.com/oraios/serena/discussions) instead).
- [x ] I have consulted the [user guide](https://oraios.github.io/serena/02-usage/) and verified that the issue cannot be resolved by adjusting configuration/following recommended workflows.
- [x ] I have looked for similar issues and discussions, including closed ones.

Issue details:

- [ x] I have provided a meaningful title and description.
- [ x] I have explained how the issue arose and, where possible, added instructions on how to reproduce it.
- [ x] I have added details on my setup: Serena version, MCP client, OS, the programming language(s), relevant configuration adjustments and project specifics.
- [ x] If the issue relates to an application of Serena to an open-source project, I have added the link.

opcode81:
@sse-virto please test the now merged solution. It would be great to get some feedback on this.

sse-virto:
[...]

PR #1981, opened 2026-09-03
fix(csharp): fold newly created files into the loaded Roslyn project

AmirF194:
Roslyn's project system only learns which files belong to a project from the `solution/open`/`project/open` notifications `_open_solution_and_projects` sends once at startup. The generic `didChangeWatchedFiles`/open-close cycle that `poll_and_notify` sends for every backend on file creation doesn't make it re-evaluate the project, so a `.cs` file created after the project is already indexed stays a standalone Miscellaneous Files document. That's the phantom `IDE0005 "Using directive is unnecessary"` you saw: Roslyn can't resolve the imported type, so it thinks the using is unused, and `dotnet build` disagrees because it isn't confused about project membership.

Added `SolidLanguageServer.notify_files_created`, a no-op hook called by `poll_and_notify` right before it opens newly created files. `CSharpLanguageServer` overrides it to resend the same solution/project notifications and wait for Roslyn's own reload-complete log line (`workspace/projectInitializationComplete` fires too early on a re-open to use for this). Every other backend keeps its current behavior via the default no-op.

Added a test that creates a `.cs` file referencing an existing type after the server has started, runs it through the real `sync_file_system_changes` path, and checks two things: no phantom "unnecessary using", and the real compiler diagnostic for the missing constructor argument (which only shows up once the file is actually part of the compilation). The new test fails without the change and passes with it. `poe lint`, `codespell`, and `poe type-check` are clean; the rest of the C# suite and `test_ls_manager.py` still pass. Didn't run the non-C# suites here since this environment only has the C#/dotnet toolchain set up.


src/solidlsp/language_servers/csharp_language_server.py
```diff
@@ -758,3 +767,19 @@ def _open_solution_and_projects(self) -> None:
...
+        `_open_solution_and_projects`); a plain `didChangeWatchedFiles`/open-close cycle
+        does not make it re-evaluate the project, so a file created after startup is
+        analyzed as a standalone Miscellaneous Files document instead of being folded into
+        the already-loaded compilation. Resend the same notifications and wait for the
+        server to confirm the reload before the file is opened.
+        """
+        self._project_reload_complete.clear()
+        self._open_solution_and_projects()
+        if not self._project_reload_complete.wait(30):
+            log.warning("Timeout waiting for project reload after new file(s) were created: %s", list(relative_file_paths))
```
opcode81, line 785: This looks like it could be a very expensive operation. It feels wrong to completely reopen the project only because a new file was created. Are you sure this is the right way to handle this?

AmirF194: You're right, and it's worse than "feels expensive": notify_files_created calls _open_solution_and_projects(), which re-scans the whole tree for every .sln/.csproj and resends solution/open plus project/open for all of them, not just the project that owns the new file. On a big solution that's the full startup reload cost again, on every batch of created files.

The targeted version would resolve the new file's nearest ancestor .csproj and resend project/open for just that one path, skipping solution/open and the other projects entirely. I didn't build that here because I can't verify it against live Roslyn without a real multi-project .csproj fixture, which is more setup than this PR currently exercises. Want it done as a follow-up in this PR, or should I open a separate issue for the scoped version and leave this as is for now?

opcode81: Even limiting this to a single project could be prohibitively expensive. This cannot be the right way to go about this.

AmirF194: Fair. project/open on this server does a full reload of whatever project it names, so scoping to one project only shrinks which project pays the cost, not the cost itself. I don't see a lighter primitive exposed here, `_open_solution_and_projects` only ever sends `solution/open` and `project/open`, nothing for adding a single file to an already-loaded project.

Two directions from here: drop this PR and document the gap on #1961, since the real fix needs an API this server doesn't offer; or restart the language server on file creation, heavier but at least explicit instead of a hidden reload. Do you know a cheaper op on this server I'm missing, or is dropping it the right call?

AmirF194:
Circling back on this: do you know a cheaper reload op on this server, or should I drop the PR and document the gap on #1961 instead?

opcode81:
I had an agent research Roslyn's behaviour, and it came to the conclusion that Roslyn actually already has an internal mechanism that will trigger a project reload when a `didChangeWatchedFiles` notification is received.
So the question is:
* Why didn't this mechanism apply in case of the reporter of #1961
* Can the behaviour be reproduced, verifying that that this is a global issue affecting all Serena/Roslyn users?

AmirF194:
On the "why doesn't it apply" question: the PR's own regression test (test_new_csharp_file_is_folded_into_compilation) already answers this empirically, verified at submission against a real Roslyn backend. It sends the same didChangeWatchedFiles(Created) plus open/close cycle poll_and_notify sends for every backend; without notify_files_created's resend of solution/open and project/open, the phantom "unnecessary using" still shows up, meaning the file stays a Miscellaneous Files document to this server despite that notification. So whatever internal reload path your agent found isn't triggered by didChangeWatchedFiles alone on this server version, at least not on its own.

On reproducibility: the test is deterministic against a real backend rather than depending on the reporter's specific environment, so it isn't a one-off.

opcode81:
OK, I think it is a reasonable experiment to just merge this.
Most likely, C# users will complain if it is excessively slow.

AmirF194:
Agreed, real C# usage will surface fast if the polling reload is too slow for large projects.

PR #1982, opened 2026-09-03
fix(kotlin): give concurrent Serena instances independent LSP storage

### LINKED ISSUE #1966: Kotlin LSP fails when multiple Serena instances open the same workspace
## Summary

When two independent Serena MCP server processes activate the same Kotlin workspace concurrently, the first instance works normally, but the second instance fails to provide language-server features.

The Kotlin LSP processes appear to use the same default on-disk index location. Because the index storage cannot be opened concurrently, requests from the second Serena instance are repeatedly cancelled.

## Steps to reproduce

1. Configure an MCP client to start Serena using the SolidLSP backend.
2. Start one Serena MCP session and activate a Kotlin workspace.
3. Invoke a symbol-related tool and confirm that the Kotlin language server is working.
4. Without stopping the first session, start another independent Serena MCP session.
5. Activate the same workspace in the second session.
6. Invoke a symbol-related tool from the second session.

## Actual behavior

The second instance cannot use Kotlin language-server functionality and repeatedly reports an error similar to:

```text
solidlsp.lsp_protocol_handler.server.LSPError: cancelled (-32800)

Further investigation indicates that both Kotlin LSP processes contend for the same index storage.

A failed startup may also leave the Kotlin LSP process running after Serena has recorded the startup exception.
```

## Expected behavior

Multiple independent Serena processes should be able to open the same Kotlin workspace concurrently without sharing a locked index database.

## Environment

- Serena: 1.7.0
- Kotlin LSP: 262.9593.0
- Backend: SolidLSP
- MCP transport: stdio
- MCP client: Codex CLI
- Operating system: Linux
- Workspace: local Kotlin project; project details omitted because they are private

## Suggested fix

Create a separate index directory for each KotlinLanguageServer instance and pass it to Kotlin LSP through:

{
  "initializationOptions": {
    "indexDir": "<instance-specific temporary directory>"
  }
}

The directory should be cleaned up when the language server stops. If startup fails, Serena should also stop the partially started language server so that its process and temporary
resources are released.

I tested this approach locally. Two Serena instances were then able to initialize Kotlin LSP concurrently against the same workspace.

## Related issue

Kotlin LSP previously tracked a related problem in:

- https://github.com/Kotlin/kotlin-lsp/issues/108

That issue covered multiple LSP instances opening different projects and was closed after an upstream fix. This report concerns multiple Serena instances opening the same workspace, where
the default index location is still shared.

AmirF194:
Went with a lock instead of always giving each instance its own directory, since you flagged that unconditional per-instance storage would hurt the normal case (losing the index cache on every restart).

`DependencyProvider` now tries to claim the existing per-project cache dir with a non-blocking file lock. If it gets the lock (the common case, including a plain restart), nothing changes: same directory, same cache, same as before. If another live instance already holds it, this one falls back to a directory named after its own pid instead of racing the first instance for the same index. The lock is held for the process's lifetime and released in `stop()`.

Fixed both places two instances were computing the same path: the `--system-path` arg for the modern intellij-server launcher, and `initializationOptions.storagePath`, which was hardcoded `None` for every instance regardless of which launcher is active.

One thing I didn't handle: if a Serena process is killed hard enough to skip `stop()` (SIGKILL, OOM), its per-pid fallback directory is never cleaned up. That only affects the concurrent-instance path, and it's the same "not officially supported" territory you described, but wanted to flag it rather than leave it looking fully closed.

`test_kotlin_dependency_provider.py` covers the lock claim, fallback and release directly; `test_kotlin_startup.py` pins `storagePath` no longer being `None`. Also ran the existing Kotlin suite end to end against the real LSP binary; the two pre-existing failures there aren't from this change and reproduce the same on main.

Fixes #1966

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,12 @@ Status of the `main` branch. Changes prior to the next official version change w
     silently returned an empty result instead of surfacing the crash. The crash is now detected
     independently via the `window/logMessage` notification tsserver already sends, and the
     affected wait now raises instead of reporting success (#1814)
+  - Fix: two Serena instances activating the same project concurrently launched their Kotlin LSP
+    processes against the same on-disk index storage location, so the second instance's requests
+    were repeatedly cancelled by the first instance's server. A Kotlin LSP process now claims that
+    storage directory via a lock; a single instance (including across restarts) still gets the
+    same directory, and a second concurrent instance gets a directory of its own instead of
+    contending for the first one's (#1966)
 
 CLI:
   - Fix `project index-file` command not using only the relevant language server to index the given file (#1965)
--- a/src/solidlsp/language_servers/kotlin_language_server.py
+++ b/src/solidlsp/language_servers/kotlin_language_server.py
@@ -23,6 +23,7 @@ import stat
 import threading
 from dataclasses import dataclass
 
+from filelock import FileLock, Timeout
 from overrides import override
 
 from solidlsp.dependency_provider import DownloadedDependency, DownloadedDependencyHashDatabase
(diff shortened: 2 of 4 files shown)


opcode81:
This looks like a good solution. Thanks for your PR.

AmirF194:
Glad the per-instance fallback landed cleanly. Thanks for reviewing it against the non-parallel case specifically.

PR #1904, opened 2026-08-21
feat(solidlsp): discover installed language server adapters

vitalyruhl:
## Summary

Allow installed external Python language-server adapters to register with
Serena and participate in the existing SolidLSP runtime.

## Motivation

Serena currently requires language-server integrations to be registered in
core.

External language-server support was previously discussed in #285 and #361.
Review of #361 highlighted that static configuration is insufficient because
language servers need implementation-specific startup, initialization, and
event handling. A direction based on real Python language-server
implementations, automatic pickup, and `register_ls`-style registration was
suggested.

This PR implements that direction using Python entry points and Serena's
existing `SolidLanguageServer` abstractions. Python entry points are the
packaging mechanism proposed by this PR; they were not prescribed in the
earlier discussion.

## Design

Installed adapter packages expose a callable through the
`serena.language_servers` entry-point group. The callback registers an ID,
filename matcher, and `SolidLanguageServer` implementation using:

`register_ls(id=..., matcher=..., implementation=...)`

Registered IDs become valid `LanguageServerKey` values.

`ProjectConfig.load()` discovers installed adapters before resolving configured
language-server IDs. The existing language-server factory, dependency
providers, manager, settings, and process infrastructure remain responsible
for creation and execution. No separate custom-LS execution path is introduced.

## Why Python entry points?

Python entry points are a standard installed-package discovery mechanism.

They:

- allow multiple installed adapters
- require no custom plugin database or filesystem scanning
- keep adapter-specific startup and initialization logic in Python
- avoid Python module paths in `project.yml`

This follows the Python-implementation and registration direction discussed
during review of #361 while avoiding the earlier `custom_lsp.yml` model.

## Safety / trust model

Installed adapter packages execute trusted Python code, like other packages in
Serena's environment.

A project configuration can reference only an already registered adapter ID.

It cannot specify:

- import paths
- arbitrary Python modules
- project-local plugins

This PR introduces no project-configured shell command.

## Failure handling

Initial discovery is serialized.

Entry-point metadata failures remain retryable.

If one adapter fails while loading or registering, its registry changes are
rolled back and discovery continues with the remaining adapters.

Unknown IDs continue to fail through normal project-configuration validation.

## Backward compatibility

Existing built-in IDs and project configurations remain valid.

Built-in language auto-detection is unchanged.

When no matching adapter packages are installed, discovery leaves existing
language-server behavior unchanged.

## Dashboard

Registered external adapters participate in the existing available,
add, and remove language-server workflow.

## External proof of concept

[...]

PR #1988, opened 2026-09-04
SolidLSP: Add language server registry, supporting externally provided implementations

opcode81:
Continuation of #1904


opcode81:
@vitalyruhl do you want to test this?

tagawa0525:
Thanks for this registry. I would like to use it for a different kind of external implementation and want to check that the shape fits before writing it.

lsp-det (https://github.com/tagawa0525/lsp-det) is a transparent proxy that sits between a client and a language server and adds one thing: a machine-readable server state (`experimental/serverState` request and `experimental/serverStateChanged` notification, with `health`, `readiness`, and what the server's index covers). It holds cross-file requests such as `references` until the server has finished indexing, and turns a crashed server into an explicit error instead of an empty result. Serena already works through it with only `ls_specific_settings.<language>.ls_base_cmd` (measured with pyright and typescript-language-server, including the tsserver-crash case where the direct path returns 0 references and the proxied path raises).

What I would register through this registry is one adapter per language that derives from the built-in one (`PyrightServer`, `TypeScriptLanguageServer`, …), prepends `lsp-det --` to the launch command, declares `experimental.serverState` in the client capabilities, and replaces the adapter's own readiness wait with a read of the server state. Everything else (dependency handling, initialize params, file matching) stays inherited.

Three questions:

1. Is deriving from a built-in adapter and registering the subclass under a new key (e.g. `python-lsp-det`) the intended use, or would you rather see `register(…, allow_override=True)` on the existing key so that `project.yml` does not change?
2. The readiness wait lives inside each adapter's `_start_server` (`PyrightServer` waits for "Found N source files", `TypeScriptLanguageServer` for `$/progress`). Would a hook such as `_wait_for_initial_readiness()` on `SolidLanguageServer`, called from `start_server`, be acceptable so that a subclass can override only that part? Without it the subclass has to copy `_start_server`.
3. Should such an adapter live in this repository (under `src/solidlsp/language_servers/`) or as a separate package that uses the entry point? I am happy to maintain it either way.

opcode81:
Hello @tagawa0525!

> It holds cross-file requests such as references until the server has finished indexing

How does it achieve this? Isn't this highly server-specific?

> Is deriving from a built-in adapter and registering the subclass under a new key (e.g. python-lsp-det) the intended use, or would you rather see register(…, allow_override=True) on the existing key so that project.yml does not change?

Both are possible, depending on whether you want your module to replace language server implementations or not.

> Would a hook such as _wait_for_initial_readiness() on SolidLanguageServer, called from start_server, be acceptable so that a subclass can override only that part? Without it the subclass has to copy _start_server.

Many server wrappers in SolidLSP already await readiness using varying sets of indicators from the server.
It wouldn't really be practical to move this into a separate method, because the event handlers currently use local closures, and readiness signals and other event handler registration cannot really be separated.

> Should such an adapter live in this repository (under src/solidlsp/language_servers/) or as a separate package that uses the entry point? I am happy to maintain it either way.

If your adapter solves a genuine problem for affected language servers, then it could be interesting to have it included directly. However, adding an intermediate proxy process is, of course, an indirection we would normally want to avoid. We would prefer a solution that is implemented directly in SolidLSP. 
Perhaps you can open an issue and provide more details on what exactly it is your solution does and how it could be added to SolidLSP.

tagawa0525:
Thank you for the answers, and for merging the PR.

**How lsp-det holds cross-file requests**

It reads the same indicators your adapters already read; for typescript-language-server, the `$/progress` tokens and the "[tsserver] Exited" log line. The difference is only in what happens with the result. lsp-det keeps it as a state — whether the server is ready, and whether it is functioning — and checks that state before every cross-file request; requests such as `references` wait while the server is not ready and fail immediately while it is broken, and single-file requests such as `hover` are never delayed.

**On the proxy**

I agree that an intermediate proxy process should be avoided, and I do not ask for it to be included in Serena. The proxy is a stopgap. What I want is for each language server to report, in a common form, whether indexing has finished and whether it is functioning. I have proposed this to rust-analyzer (discussion in rust-lang/rust-analyzer#23331, PR rust-lang/rust-analyzer#23362, which adds `ready: bool` to `experimental/serverStatus`; it is not merged). A proposal for the LSP itself is still a draft, and its form is not decided.

**How this could be done in SolidLSP without a new hook**

I withdraw the request for a `_wait_for_initial_readiness()` hook. SolidLSP already has a place on the request path that can do this. On `main` at 403ad0a5, the base class calls `_wait_for_cross_file_references_if_needed()` immediately before sending each definition, implementation and references request, and an adapter can override it. The default implementation checks `_has_waited_for_cross_file_references`, calls `sleep(2)` on the first call only, and sets the flag; the flag is never reset. The typescript-language-server adapter overrides it and, on the first call only, waits until its `_active_progress_tokens` set is empty; on later calls it returns immediately because of the flag. The flag and the `sleep(2)` came in with commit c8827191 (2025-09-18, "fix(swift-lsp): improve CI stability with enhanced delays and retry logic"); at that time the adapter did not read `$/progress`.

The change I propose is that this method consults the adapter's readiness state on every call, not only on the first one. The event handlers can stay as local closures; the method only needs to read the state they already maintain.

**What I measured**

I reproduced oraios/serena#1937 (TypeScript: `find_referencing_symbols` returns partial results from the second call on). The setup is a generated pnpm-style monorepo with one shared package, 12 packages that import it (each with its own tsconfig, so each is a separate tsserver project), and one app that imports all 12; the shared package has no tsconfig of its own, the root has no solution tsconfig, and a file of the app is kept open. The complete answer for a symbol in the shared package is 13 files: the app file and one file in each of the 12 packages. A `find_referencing_symbols` call issued within 100 ms of the previous one returns 1 of those 13 files, 30 times out of 30. The result is the same with typescript-language-server 5.3.0 + TypeScript 5.9.3 and with the reporter's 5.1.3 + TypeScript 6.0.3.

I then added 8 lines to the typescript-language-server adapter: in `_wait_for_cross_file_references_if_needed()`, call `wait_for_indexing()` whenever `_active_progress_tokens` is not empty, even after the flag is set. With this change, 35 of 36 calls return the complete 13 files. The branch is https://github.com/tagawa0525/serena/tree/request-path-consults-state. oraios/serena#1978 already proposes the same change, so I will not open a PR for it.

The cost is this. A call on which the wait is triggered takes about 0.4 s, which is the time tsserver needs to reload the 12 package projects; `main` answers the same call in 10 ms with the wrong result. Calls during which no reload is running take 10–20 ms, the same as `main`. The one miss in 36 has a known cause: tsserver's first `$/progress` notification arrives a few ms after `didOpen`, and the check sometimes runs before it.

My open PR oraios/serena#2007 adds a check at the top of the same method for a different case: it detects that tsserver has exited, so that a `references` request after the crash does not return `[]` as a success.

One limit: the request-path approach does not cover Metals (the "second call" version of oraios/serena#1858). A `references` request immediately after creating and opening a new file does not include the new reference; one second later it does. At that moment the server has not started working yet (its first `$/progress` arrives 135 ms after `didOpen`), so there is nothing in the adapter's state to consult. Closing that window requires either a timed wait or the server itself holding the request.

**What I can provide**

- The fixture generator and the probe used for the reproduction above.
- Tests, once it is decided what the base class holds and what each adapter supplies.
- A tracking issue, if you would like one.

PR #1990, opened 2026-09-05
fix: disable automatic type acquisition in TypeScript backends

### LINKED ISSUE #1989: TypeScript/VTS: automatic type acquisition remains enabled after #1639
[PR #1639](https://github.com/oraios/serena/pull/1639) intended to disable automatic type acquisition, but both backends place `disableAutomaticTypingAcquisition` under `initializationOptions.preferences`. Neither backend reads its ATA setting from that location, so the change does not reliably prevent background `@types` downloads and their associated startup delays.

The correct settings differ:

| Backend | Required setting |
|---|---|
| TypeScript Language Server | `initializationOptions.disableAutomaticTypingAcquisition = true` |
| VTS | `initializationOptions.typescript.disableAutomaticTypeAcquisition = true` |

TypeScript Language Server reads the option directly from `initializationOptions`; VTS defines it under `typescript`. The spelling also differs: **Typing** versus **Type**. See the [TypeScript implementation](https://github.com/typescript-language-server/typescript-language-server/blob/v5.3.0/src/lsp-server.ts) and [VTS schema](https://github.com/yioneko/vtsls/blob/main/packages/service/configuration.schema.json).

VTS additionally answers `workspace/configuration` requests using only user-provided initialization options. Correcting its initialization payload alone would leave those responses without the default ATA setting. Initialization, configuration responses, and configuration notifications should consistently supply the effective settings while preserving unrelated user configuration.

Proposed fix and acceptance criteria:

- Correct the setting’s location and spelling for each backend.
- Ensure VTS consistently supplies the effective configuration, including when no custom initialization options are provided.
- Preserve other user settings, including nested TypeScript settings.
- Add regression coverage using actual server startup: verify the semantic `tsserver` receives `--disableAutomaticTypingAcquisition` and no typings-installer child starts when opening a project that would otherwise trigger ATA.
- Run the existing symbol and reference tests for both backends.

Keep `useSyntaxServer` unchanged. Disabling automatic type acquisition is separate from removing the syntax worker.

This report is based on source inspection; runtime reproduction and regression validation remain part of the follow-up.

GuddXzy:
## Summary

The automatic type acquisition defaults were sent under settings neither TypeScript backend consumes. VTS also returned an empty response to the root configuration request it makes during startup, so its semantic tsserver could still launch the typings installer.

Fixes #1989.

- Put the TypeScript Language Server option at the correct initialization level.
- Use the shared initialization builder for VTS and share its effective settings across initialization, configuration requests, and change notifications.
- Preserve the legacy `initialization_options` alias when `initializationOptions` is absent, without mutating caller-owned settings during assembly.
- Follow the shared builder's per-top-level-key overrides: supplying a `typescript` block replaces the entire default block, including the ATA default. This behavior is documented in the changelog and VTS configuration docstring.

## Testing

The original six regressions failed against the original main baseline. Following review, three configuration cases failed against the previous PR implementation before simplifying VTS. The revised focused file now has eight passing cases, including actual startup of both backends: symbol lookup succeeds, the semantic tsserver receives the disabling flag, and no typings-installer process is observed. The syntax worker is excluded from the semantic-worker assertion.

- Affected TypeScript and timeout-policy suites: **49 passed** on the final run. An earlier run had 48 passed and one failure in the unchanged cross-package reference test; its isolated module passed (3 tests), then the full affected suite passed without changing that code or test.
- `poe format`, `poe type-check`, `poe lint`, and `git diff --check`: passed.
- Tested on Windows with Python 3.13.12 and Node 24.16.0, using the existing default backend versions. The all-language suite was not run locally.

## Checklist

- [x] Added a concise entry under Language Servers in `CHANGELOG.md`.

### DIFF
--- a/src/solidlsp/language_servers/vts_language_server.py
+++ b/src/solidlsp/language_servers/vts_language_server.py
@@ -8,11 +8,14 @@ import logging
 import os
 import shutil
 import threading
+from copy import deepcopy
+from typing import cast
 
 from overrides import override
 
 from solidlsp.ls import SolidLanguageServer
 from solidlsp.ls_config import LanguageServerConfig, LanguageServerId
+from solidlsp.lsp_protocol_handler.lsp_types import InitializeParams
 from solidlsp.lsp_protocol_handler.server import ProcessLaunchInfo
 from solidlsp.settings import SolidLSPSettings
 
@@ -34,8 +37,8 @@ class VtsLanguageServer(SolidLanguageServer):
     Supported entries in ``ls_specific_settings["typescript_vts"]``:
         - ``vtsls_version``: version of ``@vtsls/language-server`` to install (default: ``"0.2.9"``).
         - ``npm_registry``: custom npm registry for the managed install.
-        - ``initialization_options``: optional dict forwarded verbatim as LSP
-          ``initializationOptions``. Useful for Yarn PnP projects, e.g.::
+        - ``initialization_options``: optional LSP configuration. Automatic type acquisition
+          is disabled by default; explicit settings take precedence. Useful for Yarn PnP projects, e.g.::
 
               initialization_options:
                 typescript:
@@ -128,16 +131,11 @@ class VtsLanguageServer(SolidLanguageServer):
         """
         Returns the initialize params for the VTS Language Server.
 
-        If ``initialization_options`` is set in ``ls_specific_settings["typescript_vts"]``,
-        it is forwarded verbatim as LSP ``initializationOptions``.
+        User-provided options are copied so initialization overrides do not mutate the caller's settings.
         """
         initialize_params: dict = {
             "locale": "en",
-            "initializationOptions": {
-                "preferences": {
-                    "disableAutomaticTypingAcquisition": True,
-                },
-            },
+            "initializationOptions": deepcopy(self._initialization_options),
             "capabilities": {
                 "textDocument": {
                     "synchronization": {"didSave": True, "dynamicRegistration": True},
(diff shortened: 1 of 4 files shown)


GuddXzy:
The six new regression tests pass on Ubuntu, Windows, and macOS. The remaining failures are in unchanged C++ ccls tests and the JetBrains coordinator sleep-count test. PR #1914 reports the same JetBrains failure on other platforms, which suggests an existing test instability rather than a regression from this change. Could a maintainer rerun the failed jobs when convenient? I don’t have permission to rerun them.

src/solidlsp/language_servers/vts_language_server.py
```diff
@@ -128,16 +131,11 @@ def _create_base_initialize_params(self) -> dict:
...
+        User-provided options are copied so initialization overrides do not mutate the caller's settings.
         """
         initialize_params: dict = {
             "locale": "en",
-            "initializationOptions": {
-                "preferences": {
-                    "disableAutomaticTypingAcquisition": True,
-                },
-            },
+            "initializationOptions": deepcopy(self._initialization_options),
```
opcode81, line 138: This has brought to my attention an inconsistency which we could reconcile with this PR: The VTS language server internally defines its own way of handling user-provided initialisation options, even though a mechanism for this is now provided centrally. Notice that `DefaultInitializeParamsBuilder`, which VTS uses, already handles this. User-provided settings under key `initializationOptions` will override predefined settings on a per-key basis. This should suffice for VTS, and we should drop the additional mechanism.

This will make the changes here much simpler, as we can get rid of `self._initialization_options` and the override for `_create_initialize_params`. We can simply set the defaults here directly and `DefaultInitializeParamsBuilder` can take care of the rest.

However, the existing VTS-specific configuration option uses key `initialization_options` rather than `initializationOptions`, so to improve backward compatibility, we could set account for this: Set the new key if the old key is found in the CustomLSSettings object's `settings` dictionary.

opcode81 (changes requested):
Thanks for your PR. Please see my comment; we can simplify this and remove duplicate handling of initialization options in VTS.

src/solidlsp/language_servers/vts_language_server.py
GuddXzy: I can simplify VTS to use the shared builder and retain compatibility with the legacy key. One detail: the builder uses a shallow merge, so setting only `typescript.tsdk` replaces the entire `typescript` block and drops `disableAutomaticTypeAcquisition=True`. Should we preserve that default unless the user explicitly overrides it, or follow the builder’s whole-section replacement behavior?

opcode81: The contract for VTS was previously even less flexible: It was all or nothing; user-provided object replaces the entire `initializationOptions` object.
```
        if self._initialization_options:
            log.info("Forwarding user-provided initializationOptions to vtsls: %s", self._initialization_options)
            initialize_params["initializationOptions"] = self._initialization_options
```
So the implementation in the builder is actually slightly more flexible (per-subkey override). We just have to document the change in behaviour, but I'd regard it as an improvement.

opcode81: Ah, sorry, your question was whether we should preserve the `disableAutomaticTypeAcquisition` even if the user overrides to subkey it is in. I think we probably shouldn't. Let's keep the semantics simple.

opcode81:
Thanks for your contribution!

PR #1993, opened 2026-09-05
fix: send languageId typescriptreact for .tsx files in vtsls

chad-loder:
Same fix as #1436, for the vtsls backend.

`vts_language_server.py` doesn't override `_get_language_id_for_file`, so `.tsx`
and `.jsx` are opened with the base `typescript` id. JSX then parses as TS and
symbol ranges silently truncate at the first multi-line JSX expression.

Implementation is identical to the one #1436 added to
`typescript_language_server.py`.

### DIFF
--- a/src/solidlsp/language_servers/vts_language_server.py
+++ b/src/solidlsp/language_servers/vts_language_server.py
@@ -70,6 +70,15 @@ class VtsLanguageServer(SolidLanguageServer):
             "build",
         ]
 
+    def _get_language_id_for_file(self, relative_file_path: str) -> str:
+        # JSX is parsed as TS without this, which silently truncates symbol
+        # ranges at the first multi-line JSX expression.
+        if relative_file_path.endswith(".tsx"):
+            return "typescriptreact"
+        if relative_file_path.endswith(".jsx"):
+            return "javascriptreact"
+        return self.language_id
+
     @classmethod
     def _setup_runtime_dependencies(cls, config: LanguageServerConfig, solidlsp_settings: SolidLSPSettings) -> str:
         """


src/solidlsp/language_servers/vts_language_server.py
```diff
@@ -70,6 +70,15 @@ def is_ignored_dirname(self, dirname: str) -> bool:
             "build",
         ]
 
+    def _get_language_id_for_file(self, relative_file_path: str) -> str:
```
AmirF194, line 73: I checked this against the identical fix already merged for TypeScriptLanguageServer (commit 37c837ff, PR #1436), which shipped a regression test alongside the code change at test/solidlsp/typescript/test_typescript_basic.py:64-65, `test_tsx_symbol_range_not_truncated_by_jsx`, parametrized on `LanguageServerId.TYPESCRIPT`.

VtsLanguageServer and TypeScriptLanguageServer share the same TypeScript file matcher in ls_config.py's `get_source_fn_matcher` (the `TYPESCRIPT | TYPESCRIPT_VTS` case) and the same `typescript/test_repo` fixture, so the existing `jsx_component.tsx` fixture already covers this backend too. Extending that test's parametrize list to `[LanguageServerId.TYPESCRIPT, LanguageServerId.TYPESCRIPT_VTS]` would exercise this exact fix, the same way the original fix for the sibling backend was tested.

I have not run this myself; `LanguageServerId.TYPESCRIPT_VTS` is not referenced anywhere in test/ today, so I cannot confirm vtsls resolves cleanly in CI. It is also marked experimental in ls_config.py, so the missing coverage may be intentional rather than an oversight.

opcode81:
Changelog entry is missing; otherwise looks good!

opcode81:
@chad-loder ping

PR #1995, opened 2026-09-06
fix(project): prune the .gitignore search with ignored_paths

### LINKED ISSUE #1991: ignored_paths does not prune the .gitignore discovery walk at startup (35 s on a large tree)
Preconditions:

Issue details:

## Summary

`ignored_paths` does not prune the `.gitignore` discovery walk that runs at project activation. On a large tree that walk alone took 35 seconds on every start, even though the directories it spent that time in are listed in `ignored_paths`. The only way to prune the walk today is a `.gitignore` file at the project root.

## What happens

At activation, `Project._gather_ignorespec` builds a `GitignoreParser` with only the project root:

https://github.com/oraios/serena/blob/13ac8c5b1d51873bd148aea440dcb22f85d3a439/src/serena/project.py#L93

`GitignoreParser._iter_gitignore_files` then walks every directory under the root. It skips a directory only when it matches a `.gitignore` file it has already loaded. It never sees the global or project `ignored_paths`:

https://github.com/oraios/serena/blob/13ac8c5b1d51873bd148aea440dcb22f85d3a439/src/serena/util/file_system.py#L237-L273

So `ignored_paths` correctly excludes files from indexing and from `gather_source_files`, but the walk that discovers `.gitignore` files still descends into every ignored directory, including `.git`, `node_modules` and build output.

## Expected

Directories matched by `ignored_paths` are skipped by the `.gitignore` discovery walk, in the same way directories matched by an already loaded `.gitignore` are skipped. One way to do that is to pass the combined `ignored_paths` patterns into `GitignoreParser` as an initial spec.

## Reproduction

Any large tree with a directory you do not want scanned. Mine is a folder holding all the cert-manager repositories, with several worktree checkouts each (6.5 GB, 34,639 Go files, 36 checkouts of https://github.com/cert-manager/cert-manager).

1. Set `ignored_paths` in `.serena/project.yml` to exclude the extra checkouts, `.git/`, `_bin/` and `node_modules/`.
2. Start the MCP server, or run `serena project index`.
3. Read the log line `Loading of .gitignore files completed in N seconds`.

<details>
<summary>Measurements</summary>

Serena MCP server log, before adding a root `.gitignore`:

```
serena.util.file_system:stop:336 - Loading of .gitignore files completed in 35.845 seconds
serena.project:stop:336 - Gathering ignore spec for project cert-manager completed in 36.097 seconds
```

After adding a root `.gitignore` with the same patterns (the parent folder is not a git repository, the file exists only for Serena):

```
serena.util.file_system:stop:336 - Loading of .gitignore files completed in 1.270 seconds
serena.project:stop:336 - Gathering ignore spec for project cert-manager completed in 1.276 seconds
```

A direct check of the walker with the root `.gitignore` in place:

```
$ uvx --from serena-agent python -c '
import time
from serena.util.file_system import GitignoreParser
t = time.monotonic()
p = GitignoreParser("/home/richard/projects/github.com/cert-manager")
print(f"{time.monotonic()-t:.1f}s, {len(p.ignore_specs)} .gitignore files")'
1.0s, 21 .gitignore files
```

Without it the same walk visits 155 `.gitignore` files and every `.git` and `node_modules` directory.

</details>

## Workaround

Put the patterns in a `.gitignore` at the project root. `_iter_gitignore_files` loads that file first and then prunes on it. Note that a pattern with an inner slash is root-anchored, so nested directories need `**/`, for example `**/.claude/worktrees/`.

## Setup

- Serena 1.7.0 (`uvx --from serena-agent`), LSP backend, Go (`gopls`)
- Claude Code 2.1.261, started through `headroom wrap claude`
- Fedora Linux 44, Python 3.14
- Project config: default `project.yml` generated by Serena, `ignore_all_files_in_gitignore: true`

Related: #1624 hit the same walker from a different direction (an unreadable directory listed in `ignored_paths` still crashed the walk).

[Claude Fable 5.1]

ebarkhordar:
`Project._gather_ignorespec` builds the combined `ignored_paths` list and then constructs `GitignoreParser(self.project_root)` on the next line, so those patterns never reach the discovery walk. `_iter_gitignore_files` prunes only on specs it has already loaded from `.gitignore` files, which is why a root `.gitignore` works as a workaround and `ignored_paths` does not.

`GitignoreParser` now takes `additional_ignore_patterns`, and `_gather_ignorespec` passes the list it has just built. Matching goes through `match_path`, which `Project._is_ignored_relative_path` already applies to these same patterns, so a directory the walk skips is one the project independently considers ignored.

The patterns prune the search and nothing else. `get_ignore_specs` returns specs read from files and each carries its `file_path`, so there is no honest way to report a configured pattern as one of them; `should_ignore` is unchanged as well.

One consequence worth naming: a `.gitignore` nested inside an ignored directory is no longer read. Its patterns could only ever match paths under a directory that is already ignored, and `gitignore(5)` is explicit that patterns on files inside an excluded directory have no effect. The existing pruning has always behaved this way for the patterns it discovers itself.

Verified in `python:3.11-slim`, the version the test matrix pins:

- On a synthetic tree with `vendor/` in `ignored_paths`, the walk entered 1401 directories under `vendor/` and loaded a `.gitignore` from inside one of them. With the change it enters none, and 0.130s becomes 0.001s.
- Three new tests. Two of them fail on `main` and pass here. The third is green either way by design, since it pins that the default walk stays unpruned, so I checked it can fail at all: making the prune unconditional turns it red.
- `test/serena/util/`, `test/serena/config/` and `test/solidlsp/test_is_ignored_path_missing.py` give 171 passed against 168 on `main`, with the same three `test_exception.py` tkinter failures on both sides.
- `poe lint`, `poe type-check` and `codespell` are clean.

I have not measured this against a large real repository, so the timings above come from the synthetic tree rather than from the reporter's 6.5 GB checkout.

Fixes #1991

## Checklist


ebarkhordar:
One design question worth settling before this goes further. `additional_ignore_patterns` prunes the `.gitignore` discovery walk and nothing else: it does not feed `should_ignore` and it is not returned by `get_ignore_specs`. That is deliberate, since `_gather_ignorespec` already holds those patterns and handing them back would double them in the combined spec, but it does make the parameter narrower than its name reads.

The other seam is a `should_descend` callable that `_gather_ignorespec` passes in, which keeps `GitignoreParser` from knowing anything about project config. Tell me which one you want and I will switch it.

Issue #1826, opened 2026-08-08
project health-check reports PASS and exits 0 while its own log records a tool failure and a degenerate probe


CoderNoveau:
### Summary

`serena project health-check` prints `✅ Health check passed - All tools working correctly` and **exits 0** while its own log, in the same run, records a tool failure and a result it explicitly flags as suspicious.

Version: **Serena 1.6.1**. Project: a ~5,000-file TypeScript monorepo. Reproduced twice — once with no LSP cache, once immediately after a successful `serena project index .` (4,975 files, 83s) — with **identical output both times**.

### What it prints

```
✅ Health check passed - All tools working correctly     (exit 0)
```

### What the same run logs

```
INFO    serena.cli:health_check:998  - FindSymbolTool found 1 matches for symbol default
INFO    serena.cli:health_check:1005 - FindReferencingSymbolsTool found 0 references for symbol default
WARNING serena.cli:health_check:1020 - SearchForPatternTool failed for pattern default: Expecting value: line 1 column 1 (char 0)
WARNING serena.cli:health_check:1030 - Both FindReferencingSymbolsTool and SearchForPatternTool found no matches - this might indicate an issue
INFO    serena.cli:health_check:1032 - Health check completed successfully
```

### Two issues, which compound

**1. The verdict contradicts the tool's own observations.** `SearchForPatternTool` raised a JSON decode error, and the check emitted its own warning that the result "might indicate an issue" — then reported success and exited 0. Any caller that branches on the exit code (which is the correct thing to do in automation) receives the opposite of what the tool observed.

**2. The probe symbol is degenerate.** It probes the symbol `default`. In a TypeScript codebase that resolves to `export default`, which legitimately has **no named references** — so `FindReferencingSymbols → 0` is an expected result for that symbol, not a signal. The check therefore cannot distinguish a broken language server from a correct answer about a badly chosen probe. This is why indexing changed nothing in the output: the check was never sensitive to the thing it appears to test.

### Why it matters

Serena's documented failure mode is a *confident empty* rather than an error — a wrong project or an unhealthy LSP returns `[]`, not an exception. A health-check is exactly the tool that should discriminate that case. As written it returns a **confident green** instead, so a newly provisioned environment that runs the documented check and reads the exit code will conclude Serena is working when its reference tools are not yet usable.

### Suggested fix

- Make the exit code a function of the run's own observations: non-zero when any tool raises, or when a control returns empty.
[...]

PR #1998, opened 2026-09-06
Fix: health-check reported success after reference search failed

gladkia:
## The problem

`project health-check` runs three tools, but only two of them can fail the check. An exception from `FindReferencingSymbolsTool` is caught and logged as a warning, while the verdict tests `find_symbol_data` alone:

```python
                except Exception as e:
                    log.warning("FindReferencingSymbolsTool failed for symbol %s: %s", symbol_name, str(e))

                # Verify tools worked as expected
                if not find_symbol_data:
                    raise ProjectCommands._HealthCheckFailure("FindSymbolTool returned no results")
```

So on a project whose reference search has just blown up, the command prints `✅ Health check passed - All tools working correctly` and exits 0. Captured from a run against the unpatched code:

```
WARNING serena.cli:health_check:1023 - FindReferencingSymbolsTool failed for symbol hello: language server went away
INFO    serena.cli:health_check:1029 - Health check completed successfully
```

A false green here is worse than no check: the command exists to answer whether the project's tooling works, and it is used that way — `.github/workflows/python-versions.yml` gates each interpreter on it, and callers script on the exit code.

This also contradicts the principle stated a few lines below, in the outer handler:

> expected failures and unexpected exceptions are reported alike: both mean the project's tooling is not functional, which is the single thing this command is asked to determine

## The change

A reference-search **failure** now fails the check.

The number of references is deliberately still not asserted — a symbol nobody calls is a legitimate result and stays a pass. Only an exception from the tool is treated as "the language server is not functional".

## Tests

Three cases in `test/serena/test_cli_project_commands.py`, with the language-server dependencies replaced by doubles, so they need no running server and add ~0.1 s:

| case | expected |
|---|---|
| all tools answer | exit 0, "Health check passed" |
| symbol has no references | exit 0, "Health check passed" |
| reference search raises | exit 1, "Health check failed", no "All tools working correctly" |

The third fails on the unpatched code (`1 failed, 2 passed`), which is what makes the first two meaningful.

## Checks run

- `poe lint` — clean, `poe type-check` — clean
- `pytest test/serena/test_cli_project_commands.py` — 28 passed
- `serena project health-check test/resources/repos/python/test_repo` — passes with the change (9 references found for `UserService`), i.e. the workflow this fix makes stricter is green on your own fixture

## One thing worth naming

This makes `python-versions.yml` genuinely stricter. If that job starts failing on some interpreter after this merges, it is the check reporting a condition it previously hid, not a new defect introduced here.

---

### DIFF
--- a/src/serena/cli.py
+++ b/src/serena/cli.py
@@ -1020,7 +1020,12 @@ class ProjectCommands(AutoRegisteringGroup):
                         find_refs_data = json.loads(find_refs_result)
                         log.info("FindReferencingSymbolsTool found %d references for symbol %s", len(find_refs_data), symbol_name)
                 except Exception as e:
-                    log.warning("FindReferencingSymbolsTool failed for symbol %s: %s", symbol_name, str(e))
+                    # A symbol with no references at all is a legitimate result, so the number of
+                    # references is not asserted - but a *failure* of the reference search means the
+                    # language server is not functional, which is the single thing this command is
+                    # asked to determine. Logging it as a warning let the command print
+                    # "All tools working correctly" and exit 0 after the search had already failed.
+                    raise ProjectCommands._HealthCheckFailure(f"FindReferencingSymbolsTool failed for symbol {symbol_name}: {e}") from e
 
                 # Verify tools worked as expected
                 if not find_symbol_data:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,12 @@ Status of the `main` branch. Changes prior to the next official version change w
     which could leave grandchildren as zombies; cleanup now waits for the discovered descendants (#1464)
   - Fix: `read_only` restriction in project definition was not applied to base tool set when in single-project context (#1938)
 
+* CLI:
+  - Fix: `project health-check` reported `Health check passed - All tools working correctly` and
+    exited 0 even when `FindReferencingSymbolsTool` had raised, because that failure was logged as
+    a warning while the verdict checked `FindSymbolTool` only. A reference-search failure now fails
+    the check; a symbol with no references is still a pass
+
 * Memories:
   - Fix: `save_memory`/`edit_memory` wrote directly to the memory file with `open(path, "w")`, which
     truncates it before the new content is written; a crash, OOM kill, or full disk partway through
--- a/test/serena/test_cli_project_commands.py
+++ b/test/serena/test_cli_project_commands.py
@@ -1,10 +1,15 @@
 """Tests for CLI project commands (create, index)."""
 
+import json
 import os
 import shutil
 import tempfile
 import time
+from collections.abc import Iterator
+from contextlib import contextmanager
 from pathlib import Path
+from types import SimpleNamespace
+from typing import Any
 
 import pytest
 from click import Command, Option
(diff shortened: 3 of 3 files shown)


opcode81:
The exception is intentionally caught; see my comment here: https://github.com/oraios/serena/issues/1826#issuecomment-5227875868

src/serena/cli.py
```diff
@@ -1020,7 +1020,12 @@ def health_check(project: str) -> None:
                         find_refs_data = json.loads(find_refs_result)
                         log.info("FindReferencingSymbolsTool found %d references for symbol %s", len(find_refs_data), symbol_name)
                 except Exception as e:
-                    log.warning("FindReferencingSymbolsTool failed for symbol %s: %s", symbol_name, str(e))
+                    # A symbol with no references at all is a legitimate result, so the number of
+                    # references is not asserted - but a *failure* of the reference search means the
+                    # language server is not functional, which is the single thing this command is
+                    # asked to determine. Logging it as a warning let the command print
+                    # "All tools working correctly" and exit 0 after the search had already failed.
+                    raise ProjectCommands._HealthCheckFailure(f"FindReferencingSymbolsTool failed for symbol {symbol_name}: {e}") from e
```
ebarkhordar, line 1030: I ran this against a6c7600 in a clean `python:3.11-slim` container, using the doubles from your own new test file.

The fix does what it says: a reference-search failure now fails the check. But the `if not find_symbol_data` verdict at `src/serena/cli.py:1031` sits below this raise, and both tools resolve the same name path through `symbol_retriever.find(...)` with `substring_matching=False` and the same `within_relative_path`. So when FindSymbolTool returns nothing, `find_unique` raises `ValueError("No symbol matching 'hello' found")` (`src/serena/symbol.py:785`) at line 1028 first, and the check names the wrong tool.

Same doubles on both commits. Only the third row differs:

```
FindSymbolTool EMPTY, refs raise
  9f9db76 (main)     exit=1  Health check failed: FindSymbolTool returned no results
  a6c7600 (this PR)  exit=1  Health check failed: FindReferencingSymbolsTool failed for symbol hello: No symbol matching 'hello' found
```

The row above it, FindSymbolTool OK and refs raising, is the case you are fixing: exit=0 on main, exit=1 here.

Moving the two `if not find_symbol_data` lines up to just after `find_symbol_data = json.loads(find_symbol_result)`, above Test 3, keeps both verdicts accurate.

It is reachable because `symbol_name` comes from the overview tool, so an empty FindSymbolTool result means those two disagree, which is the degraded state this command exists to catch. I did not run it against a real language server.

gladkia:
Thanks for the thorough reviews!

### 1. Tool ordering and verdict accuracy (@ebarkhordar)
Great catch and reproduction — you are completely right. When `FindSymbolTool` returned no matches, `FindReferencingSymbolsTool` proceeded to call `symbol_retriever.find_unique(...)`, which raised `ValueError("No symbol matching '<name>' found")`. That caught exception caused the check to fail naming `FindReferencingSymbolsTool`, shadowing the earlier empty `FindSymbolTool` result.

I have pushed an update that:
- Moves the `if not find_symbol_data:` check immediately after Test 2 (before Test 3 is executed).
- Adds a test double case in `test_cli_project_commands.py` (`test_health_check_fails_with_find_symbol_error_when_symbol_search_empty`) verifying that an empty symbol search fails immediately attributing the failure to `FindSymbolTool returned no results`, without calling or blaming `FindReferencingSymbolsTool`.

---

### 2. Rationale on catching reference search exceptions (@opcode81)
Regarding the reference to #1826: in that issue, the discussion centered around `SearchForPatternTool` with a guessed keyword (`"default"`) and whether 0 matches should fail the check. 

Here, two key differences are worth highlighting:
1. **The probe symbol is not guessed blindly**: `symbol_name` is taken directly from the output of `GetSymbolsOverviewTool` for `target_file` (preferring functions, classes, and methods). It is a known, existing symbol in the file AST.
2. **0 references is still considered a PASS**: If `FindReferencingSymbolsTool` returns 0 references (`len(find_refs_data) == 0`), the check passes with exit code 0. An unreferenced function or symbol is completely legitimate and is explicitly kept as a pass.
3. **Only an unhandled exception / crash fails the check**: If `FindReferencingSymbolsTool` actually raises (e.g. language server crashes, connection dropped, or raises an internal error), reference search is fundamentally broken for the project.

Reporting `✅ Health check passed - All tools working correctly` and exiting 0 when the reference tool just threw an unhandled exception leaves automation and users believing semantic reference navigation works when it doesn't. Failing on exception aligns with the design principle stated right below in lines 1036-1037: *"both mean the project's tooling is not functional, which is the single thing this command is asked to determine"*.

ebarkhordar:
Thanks, that is the ordering I was after. I read 07507a2a: the empty-result raise now sits before Test 3, so an empty symbol search is attributed to FindSymbolTool instead of the reference search, and the new test pins it. I read the diff rather than running the suite.

opcode81:
Making the test fail when the reference search raises an exception is fine. But please drop the tests. We do not need tests to cover such a simple logic change.

gladkia:
Dropped the unit tests as requested. The PR now only contains the logic fix in `src/serena/cli.py` and the changelog entry.

opcode81:
Thanks for your contribution.

PR #2009, opened 2026-09-10
fix(solidlsp): bump the document version on a full-text didChange after re-open

### LINKED ISSUE #2005: Full-document didChange on re-open reuses the document's version number

When a file is re-opened while it is already open in the language server and its mtime changed on disk, `LSPFileBuffer._open_in_ls` sends a full-text `textDocument/didChange` with `version: self.version`, unchanged (`src/solidlsp/ls.py:141-158`). The range-based `didChange` sent for edits increments the version (`ls.py:1378, 1420`).

LSP's `VersionedTextDocumentIdentifier` says the version "will increase after each change". The three servers I checked (pyright, typescript-language-server, rust-analyzer) all apply the new text regardless of the number, so the analysis itself is right. What breaks is everything keyed on the version:

- `publishDiagnostics` carries the document version the diagnostics were computed for (pyright: `languageServerBase.ts` `sendDiagnostics`; rust-analyzer: `main_loop.rs`, taken from `mem_docs`). After the reused version, diagnostics for the old text and for the new text carry the same number, so a client cannot tell which text they describe, and a client that drops diagnostics for an outdated version keeps the old ones.
- rust-analyzer rejects `codeAction/resolve` whose `data.version` differs from the current version ("stale code action", `handlers/request.rs`) and ignores outdated inlay-hint resolve data. With the number reused, resolve data computed from the old text passes those checks.

Serena's own diagnostics wait does not read the version (it counts publishes), so today this is a conformance problem rather than a visible bug in Serena; it becomes visible for any client that does read it.

Suggestion: `self.version += 1` before building the notification, as the edit path does.

Preconditions:

Issue details:

L4XB:
## Summary

Fixes #2005.

When a file that is already open in the language server is re-opened after its mtime changed on disk, `LSPFileBuffer._open_in_ls` sends a full-text `textDocument/didChange` with `version: self.version` unchanged, while the range-based `didChange` sent for edits (`insert_text_at_position`, `delete_text_between_positions`) increments it. LSP's `VersionedTextDocumentIdentifier` says the version "will increase after each change"; reusing the number makes `publishDiagnostics` for the old and the new text indistinguishable to a client that keys on it, and lets rust-analyzer's version-checked `codeAction/resolve` / inlay-hint data computed from the old text pass as current.

## Change

`src/solidlsp/ls.py`: `self.version += 1` before the full-document `didChange`, mirroring the edit path. `didOpen` is unchanged.

## Tests

`test/solidlsp/test_file_buffer_reopen_version.py`, with a `MagicMock` language server like `test_rename_didopen.py`:

- `didOpen` carries version 0; after the file changes on disk (content + mtime) `ensure_open_in_ls()` sends one full-text `didChange` with version 1 and the new text, a second change sends version 2 (fails on `main`: version stays 0);
- re-opening without a disk change sends nothing and leaves the version alone.

`pytest test/solidlsp/test_file_buffer_reopen_version.py test/solidlsp/test_rename_didopen.py` → 3 passed; `ruff check` / `ruff format --check` clean.


L4XB:
I have read the CLA Document and I hereby sign the CLA

PR #2010, opened 2026-09-10
fix: report symbol information skipped after budget exhaustion

### LINKED ISSUE #2006: find_symbol with include_info silently drops info when symbol_info_budget is exceeded

`request_info_for_symbols` stops issuing hover requests once `symbol_info_budget` (default 10 s, `serena_config.py:944`; the docstring in `symbol.py` still says 5 s) is spent and sets `info = None` for the remaining symbols (`src/serena/symbol.py:687-694`). The only trace is a single `log.debug` and the perf summary, also at debug level (`symbol.py:715-721`). The tool result is the same as for a symbol that genuinely has no docstring or signature, so the agent cannot tell a budget cut from an absence.

Suggestion: report the cut in the tool result (a count of symbols whose info was skipped, or a note on each affected symbol), and log it at INFO or WARNING.

Preconditions:

Issue details:

0206pdh:
When `find_symbol(include_info=True)` exhausted its hover budget, skipped symbols were indistinguishable from symbols whose language server returned no information. Return an explicit budget-exhaustion note in each skipped symbol's existing `info` field and log the skipped count at INFO level. The shared batch retrieval also exposes the note to `find_implementing_symbols`.

The JSON structure and budget enforcement are preserved. Genuine missing hover information still returns `None`. Correct the batch method's default-budget documentation from 5 to 10 seconds and record the fix under Unreleased in CHANGELOG.

## Validation

- Before the fix: deterministic tool-output regression test had 2 passed, 1 failed; the failure demonstrated identical output for unqueried and undocumented symbols.
- After the fix: 70 related unmarked tests passed, including all 3 regression cases and assertions on the note and INFO log.
- 24 Python-marked symbol tests passed across pyright, ty, pyrefly, and basedpyright. These start real language servers; budget timing itself is mocked.
- `python -m uv run poe format` passed.
- `python -m uv run poe type-check` passed.
- `git diff --check` passed.

Commands for the related tests:

```powershell
python -m uv run poe test test/serena/test_symbol.py test/serena/test_symbol_info_budget_reporting.py -m python --tb=short
python -m uv run poe test test/serena/test_symbol.py test/serena/test_symbol_info_budget_reporting.py -m 'not python' --tb=short
```

Validation was performed on Windows with Python 3.13.2. The full repository test suite and MCP client end-to-end behavior were not tested.

Closes #2006


src/serena/symbol.py
```diff
@@ -690,10 +692,7 @@ def request_info_for_symbol_batch(
                     # symbol_info_budget_seconds=0 disables the budget mechanism (the first inequality)
                     if 0 < symbol_info_budget_seconds <= hover_spent_seconds:
                         skipped_due_to_budget += 1
-                        info = None
-                        # log once when budget exceeded
-                        if skipped_due_to_budget == 1:
-                            log.debug("Skipping further hover operations due to budget exceeded")
+                        info = "[Symbol information omitted because symbol_info_budget was exhausted.]"
```
opcode81: Why do you think this is necessarily an improvement?
LLMs were already informed in the tool description that they would receive partial results if the budget is exceeded. So why is adding so many tokens to the output preferable?

0206pdh: Fair point — moved this to a single count instead of per-symbol prose. `request_info_for_symbol_batch` now returns a `SymbolInfoBatchResult` (`info_by_symbol` + `skipped_due_to_budget`); skipped symbols keep `info=None` as before, and `FindSymbolTool`/`FindImplementationsTool` prepend one line — `Note: symbol_info_budget exhausted; info omitted for N symbol(s).` — to the JSON result, only when the budget was actually hit. Cost is now O(1) per batch regardless of how many symbols got skipped. #2006 asked for either a count or a per-symbol note; this keeps the count option, so the ambiguity it reported is still resolved without the token overhead. Pushed as 4d680c85.

PR #2013, opened 2026-09-10
fix: do not load the anthropic package on import

### LINKED ISSUE #2012: Unconditional anthropic import adds seconds to CLI/MCP startup with CHAR_COUNT
Serena's startup time of 2–3 seconds can allow Codex to complete its initial tool discovery before Serena is fully ready. This means that Serena's tools might not appear in the initial tool list, leading to delays and potential issues. Serena loads the Anthropic Python library even when Anthropic features are not used, adding 2.4–3.6 seconds to the startup time on a Windows machine.

**Proposed Change:** Load the library only when the Anthropic-based token counting feature is selected. This change does not affect Claude support.

### Why It Matters

The problem is not just about a slightly longer wait but about missing tools at the start of a new Codex chat. This can lead to lengthy diagnosis and recovery processes before work can begin. From my experience, unavailable Serena tools often trigger this recovery process at the start of new chats.

During one observed Desktop session, Codex missed adding Serena to the initial tool list because the server was still booting up. The extra import time might contribute to this missed discovery window. Users have reported this recurring recovery burden, and the import overhead has been measured. However, the import has not been conclusively identified as the cause of all such incidents.

### How to Reproduce

In a Python environment with `serena-agent==1.7.0`, run the following command:

```bash
python -c "import sys; import serena.cli; print('anthropic' in sys.modules)"
```

**Current Result:** Prints `True`, indicating that the Anthropic library has already been loaded before any token counter is selected.

**Intended Result:** Prints `False`. The library should only load if the Anthropic token counter is actually used.

No project, language server, API key, or API request is necessary to reproduce this.

### Evidence

With `CHAR_COUNT` (local character-based token counting), a small local change deferred the unnecessary import. In representative before-and-after tests, importing Serena's CLI took **5.13 seconds before and 2.89 seconds after**. Serena's MCP tools continued to function after the change.

**Measurement Details and Environment:**

- Operating System: Windows
- Python Version: 3.13.13
- Serena Version: 1.7.0
- Anthropic Python Library Version: 0.117.0

MCP launches used `uvx` through the Codex CLI, with a context of `codex`, an LSP backend, and a C# project. The `CHAR_COUNT` feature was used. Dashboard and GUI log windows were disabled.

Measurements were taken using `PYTHONPROFILEIMPORTTIME=1` in the Serena process. Import-only profiling is also possible with the following command:

```bash
python -X importtime -c "import serena.cli" 2> importtime.txt
```

| Cumulative Import Time | Before | After |
|------------------------|--------|-------|
| `anthropic`            | 2.424469 s | Not imported |
| `serena.analytics`     | 2.437530 s | 0.014110 s |
| `serena.cli`           | 5.129544 s | 2.891129 s |

These times are nested, not additive, and do not measure the full MCP readiness time. Other unpatched runs measured 3.29–3.59 seconds for `anthropic`. Results may vary depending on the machine and load. No comparison with older Serena versions has been made to determine when this behavior began.

The Desktop observation noted that the MCP server was omitting optional tools, specifically mentioning `serena`. Tools became available later. This is distinct from the PyPI version check in issue #1774 and the C# discovery in issue #1999.

### Technical Cause, Suggested Fix, and Regression Check

The module-level import in `src/serena/analytics.py` loads the SDK:

```python
from anthropic.types import MessageParam, MessageTokensCount
```

This import is local Python import overhead and does not involve waiting for a Claude API response. The `AnthropicTokenCount.__init__` method already imports `anthropic` on demand, but the module-level import bypasses this deferral.

The tested local change was:

```diff
-from anthropic.types import MessageParam, MessageTokensCount
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from anthropic.types import MessageTokensCount
@@
     def _send_count_tokens_request(self, text: str) -> MessageTokensCount:
+        from anthropic.types import MessageParam
+
         return self._anthropic_client.messages.count_tokens(
```

`from __future__ import annotations` is already present. This check failed before the change and passed afterward:

```python
import sys
import serena.cli
from serena.analytics import RegisteredTokenCountEstimator, ToolUsageStats

stats = ToolUsageStats(RegisteredTokenCountEstimator.CHAR_COUNT)
stats.record_tool_usage("check", "abcdefgh", "abcd")
entry = stats.get_stats("check")
assert (entry.num_times_called, entry.input_tokens, entry.output_tokens) == (1, 2, 1)
assert not any(
    name == "anthropic" or name.startswith("anthropic.")
    for name in sys.modules
), "anthropic loaded during CHAR_COUNT startup"
```

A fresh Codex CLI session also successfully called `initial_instructions`, retrieved a C# symbols overview, and obtained diagnostics. The Anthropic API-backed counter was not tested against the live API; its on-demand behavior should be covered by upstream tests.

L4XB:
Fixes #2012.

`serena.analytics` imported `anthropic.types` at module level for two type annotations, so every `import serena.cli` / `import serena.agent` loaded the whole anthropic package (2.4–3.6 s on the reporter's Windows machine) even with the default `CHAR_COUNT` estimator. As the issue describes, that delay can push MCP startup past a client's initial tool-discovery window.

**Change**

- The `anthropic.types` import moves behind `TYPE_CHECKING` (the module already uses `from __future__ import annotations`).
- `AnthropicTokenCount` already imported `anthropic` lazily in `__init__`; the request body is now a plain dict, which is exactly what `MessageParam(...)` produced at runtime (it is a `TypedDict`).
- CHANGELOG entry added.

**Verification**

```
$ python -c "import sys, serena.cli; print('anthropic' in sys.modules)"
```
prints `True` on `main` and `False` with this change. On my machine `import serena.cli` drops from 1.44 s to 0.36 s (macOS, warm cache).

`test/serena/test_analytics_imports.py` runs that check in a fresh interpreter for `serena.cli`, `serena.agent` and `serena.analytics`, and checks that the Anthropic estimator still sends the expected `count_tokens` request. `ruff format/check` and `ty check` pass on the touched files.

## Checklist


AmirF194 (commented):
I ran this against `8f121d51` in a clean container, before and after.

`python -c "import sys, serena.cli; print('anthropic' in sys.modules)"` prints `True` on `main` (403ad0a) and `False` on this branch, with the exact core dependency set pinned in `pyproject.toml` installed. `grep -rn anthropic src/serena` shows `analytics.py` is the only file that touches the package, and the only other reference left is the lazy `import anthropic` inside `AnthropicTokenCount.__init__`, unchanged by this diff. `test/serena/test_analytics_imports.py` also passes (4/4) in the same container.

```
$ PYTHONPATH=src python3 -c "import sys, serena.cli; print('anthropic' in sys.modules)"
main:  anthropic loaded: True
PR:    anthropic loaded: False
$ pytest test/serena/test_analytics_imports.py -q
4 passed in 2.49s
```

No gaps I could find: the `MessageParam` TypedDict swap and the `TYPE_CHECKING` guard both check out against the only call site.

L4XB:
I have read the CLA Document and I hereby sign the CLA

L4XB:
Thanks for running it in a clean container rather than reading the diff — the `'anthropic' in sys.modules` check before and after is exactly the thing that would have caught it if the `TYPE_CHECKING` guard had leaked.

Two notes on the current state:

The CLA was the only thing outstanding on my side and I have now signed it. The check still reads `pending`, which I think is cla-assistant not having re-run rather than a real refusal; it may need a push or a maintainer recheck.

The branch is six commits behind `main` (`18fa47bf`). I have deliberately not rebased, because that would move it off `8f121d51` and invalidate what you just verified. Say the word and I will rebase, or a maintainer can merge as-is — `mergeable` is clean either way.

AmirF194:
Good, that matches what I found. The rebase timing isn't mine to call, that's up to a maintainer.

opcode81:
Thanks for your contribution.

PR #2016, opened 2026-09-11
Fix string-keyed language server settings after registry resolution

vitalyruhl:
## Summary

Testing #1988 with the external QuickScript adapter exposed a settings regression: project/global configuration provides string keys, but `SolidLSPSettings.get_ls_specific_settings()` looks up only the new identifier object. A configured `ls_path` is therefore lost before the language server starts. Built-in language servers are affected too.

This draft targets **`ls-registry`**, so its diff contains only the proposed follow-up fix for #1988. It follows the maintainer's testing request in https://github.com/oraios/serena/pull/1988#issuecomment-5583110089 and the earlier work in #1904.

The lookup retains identifier-keyed mappings first and falls back to `ls_id.get_key()` only when absent. Explicit empty mappings keep their precedence. The accepted key types are reflected in the annotation and method documentation. Five behavior tests and the required changelog entry are included.

## Test results

Tested against `2f7946ded11a0e2dd467da8042c3521b3d209642` plus this fix, on Windows with Python 3.13.15 and pytest 8.4.1.

| Check | Before fix | After fix |
| --- | --- | --- |
| New settings regression tests | 2 failed, 3 passed | 5 passed |
| Affected settings, registry, manager, and config tests | 60 existing tests passed | 65 passed, including 5 new tests |
| Real external QuickScript/Python integration | 3 passed, 3 failed | 6 passed |
| Built-in settings reproduction | Object key preserved; string key lost | Both preserved |
| Formatting/lint | — | Passed |
| Source and test type checks | — | Passed |

The integration exercises project and global configured paths, configured-path precedence over an invalid environment value, definition/reference navigation, cache and restart behavior, unknown-adapter rejection, and coexistence with built-in Python. It uses the real QuickScript language-server subprocess and `jedi-language-server==0.47.0`.

The three failed integration cases became green with this settings-only runtime change. The external adapter had already been migrated to the new `solidlsp.language_server_registration` contract; neither that adapter nor the native server needed another change for this fix. The five portable regression tests are included in this PR; the six project-specific integration cases were run in an external local harness.

Focused checks, rerun on the final publication tree:

```sh
uv run --no-sync poe format
uv run --no-sync poe type-check
uv run --no-sync poe test test/solidlsp/test_settings.py test/solidlsp/test_ls_config.py test/serena/test_ls_manager.py test/serena/config/test_serena_config.py
```

All tests used isolated environments and task-local cache/fixture directories. Full all-language tests and normal installed-runtime validation were not performed. Earlier MCP smoke tests (initialize, tool listing, symbol overview, symbol body, references) and 25 native LSP tests passed during the adapter validation; they were not rerun for this settings-only fix.

## Checklist


opcode81:
Thanks a lot for spotting this! Since `LanguageServerId` is no longer a `StrEnum`, the LS-specific settings being keyed by it is actually wrong now. The key is obtained via `get_key()` only.
I will amend the fix to my commit.

PR #2022, opened 2026-09-11
Route extensionless scripts to their language by the shebang line

### LINKED ISSUE #2020: Scripts without an extension are invisible: `FilenameMatcher` matches by suffix only — route them by shebang?
Serena hands a file to a language server by its suffix (`FilenameMatcher`). A repository's control plane is often extensionless scripts with a shebang — `dev`, `bin/check-*`, a `#!/usr/bin/env python3` or `#!/usr/bin/env bash` first line — and for those every symbolic tool answers "Cannot extract symbols from file …". In our monorepo that is 26 scripts (15 Python, 11 shell) that had to be read and edited by hand, next to the code that IS served.

Let a language's matcher declare *shebang interpreters* (Python: `python`, `pypy`; Bash: `bash`, `sh`, `zsh`, `dash`, `ksh`; Ruby; Perl). `is_relevant_filename` keeps deciding by extension when there is one; for an existing file WITHOUT an extension it reads the first line, resolves `env` (and its `-S` / `VAR=value` arguments), strips a trailing version (`python3.12` → `python`) and matches. A bare filename (not a path to an existing file) is never sniffed; the verdict is cached per path and mtime.

Measured on our repository: `dev` (bash) — 60+ functions in `get_symbols_overview`, a function's body, its callers; `bin/adherence-report` (python3) served through pyrefly. This changes which files are considered sources (extensionless files with a matching shebang become visible), hence the issue before the PR. I have the change ready with unit tests (no language server needed).

Per-project extra extensions (`add_extensions`) — cannot express "no extension". Renaming the scripts — not ours to decide for every repository.

**Additional context**
The read is 256 bytes of one file per extensionless path, only for languages that declare interpreters.

MarioRial22:
Resolves #2020 

`FilenameMatcher` gains `shebang_interpreters`: an existing file without an extension is routed to a language when its first line is a shebang naming one of the language's interpreters — `env` resolved (with its `-S` and `VAR=value` arguments), a trailing version stripped (`python3.12` → `python`). Anything with an extension is decided by the extension alone; a bare filename is never sniffed; the verdict is cached per path and mtime. Python (`python`, `pypy`), Bash (`bash`, `sh`, `zsh`, `dash`, `ksh`), Ruby and Perl declare their interpreters.

Tests: `test/solidlsp/test_shebang_matcher.py` (15 cases, no language server needed) — the `env` forms, versioned interpreters, a foreign interpreter, the extension winning over the shebang, no shebang, a bare filename, a directory, a matcher without interpreters never reading files, and the cached verdict following the file's mtime.

Measured on a 30k-line monorepo: its extensionless `dev` (bash) and `bin/*` (python3) scripts answer overviews, bodies and references through bash-language-server and pyrefly.

## Checklist


src/solidlsp/ls_config.py
```diff
@@ -67,12 +77,56 @@ def add_extensions(self, *file_extensions: str) -> None:
...
     def is_relevant_filename(self, fn: str) -> bool:
-        if not self._case_sensitive:
-            fn = fn.lower()
+        name = fn.lower() if not self._case_sensitive else fn
         for ext in self._file_extensions:
-            if fn.endswith(ext):
+            if name.endswith(ext):
                 return True
-        return False
+        return self._matches_shebang(fn)
```
opcode81: This won't work. A "filename" is not the same as a file path. This method's contract is for a filename (optionally a full path), and you cannot read a file based on only its filename.

MarioRial22: You're right: the sniff was hanging off a method whose contract is a name, and `compute_language_server_support_composition` does pass bare basenames, so whether an extensionless script was routed silently depended on which caller asked.

Split it (b6e84aa2): `is_relevant_filename` is back to deciding by the name alone and never reads anything; the shebang sniff is a new `is_relevant_file(path)`, and the three places that actually hold a path ask that one — `Project.is_ignored_path`, `SolidLanguageServer.is_ignored_path` and the language-composition scan, which now passes the full path instead of the basename. `test_is_relevant_filename_decides_by_the_name_alone` pins the contract: the same answer for a bare name and for the full path of an extensionless script.

PR #2024, opened 2026-09-11
Add rename_file: rename or move a file and update its importers through workspace/willRenameFiles

MarioRial22:
A new editing tool, `rename_file(relative_path, new_relative_path)`: the language server's `workspace/willRenameFiles` edit is applied while the files are still at their current paths (that is what the edit's URIs address, the moved file's own imports included), the file is moved (parent directories created), and `workspace/didRenameFiles` is sent. Files only. The tool's answer says when the server proposed no edits, so the caller knows to search for the old name.

- `SolidLanguageServer.request_will_rename_files` / `notify_did_rename_files`; the file is opened before the request, which is what lets tsserver load the project it belongs to.
- `LanguageServerCodeEditor.rename_file` (refuses a missing source or an existing target; a server that includes the rename among its document changes is honoured).
- The Python servers (pyright, basedpyright, ty, pyrefly) and the TypeScript server advertise `workspace.fileOperations` (willRename/didRename).
- `test/conftest.py`: `project_with_explicit_ls_context` (also proposed in the type-hierarchy PR, byte-identical — whichever merges first).

pyrefly needs two things, both documented in the server module, the tool and the changelog: version >= 1.2.0 (1.1.1 advertises the request and answers `null`; the test pins 1.2.0 through `ls_specific_settings`, so this PR does not depend on the bump PR) and `indexing_mode: lazy-blocking` — in the default lazy-non-blocking mode pyrefly may answer before its reverse-dependency graph is built and miss importers (measured: 1 of 5 dependents right after start-up, all 5 a few seconds later); in blocking mode the index completing cancels the first request, the existing mutation retry re-asks, and the answer is complete. pyrefly rewrites absolute imports; a relative `from .models import` is left alone.

Tests (`test/serena/test_rename_file.py`, on a throwaway copy of the fixtures): pyrefly renames `test_repo/models.py` → `entities.py` and rewrites its two absolute importers; typescript-language-server moves `formatters.ts` into a new directory and rewrites `index.ts`; a missing source and an existing target are refused.

Measured on a 30k-line project: a module with 27 importers moved and rewritten in 0.5 s.

## Checklist


src/serena/code_editor.py
```diff
@@ -414,6 +414,43 @@ def rename_symbol(self, name_path: str, relative_path: str, new_name: str) -> st
...
+        if not os.path.isfile(old_abs_path):
+            raise FileNotFoundError(f"'{relative_path}' is not a file in the project (directories cannot be renamed with this method)")
+        if os.path.exists(new_abs_path):
+            raise FileExistsError(f"'{new_relative_path}' already exists")
+
+        lang_server = self._get_language_server(relative_path)
+        workspace_edit = lang_server.request_will_rename_files(relative_path, new_relative_path)
+        # The proposed edits address the files at their CURRENT paths (the file being moved included, e.g. for its own
+        # relative imports), so they are applied before anything is moved.
+        num_edited_files = self._apply_workspace_edit(workspace_edit) if workspace_edit is not None else 0
```
AmirF194, line 446: I ran this against d24f9b31 in a clean container (python:3.12-slim, `pip install --no-deps -e .`, the real `EditOperationRenameFile` class, not a reimplementation).

`EditOperationRenameFile.apply()` (code_editor.py:337-346, unchanged by this PR) does a bare `os.rename(old_abs_path, new_abs_path)`, no `os.makedirs` for the target's parent directory. `rename_file()` creates that directory itself, but only in its own fallback branch (line 442). When a server includes the rename among its `documentChanges` (the case this PR's own comment at line 440 names), `_apply_workspace_edit` reaches `EditOperationRenameFile.apply()` at line 438 first, and that path never creates the directory.

Repro, using this PR's own TypeScript scenario (`formatters.ts` to `lib/greeters.ts`, target dir does not exist yet), building the rename op a server would return and calling `EditOperationRenameFile.apply()` directly:

```
old exists before apply: True
new dir exists before apply: False
apply() raised: FileNotFoundError [Errno 2] No such file or directory: '.../formatters.ts' -> '.../lib/greeters.ts'
```

The PR's own test for this scenario only passes because typescript-language-server happens not to include a rename op in its response. A server that does (a normal way to answer `workspace/willRenameFiles`) would hit this uncaught exception on any cross-directory move, so `rename_file`'s "its directory may differ" promise would not hold for that server.

Suggested fix: move `os.makedirs(os.path.dirname(new_abs_path), exist_ok=True)` into `EditOperationRenameFile.apply()` itself, ahead of its `os.rename`, instead of only in `rename_file()`'s fallback branch.

AmirF194: Confirmed: f7924a54 moves the makedirs call into a shared _move_file used by both call sites and adds a test for the server-side rename case. Thanks for the fix.

MarioRial22: Good catch — the `makedirs` sat on the fallback move only. It now lives in `EditOperationRenameFile.apply()` (one `_move_file` helper serves both the operation and the fallback), and `test_a_rename_the_server_includes_in_its_edit_creates_the_target_directory` applies a `documentChanges` rename into a not-yet-existing directory through `_apply_workspace_edit`: it fails with your `FileNotFoundError` on the previous commit and passes on f7924a54.

PR #2025, opened 2026-09-12
Per-edit diagnostics as a project setting, and the answer names the language server they come from

MarioRial22:
`EditingToolWithDiagnostics.ENABLE_DIAGNOSTICS` is a hard-coded `False` with no way for a project to opt in, while the machinery behind it is complete and tested. This adds `edit_diagnostics` to `project.yml` (default `false`, so nothing changes for anyone): a project whose workflow wants the verdict after every edit turns it on; the global flag keeps its meaning (on regardless of the project). The answer also names, per edited file, the language server the diagnostics come from (`diagnostics_from`): a project that grades with a second type checker at its gate should not read one server's "no errors" as the gate's.

Test (`test/serena/test_serena_agent.py`): with the flag off and the setting off, an edit reports nothing; with the setting on, the same edit reports the new diagnostic and `diagnostics_from` names the server. The project template documents the key.

## Checklist


AmirF194 (commented):
I ran `ProjectConfig.load()` against this PR's head (`87f874b9`) in a clean container to check whether `edit_diagnostics: true` in `project.yml` actually reaches the loaded config.

`ProjectConfig._from_dict` (src/serena/config/serena_config.py:638-661) builds the returned instance with an explicit keyword for every other field (`read_only=data["read_only"]`, etc.) but never reads `data["edit_diagnostics"]`. Since the dataclass field defaults to `False`, any value set in `project.yml` is silently dropped and the loaded `ProjectConfig` always comes back with `edit_diagnostics=False`, regardless of what the file says.

```
$ python3 /repro.py
edit_diagnostics on loaded ProjectConfig: False
  ...
AssertionError: BUG: project.yml's edit_diagnostics: true was not honored by ProjectConfig.load
```

(repro: write a project.yml with `edit_diagnostics: true` under a temp project root, call `ProjectConfig.load(project_root, serena_config)`, assert `.edit_diagnostics is True`.)

The new test in this PR doesn't catch it because it sets `project_config.edit_diagnostics = True` directly on the Python object rather than loading it from `project.yml`, so it never exercises `_from_dict`.

Fix: add `edit_diagnostics=data["edit_diagnostics"],` to the `cls(...)` call in `_from_dict`, next to `read_only=data["read_only"],`.

AmirF194:
Confirmed: c76bdf70 adds edit_diagnostics to the _from_dict call and covers it with a test that loads project.yml directly. Thanks for the fix.

MarioRial22:
Right — `_from_dict` builds the instance with an explicit keyword per field and I had not added this one, and the test mutated the loaded config instead of loading it, which is why it stayed green. Fixed in c76bdf70 (`edit_diagnostics=data["edit_diagnostics"]`), and `test_edit_diagnostics_is_read_from_project_yml` now round-trips through `ProjectConfig.load` from a `project.yml` with the key set to `true`; it fails on the previous commit.

src/serena/config/serena_config.py
```diff
@@ -332,6 +332,11 @@ class ProjectConfig(SharedConfig, ModeSelectionDefinitionWithAddedModes):
     ls_workspace_folders: list[str] = field(default_factory=lambda: ["."])
     ls_additional_workspace_folders: list[str] = field(default_factory=list)
     read_only: bool = False
+    edit_diagnostics: bool = False
```
opcode81, line 335: This would need to be called `ls_edit_diagnostics` to make clear that applies to language server applications (the LSP backend) only.

src/serena/resources/project.template.yml
```diff
@@ -121,6 +121,13 @@ ignored_paths: []
 # Added on 2025-04-18
 read_only: false
 
+# whether editing tools report the diagnostics (warnings and errors) an edit newly introduced in the edited
+# files, as the language server sees them. The answer names the language server the diagnostics come from
+# (`diagnostics_from`), since one server's verdict is not every checker's.
+# Off by default: intermediate edits often introduce diagnostics on purpose (a signature changed here, its
+# callers in the next edit), and reporting them at every step is noise for such workflows.
+edit_diagnostics: false
```
opcode81, line 129: The docstring would have to make clear that this feature currently affects the traditional `tools` agent interface only; the REPL interface does not provide diagnostics.

opcode81 (changes requested):
Please see my comments.

@MischaPanch, do you think it is a good idea to expose the feature in this way (opt in)?

PR #2026, opened 2026-09-12
find_declaration answers for a symbol defined outside the project instead of failing on the workspace check

MarioRial22:
A definition in the standard library or an installed package used to end in `ValueError: Path … is outside of configured workspaces`: `get_relative_path` computes a `../..` path for anything on the same drive, so the location passed the skip meant for out-of-repo results and failed when the file was opened. "How is this third-party symbol typed?" is exactly the question a language server answers well, and `find_declaration` refused it.

- `SymbolLocationRequest` recognises a location outside the repository and its workspace folders (`PathWorkspaceStatus`); by default it is skipped as before — at info level, without calling it a bug.
- `DefinitionLocationRequest` keeps it, marked `external` (`relativePath` None).
- `request_defining_symbol` reads the symbol by its absolute path — the file-opening machinery accepts one as it stands — and the tool answers with the defining symbol, its body on request, and `external: true` so the caller does not take the path for a project file. Read-only by construction: the editing tools address project files.

Test: `os.path.exists` from the Python test repo resolves to typeshed's `genericpath.pyi` under pyright (`external`, an absolute `.pyi` path, `def exists(` in the body); the same measured through pyrefly (its bundled typeshed under `/tmp`).

## Checklist


src/serena/symbol.py
```diff
@@ -505,6 +519,8 @@ def to_dict(
             result["location"] = self.location.to_dict(include_relative_path=relative_path)
         elif relative_path:
             result["relative_path"] = self.relative_path
+            if self.is_external:
```
AmirF194, line 524: Ran this against 18e7c2c7 in a clean container. `find_declaration`'s new `external` result carries an absolute OS path in `relative_path` (the PR's own test asserts `os.path.isabs(...)`), and every editing tool takes `relative_path` as an argument with no special-casing.

`CodeEditor.edited_file_context`'s only guard, `FileProxy.is_external_path()` (`code_editor.py:82`), checks for the JetBrains plugin's own encoded prefix, not a plain absolute path. `_save_edited_file`'s `os.path.join(self.project_root, relative_path)` (`code_editor.py:90`) collapses to that absolute path once it is absolute, so a write lands outside the project:

```
$ python3 repro.py   # oraios/serena @ 18e7c2c7, clean python:3.12-slim container
FileProxy.is_external_path(absolute_relative_path) -> False
victim file contents after CodeEditor._save_edited_file(): 'AGENT-WRITTEN CONTENT\n'
```

`is_external` (`symbol.py:522`) already knows the symbol is external where `relative_path` is built. Rejecting an edit there, or having `edited_file_context` also reject `os.path.isabs(relative_path)`, would close this before it leaves the read tools.

AmirF194: Confirmed: 229df1cc rejects an absolute relative_path in _assert_editable_path and applies the guard at every write entry point, not just edited_file_context, with a test against the external-symbol case. Thanks for the fix.

MarioRial22: Agreed, and thanks for the repro. The refusal is now in the code editor, the one seam every edit goes through: `CodeEditor._assert_editable_path` rejects an absolute path (beside the JetBrains external prefix it already rejected), and it is called from `edited_file_context` and at the top of each symbol-editing method — before the symbol lookup, so the tool answers with the reason rather than the symbol tree's "path is ignored". The stdlib test now also asks `replace_symbol_body` for the path `find_declaration` returned and asserts the `ValueError` and an unchanged `.pyi` (229df1cc).

Issue #2003, opened 2026-09-09
A request timeout is a bare TimeoutError and bypasses the language-server restart path


tagawa0525:
Serena 1.7.1.dev0 at `701e7c84`, read from the source (line numbers refer to that commit) and reproduced with solidlsp directly where stated; Linux; Python (pyright) and TypeScript.

`Request.get_result` turns `queue.Empty` into a plain `TimeoutError` (`src/solidlsp/ls_process.py:101-107`). Everything that recovers from a broken language server keys on `SolidLSPException.is_language_server_terminated()` (`src/serena/tools/tools_base.py:382-397`), so a timed-out request never reaches the restart-and-retry branch, even when the server is hung rather than slow. The agent sees "Tool execution timed out after N seconds." (`tools_base.py:430-433`) with no indication whether the server is still healthy, and the next tool call goes to the same server.

`$/cancelRequest` is defined (`lsp_protocol_handler/lsp_requests.py`) but never sent, so the timed-out request also keeps running in the server and its late response is dropped as an unknown id.

Suggestion: raise a `SolidLSPException` subclass for the timeout (carrying the language), send `$/cancelRequest`, and let the tool layer decide whether a timeout counts as "terminated" (e.g. after a health probe such as a cheap `documentSymbol`).

Preconditions:

- [x] I have made sure it's an actual issue, not a question (use [GitHub Discussions](https://github.com/oraios/serena/discussions) instead).
- [x] I have consulted the [user guide](https://oraios.github.io/serena/02-usage/000_intro.html) and verified that the issue cannot be resolved by adjusting configuration/following recommended workflows.
- [x] I have looked for similar issues and discussions, including closed ones.

Issue details:

- [x] I have provided a meaningful title and description.
- [x] I have explained how the issue arose and, where possible, added instructions on how to reproduce it.
- [x] I have added details on my setup: Serena version, MCP client, OS, the programming language(s), relevant configuration adjustments and project specifics.
- [x] If the issue relates to an application of Serena to an open-source project, I have added the link.

opcode81:
A response that times out should never trigger the LS restart path; this makes no sense. A timeout simply means that the time we were willing to wait was exceeded (which could be entirely normal for a short timeout or time-consuming request on a large codebase); it does not imply that there is a problem in the language server that requires it to be restarted.

What is the actual problem you encountered in practice, and why do you think restarting the language server would be an appropriate measure to take?

tagawa0525:
[...]

PR #2031, opened 2026-09-13
fix(solidlsp): drop a timed-out request from the pending-request table

ebarkhordar:
`_send_request_once` puts the request into `_pending_requests`, and the only thing that ever takes it back out is `_response_handler`, when a response arrives. Against a server that is running but not answering, every timed-out request keeps its entry for the lifetime of the server. `_cancel_pending_requests` later counts those and pushes an error into requests nobody is reading.

Small correction to the issue while I'm here: the late response isn't dropped as an unknown id today, because the entry is still sitting there when it arrives. It is after this change, which is what the second test pins.

I left the `TimeoutError` itself and the restart path alone. Whether a timeout should count as "terminated" is your call and a bigger change, and `$/cancelRequest` is a separate thing again, so this is just the table.

- `test/solidlsp/test_request_timeout_cleanup.py`, three of the four fail on main
- `poe lint`, `poe type-check` and `codespell` all clean
- `test/solidlsp` and `test/serena` give the same failures before and after the change; plenty of them fail for me either way, I have no language servers installed here

Refs #2003

## Checklist


opcode81 (approved):
Thanks. Please accept the CLA, then we can merge this.

ebarkhordar:
Thanks for the review. I will get the CLA signed.

feiiiiii5 (commented):
Ran this branch and its tests locally, and checked it against my own open PR #2030 in the same file. The change is correct, minimal and well-tested; two notes for the merge decision and one minor test observation. Full disclosure: I am the author of #2030, so the interaction point below is the part of this review with the most self-interest in it — weigh it accordingly.

*(Posted with an AI agent on this account; the commands and outputs are listed at the end.)*

## What I measured

`solidlsp.__file__` = `/private/tmp/oas-serena-r7-2031/src/solidlsp/__init__.py` (Python 3.14.5), i.e. this branch's tree:

| run | result |
|--|--|
| `test/solidlsp/test_request_timeout_cleanup.py` at head `6c030da5` | rc=0, **4 passed** |
| same file, `ls_process.py` replaced by the merge-base `c6fbd1c5` copy | rc=1, **3 failed / 1 passed** — `test_timed_out_requests_do_not_accumulate`, `test_a_late_response_to_a_timed_out_request_is_discarded`, `test_cancelling_after_a_timeout_only_sees_the_live_request` |
| same, with the pop retargeted to `request_id + 1` | the same 3 fail — so the tests pin *this* id, not just "table got smaller" |
| `-m "not (java or … or ocaml)"` (expression composed from `pytest.yml`'s own `MARKERS_*` vars) | rc=0, **4 passed, 0 deselected** → lands in the LS-free `catch-all` batch, which is also where `poe lint` and `poe type-check` run |

So your "three of the four fail on base" checks out, and the surviving one is `test_answered_requests_also_leave_no_entry_behind`, which is the control — right for it to pass on base.

The lock discipline is consistent too: registration (`ls_process.py:349`), the response pop (`:424`) and `_cancel_pending_requests` (`:332`) all hold `_response_handlers_lock`, and the new pop takes the same lock. `pop(request_id, None)` also correctly tolerates the race where the response lands between the timeout and the pop.

## Merge-level note: this closes one of two leak producers

The description says the only place an entry was ever removed is `_response_handler`. True for the timeout path, and this fixes it. Worth knowing while it is fresh: on `c6fbd1c5`, `StdioLanguageServer._send_payload` catches `(BrokenPipeError, ConnectionResetError, OSError)` and returns silently, and does **not** catch `ValueError` — so a stdin already closed by `_stop` raises straight out of `_send_request_once` at `:352`, past the `except TimeoutError`, and leaves the entry registered with no timeout ever firing. That is the case #2030 addresses (fail fast with `LanguageServerTerminatedException`, which also clears the table via `_cancel_pending_requests`).

The two are complementary rather than overlapping, and they don't collide: `git merge-tree --write-tree r7/2030 r7/2031` merges cleanly (`a940ec19`). With #2031 alone the timeout producer is closed and the raise producer still leaks; with #2030 alone the raise producer is closed and every timeout still accumulates. If only one can land soon, #2031 is the smaller and safer of the two.

## Blocking, and one minor

- **CLA is still unsigned.** The cla-assistant badge on this PR reads `not_signed` and your 09-16 comment said you would sign it; nothing has changed on that check since. `license/cla` is a hard gate here for SolidLSP-only changes too, so it is worth clearing before a maintainer spends time on the review.
- Minor: in `test_cancelling_after_a_timeout_only_sees_the_live_request`, the intermediate `assert list(server._pending_requests) == [99]` is a statement about the dict the two preceding lines just built, so it holds whatever the fix does. The last line (`live.get_result(...).is_error()`) is what actually pins the behaviour. Dropping the middle assert, or asserting that the timed-out id is absent *before* inserting 99, would keep the intent and remove a line that can never fail.

## Commands

```
git fetch up pull/2031/head:refs/heads/r7/2031 && git worktree add /private/tmp/oas-serena-r7-2031 r7/2031
uv sync --extra dev
.venv/bin/python -m pytest test/solidlsp/test_request_timeout_cleanup.py -q -p no:cacheprovider
# pre-fix run: src/solidlsp/ls_process.py temporarily replaced with
#   git show c6fbd1c5:src/solidlsp/ls_process.py   (restored byte-identical afterwards, verified)
git merge-tree --write-tree r7/2030 r7/2031
```

I did not run the full suite or start a language server; these tests are the LS-free ones, and CI's `catch-all` jobs on all three OSes are already green on this head. Nothing on this PR was pushed or changed by me.

feiiiiii5:
Correction on the framing of my review above: this PR already has an **approval from @opcode81 dated 2026-09-16** (review 5217428490). I queried `review_comments` on the PR object, got `0`, and wrote as though nobody had looked at it yet — that field counts inline review comments, not reviews, so it says nothing about whether the PR was reviewed. Apologies for the misleading lead-in.

The measurements themselves are independent of who reviewed what and still stand. One refinement to my "blocking" item, now checked directly rather than from the badge: `license/cla` on `6c030da5` is **pending** (the only status on that commit, `state=pending`, target `cla-assistant.io/oraios/serena?pullRequest=2031`), so that looks like the one thing between this approval and a merge, alongside `@ebarkhordar`'s 09-16 note that the signature was still to come.

The part I would still flag for whoever merges: since #2030 touches the same file, the note about which leak producer each PR closes may be useful at merge time — `git merge-tree` says they combine cleanly.

ebarkhordar:
The intermediate assert is read correctly, and I am leaving it in: this PR is approved at this head, and pushing a change to a line that cannot fail spends that approval for nothing.

If @opcode81 would rather it were gone before merge, I will drop it.

opcode81:
@ebarkhordar the CLA is not signed yet.

ebarkhordar:
Sorry, not yet. I'll get it done.

PR #2033, opened 2026-09-14
Fix: C# project discovery ignores configured ignore patterns

### LINKED ISSUE #1999: C# project discovery in Serena bypasses .gitignore, causing slow startup
Preconditions:

Issue details:

## Summary

Even with `ignore_all_files_in_gitignore: true` set, Serena's C# startup process scans directories that should be ignored when searching for `.sln` or `.csproj` files.
This search process takes place within Serena's Python code before any discovered projects are passed to Roslyn. In my repository, this process causes a startup delay of several minutes, sometimes leading to a timeout of the `initial_instructions`.
Relation to #1505
This appears to be the same behavior reported in #1505.
That issue was closed with the explanation that Roslyn does not seem to support the configuration of ignored folders.
However, I would like to clarify the boundaries of the process in question here.
This report concerns the project discovery performed by Serena itself, not the file discovery that occurs internally within Roslyn.

## Root cause

The problem is in Serena’s C# project-discovery code, specifically these functions in `src/solidlsp/language_servers/csharp_language_server.py`:
 
* breadth_first_file_scan(root_dir) traverses directories using os.listdir(), os.path.isdir(), and os.path.isfile(). It skips dot-prefixed names, but never checks .gitignore or Serena’s configured ignore patterns.
* find_solution_or_project_file(root_dir) uses that traversal to locate a solution or project during initialization.
* _open_solution_and_projects() uses the same traversal to find a solution and collect .csproj files, then explicitly sends them to Roslyn through solution/open and project/open notifications.

The underlying issue is that loading ignore patterns does not automatically apply them to filesystem operations. Each traversal must explicitly consult the ignore matcher. Serena has that mechanism for its normal searches, but this C# discovery helper bypasses it.
As a result, ignore_all_files_in_gitignore: true can be correctly configured and the .gitignore files successfully loaded, while C# startup still traverses ignored directories and sends projects found inside them to Roslyn.

RizgarOzan:
Fixes #1999

## What
C# solution/project discovery (`_open_solution_and_projects`) now respects the language server's ignore spec (`ignored_paths`, plus `.gitignore` patterns when `ignore_all_files_in_gitignore` is on). `breadth_first_file_scan` takes an optional `is_ignored_path` predicate (paths relative to the scan root): ignored directories are not traversed and ignored files are not yielded. The default ignores nothing, so other callers are unchanged.

I intentionally did not reuse `self.is_ignored_path`, since it also skips `bin`/`obj`/`packages`, which would stop discovering projects laid out as `packages/Foo/Foo.csproj`.

## How I tested
- Repro (`src/App.csproj` + `vendor/huge/Vendored.csproj`, with `vendor/` ignored), recording what `_open_solution_and_projects` sends to Roslyn:
  - before: `project/open ['src/App.csproj', 'vendor/huge/Vendored.csproj']`
  - after: `project/open ['src/App.csproj']`
- New unit test `test_breadth_first_file_scan_skips_ignored_paths`.
- `pytest test/solidlsp/csharp/test_csharp_basic.py -m csharp` (starts the real Roslyn server) → 17 passed.
- `uv run poe format`, `uv run poe lint`, `uv run poe type-check` → all checks passed.

Not changed: `DependencyProvider.create_launch_command()` still calls `find_solution_or_project_file` without the ignore spec, but its result is only logged and it runs in `SolidLanguageServer.__init__` before the ignore spec exists, so fixing it would mean reordering the base-class init — out of scope for this fix.

## Checklist


RizgarOzan:
The `niche (ubuntu-latest)` failure is unrelated to this change: the two failing tests are R language server cross-file tests (`test/solidlsp/r/test_r_basic.py`), and the same job also fails on `main` (run 34719503298). All C#/catch-all jobs pass.

PR #2036, opened 2026-09-15
Fix: shell commands outlive the tool call that was meant to bound them

ChiFungHillmanChan:
## The issue

Partially addresses #1251 (finding 2, "Shell execution with no timeout"). `execute_shell_command`
runs the command with no deadline of its own, so a command that does not terminate keeps running
after the tool call that was meant to bound it has already ended. The agent is told the call timed
out; the shell process, and the thread waiting on it, stay alive for the remaining lifetime of the
server.

Only finding 2 is addressed here, to keep the PR to a single logical change. Finding 1 (memory path
traversal) is already fixed on `main` by `3ab9ff38` and the containment check in
`MemoriesManager._resolve_memory_path`; the issue is just still open.

## How it works now, and why that is wrong

`src/serena/util/shell.py` waits on the process with no bound:

```python
stdout, stderr = process.communicate()
```

The only deadline anywhere above it is the tool timeout at `src/serena/tools/tools_base.py:426`:

```python
timeout = self.agent.serena_config.tool_timeout
task_exec = self.agent.issue_task(task, name=self.__class__.__name__, timeout=timeout)
return task_exec.result(timeout=timeout)
```

That bounds *waiting for the result*, not the work. `Task.result`'s own docstring says so, at
`src/serena/task_executor.py:92`:

> if it is already running, the underlying thread continues to run, but its result is discarded and
> any waiter will receive a CancelledError

So on expiry the caller gets a `TimeoutError`, the future is discarded, and the command keeps
running. `DEFAULT_TOOL_TIMEOUT` is 240s, so an agent that runs a server or a process waiting on
input — which the tool's own docstring warns against, but cannot prevent — leaks one process and one
thread per occurrence.

The repository already handles this correctly one level over, for project activation commands, at
`src/serena/agent.py:1312`:

```python
except subprocess.TimeoutExpired:
    log.error(...)
    terminate_process_tree_with_kill_fallback(p, terminate_timeout=5.0, process_name="activation_command")
```

## Tested before the fix

End to end, against a real MCP server: `serena start-mcp-server --transport stdio --tool-timeout 8`,
a real `initialize` handshake, and a real `tools/call` for `execute_shell_command` running a command
that takes 40s and writes a marker file at the very end. Whether the process survived is read from
`ps`, not from anything Serena reports. Same script in both states.

At merge base `18fa47bf`:

```
[  0.00s] tools/call execute_shell_command sent
[  8.01s] response: Error executing tool execute_shell_command: Tool execution timed out after 8.0 seconds.
[ 10.06s] ps check while the server is still up:
           STILL RUNNING: 95667 .../Python -c TAG='SERENA_E2E_95602'; import time,pathlib; time.sleep(40.0); ...
[ 12.14s] ps check after the server exited: STILL RUNNING (orphaned)
[ 43.01s] marker file: EXISTS -> the command ran to completion
VERDICT: LEAK
```

The command outlives not just the tool call but the MCP server process itself: after the server is
sent SIGTERM and exits, the command is still running, and it goes on to complete its work 30 seconds
later. Nothing in the session can reach it any more.

## Tested after the fix

At `d15a18d9`, same script:

```
[  0.00s] tools/call execute_shell_command sent
[  4.05s] response: Error executing tool execute_shell_command: TimeoutExpired: Command '...' timed out after 4.0 seconds
[  6.11s] ps check while the server is still up:
           no matching process -> the command was terminated
[  8.16s] ps check after the server exited: gone
[ 43.01s] marker file: absent -> the command never finished
VERDICT: NO LEAK
```

The agent also gets a better error: the specific command that timed out, instead of
`Tool execution timed out after 8.0 seconds`.

A second, narrower rig (`repro.py`) drives `ExecuteShellCommandTool.apply()` through the real
`TaskExecutor` deadline without MCP in the way, and agrees: LEAK at `18fa47bf`, NO LEAK at `d15a18d9`.

### Test suite

`test/serena` minus the seven modules that need a language server, run identically at both commits
and on both interpreters (CI pins 3.11; 3.13 is the local default):

```
uv run pytest test/serena \
  --ignore=test/serena/test_cli_project_commands.py --ignore=test/serena/test_ls_file_sync.py \
  --ignore=test/serena/test_symbol.py --ignore=test/serena/test_symbol_editing.py \
  --ignore=test/serena/test_serena_agent.py --ignore=test/serena/test_ls_manager.py \
  --ignore=test/serena/test_project_server.py -q
```

| commit | Python 3.11 | Python 3.13 |
|---|---|---|
| `18fa47bf` (merge base) | 580 passed | 580 passed |
| `d15a18d9` (this branch) | 597 passed | 597 passed |

The difference is exactly the 17 tests added here; no existing test changed behaviour.

### The rest of the repo's gate

| check | result |
|---|---|
| `uv run poe lint` (ruff format + ruff check) | green |
| `uv run poe type-check` (ty on `src` and `test`) | green |
| `codespell` on a clean checkout | exit 0, same as merge base |
| `uv sync --extra dev --locked` | resolves; `uv.lock` untouched by this branch |
| `uv run serena project health-check test/resources/repos/python/test_repo` | `Health check passed` |

The full CI matrix is five marker batches across Ubuntu, Windows and macOS; these changes land in the
`catch-all` batch, whose 1362 tests include language servers I cannot install locally (down to
Wolfram, which needs Mathematica). What is run above is that batch minus its language-server modules.

### Mutation and negative-case evidence

Six deliberate mutations of the fix, each caught:

| mutation | caught by |
|---|---|
| drop the terminate call, just re-raise | the two kill tests + the tool wiring test |
| swallow `TimeoutExpired` instead of re-raising | three tests, including the success case |
| derive no margin (`return tool_timeout`) | the deadline-ordering test, at 240/20/10s |
| flat `max(t-5, 5)` floor | the same test at 5/4/2s |
| tool stops passing the timeout through | the tool wiring test only |
| terminate the group leader, not the tree | the compound-command test only |

The last two each escape every other test, which is why both exist.

**The new tests cannot be run against the merge base.** `test_cmd_tools.py` fails at import
(`ImportError: cannot import name 'TOOL_TIMEOUT_MARGIN'`), which aborts collection; running
`test_shell.py` alone gives `3 failed, 1 passed`, all three with
`TypeError: execute_shell_command() got an unexpected keyword argument 'timeout'`. They fail on the
signature, not on the behaviour, so they are not the before/after evidence — `e2e.py` above is, and
it calls only `apply()`, whose signature is identical in both states.

The one test that passes at the merge base is the negative case,
`test_without_timeout_the_command_runs_to_completion`: omitting the timeout must keep today's
blocking behaviour, and it does, in both states. The fix is not buying success with false positives.

## What the fix does

`execute_shell_command` takes an optional `timeout`. On expiry it terminates the command's process
tree with a kill fallback and re-raises `subprocess.TimeoutExpired`, so the failure reaches the agent
as a specific error instead of a generic tool timeout.

`ExecuteShellCommandTool` derives that deadline from `tool_timeout`, the same way the language server
timeout is derived at `src/serena/project.py:511`. The margin is 5s, falling back to half the tool
timeout when the configured timeout is smaller than the margin, so the command's deadline is strictly
below the tool's for every configured value. `tool_timeout` of `None` or negative still means no
timeout, preserving today's behaviour for anyone who has opted out.

## What it does not do

- **Only the tool path gets a deadline.** The other `execute_shell_command` callers, all in
  `src/serena/config/client_setup.py`, keep the current unbounded behaviour: five fixed `--version`
  / `--help` probes, plus `_run_shell_command` at line 37, which runs client-registration commands
  from an interactive `serena` CLI session where a hang is visible to the user rather than silently
  held by the server. None of them takes an agent-supplied command, and giving the CLI path a
  timeout policy is a separate decision.
- **The margin is a choice, not a measurement.** 5s matches `project.py:511`; the proportional
  fallback below it is mine, and a maintainer may prefer an explicit config key instead of deriving
  from `tool_timeout` at all.
- **Everything above was measured on macOS only.** CI also runs Ubuntu and Windows. The
  process-tree test uses POSIX shell grouping and skips itself on Windows, where
  `terminate_process_tree_with_kill_fallback` signals a process tree rather than a process group;
  that path is unexercised here.
- **The language-server test modules were not run**, so a regression reachable only through a live
  language server would not have been caught locally. Nothing in the diff is on that path, but I am
  stating the gap rather than implying full-suite coverage.
- Findings 3-6 of #1251 are untouched.


src/serena/util/shell.py
```diff
@@ -40,7 +45,13 @@ def execute_shell_command(command: str, cwd: str | None = None, capture_stderr:
         **subprocess_kwargs(),
     )
 
-    stdout, stderr = process.communicate()
+    try:
+        stdout, stderr = process.communicate(timeout=timeout)
+    except subprocess.TimeoutExpired:
+        # communicate() does not terminate the process when the timeout expires, so without this the command
+        # would keep running - and keep holding the calling thread - for the remaining lifetime of the server
+        terminate_process_tree_with_kill_fallback(process, terminate_timeout=5.0, process_name="Shell command")
```
AmirF194, line 53: I ran this against `d15a18d9` in a clean container (python:3.11-slim, deps installed from pyproject.toml's pinned versions, no full editable install), driving `ExecuteShellCommandTool.apply()` directly the way `tools_base.py` does.

`terminate_process_tree_with_kill_fallback` is called with a fixed `terminate_timeout=5.0` here, the same value as `TOOL_TIMEOUT_MARGIN`. When the spawned process does not exit on SIGTERM (a slow shutdown hook, a JVM child, anything blocked on I/O), termination itself can take up to `terminate_timeout` plus the kill-fallback's own 2s wait (`subprocess_util.py:430-432`) before `execute_shell_command` actually raises. That consumes the whole margin the tool-level deadline was supposed to have as headroom, so the exception can still land at or after the outer `tool_timeout`, which is the race this PR sets out to close.

```
$ python3 repro_margin.py
-- positive control: process terminates immediately on SIGTERM --
tool_timeout=  10.0  command_timeout=  5.00  margin= 5.00  sigterm_resistant=False  elapsed_to_raise=  5.01s  OUTER_DEADLINE_EXCEEDED=no

-- fault case: process ignores SIGTERM, only SIGKILL removes it --
tool_timeout=  10.0  command_timeout=  5.00  margin= 5.00  sigterm_resistant= True  elapsed_to_raise= 10.01s  OUTER_DEADLINE_EXCEEDED=YES
```

(`sigterm_resistant` spawns `signal.signal(signal.SIGTERM, signal.SIG_IGN); time.sleep(30)` as the command.)

`TOOL_TIMEOUT_MARGIN` would need to cover the termination sequence's own worst case, not just the command's, so something like `terminate_timeout=5.0 + 2.0` for the margin, or a smaller `terminate_timeout` passed into `terminate_process_tree_with_kill_fallback` at `shell.py:53` so its worst case fits inside the existing 5s margin. I did not test the Windows path (`terminate_process_tree_with_kill_fallback` signals the tree rather than a process group there, per the PR description).

Issue #2001, opened 2026-09-08
trusted_project_path_patterns: a "<project>/**" entry cannot trust the project it names, and the project's ls_specific_settings are then silently discarded


tim-talostitle:
### Summary

A `trusted_project_path_patterns` entry of the form `/path/to/project/**` does **not** match
`/path/to/project` itself, so the project root is untrusted — and an untrusted project's
`ls_specific_settings` are then **silently discarded**. The only signal is a log warning that does not
mention which settings were lost or what they controlled.

`/path/to/project/**` is the pattern a user naturally writes to mean *"this project"*.

**Serena 1.7.0**, Python 3.14.5, macOS (Darwin 25.6.0).

### Mechanism

Trust is checked against the project **root**:

- `Project.is_trusted()` → `SerenaConfig.is_trusted_project_path(self.project_root)`
  (`serena/project.py:162-168`, `serena/config/serena_config.py:1431-1442`)

`GlobMatcher._translate_glob_to_regex` renders a trailing `/**` as `/.*`
(`serena/util/text_utils.py:242-289`), and the result is `\Z`-anchored — so the separator is required
and the pattern cannot match its own base directory.

An untrusted project's settings are then dropped in
`Project._create_language_server_manager` (`serena/project.py:519-527`):

```python
ls_specific_settings = dict(self.serena_config.ls_specific_settings)
if self.project_config.ls_specific_settings:
    if self.is_trusted():
        ls_specific_settings.update(self.project_config.ls_specific_settings)
    else:
        log.warning(
            f"Project path {self.project_root} is not trusted, ignoring LS-specific settings from project configuration. ..."
        )
```

### Reproduction

```python
import solidlsp  # import FIRST: serena.util.text_utils and solidlsp.ls import each other,
                 # and importing text_utils directly raises a partially-initialized ImportError
from serena.util.text_utils import GlobMatcher

ROOT = "/Users/example/projects/my-project"

print(GlobMatcher(ROOT + "/**").matches(ROOT))            # False  <-- the bug
print(GlobMatcher(ROOT + "/**").matches(ROOT + "/a.py"))  # True
print(GlobMatcher(ROOT).matches(ROOT))                    # True
print(GlobMatcher(ROOT + "/**").matches("/nope"))         # False  (control: not always-true)
```

The fabricated-root control matters: without it, the `False` on line 1 would not distinguish the bug
from a matcher that never matches anything.

### Real-world impact

[...]

PR #2037, opened 2026-09-15
Docs: show how to trust a single project in trusted_project_path_patterns

ChiFungHillmanChan:
## The issue

#2001. A user who wants to trust one project naturally writes
`trusted_project_path_patterns: ["/path/to/project/**"]`, which trusts nothing — the project stays
untrusted and its `ls_specific_settings` are dropped with a warning that does not say what was lost.
The reporter measured the consequence on a 13-module Maven reactor: 50 files went into the shared
`~/.m2` instead of the per-project repository they had configured.

@opcode81 answered that `/path/to/project/**` correctly means "subdirectories only", and that the
fix is to document the single-project case:

> I believe the best way to address this is to add an additional example for trusting a single project.
> This will resolve the issue for users unfamiliar with glob pattern semantics.

This is that change. It does not alter matching behaviour.

## How it works now, and why the example was not enough

Trust is decided by the project's **root path** and nothing else:

```python
# src/serena/project.py:170
def is_trusted(self) -> bool:
    return self.serena_config.is_trusted_project_path(self.project_root)

# src/serena/config/serena_config.py:1456-1460
project_root_str = str(project_root)
for pattern in self.trusted_project_path_patterns:
    if GlobMatcher(pattern).matches(project_root_str):
        return True
```

`is_trusted()` has two call sites — `src/serena/agent.py:1283` (the project activation command) and
`src/serena/project.py:523` (`ls_specific_settings`) — and both pass the root. So the only string a pattern is
ever tested against is the project root itself, and `<root>/**` cannot match it.

The template's example listed only parent-directory globs:

```yaml
#   trusted_project_path_patterns:
#     - /home/user/projects/**
#     - C:\Users\Anna\Dev\work\**
```

Both entries are of the "everything below this directory" shape, so a reader wanting one project has
nothing to copy and reasonably extends the shape they can see.

## Measured, before

The three cases the example now covers, against `GlobMatcher` on `18fa47bf`:

```
False  does '<project>/**' trust the project root itself?
True     ... does it match a path inside?
True   does a bare '<project>' trust the root?
False    ... and a path inside?
False  control: unrelated path must NOT match
```

The last line is the control: without it, the `False` on line 1 would not distinguish this from a
matcher that never matches anything. (Adapted from the reporter's own reproduction in #2001; the
control is theirs.)

## Measured, after

Behaviour is unchanged — this is a documentation change — so the same five readings hold. The new
tests therefore also pass at the merge base, which is the point: they pin behaviour that is already
correct so the documented example cannot drift away from it. They are not before/after evidence, and
the mutation runs below are what establish that they can fail at all.

`test/serena` minus the seven modules that need a language server, at both commits and on both
interpreters (CI pins 3.11):

| commit | Python 3.11 | Python 3.13 |
|---|---|---|
| `18fa47bf` (merge base) | 580 passed | 580 passed |
| `157da220` (this branch) | 584 passed | 584 passed |

The difference is exactly the 4 tests added. Narrowed to the file they live in,
`test/serena/config/test_serena_config.py` goes 55 → 59.

| check | result |
|---|---|
| `uv run poe lint` (ruff format + ruff check) | green |
| `uv run poe type-check` (ty on `src` and `test`) | green |
| `codespell` on a clean checkout | exit 0, same as merge base |
| `uv sync --extra dev --locked` | resolves; `uv.lock` untouched |
| `uv run serena project health-check test/resources/repos/python/test_repo` | `Health check passed` |
| template still parses | `yaml.safe_load` → `trusted_project_path_patterns: []` |

`ruff format` was run only after confirming `test_serena_config.py` is already conforming on `main`,
so the formatter touched none of anyone else's lines; the test diff is `+34 -0`.

Four mutations of the matching behaviour, each caught:

| mutation | caught by |
|---|---|
| make `/**` also match its own base directory | the trailing-glob test and the parent-glob test |
| trust everything | three of the four, including the control |
| trust nothing | the bare-root test and the parent-glob test |
| strip `/**` so it is treated as a bare root | the trailing-glob and parent-glob tests |

## What the change does

Adds the bare-root entry to the template's example, annotates both forms with what they trust, and
states which path a pattern is matched against:

```yaml
# A pattern is matched against the project's root path itself. To trust one specific project, therefore
# give its root path as-is: appending "/**" to it matches only paths *below* the root and consequently
# matches no project root at all.
# Example:
#   trusted_project_path_patterns:
#     - /opt/dev/projects/my_trusted_project   # trusts exactly this project
#     - /home/user/projects/**                 # trusts every project located below this directory
#     - C:\Users\Anna\Dev\work\**
```

The four tests pin the bare-root, trailing-glob, parent-glob and control cases, so the example cannot
go stale without a test failing.

## What it does not do

- **No behaviour change.** Matching, `GlobMatcher` and `is_trusted_project_path` are untouched; only
  the template comment and tests change. The template still ships `trusted_project_path_patterns: []`.
- **The warning still does not say what was lost.** #2001's second complaint is that
  `src/serena/project.py:526`'s message names neither the discarded settings nor what they controlled. That is a
  separate change and is deliberately not in this PR; happy to open one if it is wanted.
- **Only the template is edited.** `docs/02-usage/050_configuration.md` links to the template rather
  than repeating the example, so there is no second copy to keep in sync — but I have not checked
  the rendered docs site build.
- **`docs/02-usage/070_security.md` is untouched.** It describes what trust means, not how to write
  patterns, so the example does not belong there.


opcode81:
Thanks.

Issue #2029, opened 2026-09-13
Project registry grows unboundedly (autoregister hardcoded, no unregister) and is fully re-validated on every startup, contributing to MCP discovery races


Serg2000Mr:
Serena autoregisters every path passed to `activate_project` into `serena_config.yml → projects:` — including temporary working copies, tool bin-dirs and per-task worktrees. `autoregister=True` is hardcoded in the CLI (`serena/config/cli.py`, `get_registered_project(..., autoregister=True)`), there is no opt-out, no `serena project remove/unregister` command, and stale entries produce warnings on every start.

On every startup, every serena process (hosts typically run one per session) re-reads and validates the `project.yml` of all registered entries: measured ~1.0–1.5 s for 26 entries on Windows, growing linearly. This stacks with the slow-start discovery race reported in #2012: hosts with a fixed tool-discovery window (ZCode ~15 s, measured 15.0–15.1 s) lose serena's tools whenever startup exceeds it; users maintaining dozens of per-task worktrees make the miss a certainty over time.

**Requested:**
1. Lazy registry handling — load only the path list at startup, validate/load each project's `project.yml` on `activate_project` instead of at every server start.
2. A `serena project remove/unregister` command (or automatic cleanup of non-existent paths) — today the only way to clean the registry is hand-editing `serena_config.yml`.
3. An opt-out of autoregister for temporary paths.

opcode81:
> 1. Lazy registry handling — load only the path list at startup

This is not possible, because projects can be referred to by name, and the name is stored in the project's configuration file only.
Referencing a project by name is a core part of Serena's workflow (when not using a single-project context).

> 2. A serena project remove/unregister command (or automatic cleanup of non-existent paths) — today the only way to clean the registry is hand-editing serena_config.yml.

The configuration file is intended to be hand-edited; this is part of the workflow, and we are not currently planning to support alternative ways of configuring Serena in general.
However, if you find value in a remove command, we would accept a PR that adds it.

> 3. An opt-out of autoregister for temporary paths.

I suppose you would want this as an option when launching the MCP server (e.g. with --project-from-cwd), activating the project without adding it to the project list?
This cannot be done without refactoring the project handling. The method that currently creates a non-existing project at startup is `SerenaConfig.add_project_from_path` and it would obviously make no sense to add an option to this method which achieves that the project is not actually added, as it is the creation that is the side-effect here, not the addition.

Serg2000Mr:
[...]

PR #2039, opened 2026-09-15
feat(cli): add 'project remove' to unregister a project

ChiFungHillmanChan:
## The issue

#2029. The reporter's `serena_config.yml` had accumulated `projects:` entries for temporary working
copies, tool bin-dirs and per-task worktrees, and asked for three things. @opcode81 answered each,
declining the first and third and inviting the second:

> However, if you find value in a remove command, we would accept a PR that adds it.

This is that command. It does not change how the registry is loaded or persisted.

## How it works now

The project registry can be **added** to from both sides — `project create` and `project index` on the
CLI, `activate_project` on the MCP side — but it can only be **removed** from on the MCP side, by the
`remove_project` tool (`src/serena/tools/config_tools.py`). The CLI has no counterpart:

```
$ serena project --help          # on main, 704e8c3d
Commands:
  create           Create a new Serena project configuration.
  health-check     Perform a comprehensive health check of the project's...
  index            Index a project by saving symbols to the LSP cache.
  index-file       Index a single file by saving its symbols to the LSP...
  is_ignored_path  Check if a path is ignored by the project configuration.
```

So from a terminal, hand-editing `serena_config.yml` is the only way to drop an entry, which is what
the issue reports.

## What the change adds

`serena project remove <PROJECT_NAME|PROJECT_PATH>`, using the same `PROJECT_TYPE` argument as
`project index`, so a project can be addressed either way. Only the registry entry is removed; the
project's own files, including its `.serena/project.yml`, are left on disk.

Unlike `project index`, the argument is required and does **not** default to `os.getcwd()`. A bare
`serena project remove` in a project directory silently unregistering that project seemed like the
wrong default for the one command here that takes something away.

The persistence machinery is not new — `SerenaConfig._persist_projects` already handles removals, and
`SerenaConfig.remove_project(name)` already exists for the MCP tool. What the command adds alongside
itself is `SerenaConfig.remove_registered_project(entry)`, which mirrors `add_registered_project` and
removes **the given entry** rather than the first entry whose name matches.

That distinction is the one design decision here. Resolving the argument to an entry and then deleting
by name would delete the wrong project whenever two registered directories carry the same
`project_name`, which is possible today and which `test_remove_by_path_picks_the_entry_at_that_path_when_names_collide`
pins. Making the command remove by name is one of the mutations listed below, and that test is what
catches it.

## Tested

New file `test/serena/test_cli_project_remove.py`, 7 tests. They drive the command through
`CliRunner` against a temporary configuration file (`_determine_config_file_path` monkeypatched, the
pattern `TestSerenaConfigLoadSave` already uses), so a test run never touches the user's real
registry.

All 7 fail on `704e8c3d` with `AttributeError: type object 'ProjectCommands' has no attribute 'remove'`
and pass on this branch.

Regression, same command on both:

```
pytest test/serena/test_cli_project_commands.py test/serena/config/test_serena_config.py
  704e8c3d      84 passed

pytest test/serena/test_cli_project_commands.py test/serena/config/test_serena_config.py \
       test/serena/test_cli_project_remove.py
  this branch   91 passed
```

The difference is exactly the seven tests added. The new file also passes on Python 3.11, which is
what CI pins.

Five mutations of the change, each caught:

| # | mutation | result |
|---|---|---|
| 1 | remove by name instead of by entry | 1 failed |
| 2 | drop the `_persist_projects()` call | 3 failed |
| 3 | return quietly instead of failing on an unknown project | 1 failed |
| 4 | let an ambiguous name raise through as a traceback | 1 failed |
| 5 | delete the project's `.serena` directory too | 1 failed |

End to end against a real, isolated Serena home (`SERENA_HOME` pointed at a temp dir), using the real
`serena` entry point rather than `CliRunner`:

```
### 2. register two projects
registry: ['alpha', 'beta']

### 3. remove by NAME
Removed project 'alpha' (/tmp/.../alpha) from the project registry.
registry: ['beta']

### 4. remove by PATH
Removed project 'beta' (/tmp/.../beta) from the project registry.
registry: []

### 5. the project files are still there
/tmp/.../beta/.serena/project.yml
beta project.yml: PRESENT

### 6. the project can be picked up again
-- 'project create' refuses, because project.yml is still on disk (by design):
Error: Project already exists: Project file /tmp/.../beta/.serena/project.yml already exists.
-- but the auto-registration 'project index' and 'start-mcp-server --project' use does re-register it:
auto-registered: True
registry: ['beta']

### 7. unknown project
Error: No registered project found for 'no_such_thing'; registered project names: ['beta']
exit code: 1
registry unchanged: ['beta']
```

Gate: `poe lint` (419 files formatted, all checks passed), `poe type-check` (both `ty` runs pass),
`codespell` clean on the changed files. `src/serena/cli.py` was already formatter-clean on `main`, so
`ruff format` touched only the added lines; the diff has 0 deleted lines in `src/` and `test/`.

## What it does not do

- **It does not add `project list`.** Listing the registry is still done by reading
  `serena_config.yml` (or `serena config edit`). The issue did not ask for one and I would rather
  keep this to the command that was invited.
- **It leaves `SerenaConfig.remove_project(name)` alone.** That is what the `remove_project` MCP tool
  calls, and it resolves by name by design. The new method sits beside it rather than replacing it,
  so no MCP tool signature or docstring changes and the tool surface is unchanged.
- **It does not address the stale-entry part of the issue separately, because that already works.**
  Measured on `main`: a `projects:` entry whose directory no longer exists is skipped at load with a
  warning, and because `_persist_projects` rebuilds the list from a fresh load, the next registry
  write drops it from the file. So removing any one project also flushes the dangling entries. This
  PR does not change that behaviour and does not add a test pinning it.
- **It does not touch autoregister**, which is the issue's third request and the one @opcode81 said
  would need project-handling to be refactored.

## Checklist

Refs #2029


opcode81:
Thanks for your contribution.

PR #2040, opened 2026-09-15
fix(dashboard): marshal tray menu updates onto the macOS main thread

### LINKED ISSUE #2038: macOS: tray_manager (the macOS default) crashes with SIGTRAP — NSStatusItem.setMenu_() called off the main thread from Flask/alive-check threads
## Summary

On macOS, the default dashboard interface (`tray_manager`) crashes the tray-manager process with
`EXC_BREAKPOINT` / `SIGTRAP` within ~2s of every Serena agent start. The crash happens *inside* the
`/register` request handler, so the agent logs:

```
WARNING serena.dashboard:register_instance:1312 - Failed to register with tray manager: Remote end closed connection without response
```

The tray icon never becomes usable, and every Serena MCP start leaves a `Python-*.ips` crash report
in `~/Library/Logs/DiagnosticReports/`.

## Root cause

`SerenaDashboardTrayManager._update_menu()` calls `pystray.Icon.update_menu()` from threads that are
not the main thread:

* the Flask request handlers for `/register`, `/update_project`, `/unregister` (`dashboard.py`, `_setup_routes`)
* `_alive_check_loop()`, which runs on its own `threading.Thread`

pystray's macOS backend (`pystray/_darwin.py`) implements that synchronously on the calling thread:

```python
def _update_menu(self):
    callbacks = []
    nsmenu = self._create_menu(self.menu, callbacks)
    if nsmenu:
        self._status_item.setMenu_(nsmenu)   # <-- AppKit, main-thread-only
```

`-[NSStatusItem setMenu:]` is main-thread-only. On macOS 26/27 AppKit now enforces this with a hard
barrier assertion instead of tolerating it, so the process is killed outright.

Faulting thread backtrace from the crash report:

```
BoardServices        -[BSServiceMainRunLoopQueue assertBarrierOnQueue]
FrontBoardServices   -[FBSScene _updateClientSettings:]
FrontBoardServices   -[FBSScene updateClientSettingsWithBlock:]
AppKit               -[NSSceneStatusItem _updateSupportsExpandedInterfaceSession]
AppKit               -[NSStatusBarButtonCell setStatusMenu:]
AppKit               -[NSStatusBarButton setStatusMenu:]
AppKit               -[NSStatusItem setMenu:]
libffi.dylib         ffi_call_int
_objc.cpython-313-darwin.so  PyObjCFFI_Caller_SimpleSEL
Python               _PyEval_EvalFrameDefault
Python               thread_run                      <-- not the main thread
libsystem_pthread.dylib      _pthread_start
```

`exception: {"type":"EXC_BREAKPOINT","signal":"SIGTRAP"}`, `termination: Trace/BPT trap: 5`.

## Minimal reproduction

Independent of Serena — this is ~20 lines of pystray:

```python
import threading, time
from PIL import Image
import pystray
from AppKit import NSApplication, NSApplicationActivationPolicyAccessory

img = Image.new("RGBA", (22, 22), (255, 0, 0, 255))
nsapp = NSApplication.sharedApplication()
nsapp.setActivationPolicy_(NSApplicationActivationPolicyAccessory)

items = [pystray.MenuItem("one", None, enabled=False)]
icon = pystray.Icon("repro", img, "repro", menu=pystray.Menu(lambda: tuple(items)),
                    darwin_nsapplication=nsapp)

def poke():
    time.sleep(2)
    print("BG THREAD: calling update_menu()", flush=True)
    icon.update_menu()          # exactly what _update_menu() does from the Flask thread
    print("BG THREAD: survived", flush=True)
    icon.stop()

threading.Thread(target=poke, daemon=True).start()
icon.run()
```

Result: `BG THREAD: survived` never prints; process exits **133** (128 + SIGTRAP).

## Verified fix

Marshaling the call onto the main thread makes it exit cleanly (0):

```python
from Foundation import NSObject

class _MainThreadHelper(NSObject):
    def run_(self, _arg):
        icon.update_menu()

helper = _MainThreadHelper.alloc().init()
# from any background thread:
helper.performSelectorOnMainThread_withObject_waitUntilDone_("run:", None, True)
```

Suggested change: have `SerenaDashboardTrayManager._update_menu()` marshal to the main thread on
Darwin (or fix it in `pystray._darwin.Icon._update_menu`, which is arguably where it belongs —
`update_menu()` is documented as callable at any time and gives no thread-safety caveat).

Note `_open_dashboard`/`_run_viewer` are reached from the same Flask threads, so any other AppKit
work on that path is likely to need the same treatment.

## Workaround

Setting `web_dashboard_interface: browser` in `~/.serena/serena_config.yml` avoids the tray manager
entirely and stops the crashes. As a side effect it also removed ~1.2s from MCP startup here
(1.61s of tray init → 0.37s), since the tray was being spawned and killed on every launch.

## Environment

| | |
|---|---|
| serena-agent | 1.7.0 (current release) |
| pystray | 0.19.5 (current release) |
| Python | 3.13.15 (Homebrew framework build) |
| macOS | 27.0 (26A428) |
| Hardware | Mac16,12, arm64 |
| Client | Claude Code, stdio MCP |

`web_dashboard_interface` was unset, so `DashboardManager.Mode.from_platform()` selected
`TRAY_MANAGER` — which the shipped config comments describe as "EXPERIMENTAL … tested on Windows
only … yet untested" on macOS. Since it is the macOS *default*, every macOS user on 26/27 should be
hitting this.

Possibly related: #1825 (macOS tray, stranded instances).

MannXo:
Fixes #2038.

## Root cause

`SerenaDashboardTrayManager._update_menu()` called `pystray.Icon.update_menu()` directly from whatever thread happened to be running. Four call sites reach it off the main thread: the Flask request handlers for `/register`, `/update_project` and `/unregister`, and `_alive_check_loop`.

pystray does no marshalling of its own. `Icon.update_menu()` in `pystray/_base.py` is a one-line call to `self._update_menu()`, and the Darwin backend's `_update_menu` goes straight to AppKit:

```python
# pystray/_darwin.py
def _update_menu(self):
    callbacks = []
    nsmenu = self._create_menu(self.menu, callbacks)
    if nsmenu:
        self._status_item.setMenu_(nsmenu)
```

`NSStatusItem.setMenu_()` is main-thread-only. On macOS versions that enforce it, the process traps with SIGTRAP inside the request handler, which is why the agent logs `Failed to register with tray manager: Remote end closed connection without response` and the tray icon never becomes usable.

## Fix

Menu refreshes are marshalled onto the main thread via `PyObjCTools.AppHelper.callAfter`, which is PyObjC's supported API for this and is asynchronous, so no Flask handler blocks on the main run loop. Non-Darwin platforms keep calling straight through.

The helper is separate from `_update_menu` because `_open_dashboard` and `_run_viewer` are reached from the same Flask threads, so there is one obvious home for the next call that needs it.

No new dependency. `PyObjCTools` comes from `pyobjc-core`, which is already transitively required on macOS through `pystray` -> `pyobjc-framework-Quartz` -> `pyobjc-core`, and `dashboard.py` already imports `AppKit` in the Darwin branch of `run()`.

## What I verified, and what I could not

**I could not reproduce the crash.** I am on macOS 26.6.2 and the reporter is on 27.0. Everything else in our environments matches, pystray 0.19.5 and Python 3.13 and arm64 included. I ran the reporter's own minimal repro and a second one driving the real `SerenaDashboardTrayManager` through its HTTP routes. Both exit 0 here, `/register` returns 200, and no crash report appears in `~/Library/Logs/DiagnosticReports/`. The reporter gets exit 133 on 27.0. So the enforcement that turns this into a hard trap appears to land in macOS 27, and I cannot demonstrate the crash or its absence on this machine.

**What I did verify** is that the marshalling does what it claims. I instrumented `update_menu` to record `NSThread.isMainThread()` and drove all three HTTP routes against a live tray icon:

```
before the fix:  update_menu invocations: 2   isMainThread: [False, False]
after the fix:   update_menu invocations: 3   isMainThread: [True, True, True]
```

So on `main` the AppKit call genuinely happens on Flask worker threads, and with this change it genuinely happens on the main thread. That is the property the fix is for, independent of which macOS version chooses to punish the violation.

@Baltimoreteacher1, since you are on 27.0 and can reproduce, would you confirm this branch clears it on your machine?

## Not changed

`_alive_check_loop` also calls `self._tray_icon.stop()`, and `_open_dashboard` runs on the same Flask threads. Both are the same class of AppKit-off-the-main-thread as the reported bug. I left them alone to keep the PR to one logical change and because I cannot test the crash path, so widening it would mean shipping more untested AppKit changes. Happy to follow up on either if you want them in scope.

## Checklist


opcode81:
Thanks a lot for your contribution. Please accept the CLA, then we can merge this.

MannXo:
> Thanks a lot for your contribution. Please accept the CLA, then we can merge this.

thank you.done

PR #2041, opened 2026-09-15
fix(tools): save edited source files atomically

### LINKED ISSUE #1958: Protect memory and source files from partial writes

ChiFungHillmanChan:
## The issue

#1958 lists three direct `open(path, "w")` writes that can destroy a file if the process stops
partway through, because that call truncates the file before the new content is complete:

- `MemoryManager.save_memory()`
- `MemoryManager.edit_memory()`
- `CodeEditor._save_edited_file()`

The first two were fixed in merged #1969, which is where `write_file_atomic` comes from. The
reporter proposed the rest as a follow-up in the issue itself:

> Source-file writes can use it after tests confirm that executable files, symlinks, special file
> permissions and metadata, and language-server notifications still behave correctly. If that makes
> one change too broad, memory writes can be the first PR and `CodeEditor` a follow-up.

This is that follow-up. The analysis of the problem is theirs; what this adds is the measurement of
the three conditions they attached to it.

## How it works now, and why that is wrong

`src/serena/code_editor.py:92`, reached by every editing tool through `edited_file_context()`:

```python
def _save_edited_file(self, edited_file: "CodeEditor.EditedFile") -> None:
    abs_path = os.path.join(self.project_root, edited_file.relative_path)
    new_contents = edited_file.get_contents()
    with open(abs_path, "w", encoding=self.encoding, newline=self.newline) as f:
        f.write(new_contents)
```

`open(abs_path, "w")` empties the file first. If Serena is killed, runs out of memory, or the disk
fills between the truncate and the final byte, the user's source file holds neither its old content
nor the new one. `JetBrainsCodeEditor._save_edited_file` calls `super()` before refreshing the file
in the IDE, so both backends go through this one line.

## Checking the three conditions the issue attached

Measured on `704e8c3d` before changing anything, comparing the current call against a bare
`write_file_atomic`:

| | `open(path, "w")` (today) | `write_file_atomic` as merged |
|---|---|---|
| symlink survives the write | **yes** | **no** — replaced by a regular file |
| symlink's target gets the new content | **yes** | **no** — keeps the old content |
| executable bit (`0o755`) | preserved | preserved |

So the executable-bit and permission condition was already met — `write_file_atomic` restores the
original mode after `mkstemp`. **The symlink condition was not.** `os.replace` onto the link path
replaces the link itself, so an edit to a symlinked source file would have silently stopped reaching
the file the link points at. That is a quieter failure than the one being fixed, so it had to be
closed before the call site could change.

On the third condition, language-server notifications: Serena tells the server about a change by
path and content (`did_change_text_document`, and `did_change_watched_files` in
`ls_manager.py:351-361`), never by file descriptor or inode, so swapping the inode underneath is not
visible to that path.

## What the fix does

1. `write_file_atomic` resolves the path first, so it follows symlinks exactly as the `open()` call
   it replaces did, and the temporary file is created in the destination's **real** directory, which
   is where it has to be for the rename to stay atomic. This also removes the same latent hazard from
   the memory writes that already use the helper.
2. `CodeEditor._save_edited_file` calls it instead of writing in place.

## Tested

Two-state, same command both times:

```
pytest test/serena/test_code_editor_atomic_writes.py test/serena/util/test_file_system.py
  704e8c3d src + these tests   4 failed, 43 passed
  this branch                  47 passed
```

(The base reading was taken by restoring only `src/serena/code_editor.py` and
`src/serena/util/file_system.py` from `704e8c3d`, confirmed by grepping that neither `write_file_atomic`
nor the path resolution was present, then restoring them.)

The four failures on base are the ones that describe the defect: the interrupted-save test and the
three symlink tests. The other nine new tests **pass on both** — they are the guard tests for the
conditions above (exec bit, encoding, newline, plain-file control), so the fix is not bought at their
expense.

The interrupted-save test injects a crash into the temp-file write. On a non-atomic implementation
nothing raises at all, so on base it fails with `DID NOT RAISE` — which is precisely the issue's
point: there is no intermediate copy, the real file was already truncated.

Regression, wider set:

```
pytest test/serena/util/test_file_system.py test/serena/test_file_tools.py \
       test/serena/test_edit_marker.py test/serena/test_ls_file_sync.py \
       test/serena/test_symbol_editing.py [+ test/serena/test_code_editor_atomic_writes.py]
  704e8c3d      1 failed, 79 passed, 1 skipped
  this branch   1 failed, 92 passed, 1 skipped
```

The extra passes are exactly the tests added (seven in the new file, six appended to
`test_file_system.py`). The one failure is the same in both:
`test_symbol_editing.py::test_go_symbol_replacement_no_double_keyword`, which needs `gopls`; it is
not installed on this machine and the test fails identically on `704e8c3d`.

Five mutations of the change, each caught by the test written for it:

| # | mutation | caught by |
|---|---|---|
| 1 | put the plain `open(path, "w")` back | interrupted-save test |
| 2 | drop the symlink resolution | all 4 symlink tests |
| 3 | drop `newline=self.newline` from the call | newline test |
| 4 | hardcode `encoding="utf-8"` | encoding test |
| 5 | resolve the path only when it is *not* a link | all 4 symlink tests |
| 6 | `os.path.abspath` instead of `os.path.realpath` | all 4 symlink tests |

Gate: `poe lint` (419 files formatted, all checks passed), `poe type-check` (both `ty` runs pass),
`codespell` clean on the changed files. Python 3.13 and 3.11.

The test diff is a pure insertion: `@@ -0,0 +1,166 @@` for the new file and `@@ -920,3 +920,61 @@`
for the appended class, zero deleted lines under `test/`. The only two deleted lines in the whole
diff are the `open(...)`/`f.write(...)` pair being replaced.

Note `src/serena/util/file_system.py` is stored with CRLF line endings; the added lines match, so the
diff is 5 added lines and nothing else.

## What it does not do

- **It does not touch the two `MemoryManager` call sites**, which already use the helper. They do
  inherit the symlink fix, which is a behaviour change for a symlinked memory file: it is now written
  through to its target instead of being replaced. That direction matches what those call sites did
  before #1969.
- **It does not change `create_text_file`.** `CreateTextFileTool.apply`
  (`src/serena/tools/file_tools.py:87`) writes with `Path.write_text()`, which truncates the same
  way, and it overwrites existing files, so it has the same exposure. It is not one of the three
  sites #1958 lists, so I left it out rather than widen the scope on my own; say the word and I will
  fold it in here or send it separately. (It calls `.resolve()` on the path first, so it is already
  symlink-correct.)
- **It does not add a `fsync`.** `os.replace` gives atomicity against a process dying; surviving a
  power loss would additionally need the temp file and its directory flushed. That is a separate
  trade-off and the merged helper did not make it either.
- **It does not coordinate concurrent writers.** Per #1637 that is out of scope: the last complete
  write still wins. This only prevents an interrupted write from leaving a partial file.
- **Two of these tests were wrong on Windows in the first push, and are fixed in the second commit.**
  `catch-all (windows-latest)` failed on `98bc05aa`: the executable-bit test read back `0o666` where
  it asserted `0o755` (Windows does not model POSIX permission bits, `chmod` there only toggles the
  read-only flag), and the encoding test got `b"caf\xe9\r\n"` because it left `newline` unset, which
  is `LineEnding.NATIVE` and so `newline=None`, under which Python translates `"\n"` to `os.linesep`.
  Neither was a problem in the change itself: the sibling test that pins `"\r\n"` explicitly passed
  on that same Windows job. The first is now skipped on win32 and the second pins its line ending.
- **Tests skip where the platform cannot support them rather than failing.** The symlink tests use the
  same idiom `test_memories_manager.py:375` already uses, and the non-ASCII filename test guards the
  same way. Verified by forcing each to raise: `42 passed, 5 skipped`, nothing failing. I have no
  Windows machine, so those are simulations of the condition, not runs on Windows.

## Checklist

Resolves #1958


opcode81:
Thanks for your contribution.

PR #2042, opened 2026-09-15
perf(search_text): precompute line offsets for O(log n) line lookup

davalillo:
## Checklist

## The problem

`search_for_pattern` resolves each match's line number through `TextUtils.get_line_from_index`, which walks a `TextStepper` from index 0 up to the given offset. For a file with `n` lines and `m` matches, that is O(n) per match and **O(n·m) total**. On a large file with many matches, match-coordinate resolution dominates the tool time by a wide margin — and `search_for_pattern` is one of the most frequently invoked tools in agent sessions.

Reproduction (synthetic file, generated with a fixed seed — 12,000 lines, ~450 KB, ~1830 matches of a string pattern):

```
uv run python scripts/profile_search_text.py
# resolve line numbers for all matches, best of 5
before: ~4450 ms
after:  ~2.4 ms   (~1800x)
```

The benchmark script is included in this PR (`scripts/profile_search_text.py`): it regenerates the
same synthetic content deterministically (no real-world code), asserts that both resolution paths
produce identical results, and prints the timings above. Optional CLI args (`num_lines`) scale it
down for a quick run (1000 lines: 33 ms → 0.19 ms).

The one-off precompute required by the fix costs ~1.1 ms, and small files do not regress (100 lines, 5 matches: 70 µs → 11 µs).

## The fix

`search_text` precomputes the character offset at which each line starts in a single O(n) pass (`_compute_line_starts`), then resolves each match's start/end coordinates via binary search (`bisect_right`, O(log n) per match) through a new `_line_col_at_index` helper.

The table mirrors `TextStepper` line semantics exactly — `\n`, `\r\n` and bare `\r` are all line separators, and one subtlety is pinned explicitly: an index pointing at the `\n` of a `\r\n` pair resolves to **(next line, 0)**, the same rule `TextUtils.get_line_col_from_index` implements. Equivalence with `TextUtils.get_line_col_from_index` was verified over 416 random and edge-case contents (mixed line endings, empty content, boundary indices) with zero mismatches; the search results are byte-for-byte identical to the previous implementation.

## Tests

13 new tests in `test/serena/test_text_utils.py`:

- a parametrized equivalence test pinning `_line_col_at_index` against `TextUtils.get_line_col_from_index` across `\n` / `\r\n` / bare `\r` / mixed / empty contents
- CRLF, bare-`\r` and boundary (first/last line) line-number tests through the public `search_text` API
- a multiline-match range test (`start_line`/`end_line` of a multi-line match)

## Compatibility with #1899

[PR #1899](https://github.com/oraios/serena/pull/1899) (interruptible agent regexes) touches the same file but a different region: it routes matching through `compile_pattern` at the `re.compile` call site in `search_text` and in `ContentReplacer`. This PR only touches the post-match line-coordinate resolution that follows. The two changes are complementary and rebase cleanly onto each other; whichever merges first, the other rebases without conflict beyond the import block.

## Benchmark methodology

Synthetic content generated programmatically with a fixed seed (no proprietary code): 12,000 lines of pseudo-code, ~450 KB, pattern matching ~1830 times. Timing is best-of-5 with `time.perf_counter`, regex excluded from both paths (identical in both), measuring only the line-coordinate resolution that differs between the two implementations. The exact generator and measurement code ship in `scripts/profile_search_text.py`.


src/serena/util/text_utils.py
```diff
@@ -150,6 +151,12 @@ def search_text(
     lines = TextUtils.split_lines(content)
     total_lines = len(lines)
 
+    # Precompute line start offsets once (O(n)) so that each match's line number
+    # can be resolved via binary search (O(log n)) instead of re-scanning the
+    # text from the beginning with TextStepper (O(n) per call).
+    # The semantics mirror TextStepper: \n, \r\n, and \r are all line separators.
+    line_starts = _compute_line_starts(content)
```
opcode81: Turn this into an abstraction for computing text coordinates based on cached line start coordinates.

Make sure your agent is reading the software design principles for this project in the `critical_info` memory.

src/serena/util/text_utils.py
```diff
@@ -187,6 +194,66 @@ def search_text(
     return matches
 
 
+def _compute_line_starts(content: str) -> list[int]:
```
opcode81: This function is unnecessarily duplicating logic which is already present in `TextStepper`. Use `TextStepper` to compute the line start indices via `step_line` and `prev_line_start_idx`/`line_start_idx`.

src/serena/util/text_utils.py
```diff
@@ -187,6 +194,66 @@ def search_text(
...
+            break
+        next_start = cr + 2 if cr + 1 < n and content[cr + 1] == "\n" else cr + 1
+        if next_start > n:
+            break
+        starts.append(next_start)
+        last = next_start
+    return starts
+
+
+def _line_col_at_index(content: str, index: int, line_starts: list[int]) -> tuple[int, int]:
```
opcode81: This function becomes a public method of the abstraction to be introduced.

opcode81 (changes requested):

src/serena/util/text_utils.py
davalillo: Well, to be honest, I hadn’t taken those restrictions into account. I’m going to check all the other PRs as well to verify the generated code.

Thanks for letting me know.

davalillo: Done — the rework introduces an encapsulated `TextCoordinates` abstraction in `solidlsp/ls_utils.py`. It privately caches the table of line start offsets and exposes a single public method, `line_col_at_index(index) -> LineCol`. The module-level helpers are gone, and `search_text` resolves match coordinates exclusively through that public API.

One design note: I deliberately returned a `LineCol` dataclass rather than a tuple, since the `critical_info` memory explicitly calls for dataclasses instead of tuples for simple data storage. I'm aware line/column coordinates are otherwise represented as tuples throughout the codebase (e.g. `TextUtils.get_line_col_from_index`), so if you'd rather keep that convention here for consistency, I'm happy to switch — your call.

The tests were reworked per the testing principle as well: the parametrized test importing the private helpers is removed; coverage is now behaviour-anchored (public-contract tests of `TextCoordinates` in `test/solidlsp/test_ls_utils.py`, plus absolute line-number assertions through the public `search_text` API). Equivalence with `TextUtils.get_line_col_from_index` was verified across LF/CRLF/bare-CR/mixed/empty contents with zero mismatches.

FYI: while working on this, I noticed `msl_lsp_server._get_line_col` recomputes coordinates via `text[:pos].split("\n")` per call (O(n) inside loops over many call sites), with `\r` semantics that differ from `TextStepper`. It looks like a natural follow-up adopter of `TextCoordinates` — happy to open an issue or a separate PR if you prefer.

src/serena/util/text_utils.py
davalillo: Fixed — `TextCoordinates` builds its line start table in a single pass via `TextStepper` (`step_line()` / `line_start_idx`); no line-ending logic is reimplemented anywhere, and the helper functions that duplicated it are removed.

src/serena/util/text_utils.py
davalillo: Done — the lookup is now the public method `TextCoordinates.line_col_at_index`, returning a `LineCol` dataclass instead of a raw tuple.

opcode81:
Thanks for your contribution.

PR #2043, opened 2026-09-15
perf(ls): warm file buffer cache to avoid redundant didOpen/didClose cycles

davalillo:
## Checklist

## The problem

Serena's symbolic editing tools (e.g. `replace_symbol_body`, `insert_after_symbol`) go through `EditedFileContext`, which wraps each operation in `SolidLanguageServer.open_file()`. Every operation therefore performs a full open/close cycle against the language server:

1. `didOpen` — sends the **entire file content** over stdin and forces the server to create the document from scratch and parse it in full
2. `didClose` — discards the document the server just built

For a sequence of operations on the same file (the common case: several symbolic edits in one editing session), the document the server just parsed is thrown away only to be rebuilt milliseconds later. Each symbolic edit on a large file re-transfers and re-parses the whole file.

Reproducible measurement (`scripts/profile_warm_buffers.py`, synthetic content with a fixed seed, mocked LSP notifications counting both notification types and bytes sent):

```
uv run python scripts/profile_warm_buffers.py 20 5000

workload: 20 sequential edits on the same 5000-line file (242,224 bytes per full read)

notification counts (fewer didOpen/didClose is better):
  upstream (no warm cache): didOpen=20, didChange=0, didClose=19
  warm file buffer cache:   didOpen=1,  didChange=10, didClose=0

bytes sent to the LS (didOpen+didChange payloads): upstream=4,845,605  warm=2,665,084
bytes avoided: 2,180,521 (45% reduction)
```

## The fix

`open_file()` now moves closed buffers into a **warm cache** (30 s TTL) instead of closing them immediately. The LSP document stays open on the server side, so a rapid reopen within the TTL:

- **reuses** the existing buffer — `ensure_open_in_ls()` sends `didChange` only if the file's mtime changed on disk, and **nothing at all** if it did not
- **evicts lazily**: expired warm buffers are closed (deferred `didClose`) on the next `open_file()` call
- **shuts down cleanly**: `stop()` closes all warm buffers so no stale documents remain open on the server side
- **stays correct under external edits**: `invalidate_warm_buffer()` (wired into the `workspace/didChangeWatchedFiles` handling in `LanguageServerFileChangeNotifier`) evicts the buffer so the next open sends a fresh full `didOpen` with current content

One implementation subtlety worth flagging for review: lazy eviction runs **before** the reuse decision in `open_file()`. Evicting after the reuse branch would silently reuse an expired buffer and never send its deferred `didClose` — the notification-counting tests catch this ordering bug.

## Behavioral change (one observable contract)

`open_file()`'s context manager no longer sends `didClose` on exit: the close is deferred to TTL expiry. `test_rename_didopen.py` pinned the old contract (`["didOpen", "rename", "didClose"]`); the pin now reflects the new behavior (`["didOpen", "rename"]` + buffer retained warm). Everything else is unchanged: buffers are still updated via `didChange` when the file changed on disk, and `didOpen` is still sent whenever there is no warm buffer (TTL expired, external invalidation, first open, or server restart).

## Tests

5 new tests in `test/solidlsp/test_warm_file_buffers.py` (pattern follows the existing `test_rename_didopen.py` mock style):

- reopen within TTL sends **no** second `didOpen` and no `didClose`; TTL expiry triggers the deferred close on the next open
- reopen after a disk change sends `didChange`, not a second `didOpen`
- `invalidate_warm_buffer()` evicts and forces a fresh full cycle
- `stop()` closes every warm buffer
- lazy eviction ordering (the bug described above, pinned as a regression test)

## Note on #1528

PR #1528 (parallel `project index`) restructures the same `open_file()` method around an RLock. This PR is based on the current `main` shape of `open_file()` and applies cleanly on top of it; if #1528 merges first, I will rebase this PR onto the restructured version (the warm-cache logic composes with the RLock bookkeeping: eviction/reuse decisions move inside the locked section, I/O stays outside).

## Benchmark methodology

Fully synthetic (no real-world code): a seeded generator produces pseudo-code content; a mocked `server.notify` counts `didOpen`/`didChange`/`didClose` notifications and accumulates the text payload bytes each carries. The workload alternates symbolic edits (mtime changes → `didChange` path) with pure reopens (no notification at all). The upstream baseline is modeled with `TTL=0`.


opcode81:
Which actual problem does this solve? You don't mention any problems being caused by the (repeated) opening behaviour.
Such optimisations come at the expense of complexity, and we should optimise only when there is an actual need to.

davalillo:
I’m working on a branch with an MQL4/5 LSP. The thing is, I’m using it to refactor code files containing over 20k lines. I’ve experienced crashes and timeouts, so I’m currently looking into the errors I’m encountering. I’m going to try to put together a test using a large amount of MQL code so I can show you the problem I’m trying to solve.

davalillo:
Fair point, and I agree. I spent more time on this and it just kept getting messier: the cache interacts badly with how the different language servers handle document lifecycle, and every fix I made for one case broke another. That's a lot of complexity for an optimization I haven't actually proven is needed.

I'm going to close this PR and go back to the real problem I mentioned, the big MQL files with the crashes and timeouts. If profiling shows the repeated open/close cycling is actually the bottleneck there, I'll come back with real numbers and a much narrower proposal.

One small thing I'd still like to keep, unrelated to the cache: _open_in_ls() can send a didChange with a version that doesn't increase, and servers that validate document versions will silently drop the update. It's a one-line fix, so I can open a tiny PR for it if you think it's worth it.

PR #2044, opened 2026-09-15
fix(serena): non-blocking stderr handler to prevent logging deadlock in MCP servers

davalillo:
## Checklist

## How this was found (real usage, not a code-reading exercise)

This fix comes from a production failure during a real, extended agent session. The setup: Serena running locally as an MCP server in stdio mode (`start-mcp-server --context ide --project-from-cwd`), launched by **opencode** as the MCP client host, working for hours on a large real codebase (a >11k-line source file with hundreds of tool calls in the session).

Mid-session, the agent received:

```
ERROR ... serena.tools.tools_base:apply_ex:432 - Tool execution timed out after 240 seconds.
```

The confusing part — and what made us dig: the log showed the tool (`search_for_pattern`) had **completed successfully in 19 ms** and logged its full result. The result just never reached the client. The timestamp arithmetic gave the first clue: the timeout fired **exactly 240.000 s** (tool_timeout) after the request, meaning the client had waited on something that never returned.

Further forensics on the still-hung live process (`/proc/<pid>` — no debugger needed):

| Observation | Meaning |
|---|---|
| log file on disk: mtime frozen at the exact second the tool completed | the file-handler pipeline stopped receiving records mid-session |
| one thread blocked in `sock_alloc_send_pskb` on **fd 2** — and fd 2 is a **socketpair** (stderr), its peer owned by nobody | the stderr buffer (64 KiB) was full and **nobody was draining it** |
| other threads in `futex_wait` | threads waiting on a lock held by the blocked writer |
| CPU deltas of 0 ticks across all threads over 3 s | a true deadlock, not slow work |
| ~50 TCP connections to the dashboard port in `CLOSE_WAIT` | the same stalled logging pipeline, accumulating dead connections |

**Root cause chain:** opencode (like some other MCP client hosts) never reads the spawned server process' stderr. Serena's `start-mcp-server` installs a plain blocking `logging.StreamHandler(stream=sys.stderr)`. After ~64 KiB of accumulated log records, every `emit()` blocks inside `write()` — **while holding the logging module's global lock**. The next logging call from any other thread (in this incident, the post-completion `save_all_caches()` inside the task, right after the successful result was logged) hangs forever. The main thread waits on the task's future → `tool_timeout` at exactly `tool_timeout` seconds.

**Why this matters beyond this incident:** it is a time bomb for every MCP user whose client host doesn't drain stderr. The server doesn't fail at startup — it freezes later, after enough log volume, in a way that looks like a "tool timed out" (misleading: the tool finished).

## The fix

`NonBlockingStderrHandler` (new file, used by both `start-mcp-server` and `project-server`):

- the **write strategy is determined exactly once at construction time** via a small private `_RecordWriter` hierarchy (stream-without-fileno / plain fd / non-blocking fd), so `emit` never branches on a nullable fd
- on POSIX, stderr pipes/socketpairs are made **non-blocking** (`fstat` detects them, `O_NONBLOCK` via `os.set_blocking`); when the buffer is full (`BlockingIOError`) or only part of the message fits, the record is **dropped** instead of blocking the logging lock
- a tty or regular file is written normally (no bounded buffer a non-draining consumer could fill)
- the authoritative streams stay lossless: the log file (`mcp_*.txt`) and the dashboard's in-memory buffer continue to be synchronous and complete
- **no background writer thread** — deliberately. A first implementation used a queue + daemon writer (the same pattern as `MemoryLogHandler`), and the tests caught it: a daemon thread stuck in a blocking `write()` deadlocks CPython's interpreter shutdown (`PyThreadState_Clear` waits for the thread's frame). The process would hang at exit instead of mid-session — trading one hang for another. Writing synchronously on a non-blocking fd has no such trap.

### Windows limitation (explicit, not hidden)

On Windows, anonymous pipes are neither reported as FIFOs by `fstat` nor can they be made non-blocking (`os.set_blocking` does not exist), so stderr writes there remain blocking, exactly as with the plain `StreamHandler` — the deadlock scenario this PR fixes is *possible* on Windows. This is documented in the handler's docstring and the CHANGELOG. Practical mitigation: the dominant Windows MCP client hosts redirect the server's stderr to a log file (an unbounded sink). Details and the alternative we considered and rejected (`SetNamedPipeHandleState(PIPE_NOWAIT)` via ctypes — works, but Microsoft documents it as deprecated) are in [this PR comment](https://github.com/oraios/serena/pull/2044#issuecomment-5697068739); if the maintainers prefer full Windows coverage, we're happy to add the ~15-line Windows branch.

## Tests

4 behaviour-anchored tests in `test/serena/test_non_blocking_stderr_handler.py`:

- **survival under an undrained stderr** (subprocess, stderr connected to a raw `os.pipe()` whose read end the parent never reads — exactly the client-host scenario): a workload that floods stderr after its work is done completes and the process exits 0; with the old blocking handler this hangs and the parent must kill it. Success is reported via a file marker, so the harness needs neither `os.set_blocking` nor stdout polling (both unavailable/fragile on Windows)
- **full-pipe drop contract**: 5000 records emitted against a permanently full pipe all return from `emit()` without blocking
- ordered delivery when the pipe is drained; plain writes for streams without a usable fileno (e.g. `StringIO`)

Survival and full-pipe tests are `skipif(win32)` (the mechanism cannot exist there, per the limitation above); the delivery tests run on all platforms. Per the repo's testing principles, tests assert only externally observable behaviour — the earlier iteration's race-signature test of the *pre-fix* deadlock and the asserts on the private non-blocking flag were removed (the race test failed nondeterministically on macOS CI; behaviour-preserving refactorings must not be able to break tests).

## Scope

`src/serena/cli.py` (2 call sites), the new handler, tests, CHANGELOG. No behavior change other than: on POSIX, stderr becomes best-effort under a non-draining consumer. Normal usage (terminal, or a client that drains stderr) is byte-identical to before.

## CI

All workflow jobs green on the current head, with one known exception: the `niche (ubuntu)` job fails in the 2 pre-existing R language-server tests (`test_find_definition_across_files`, `test_find_references_across_files` returning empty results) — these fail identically on `main` (see run 34983624289 of an unrelated docs-only PR), so they are unrelated to this change.


davalillo:
## Windows casuistics (from `f6ef7183`)

The non-blocking mechanism **does not exist on Windows**, and we deliberately did not emulate it with deprecated APIs. Details for the maintainers to weigh in:

**Why Windows cannot get the same mechanism:**

1. `os.set_blocking` does not exist on Windows (`AttributeError`) — this is what broke the first iteration's tests there.
2. `fstat` on Windows does not report anonymous pipes as FIFOs (`stat.S_ISFIFO` is false), so the handler cannot even detect the "bounded buffer" case via `st_mode`.
3. There is no POSIX `O_NONBLOCK` equivalent for anonymous pipes on Windows.

**Consequence:** on Windows the `NonBlockingFdRecordWriter` degrades to plain blocking writes — identical to a plain `StreamHandler`, i.e. the deadlock scenario this PR fixes remains *possible* on Windows. This is documented explicitly in the handler's docstring and the CHANGELOG rather than hidden behind a silent fallback. Practical mitigation: the dominant Windows MCP client hosts redirect the server's stderr to a log file (an unbounded sink), so the real-world exposure there is much lower than on POSIX.

**Alternative we considered and rejected — for the maintainers to decide:** calling `SetNamedPipeHandleState(handle, PIPE_NOWAIT)` via `ctypes` on the *write* end does work for anonymous pipes (when the buffer is full, `WriteFile` fails immediately with `ERROR_NO_DATA` → Python raises `OSError` → the handler's existing catch drops the record), which would honor the non-blocking contract on Windows too and let the full-pipe tests run unskipped in the Windows CI job. We did **not** include it because `PIPE_NOWAIT` is [documented as deprecated](https://learn.microsoft.com/en-us/windows/win32/api/namedpipeapi/nf-namedpipeapi-setnamedpipehandlestate) and we don't want to build a bug-fix on deprecated functionality. If you prefer full Windows coverage, we're happy to add a Windows branch to `_create_writer` (~15 lines of ctypes) — say the word.

**Test changes in this push** (aligned with the repo's testing principles and fixing the other CI failures):

- dropped the "documents the pre-fix deadlock" test: its exact race signature (`SURVIVED` in stdout **and** `rc == -9`) depends on thread scheduling and failed nondeterministically on macOS — the deterministic survival test (exit 0 + marker written) is what remains
- dropped the asserts on the private `_nonblocking_fd` flag (implementation-coupled); the full-pipe test verifies the observable guarantee (never blocks) instead
- the subprocess harness now reports success via a file marker with `stdout=DEVNULL`, so it no longer needs `os.set_blocking` or non-blocking stdout polling (both unavailable/fragile on Windows)
- survival and full-pipe tests are `skipif(win32)` with a reason pointing at the docstring; delivery tests (drained pipe, StringIO) still run everywhere

**Note on the `niche` job:** the 2 R language-server failures (`test_find_definition_across_files`, `test_find_references_across_files` returning empty results) are pre-existing on `main` — they fail identically in runs of unrelated PRs (e.g. the docs-only run 34983624289), so they are unrelated to this PR.

src/serena/util/non_blocking_stderr_handler.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+
+        def __init__(self, fd: int) -> None:
+            self._fd = fd
+
+        def write(self, msg: str) -> None:
+            # a partial write means the buffer filled up mid-message; the remainder is
+            # dropped rather than retried (a retry loop could still block if the consumer
+            # never drains)
+            os.write(self._fd, msg.encode(errors="replace"))
```
AmirF194, line 96: Ran this against `f6ef7183` in a clean container.

`_FdRecordWriter.write` (`non_blocking_stderr_handler.py:94`) calls `os.write(self._fd, ...)` and drops the return value. On a non-blocking pipe with some free space but less than the message length, Linux writes a partial amount and returns that count, no exception. So the record is not "dropped instead of blocking" as the docstring says: it gets truncated and written anyway, no trailing newline, corrupting whatever the client reads next.

Repro: pipe left with ~2000 bytes free, then `handler.emit()` a 5055-byte formatted record.

```
delivered 959 of 5055 bytes for the record, no newline
```

`test_emits_to_full_pipe_without_blocking` only covers a permanently-full pipe (clean EAGAIN), not this partially-free case.

Fix would need to check `os.write`'s return value and drop the record on a short write rather than leaving the partial bytes on the stream.

davalillo:
Thanks for the careful review and the clean repro — confirmed. I reproduced your exact numbers (959 of 5055 bytes into ~2000 free bytes).

**Mechanism:** POSIX only guarantees atomic (all-or-nothing) behaviour for non-blocking writes of at most `PIPE_BUF` bytes. Our formatted records (timestamp, level, logger, message) regularly exceed 4096, so when the pipe had *some* free space but less than the record length, `os.write` wrote a partial amount and returned the count — no `BlockingIOError` — and `_FdRecordWriter.write` discarded the return value, leaving a truncated, newline-less record on the stream.

**Why checking the return value isn't sufficient:** by the time `os.write` returns, the partial bytes are already on the stream and cannot be unsent — the client would still read a corrupted record. So instead of detecting the short write, the fix prevents it: `_NonBlockingFdRecordWriter` now truncates the encoded record to the fd's atomic write size (`os.fpathconf(fd, "PC_PIPE_BUF")`, falling back to the POSIX minimum of 512) before writing, reserving room for a `"... [truncated]\n"` suffix. A write of ≤ `PIPE_BUF` bytes into a non-blocking pipe is guaranteed all-or-nothing: it either fits fully or raises `EAGAIN` (caught in `emit`, record dropped). Partial writes become impossible on pipes; on stderr socketpairs the never-draining scenario is a permanently full buffer (already handled), so the residual risk window there is negligible.

Trade-off, consistent with the handler's contract: a record longer than `PIPE_BUF` loses its tail when the buffer is partially free, instead of corrupting the stream. stderr remains best-effort; the log file and the dashboard buffer stay lossless.

Added `test_partial_free_pipe_drops_record_instead_of_truncating` (Linux pipe with `F_SETPIPE_SZ`, filled to leave < `PIPE_BUF` free, asserting the read end contains exactly the filler bytes — record dropped via `EAGAIN`, never written partially). Pushed as 518a3e5a.

src/serena/util/non_blocking_stderr_handler.py
AmirF194: I read the fix at dc3a05ea, did not re-run it. The truncate-before-write approach is correct: a cut that lands inside a multi-byte UTF-8 sequence decodes to exactly one U+FFFD (3 bytes) regardless of how many of the original 1-3 continuation bytes were kept, so the 3-byte reservation covers the worst case with one byte to spare, never exceeding the atomic write size. The fpathconf failure path falls back to 512, which is safe on sockets too since PIPE_BUF is a POSIX floor, not a pipe-only guarantee.

Issue #2045, opened 2026-09-16
Dart language server burns ~1 CPU core continuously at idle when activated on a monorepo root (rootUri hardcoded to project root; no idle recycling)


uhika:
### Environment
- Serena 1.7.0 (stdio MCP, LSP backend)
- Bundled DartLanguageServer, Dart SDK 3.7.1
- macOS arm64

### Symptom
On a large **monorepo root** (several sub-projects in different languages: Vue/TS, Java, C++, and one Flutter/Dart sub-project), the `dart language-server` process burns ~1 CPU core **continuously for hours with zero requests issued**. Measured: 40–48 CPU-seconds per 60s wall-clock at idle, %CPU fluctuating 120–150%, RSS 350–490 MB (peak ~1.1 GB).

All other language servers in the same setup (vue/ts/java/cpp/clangd) consume **~0 CPU at idle** (0.03–0.05s per 60s). So the Dart server is a clear outlier.

Note: activating the *same* Dart SDK via a client that roots it at the real Flutter sub-project directory (which has `pubspec.yaml`) shows **no** such behavior — it converges and idles.

### Steps to reproduce
1. Monorepo root with: no `pubspec.yaml` at root; a few hundred `.dart` files scattered under one sub-project that *does* have `pubspec.yaml`; hundreds of thousands of files in total (incl. `node_modules`, build dirs).
2. `project.yml`: `language_servers: [vue, typescript_vts, java, cpp, dart]`.
3. Start serena MCP with `--project-from-cwd` at the monorepo root.
4. Wait a few minutes without issuing any request; the `dart language-server` process stays at high CPU.

### Root cause (confirmed against SDK/source)
- solidlsp unconditionally sets `rootUri` = project root in the `initialize` request (`DefaultInitializeParamsBuilder._apply_updates`), plus `workspaceFolders` from `ls_workspace_folders`.
- The Dart analysis server adds **both** `workspaceFolders` and `rootUri` to its analysis roots (no dedup). So the entire monorepo root becomes an analysis root **regardless of `ls_workspace_folders`**.
- The analysis server then recursively enumerates the whole tree, installs a recursive FSEvents watcher on it, and background-analyzes every `.dart` file. Sub-projects without `pubspec.yaml` fall back to `BasicWorkspace` (no package resolution), so `package:` imports fail and each file is analyzed in isolation — pure waste.
- Active builds in the monorepo continuously generate FS events that keep the watcher busy, so CPU never settles.
[...]

Issue #1976, opened 2026-09-02
Java: allow project import settings without losing defaults or reusing stale state


TyceHerrman:
**Is your feature request related to a problem? Please describe.**

Serena does not provide a documented way to change these Java language-server settings:

- Maven project import enablement
- Gradle project import enablement
- `java.configuration.updateBuildConfiguration`
- Java autobuild

| Setting | Why it helps |
| --- | --- |
| Maven import | Ensures JDTLS reads `pom.xml`, dependencies, source folders, and generated code. It can be disabled for non-Maven projects to avoid unnecessary work. |
| Gradle import | Does the same for Gradle. In a Maven-only repository such as m2e-core, disabling it avoids irrelevant Gradle detection and import work. |
| `updateBuildConfiguration` | Controls what happens after `pom.xml` or `build.gradle` changes. Automatic updates keep JDTLS's classpath and dependencies current without requiring an IDE prompt that Serena cannot conveniently show. |
| Autobuild | Controls whether JDTLS recompiles after edits. Enabling it gives fresher errors and warnings; disabling it can reduce CPU and background work in large repositories. |

Using `initializationOptions` directly is not a safe workaround. Serena merges only the top level. Setting `settings.java` therefore replaces Serena's complete built-in `settings` section and can remove defaults such as registered Java runtimes.

Serena also does not consider these direct overrides when choosing a cached JDTLS workspace. It may reuse a workspace created with different import settings, leaving old project state or making a new setting appear to do nothing.

**Describe the solution you'd like**

Add Serena settings for the four behaviors above. This could use dedicated Serena settings or safely merge only the requested values into `initializationOptions.settings.java`. Merging is more flexible, but Serena would need clear rules for which values require a new cached workspace.

Suggested acceptance criteria:

- Maven import, Gradle import, `updateBuildConfiguration`, and autobuild are configurable and documented.
- Changing one setting does not remove unrelated defaults or registered JDK runtimes.
- Changing a setting that affects saved project state causes Serena to use a fresh JDTLS workspace. This likely includes both import toggles and `updateBuildConfiguration`.
- Settings that apply only while JDTLS is running, such as autobuild, do not create a new workspace unless they also affect saved state.
- Invalid values, such as a non-boolean import setting, produce a clear error.
- Tests cover both the settings sent to JDTLS and Serena's choice of cached workspace.

**Describe alternatives you've considered**

[...]

Issue #1971, opened 2026-09-02
Allow read-only and editing connections to the same Serena server


TyceHerrman:
**Is your feature request related to a problem? Please describe.**

I'd like tasks to share Serena server and have each connection use appropriate permissions. for example, an implementation task and a review task. The implementation task could change code and memories and the reviewer could inspect files and follow references without making changes. These would be separate sessions of the same agent.

Sharing a server for the same working copy avoids starting duplicate language servers. I want to keep different Git worktrees on separate Serena servers because their files and state can differ, and Serena's active project is shared across connections.

Telling a reviewer “don't edit” or hiding editing tools in its client does not stop it from requesting an editing tool directly. I want Serena itself to enforce the difference.

**Describe the solution you'd like**

Add optional access tokens that let Serena distinguish between connections allowed only to read and those also allowed to edit. The application running Serena would supply the token checker through the MCP Python SDK's existing support, rather than Serena introducing its own login system.

When this feature is enabled, Serena should:

- Show each connection only the tools it is allowed to use.
- Reject a forbidden tool call before it runs, even if the client requests it directly.
- Check the permissions on every request, not just when a session starts. Missing, invalid, or revoked tokens must not grant access, nor should tokens with no Serena permissions.

Serena already [classifies tools as editing or non-editing](https://github.com/oraios/serena/blob/main/src/serena/mcp.py). These permissions would restrict Serena tools only, not an agent's terminal or other ways of editing files.

The application would still issue and revoke tokens, manage worktrees, and decide how many editing connections to allow.

**Describe alternatives you've considered**

Separate servers for every client duplicate startup and state. Prompts and hidden tools do not enforce permissions. A separate service in front of Serena could check access, but would also need to understand Serena's tool calls and maintain its own permission rules.

**Additional context**

Two local tests passed using my own `serena-worktree-server` launcher, a modified Serena server, and `fastmcp-remote`. They started real processes and checked that:

- Serena starts and lists tools successfully.
- Two readers can share the server with a writer. Readers see only non-editing tools, and a direct editing request is rejected without creating a file.
- The writer can edit, invalid tokens are rejected, and its token stops working after disconnect.
- A replacement writer can connect without restarting Serena while readers remain connected.

My launcher also rejected a second simultaneous writer; that policy is outside this request. Another 182 tests for the launcher and worktree setup passed across two runs.

tldr - requesting optional token checking and tool permissions enforced by Serena itself.

opcode81:
[...]

Issue #1952, opened 2026-08-30
replace_symbol_body removes separator whitespace before the following GDScript function


quakquak86:
### Setup
- Serena v1.7.0
- Windows 11
- MCP over stdio
- GDScript / Godot 4.6.3 language server

### Minimal reproduction
Given a GDScript file with two adjacent top-level functions separated by a blank line:

```gdscript
func first():
    print("old")

func second():
    print("keep")
```

Call `replace_symbol_body` for `first` and replace its body.

### Actual behavior
In an isolated fixture, Serena v1.7.0 reproducibly removed the blank separator line immediately before the following function. The edit therefore changes inter-symbol trivia outside the body that was requested to be replaced.

This is especially risky for automated edits because the tool contract suggests a symbol-body replacement, while formatting/separators belonging to the following symbol can be modified as a side effect.

### Expected behavior
`replace_symbol_body` should replace only the target symbol body/range and preserve separator whitespace/trivia before the next GDScript symbol.

### Workaround
I currently exclude `replace_symbol_body` from the Serena context for GDScript and use guarded textual replacement (`replace_in_files` with dry-run, occurrence selection/count checks), followed by a disk check and diagnostics.

I searched open and closed Serena issues for `replace_symbol_body`, GDScript, separator and blank-line corruption and did not find a matching report. The affected project is private; the reproduction above is self-contained.

opcode81:
This was fixed by #1974

PR #2049, opened 2026-09-17
Fix unclaimed issues: Dart rootUri, GDScript ranges, Bloop cleanup, Java import settings, TS monorepo docs, connection tokens (#2045 #1952 #1939 #1816 #1976 #1971)

Yu-0312:
## Summary

Implements several open issues that had **no open PR claim** (verified against the open-PR list and issue cross-references). Each change is scoped to one issue; tests cover the new behaviour without requiring live language servers.

| Issue | Change |
|-------|--------|
| [#2045](https://github.com/oraios/serena/issues/2045) | Dart: omit deprecated `rootUri`/`rootPath` by default (`ls_specific_settings.dart.set_root_uri`, default `false`) so a monorepo root is not analysed as an extra analysis root |
| [#1952](https://github.com/oraios/serena/issues/1952) | GDScript: clamp symbol range ends so `replace_symbol_body` no longer deletes the blank line before the next function |
| [#1939](https://github.com/oraios/serena/issues/1939) | Docs: TypeScript monorepo cross-package references require tsconfig project references |
| [#1816](https://github.com/oraios/serena/issues/1816) | Scala: terminate Bloop daemons this session spawned after `stop()` (`terminate_bloop_on_stop`, default `true`) |
| [#1976](https://github.com/oraios/serena/issues/1976) | Java: `maven_import_enabled`, `gradle_import_enabled`, `update_build_configuration`, `autobuild_enabled` via `ls_specific_settings.java`; new keys included in the JDTLS workspace hash |
| [#1971](https://github.com/oraios/serena/issues/1971) | Optional HTTP `connection_access_tokens` via MCP SDK `TokenVerifier` (read/edit scopes) |

## #2045 — Dart `rootUri`

Matches the maintainer suggestion on the issue (`set_root_uri`, default True globally, False for the Dart server). Workspace folders are still sent; only the legacy single-root fields are omitted unless opted back in.

## #1952 — GDScript body range

Godot often reports a function range ending at the start of the next top-level construct (or on the blank separator). `_clamp_trailing_blank_lines` walks the end back to the last non-blank line of the symbol. `selectionRange` is untouched. Document-symbol cache version bumped.

## #1939 — TypeScript monorepo docs

Documents why `find_referencing_symbols` is silently partial across packages without `composite` + `references`, and what to configure.

## #1816 — Bloop daemon cleanup

Bloop self-daemonizes and is re-parented to PID 1 after Metals exits, so process-tree cleanup misses it. Before `stop()`, snapshot descendant PIDs whose cmdline contains `bloop.BloopServer`; after Metals exits, terminate survivors. Only descendants of *this* process are considered. `terminate_bloop_on_stop: false` opts out.

## #1976 — Java import / rebuild settings

Defaults match the previous hard-coded values. Invalid `update_build_configuration` falls back to `interactive`. Changing any of the four keys changes the workspace hash so a stale import workspace is not reused.

## #1971 — Connection access tokens (MCP SDK)

Per the issue author: token checker via the MCP Python SDK's existing support, not a Serena login system.

1. `connection_access_tokens: {token: read|edit}` in `serena_config.yml` (empty = feature off; stdio ignores it).
2. `SerenaTokenVerifier` implements `TokenVerifier`; permission becomes an `AccessToken` scope.
3. `FastMCP(auth=..., token_verifier=...)` rejects missing/invalid tokens in SDK middleware **before any tool runs**.
4. Edit tools re-check the `edit` scope on every call (scopes first, Authorization header as fallback).
5. `required_scopes` floor is `read`.

**Limitation (also in docs):** `tools/list` is process-global on one FastMCP instance, so read-only clients still *see* editing tools; every editing call is refused at execution time. A per-connection filtered list would race across concurrent streamable-http sessions without per-session tool managers.

`AuthSettings` requires OAuth-shaped URLs for protected-resource metadata; tokens here are application-supplied static secrets and the issuer/resource URLs are placeholders so Bearer-only clients keep working.

## Testing

- 29 focused unit tests (connection access + tool gates, GDScript clamp, Dart rootUri default, Java settings + workspace hash, Bloop discovery/stop flag)
- 135 unmarked tests including `test/serena/config/` pass locally
- `ruff` clean on changed Python files
- No Godot / Dart-monorepo / JDTLS e2e in this environment (logic aligned with issue text + maintainer comments)

## Checklist

- [x] Follows `CONTRIBUTING.md` scope guidelines (one issue per logical change; this PR groups independently reviewable commits for unclaimed issues — happy to split if preferred)
- [x] `CHANGELOG.md` updated
- [x] CLA will be accepted when prompted

---

## Further ideas for maintainers (not implemented)

These are follow-ups I considered but did **not** put in this PR; happy to do them if you want.

1. **#1971 — per-connection `tools/list`.** The issue also asks to *show each connection only the tools it is allowed to use*. That needs per-session tool registration (or a `list_tools` hook that can see the request token). I kept call-time rejection as the security boundary to avoid racing the global `_tool_manager` under concurrent HTTP connections.

2. **#1971 — pluggable host-supplied checker.** Today the “checker” is a static map in `serena_config.yml` wrapped in `TokenVerifier`. A small hook so a host process can inject its own `TokenVerifier` (revocation lists, external IdP) would match “the application supplies the token checker” even more closely.

3. **#1971 — drop the header fallback** once only SDK auth middleware is used in production, to leave a single code path.

4. **#1816 — graceful Bloop shutdown** (BSP exit / `bloop` CLI) instead of `terminate`/`kill`, so a Bloop shared with another Metals client is less likely to be stopped.

5. **#2045 — omit `rootUri` for other servers** when it is identical to the sole `workspaceFolders` entry (LSP 3.6+). Scoped to Dart here because that is the only confirmed double-root burn.

6. **#1976 — e2e** with the repo Java fixture to assert the four settings actually change JDTLS import behaviour, not only the initialize payload.

7. **Coordination:** `mcp.py` / `serena_config.py` may conflict with Serena v2 (#2035); happy to rebase after that lands.


Yu-0312:
Closing in favor of one PR per issue (CONTRIBUTING: single logical change). Replacing with split PRs.

PR #2050, opened 2026-09-17
fix(serena): Windows non-blocking stderr via PIPE_NOWAIT (#2047, follow-up to #2044)

### LINKED ISSUE #2047: Windows: stderr deadlock still possible — NonBlockingStderrHandler degrades to blocking writes (follow-up to #2044)
Preconditions:

- [x] I have actually encountered this issue in practice. Issues that are based on (AI-based) code analysis only will be closed without comment.

Issue details:

- [ ] If the issue relates to an application of Serena to an open-source project, I have added the link.

**This is a non-urgent follow-up issue.** It tracks a known, documented platform limitation left out of #2044 on purpose — nothing in the current release regresses; the deadlock this issue describes is exactly the pre-existing behavior on Windows.

## Follow-up to #2044

PR #2044 (`fix(serena): non-blocking stderr handler to prevent logging deadlock in MCP servers`) fixes the stderr logging deadlock observed in production on POSIX (see #1588 for the sibling symptom: executor thread wedged in `logging.emit` → stream write after the host failed to drain the stream). On Windows, however, the same deadlock scenario **remains possible**, and the PR ships it that way deliberately instead of emulating the missing mechanism with deprecated APIs. The limitation is documented in the handler's docstring and CHANGELOG, and analyzed for the maintainers in [this PR comment](https://github.com/oraios/serena/pull/2044#issuecomment-5697068739).

## Why Windows cannot get the same mechanism today

1. `os.set_blocking` does not exist on Windows (`AttributeError`), so the fd cannot be switched to non-blocking the POSIX way.
2. `fstat` on Windows does not report anonymous pipes as FIFOs (`stat.S_ISFIFO` is false), so the handler cannot even detect the bounded-buffer case via `st_mode`.
3. There is no `O_NONBLOCK` equivalent for anonymous pipes on Windows.

Consequence: `NonBlockingStderrHandler` degrades to plain blocking writes on Windows — identical to a plain `StreamHandler`. With an MCP client host that never drains the spawned server's stderr, the stderr buffer (a socketpair/pipe with a finite buffer) eventually fills and a `write()` blocks while holding the logging module's global lock, freezing every thread that logs — the same failure chain PR #2044 eliminates on POSIX.

Practical mitigation (why this is low urgency): the dominant Windows MCP client hosts redirect the server's stderr to a log file, which is an unbounded sink — the buffer never fills.

## Proposed resolution (for the maintainers to decide)

`SetNamedPipeHandleState(handle, PIPE_NOWAIT)` via `ctypes` on the *write* end of an anonymous pipe does work: when the buffer is full, `WriteFile` fails immediately with `ERROR_NO_DATA` → Python raises `OSError` → the handler's existing `except` drops the record. That would honor the non-blocking contract on Windows too, and would let PR #2044's full-pipe tests run unskipped in the Windows CI job. It was **not** included in the PR because `PIPE_NOWAIT` is [documented as deprecated](https://learn.microsoft.com/en-us/windows/win32/api/namedpipeapi/nf-namedpipeapi-setnamedpipehandlestate) by Microsoft and the PR did not want to build a bug-fix on deprecated functionality. The change would be a ~15-line Windows branch in `_create_writer` in `src/serena/util/non_blocking_stderr_handler.py`.

Alternatives to weigh: accept the documented limitation indefinitely, or emulate non-blocking behavior with a bounded retry loop (rejected during the PR review — it cannot avoid blocking without either dropping correctness or adding a writer thread, which PR #2044 deliberately avoids because a daemon writer deadlocks CPython's interpreter shutdown).

## Setup

- Serena: current `main` + PR #2044 branch (`fix/stderr-logging-deadlock`, commits `61ecac8a` and `f6ef7183`)
- OS: Windows (CI runners: windows-2025; the limitation also applies to user machines)
- MCP client: any Windows host that spawns `start-mcp-server`/`project-server` in stdio mode and does not drain stderr
- Languages: not applicable (the deadlock is in Serena's own logging pipeline, independent of language servers)

Yu-0312:
## Follow-up to #2044 — closes #2047

**This branch is stacked on [#2044](https://github.com/oraios/serena/pull/2044)** (`NonBlockingStderrHandler`, POSIX). That PR fixed the MCP stderr logging deadlock on POSIX and **deliberately left Windows blocking**, documenting `SetNamedPipeHandleState(PIPE_NOWAIT)` as an alternative that works but is deprecated by Microsoft.

This PR implements that Windows half so the same failure mode cannot freeze logging on Windows either ([#2047](https://github.com/oraios/serena/issues/2047), filed by the #2044 author as a non-urgent follow-up).

> **Review order:** please review/merge #2044 first (or I rebase onto main after it lands). Until then the GitHub diff against `main` includes #2044’s commits; only the last three commits on this branch are new:
> - `fix(serena): Windows non-blocking stderr via PIPE_NOWAIT (#2047)`
> - `chore: drop unused Windows error constant`
> - `fix(serena): treat Windows broken-pipe errors as non-blocking drops`

## What changes

On `win32`, when stderr’s fd is a pipe:

1. `msvcrt.get_osfhandle` + `SetNamedPipeHandleState(..., PIPE_NOWAIT)`.
2. A full or disconnected pipe surfaces as `OSError` (`ERROR_NO_DATA` / `ERROR_BROKEN_PIPE` / `ERROR_PIPE_NOT_CONNECTED`) and is mapped to `BlockingIOError`, which `emit()` **drops** instead of blocking the logging lock — same contract as POSIX.
3. If `PIPE_NOWAIT` cannot be applied (console/file handle, API failure), fall back to blocking `os.write` exactly as #2044 documented.

Authoritative streams (log file, dashboard buffer) stay lossless as in #2044. No background writer thread (same interpreter-shutdown deadlock reason as #2044).

## Testing

- Existing #2044 handler tests still pass on POSIX.
- New tests: PIPE_NOWAIT enablement (Windows-only), `ERROR_NO_DATA` → `BlockingIOError` mapping (cross-platform unit path).
- No Windows CI run in this environment for the live full-pipe path.

## Checklist

- [x] Follows `CONTRIBUTING.md` (single logical change; stacked on #2044)
- [x] `CHANGELOG.md` updated
- [x] CLA will be accepted when prompted

---

## Further ideas for maintainers (not implemented)

1. **CI:** add a Windows job that floods an undrained stderr pipe and asserts the server still exits (the survival test from #2044, unskipped on `win32`).

2. **Longer-term:** if Microsoft ever removes `PIPE_NOWAIT`, the fallback is blocking again. Alternatives were rejected in #2044 (helper thread deadlocks CPython shutdown; `SetNamedPipeHandleState` is deprecated). A future option is a tiny named-pipe helper process with its own lifetime — only worth it if Windows MCP hosts still do not drain stderr.

3. **After #2044 + this PR land:** the #2044 docstring/CHANGELOG “Windows remains blocking” note can be replaced by a pointer to this follow-up (or squashed on rebase so main has one coherent story).


Yu-0312:
Windows catch-all failed with `OSError(28, 'No space left on device')` and `winerror=None` (not ERROR_NO_DATA). Pushed a fix that also maps errno.ENOSPC / ENOBUFS / EPIPE to BlockingIOError so emit() drops the record instead of erroring.

src/serena/util/non_blocking_stderr_handler.py
```diff
@@ -0,0 +1,248 @@
...
+        record would be written partially when the buffer had some but not enough free
+        space, corrupting the stream with a truncated, newline-less record.
+        """
+
+        _TRUNCATION_SUFFIX = "... [truncated]\n"
+
+        def __init__(self, fd: int) -> None:
+            super().__init__(fd)
+            try:
+                os.set_blocking(fd, False)
```
AmirF194, line 129: I ran this against `5df0dca3` (this PR's head) in a clean container.

`_create_writer` calls `os.set_blocking(fd, False)` on the handler's stream fd (`non_blocking_stderr_handler.py:127`). `O_NONBLOCK` belongs to the open file description, not to this handler's copy of it, so it also flips blocking mode for every other piece of code still holding `sys.stderr`. That includes `show_fatal_exception_safe`'s raw `print(f"Fatal exception: {e}", file=sys.stderr)` at `exception.py:52`, called from `mcp.py:378` inside `except Exception as e: show_fatal_exception_safe(e); raise`, right after `cli.py` installs this handler. That print is unguarded, so once the pipe backs up it now raises `BlockingIOError` instead of blocking.

```
$ python3 repro.py
writer type: _NonBlockingFdRecordWriter
handler survived filling the pipe (expected, this is the PR's own point)
BYSTANDER WRITE RAISED BlockingIOError: [Errno 11] Resource temporarily unavailable
```

`repro.py` fills the pipe through the handler like the PR's own `TestNonBlockingStderrHandlerSurvival` does, then does a plain `os.write` on the same fd from outside the handler, standing in for `exception.py`'s print. That write used to just block; now it raises.

This print was hardened in `198c9415` specifically to "definitely print to stderr" as the last-resort path when agent construction fails at startup. This PR's fd mutation can take that guarantee away in exactly the client-not-draining-stderr case it targets, replacing the fatal-exception message with an unrelated `BlockingIOError`.

Wrapping `exception.py:52` in `try: ... except (BlockingIOError, OSError): pass`, the same handling `emit()` already gives its own writes, would close this.

Yu-0312:
Thanks @AmirF194 — that makes sense. `os.set_blocking` on the open file description can flip stderr for bystanders such as `show_fatal_exception_safe`, not only this handler. I'll try your suggestion and wrap that stderr print with `try/except (BlockingIOError, OSError): pass` (same drop policy as `emit()`) and push an update.

src/serena/util/non_blocking_stderr_handler.py
AmirF194: Sounds right. Wrapping the print in the same try/except emit() uses is the minimal fix and keeps the drop policy consistent between the two callers.

Yu-0312:
Addressed review from @AmirF194:

`show_fatal_exception_safe` now wraps the stderr `print` in `try/except (BlockingIOError, OSError)`, matching `emit()`'s handling of non-blocking writes. The fatal exception is still `log.error`'d first, so the message remains available when the pipe is full.

Added `test/serena/test_exception_stderr_nonblocking.py` (BlockingIOError and OSError paths). Local: 2 passed.

Yu-0312:

PR #2051, opened 2026-09-17
fix(dart): omit rootUri/rootPath (#2045)

### LINKED ISSUE #2045: Dart language server burns ~1 CPU core continuously at idle when activated on a monorepo root (rootUri hardcoded to project root; no idle recycling)

Yu-0312:
## Summary

Closes [#2045](https://github.com/oraios/serena/issues/2045).

On a monorepo root that is not a Dart package, the Dart analysis server burned ~1 CPU core at idle because Serena always sent both `rootUri` **and** `workspaceFolders`, and Dart adds both to its analysis roots without de-duplication.

This matches the approach discussed on the issue (`set_root_uri`, default True globally, False for the Dart server).

## Change

- `DefaultInitializeParamsBuilder(..., set_root_uri: bool = True)` — when False, omit `rootPath`/`rootUri`.
- Dart server defaults `set_root_uri` to **false**; opt back in with `ls_specific_settings.dart.set_root_uri: true`.
- Workspace folders are still sent as before.

## Tests

- `test_initialize_params_root_uri.py`: builder omits/includes rootUri
- `test_dart_root_uri.py`: Dart default omits; opt-in restores

## Checklist
- [x] `CHANGELOG.md` updated
- [x] Single logical change

---

1. Other language servers could omit `rootUri` when it is identical to the sole `workspaceFolders` entry (LSP 3.6+). Scoped to Dart here because that is the only confirmed double-root burn.
2. A live monorepo Dart e2e (idle CPU before/after) would be ideal; only unit tests here.


Yu-0312:

src/solidlsp/language_servers/dart_language_server.py
```diff
@@ -54,6 +55,12 @@ class DartLanguageServer(SolidLanguageServer):
     You can pass the following entries in ``ls_specific_settings["dart"]``:
         - dart_sdk_version: Override the pinned Dart SDK version downloaded by Serena
           (default: the bundled Serena version).
+        - set_root_uri: Whether to send the deprecated ``rootUri``/``rootPath`` fields in
+          the LSP ``initialize`` request (default: false). The Dart analysis server adds
+          both ``rootUri`` and every ``workspaceFolders`` entry to its analysis roots
+          without de-duplication, so on a monorepo root that is not a Dart package the
+          whole tree is analysed and the server burns CPU at idle (oraios/serena#2045).
+          Set to true only if a single-package project relies on ``rootUri`` alone.
```
opcode81: This documentation does not belong here. Dart-specific aspects should be explained in the Dart implementation, not here.

src/solidlsp/language_servers/dart_language_server.py
```diff
@@ -73,6 +80,11 @@ def __init__(self, config: LanguageServerConfig, repository_root_path: str, soli
         # via either notification it sends for this (see _start_server).
         self.analysis_complete = threading.Event()
 
+    def _create_initialize_params_builder(self) -> InitializeParamsBuilder:
+        dart_settings = self._custom_settings or {}
+        set_root_uri = bool(dart_settings.get("set_root_uri", False))
```
opcode81: Is there a use case for a user overriding this? Why are we making this an option?
Also, note that `self._custom_settings` is always defined, it is never None.

src/solidlsp/language_servers/dart_language_server.py
Yu-0312: Fixed — removed that docstring entry; the Dart-specific rationale now lives as a comment inside `_create_initialize_params_builder`, and the generic `initialize_params.py` comment no longer mentions Dart.

src/solidlsp/language_servers/dart_language_server.py
Yu-0312: Good question — I checked the analysis-server source (`pkg/analysis_server/lib/src/lsp/handlers/handler_initialize.dart`): it derives its analysis roots from `workspaceFolders` **plus** `rootUri` (falling back to `rootPath` only when `rootUri` is absent). Since Serena always sends `workspaceFolders` (`ls_workspace_folders` defaults to `["."]`), omitting `rootUri` is always safe and re-enabling it would only add a duplicate analysis root — so there is no user-facing case for the option. Removed `ls_specific_settings["dart"].set_root_uri`; the Dart server now always omits `rootUri`/`rootPath`. Also dropped the redundant `self._custom_settings or {}`.

PR #2052, opened 2026-09-17
fix(gdscript): clamp symbol ranges so replace_symbol_body keeps separators (#1952)

### LINKED ISSUE #1952: replace_symbol_body removes separator whitespace before the following GDScript function

Yu-0312:
## Summary

Closes [#1952](https://github.com/oraios/serena/issues/1952).

`replace_symbol_body` deleted the blank separator immediately before the next GDScript function because Godot reports a symbol range ending at the start of the next top-level construct (or on the blank line).

## Change

In `GodotLanguageServer._fix_symbol_ranges`, after the existing off-by-one end-column fix, `_clamp_trailing_blank_lines` walks the range end back to the last non-blank line of the symbol. `selectionRange` (the identifier) is not clamped. Document-symbol cache version bumped to 2.

## Tests

`test_godot_range_clamp.py`: clamp unit cases + end-to-end `_fix_symbol_ranges` on a two-function fixture; `selectionRange` preserved.

## Checklist
- [x] `CHANGELOG.md` updated
- [x] Single logical change

---

1. A Godot editor e2e (`replace_symbol_body` on two adjacent functions) would pin this against a live LS; only pure-function tests here.
2. If Godot ever reports correct ends, the clamp becomes a no-op (safe to keep or gate on Godot version like `_fix_range_end`).


Yu-0312:

opcode81:
#1974 already fixed the underlying issue.

PR #2053, opened 2026-09-17
docs: explain TypeScript monorepo project references (#1939)

### LINKED ISSUE #1939: Docs: TypeScript monorepos silently under-report cross-package find_referencing_symbols without tsconfig project references
## Setup

Serena 1.7.0, MCP client Claude Code, macOS 26.5 arm64. Language: TypeScript, `typescript-language-server` 5.1.3, workspace TypeScript 6.0.3. pnpm monorepo: 2 apps (Next.js, Expo) and 15 packages, 1406 tracked `.ts`/`.tsx` files.

**This is not a Serena bug** — it is tsserver behaviour Serena inherits. Filing it as a documentation request, because the failure is silent and hits what I suspect is a large share of Serena's TypeScript users.

## What happens

In a monorepo where packages import each other through `package.json` `exports`, `find_referencing_symbols` on an exported symbol returns only the references **inside its own package**, with nothing to indicate the rest were not searched.

Measured against grep ground truth before changing anything:

| Symbol | Real consumer files | `find_referencing_symbols` |
|---|---|---|
| `formatDisplay` (leaf util package) | 11 | **1** — itself |
| `BottomNav` (shared UI, platform-split `.web`/`.native`) | 2 web / 2 native | **1 / 1** — its own colocated test |
| `getAdminClient` (package referenced by its consumer) | 82 | 82 ✅ |
| `requireFamilyMemberHono` (same-package) | 25 | 24 ✅ |

The two that worked were the only two packages in the repo whose consumer declared a tsconfig `references` edge to them. That is the whole difference.

## Cause

tsserver resolves a bare cross-package import through `exports` → `types` → `dist/*.d.ts`, and only walks back from that declaration to the source across a `references` edge pointing at a `composite` project. [microsoft/TypeScript#30823](https://github.com/microsoft/TypeScript/issues/30823) (open) states it: *"if the file is not open or declaration maps are not present it cannot find all references in sibling projects"*. A solution-style root `tsconfig.json` is not sufficient on its own — two packages both listed there still could not see each other, cf. [microsoft/TypeScript#56268](https://github.com/microsoft/TypeScript/issues/56268) ("working as intended").

Ruled out first, in case it was Serena's: raised `indexing_start_grace` to 30 s and `indexing_timeout` to 180 s, confirmed both applied in the server log, re-ran — identical answers. Structural, not timing. (The timing problem is real but separate; filed as #1937.)

## What fixed it

1. `composite: true` + `declarationMap: true` on every package another package or app imports.
2. A `references` edge in every consumer's tsconfig mirroring its workspace deps.

That fixed package→package. Apps stayed blind, because an app can never be listed in the root solution: `references` requires `composite`, `composite` forbids `noEmit`, and Next.js and Expo both require `noEmit`.

3. So each app got a second config — `apps/<app>/tsconfig.ide.json` — extending the app's own tsconfig with `composite` + `emitDeclarationOnly` into a gitignored dir, restating `references` (it is **not** inherited through `extends`), and the root solution lists those files instead of the apps. Nothing builds them; the framework build and the typecheck script both read `tsconfig.json`, and no script runs `tsc -b`.

Result on the same symbols: `formatDisplay` 1 → **11 of 11**, including 5 in the web app and 1 in the native app. `BottomNav` resolves its web consumer from the `.web` twin and its native consumer from the `.native` twin. No regression on the two that already worked.

## The request

A short "TypeScript monorepos" note in the language-support or configuration docs saying that `find_referencing_symbols` under-reports across package boundaries unless the repo has real project references, and that the shortfall is silent. Two sentences would have saved a day of measurement here. Happy to open a PR with the above written up properly if that is welcome.

A second thought, lower confidence: Serena could *detect* this rather than document it — on activation, if a TypeScript project has multiple `tsconfig.json` files and no `references` edges between them, log a warning that cross-package reference search will be incomplete. That is the kind of thing a user never discovers on their own, because the wrong answer looks like a right one.

Yu-0312:
## Summary

Closes [#1939](https://github.com/oraios/serena/issues/1939) (documentation).

In TypeScript monorepos, `find_referencing_symbols` silently under-reports cross-package consumers unless packages use [project references](https://www.typescriptlang.org/docs/handbook/project-references.html). That is tsserver behaviour ([microsoft/TypeScript#30823](https://github.com/microsoft/TypeScript/issues/30823)), not a Serena bug — but it fails quietly and is easy to misread as a Serena defect.

## Change

Docs only: new subsection under TypeScript configuration explaining the failure mode, what to configure (`composite` + `references`, declaration maps), and how to sanity-check short results.

## Checklist
- [x] Documentation-only
- [x] `CHANGELOG.md` updated

---

1. Surface a one-line warning in the tool result when the project has packages but no `references` edges (heuristic; may be noisy).
2. Link this section from the TypeScript language-server troubleshooting page if one is added later.


Yu-0312:

PR #2054, opened 2026-09-17
fix(scala): terminate orphaned Bloop daemons on stop (#1816)

### LINKED ISSUE #1816: Scala LS: Bloop build-server daemon survives ScalaLanguageServer.stop(), reparented to PID 1
ScalaLanguageServer has no stop() override, so a per-project detail matters here: Metals launches a Bloop build-server daemon (bloop.BloopServer) that self-daemonizes, and it survives Serena's shutdown as an orphan re-parented to PID 1.

Reproduced against current HEAD (c7af2c09ef45faa4367c0e2a9f770fb73a62a612) in a clean eclipse-temurin:21-jdk Docker container (JDK 21.0.11, sbt 1.10.1, metals 1.6.4, bloop-frontend 2.0.17, using the repo's own test/resources/repos/scala fixture with sbt-bloop already configured):

1. Ran `sbt -Dbloop.export-jar-classifiers=sources bloopInstall` then `sbt compile` to set up `.bloop/` the way the setup guide describes.
2. Drove the real `ScalaLanguageServer` class directly (not a mock): `ls = ScalaLanguageServer(config, repo_root, settings); ls.start(); time.sleep(25)`.
3. `ps -eo pid,ppid,cmd` at that point shows the Bloop JVM (pid 1651) as a child of the Metals process (pid 1554, ppid 1651->1554).
4. Called `ls.stop()`. Log confirms the top-level process exits cleanly: `LS[scala] terminated successfully with exit code 0.`
5. `ps` immediately after shows pid 1651 still running, now with ppid 1 (re-parented to init), RSS 209MB after under 2 minutes on this minimal 2-file fixture.

Root cause as far as I can tell: `_signal_process_tree` (subprocess_util.py) calls `parent.children(recursive=True)` at the moment of shutdown to find what to terminate. Bloop daemonizes itself (the classpath includes `libdaemon`, and the process argv is `bloop.BloopServer daemon:...`), which is the same "invisible to normal process-tree cleanup" shape already described for the Kotlin/Gradle-daemon case (#1081, with a fix proposed in #1082).

One difference from the Kotlin case that I think matters for how a fix should look: this repo's own scala_setup_guide_for_serena.md documents Bloop as an intentionally *shared* daemon ("All instances share a single Bloop process (port 8212)... Compilation Results: Shared via Bloop, no duplicate compilation"), and #1769 already has Serena deliberately declining to auto-answer Metals' own "kill the old Bloop process?" prompt, specifically because killing a process on the user's machine isn't something to do unprompted. If `stop()` unconditionally kills any Bloop process it can find (mirroring #1082's Gradle-daemon approach), it could tear down a build server that a concurrent VS Code Metals session (or a second Serena project) is still using. So this probably isn't a same-shape one-liner: whatever fix lands here likely needs some notion of "am I the last user of this daemon" rather than "kill on sight."

Filing this as a repro rather than a PR since the right fix depends on a decision I don't think I should make unilaterally: whether Serena should track daemon ownership/refcounting before killing Bloop, or take a different approach entirely (e.g., a documented "don't leave orphans" opt-in). Happy to send a PR once there's a preferred direction.

Setup: Serena HEAD c7af2c09ef45faa4367c0e2a9f770fb73a62a612, Ubuntu container (eclipse-temurin:21-jdk base), Python 3.x editable install, no MCP client involved (drove ScalaLanguageServer directly in a script to isolate the language-server layer from any client-specific behavior).

Yu-0312:
## Summary

Closes [#1816](https://github.com/oraios/serena/issues/1816).

Metals launches a Bloop build-server daemon (`bloop.BloopServer daemon:…`) that self-daemonizes. After `ScalaLanguageServer.stop()` exits cleanly, Bloop is re-parented to PID 1 and is invisible to process-tree cleanup — orphaned, still using RAM.

## Change

- Before `super().stop()`, snapshot PIDs of processes whose cmdline contains `bloop.BloopServer` and that are descendants of *this* process.
- After Metals exits, terminate any that are still alive (terminate → wait → kill).
- `ls_specific_settings.scala.terminate_bloop_on_stop` (default `true`) opts out when another Metals client is expected to keep using Bloop.

Only descendants of this Serena process are considered, so a Bloop started by another client (e.g. VS Code Metals) is left alone.

## Tests

`test_scala_bloop_cleanup.py`: discovery filters non-descendants / non-Bloop; `stop()` respects the flag; terminate skips PIDs that are no longer Bloop.

## Checklist
- [x] `CHANGELOG.md` updated
- [x] Single logical change

---

1. Prefer a graceful Bloop/Metals BSP exit over `terminate`/`kill` so a Bloop shared under the same process tree is less likely to be stopped.
2. Document the multi-Serena / shared-Bloop caveat next to the other Scala `ls_specific_settings` in the configuration guide if you want it more discoverable.
3. Integration test with `test/resources/repos/scala` asserting no `bloop.BloopServer` remains after `stop()` would need a Metals+Bloop environment in CI.


Yu-0312:

PR #2056, opened 2026-09-17
feat(access): optional HTTP connection tokens via MCP TokenVerifier (#1971)

### LINKED ISSUE #1971: Allow read-only and editing connections to the same Serena server

Yu-0312:
## Summary

Closes [#1971](https://github.com/oraios/serena/issues/1971).

Optional per-connection permissions for HTTP MCP transports (`sse` / `streamable-http`): one server, some connections read-only, others may edit — enforced by Serena, not by hiding tools in the client.

Per the issue author: the token checker goes through the **MCP Python SDK’s existing support**, not a Serena login system.

## Design

```yaml
# serena_config.yml
connection_access_tokens:
  "reader-secret-token": read
  "writer-secret-token": edit
```

1. **`SerenaTokenVerifier`** implements the SDK `TokenVerifier` protocol. Permission is carried as an `AccessToken` scope (`read` / `edit`).
2. **`FastMCP(auth=AuthSettings(...), token_verifier=...)`** — missing/invalid tokens fail in SDK auth middleware **before any tool runs**.
3. **Edit tools** re-check for the `edit` scope on every call (`get_access_token()` scopes first; `Authorization` header fallback).
4. `required_scopes` floor is `read`.
5. **stdio** ignores the setting (single local client already controls the process). Empty mapping = feature off (previous behaviour).

Uses Serena’s existing `tool.can_edit()` classification. Restricts Serena MCP tools only (not the agent’s terminal).

## Known limitation (documented)

`tools/list` is process-global on one FastMCP instance. Read-only clients still *see* editing tools in the list; every editing call is refused at execution time. A per-connection filtered list would race the shared `_tool_manager` under concurrent streamable-http sessions without per-session tool managers.

`AuthSettings` requires OAuth-shaped URLs for protected-resource metadata; tokens here are static application-supplied secrets and the issuer/resource URLs are placeholders so Bearer-only clients keep working.

## Tests

`test_connection_access.py`: token→permission maps; missing/invalid denied; tool-level gates with real FastMCP `FuncMetadata`; scope preference over headers.

## Checklist
- [x] `CHANGELOG.md` updated
- [x] Single logical change
- [x] Feature is opt-in (empty config = no behaviour change)

---

1. **Per-connection `tools/list`** — the issue also asks to *show each connection only the tools it is allowed to use*. Needs per-session tool registration or a `list_tools` hook that can see the request token.
2. **Pluggable host checker** — a hook so a host process can inject its own `TokenVerifier` (revocation lists, external IdP) matches “the application supplies the token checker” even more closely than a static YAML map.
3. **Single code path** — drop the header fallback once only SDK middleware is used in production.
4. **Conflict note** — `mcp.py` / `serena_config.py` may conflict with Serena v2 (#2035); happy to rebase after that lands.


Yu-0312:

PR #2066, opened 2026-09-18
fix(lsp): use typescriptreact/javascriptreact languageIds for .tsx/.jsx in Angular, Vue and Svelte

Yu-0312:
## Summary

Angular, Vue, and Svelte language-server adapters still open `.tsx` / `.jsx` with plain `typescript` / `javascript` language IDs. JSX then parses as TS and symbol ranges silently truncate at the first multi-line JSX expression (same failure mode as the typescript-language-server fix).

This is the same fix as #1436 (merged for typescript-language-server) for backends that #1993 (vtsls-only) does not cover.

## Changes

- `angular_language_server.py`, `vue_language_server.py`, `svelte_language_server.py`: send `typescriptreact` / `javascriptreact` for `.tsx` / `.jsx`
- CHANGELOG entry

## Related

- Merged: #1436 (typescript LS)
- Open: #1993 (vtsls only — different backend)

## Test plan

- Diff mirrors the pattern already used in `typescript_language_server.py`
- Existing LS tests for those backends (where present) still construct documents via the same language-id helper

## Checklist

- [x] CHANGELOG updated
- [x] Single logical change


src/solidlsp/language_servers/angular_language_server.py
```diff
@@ -126,9 +126,15 @@ def _get_or_install_core_dependency(self) -> str:
     @override
     def _get_language_id_for_file(self, relative_file_path: str) -> str:
         ext = os.path.splitext(relative_file_path)[1].lower()
-        if ext in (".ts", ".tsx", ".mts", ".cts"):
+        # JSX must use *react language IDs; plain typescript/javascript parse JSX as TS
+        # and silently truncate symbol ranges at multi-line JSX expressions (#1436).
+        if ext == ".tsx":
```
AmirF194, line 131: I compared this PR against #1436, the merged fix for the identical bug in typescript_language_server.py that this PR's own comments cite.

This adds no test, and #1436 did. `test/solidlsp/typescript/test_typescript_basic.py::test_tsx_symbol_range_not_truncated_by_jsx` opens `jsx_component.tsx` through a real language server and asserts the symbol range is not truncated at the first multi-line JSX expression, which is exactly the regression this PR fixes for Angular, Vue and Svelte. None of `test/resources/repos/{angular,vue,svelte}/test_repo` contains a `.tsx` or `.jsx` file today, so nothing in the existing suite can catch a future revert of this branch.

```
$ find test/resources/repos/vue test/resources/repos/angular test/resources/repos/svelte -iname '*.tsx' -o -iname '*.jsx'
(no output)
$ gh pr view 2066 -R oraios/serena --json files --jq '.[].path'
CHANGELOG.md
src/solidlsp/language_servers/angular_language_server.py
src/solidlsp/language_servers/svelte_language_server.py
src/solidlsp/language_servers/vue_language_server.py
```

Each of the three frameworks already has a working `test_repo` fixture and a `language_server` pytest fixture (`test_angular_basic.py`, `test_vue_basic.py`, `test_svelte_basic.py`), so the missing piece is small: drop a minimal `.tsx` (or `.jsx`) file with a multi-line JSX expression into each fixture repo and add one test mirroring `test_tsx_symbol_range_not_truncated_by_jsx`, asserting the symbol's end line is not truncated at the JSX block.

Yu-0312:
The only failure is  on macOS native — a ccls/C++ language-server integration test. This PR only changes Angular/Vue/Svelte languageIds for / and does not touch C++ or ccls. Re-running the failed job to confirm flake.

Yu-0312:
The only failure is `test_get_document_symbols[cpp_ccls]` on macOS native — a ccls/C++ language-server integration test. This PR only changes Angular/Vue/Svelte languageIds for `.tsx`/`.jsx` and does not touch C++ or ccls. Likely environment flake on the runner; unrelated to the languageId fix.

Yu-0312:
Addressed review from @AmirF194:

1. **Unit tests** (`test/solidlsp/test_jsx_language_ids.py`): assert `.tsx`/`.jsx` → `typescriptreact`/`javascriptreact` on **both** primary and companion adapters for Angular, Vue, and Svelte (24 cases). These run without starting a language server, so a revert fails even when LS integration tests are deselected.
2. **Fixtures**: added `jsx_component.tsx` (multi-line JSX + `trailingHelper`) under `test/resources/repos/{angular,vue,svelte}/test_repo`.
3. **Integration tests**: `test_tsx_symbol_range_not_truncated_by_jsx` in `test_angular_basic.py`, `test_vue_basic.py`, and `test_svelte_basic.py`, mirroring `#1436` / `test_typescript_basic.py` — asserts symbol end line is not truncated and `trailingHelper` remains visible.

Local unit tests: 24 passed.

src/solidlsp/language_servers/angular_language_server.py
AmirF194: Ran this against `d1894c1c` in a clean container.

Thank you for adding the test coverage, that is exactly what the earlier review asked for.

The languageId change here reaches a caller in the same file that this diff does not update. `_wrap_notify_send_for_ts_js_mirror` (svelte_language_server.py:475) forwards `textDocument/didChange` to `$/onDidChangeTsOrJsFile` only when `fb.language_id in ("typescript", "javascript")`, so the svelte language server keeps its internal TS snapshot in sync. Before this PR, a `.tsx`/`.jsx` file got languageId `typescript`/`javascript` and passed that check. After this PR it gets `typescriptreact`/`javascriptreact` and is silently excluded, so edits to a `.tsx`/`.jsx` file opened through the Svelte server no longer reach the companion TS snapshot.

```
self_stub = SvelteLanguageServer._get_language_id_for_file
base (main):  tsx -> typescript,      jsx -> javascript
this PR:      tsx -> typescriptreact, jsx -> javascriptreact

_wrap_notify_send_for_ts_js_mirror, called for didChange on each file:
  file:///a.tsx (typescriptreact)  -> mirror NOT sent
  file:///b.ts  (typescript)       -> mirror sent
  file:///c.jsx (javascriptreact)  -> mirror NOT sent
  file:///d.js  (javascript)       -> mirror sent
```

Extending the tuple at line 475 to `("typescript", "javascript", "typescriptreact", "javascriptreact")` would restore the mirror for these files.

Yu-0312:
Thanks — confirmed against `d1894c1c` and reproduced the mirror drop.

Your diagnosis is correct: after this PR `.tsx`/`.jsx` open as `typescriptreact`/`javascriptreact`, so `_wrap_notify_send_for_ts_js_mirror` no longer forwarded `textDocument/didChange` → `$/onDidChangeTsOrJsFile`.

We did **not** only extend the languageId tuple. That whitelist is exactly what desynced from `_get_language_id_for_file` here; a future languageId change would hit the same bug again.

Instead, the mirror now uses the class’s existing TS/JS predicate `_is_ts_file(uri)` (same helper already used for references/definition/rename/document-symbols/diagnostics in this file). Semantics stay: mirror every open TS/JS buffer, skip `.svelte`/other non-TS/JS, skip when there is no open file buffer.

```python
fb = self.open_file_buffers.get(uri)
if fb is None or not _is_ts_file(uri):
    return
```

Also added unit tests that exercise the wrap without starting a language server: `.tsx`/`.jsx` with `typescriptreact`/`javascriptreact` must mirror; `.svelte`/`.vue` and missing buffers must not. Reverting to the old languageId whitelist fails those cases (`33 passed` on the optimized version).

CHANGELOG updated to mention the Svelte mirror follows TS/JS extensions.

Pushed as `fff36758` on this branch.

src/solidlsp/language_servers/angular_language_server.py
AmirF194: Keying off `_is_ts_file` is better than extending the tuple, it closes the whole class of bug instead of just this instance. Verified the fix against `fff36758`: the mirror predicate now matches `_get_language_id_for_file`'s extension logic instead of duplicating its output, so a future languageId change can't desync them again. Thanks for tracking it down.

PR #2070, opened 2026-09-18
fix(tools): expand unmatched-group $!N backreferences to the empty string

davalillo:
## The problem

A `$!N` backreference in a regex-mode replacement template refers to the Nth group of the search expression. When that group **exists but did not participate in the match** — for example, it sits inside an optional construct that was skipped — the expansion emitted the literal template text instead of the empty string.

### How it was observed

While porting an MQL codebase, the agent replaced 15 call sites in a header with a replacement template containing `EA_INPUT$!1(...)`, where the needle was `EA_INPUT(?:\((\w*)\))?`. The optional group never participated on most sites (no parentheses in the source), so every replacement came out containing the literal `EA_INPUT$!1(...)` — and had to be fixed by hand.

### Minimal reproduction

```python
from serena.util.text_utils import ContentReplacer

replacer = ContentReplacer(mode="regex", allow_multiple_occurrences=False)
replacer.replace("EA_INPUT\n", r"EA_INPUT(?:\((\w*)\))?", r"EA_INPUT$!1(...)")
# before: 'EA_INPUT$!1(...)\n'   (the literal '$!1' emitted)
# after:  'EA_INPUT(...)\n'
```

A reference to a group the expression does not define at all raises a raw `IndexError` — which also crashed **literal-mode** replacements whose template contained `$!N`, since literal mode compiles an escaped pattern without any groups.

## The fix

In both backreference expansion sites in `serena/util/text_utils.py`:

- a group that exists but did not participate (`group(n) is None`) expands to the empty string;
- a reference to a group the expression does not define raises a clear `ValueError` naming the problem instead of a raw `IndexError`;
- in literal mode, the replacement template is now used **verbatim** (`$!N` sequences need no escaping). The expansion ran in both modes, so a literal-mode replacement containing `$!N` failed with a backreference error instead of writing the text; the expansion is now gated on the mode, mirroring the pre-existing gate in `MultiFileContentReplacer.find_occurrences`, and the ambiguity validation still applies in both modes.

## Testing

- `test/serena/test_text_utils.py`: new `TestBackreferenceExpansion` (six cases: both expansion paths in `ContentReplacer`, the literal-mode verbatim passthrough, and the two behaviours through `MultiFileContentReplacer`).
- `poe format` and `poe type-check` pass.


davalillo:
Follow-up commit b15b41ac completes the mode semantics of this PR.

While verifying the fixed semantics in daily use, the mirror case of the original bug surfaced: a **literal-mode** replacement whose replacement string contained a `N` sequence crashed with `ValueError: Backreference $\!1 refers to a group that does not exist`. `ContentReplacer` ran the backreference expansion on the template in both modes, and a literal needle (an escaped pattern) defines no groups at all — so simply trying to *write* the text `1` (e.g. documenting the convention itself in a memory or a file) was impossible through the tool path.

This PR already fixes the regex-mode semantics; the follow-up gates the expansion on the mode so literal-mode replacements pass through verbatim, mirroring the pre-existing mode gate in `MultiFileContentReplacer.find_occurrences`. The ambiguity (multi-line over-match) validation is untouched and still applies in both modes.

- `test_nonexistent_group_reference_in_literal_mode_raises_clear_error` asserted the crash; it is inverted to `test_literal_mode_repl_is_verbatim` (literal needle + `1 $!2` in the replacement → written as-is).
- The regex-mode tests (participating group expands, non-participating group → empty string, undefined group → clear `ValueError`) and both `MultiFileContentReplacer` tests are unchanged and still pass.
- CHANGELOG entry tail corrected accordingly.

opcode81:
Thanks!

PR #2071, opened 2026-09-18
fix(tools): native line endings preserve the file's dominant line ending

davalillo:
## The problem

With `line_ending: native` (the default; `newline=None` passed to the underlying `open()`), a rewrite of an existing file translated embedded `\n` to the platform default. On Linux, editing a CRLF source file rewrote its line endings to LF — in full when the replacement spanned the file, in mixed form when it touched only a part, which is arguably worse.

### How it was observed

While porting an MQL codebase (Windows sources, CRLF), every Serena edit of a `.mqh`/`.mq5` file came out with LF-mixed line endings that had to be repaired manually after each edit (`unix2dos` as a post-edit step).

The root cause is in the atomic write helper introduced for #1958 (`serena/util/file_system.py::write_file_atomic`): it forwards `newline=None` straight to `open()`, and `open()` translates `\n` to `os.linesep` — the platform default — regardless of what the file used before.

### Minimal reproduction

```python
from serena.util.file_system import write_file_atomic

path.write_bytes(b"a\r\nb\r\n")
write_file_atomic(str(path), "a\nb\nnew\n", encoding="utf-8", newline=None)
path.read_bytes()
# Linux, before: b'a\nb\nnew\n'        (line endings silently rewritten)
# after:         b'a\r\nb\r\nnew\r\n'  (preserved)
```

A second, related corruption: a caller passing content lines that already end in `\r\n` (e.g. retrieved untranslated from a language server or a plugin) together with `newline="\r\n"` produced `CRCRLF` line breaks on disk.

## The fix

In `write_file_atomic`:

- With `newline=None` and an existing file, the file's **dominant line ending is preserved** (CRLF-dominant → `\r\n`, LF-dominant → `\n`). Files without a dominant ending (new files, content without line breaks) keep the platform default, exactly as before.
- Content is normalized to LF in memory before translation, so a `crlf`-mode write can no longer double already-CRLF content lines to `CRCRLF`.

Additionally, `create_text_file` (`serena/repl/api/fs_api.py`) is routed through the same atomic helper: it wrote with a plain `Path.write_text`, which truncates the file before the new content is complete (#1958 fixed the editing tools but left this path writing non-atomically).

## Testing

- `test/serena/test_code_editor_atomic_writes.py`: new `TestWriteFileAtomicNativeLineEnding` class plus editor-level tests for both preserved endings, the new-file platform default, and the embedded-CRLF normalization (11 tests in the module, all passing).
- `poe format` and `poe type-check` pass.


src/serena/util/file_system.py
```diff
@@ -29,13 +29,23 @@ def write_file_atomic(path: str, content: str, *, encoding: str, newline: str |
     :param path: the path to write to
     :param content: the text content to write
     :param encoding: the encoding to use for the write
-    :param newline: passed through to the underlying ``open()`` call to control newline translation
+    :param newline: passed through to the underlying ``open()`` call to control newline translation.
+        If ``None`` (native line endings) and the file exists, the file's dominant line ending is
+        preserved instead of the platform default: without this, editing a CRLF file on Linux would
+        silently rewrite its entire line endings to LF. Content lines that already contain ``\r\n``
+        are first normalized to ``\n`` in memory, so the translation cannot turn them into ``\r\r\n``.
```
opcode81: This changes the semantics of "use native line endings" to actually *not* use native line endings, which is likely to cause other problems.
Why do you think this is truly the best way to proceed?

davalillo: Fair point — you're right that this redefined what `line_ending: native` means, on the default path, silently. I've pushed a rework that keeps your semantics: `native` is again the platform default, exactly as documented, and the file's dominant ending is no longer detected at write time.

What remains from the PR is the two uncontroversial fixes it had bundled:

- `create_text_file` now writes through the atomic temp-file helper (it was still a plain `Path.write_text`, the last non-atomic write from #1958).
- The `CRCRLF` fix, narrowed so it can't change semantics anywhere else: content lines already ending in `\r\n` are normalized to `\n` only when the target translation would actually double the CR (`newline="\r\n"`, or `newline=None` on a platform whose `os.linesep` is `\r\n`). In `lf` mode and in native mode on POSIX the content now passes through verbatim, byte-exact, exactly as before.

For the underlying case that motivated the PR (CRLF sources edited on Linux), the intended path is now the explicit one: set `line_ending: crlf` in the project config. Happy to add a note about that to the `line_ending` docs if you think it's worth it.

PR #2072, opened 2026-09-18
test: cover the alternation shape that actually triggers catastrophic backtracking (follow-up to #1899)

davalillo:
Follow-up to #1899 by @GWSAnyone — credit for the root-cause analysis and the whole approach (time-bounded matching through the `regex` module). The PR is still waiting on the CLA signature, and these additions are independent of its final shape, so a separate branch seemed more useful than piling onto a branch that cannot be merged yet.

## What this adds

**1. A test for the expression shape observed in practice.** The test pattern in `TestRegexTimeout` is `(?:.*\n){0,40}?.*never_matches_anything`; the follow-up adds `test_alternation_in_quantified_group_is_aborted`, pinning the incident shape on an MQL source file of ~1550 CRLF lines: `(?:.*\n)*?(?:EA_INPUTX|NO_EXISTE)` with neither alternative present, including the CRLF line breaks. Measured on that corpus: with plain `re` (the pre-#1899 state) `re.search` blocked for over two minutes without finishing (killed via SIGALRM; no in-process way to abort), while through #1899's time-bounded matching it aborts with a clean `RegexTimeoutError` in ~0.3 s. Note that not every unanchored shape reaches the exponential state under `regex` - some are optimized into an instant non-match - so the test patterns are chosen deliberately (see the review discussion below for measurements).

**2. Review resolution.** @AmirF194's review comment is correct — `find_implementations` does not accept an agent-supplied regex on either backend, so it never routes through this module. Dropped from the module docstring.

## Verification

`poe format` and `poe type-check` pass; `pytest test/serena/test_text_utils.py` passes with the new test (12 tests in `TestRegexTimeout`).


feiiiiii5 (commented):
I tried to falsify the central worry behind this PR — that a regression test for catastrophic backtracking can pass without ever entering the pathological state — by measuring the engines directly rather than reading the test.

Method: each case in its own subprocess, killed externally after 15 s. An in-process `SIGALRM` would not do, because (as this module's own docstring says) `re` does not run signal handlers while matching, so a hang could not be reported. Python 3.11.15, `regex` 2026.9.10, macOS/arm64. `MULTILINE | DOTALL`, `search()`:

```text
engine: re
  ^func .*\n(?:.*\n){0,40}?.*never_matches_anything   on 800 LF lines   -> killed after 15s
  (?:.*\n){0,40}?.*never_matches_anything             on 800 LF lines   -> killed after 15s
  (?:.*\n)*?(?:EA_INPUTX|NO_EXISTE)                   on 1552 CRLF lines -> killed after 15s
engine: regex
  ^func .*\n(?:.*\n){0,40}?.*never_matches_anything   on 800 LF lines   -> killed after 15s
  (?:.*\n){0,40}?.*never_matches_anything             on 800 LF lines   -> completed in 0.000s, no match
  (?:.*\n)*?(?:EA_INPUTX|NO_EXISTE)                   on 1552 CRLF lines -> killed after 15s
```

Two consequences.

Good news for the test: `TestRegexTimeout.CATASTROPHIC_PATTERN` (the anchored variant) is *not* optimized away by `regex` — it still runs past 15 s with no timeout in place. So the four abort tests here do drive the engine into the state they claim to guard, rather than passing because a match happens to be cheap. The `finditer` / `search_text` / `search_files` / replacement variants all rest on that same pattern.

The description could mislead a future reader though. It says the single-alternative shape "is optimized by the engine into a fast non-match", as if alternation-vs-not were the axis. The measurements say otherwise: under `re` all three shapes explode, and under `regex` only the *unanchored* single-alternative one is optimized to an instant non-match — while the anchored committed pattern still blows up. So the variables are engine and anchoring, not the number of alternatives. Worth rewording, because a maintainer who takes the current sentence literally might conclude the committed pattern is inherently safe and drop it as redundant.

Limits of what I checked, so nobody over-reads this: I did not run serena's test suite (project deps are not installed in my environment), so I did not observe these four tests passing at `76bb84a4b1`; I measured the underlying engine behaviour that the claim rests on, with `search()` only, on synthetic content rather than the MQL file from the report. The `RegexTimeoutError` message naming `(?:.*\n){0,40}` as the example, and using `regex`'s own timeout rather than a watchdog thread, both look right for the reasons given in the module docstring.

Disclosure: an AI agent produced this comment and the measurements above without a human reviewing the text first; the numbers are pasted from the runs, and the method is stated so anyone can repeat or refute it.

davalillo:
Thanks @feiiiiii5 — this is exactly the kind of review that earns its keep: instead of reading the docstring, you measured the engines. The subprocess-plus-external-kill method is the right call too, for the reason you give (`re` does not run signal handlers while matching), and the disclosure note is appreciated.

I replicated the measurements on the project's own pin before touching anything (Linux, Python 3.13, `regex` 2026.2.28, `MULTILINE | DOTALL`, `search()` with `timeout=3` for the `regex` engine, SIGALRM kill at 4 s for `re`):

```text
engine: regex
  ^func .*\n(?:.*\n){0,40}?.*never_matches_anything   on 800 LF lines    -> aborted at 3.0 s
  (?:.*\n){0,40}?.*never_matches_anything             on 800 LF lines    -> completed 0.000 s, no match
  (?:.*\n)*?(?:EA_INPUTX|NO_EXISTE)                   on 1552 CRLF lines -> aborted at 3.0 s
  (?:.*\n)*?(?:TOKEN_THAT_DOES_NOT_OCCUR)             on 1552 CRLF lines -> completed 0.000 s, no match
  ^(?:.*\n)*?(?:EA_INPUTX|NO_EXISTE)                  on 1552 CRLF lines -> aborted at 3.0 s
engine: re
  all four unanchored/anchored shapes above           -> blocked past the 4 s alarm
```

Your two conclusions hold on this pin as well, with one refinement worth adding:

1. **The committed `CATASTROPHIC_PATTERN` is not optimized away** — it aborts under `regex` here too, so the four abort tests rest on real pathological behaviour. Agreed, nothing to change in the code.

2. **On the axis question**: you're right that my docstring overstated "alternation is what triggers it" — and the refinement cuts both ways. On the corpus, the *anchored* variant of the catastrophic pattern still blows up (as your table shows), while the *unanchored* `(?:.*\n){0,40}?.*…` single-alternative shape is optimized. But for the `*?` shape from the incident, the unanchored **single-alternative** form is optimized *while the unanchored alternation is not* (aborts at 3 s above). So neither "anchoring" nor "alternation" alone is the rule — each unanchored shape optimizes differently, and the safe statement is the one you implicitly made: don't reason from shape classes; measure, or pin the exact incident shape in a test. The docstring of `test_alternation_in_quantified_group_is_aborted` and the PR description now say exactly that (commit 387dc2fc).

Honest limits, matching your disclosure: nobody has run the full test suite against the exact `regex` version in your environment (2026.9.10 vs the project pin 2026.2.28), nor against the actual MQL file rather than synthetic content. The test pins the shape and the engine aborts it cleanly on the pin — which is what the suite needs — but I'd treat cross-version optimization differences as unverified until someone runs it there.

PR #2074, opened 2026-09-18
fix(csharp): skip .csproj files excluded by the project's ignore settings

Volksie:
### What

`CSharpLanguageServer._open_solution_and_projects` scans the whole repository root and opens **every** `.csproj` it finds, without consulting the project's ignore settings. This PR applies the existing `is_ignored_path` check at project discovery.

### Why

On repositories that vendor third-party or sample C# projects, this opens projects the language server cannot restore. The cost is paid on every server start, and the restore failures bury the diagnostics of the projects the user actually works on.

Measured on an Unreal Engine source tree: **245 projects opened, 53 of them under `Engine/Source/ThirdParty`**, which Roslyn cannot build. Each restart emitted thousands of NuGet advisory lines and ended in `The "Csc" task could not be initialized`.

The ignore settings were already being honoured everywhere else — `SolidLanguageServer.is_ignored_path` implements exactly this check, and `CSharpLanguageServer` already overrides `is_ignored_dirname` for `bin`/`obj`/`packages`/`.vs`. Project discovery was the one place that skipped them.

### Notes

- `ignore_unsupported_files=False` is required: a `.csproj` is not itself a C# source file, so the default would exclude it on file type rather than by the ignore patterns.
- Behaviour is unchanged for any project that does not set ignore patterns matching a `.csproj`, which I would expect to be almost all of them.
- The count of skipped files is logged at debug level, alongside the existing `Opened project files` line.

### Scope, and two things I deliberately left out

Both seem like separate changes, but say the word and I will fold either in.

1. **`breadth_first_file_scan` itself is not ignore-aware.** It descends into every directory that does not start with `.`, so the *walk* still visits vendored trees even though their projects are now skipped. Pruning the walk with `is_ignored_dirname` would be the larger win on big repositories, but it is a different change with a wider blast radius.
2. **No test.** The existing C# tests are server-backed integration tests, and making this path unit-testable means extracting project discovery into its own method. Happy to do that, or to add an integration test, whichever shape you prefer.

### Checks

`ruff check`, `ruff format --check` and `ty check src/serena src/solidlsp` all pass. `CHANGELOG.md` updated under *Language Servers*.

This is a SolidLSP-only change, so the MIT SPDX header on the touched file is unchanged.


opcode81:
Thanks!

Issue #2077, opened 2026-09-18
File-system freshness poll costs one os.stat per tracked file on every symbolic call


Volksie:
### Summary

`LanguageServerFileChangeNotifier.poll_and_notify` walks the project's source files and stats each one, and it runs on the symbolic tool paths via `LspApi.ls_sync_file_system_changes()`. The cost therefore scales with the size of the repository and is paid **per call**, not per change.

On a repository with ~107,000 tracked files that is a large fixed tax on every symbolic query. I would like to discuss bounding it. Opening an issue rather than a PR, because the sensible fix is a design choice rather than an obvious patch.

### Where the cost is

`poll_and_notify` (`src/serena/ls_manager.py:303`) iterates `self._project.gather_source_files()` and calls `os.stat` on each path. Its own docstring is explicit: *"The dominant cost is the directory walk plus one `os.stat` per tracked file; this is intended to be called before symbolic tool invocations rather than on a timer."*

It reaches the classic MCP tools as well as the REPL, because `LspApiMixin._api()` (`src/serena/tools/symbol_tools.py:28-32`) returns the same `LspApi` that calls `ls_sync_file_system_changes()`.

To be fair to the current code: `gather_source_files` **does** prune ignored directories during the walk (`src/serena/project.py:353-355`), so this is already much better than walking everything. The remaining issue is simply that the surviving set can still be enormous.

### Numbers

Measured on **v1.7.0** against an Unreal Engine 5.8 tree. The polling implementation has since been rewritten, so these describe the problem's scale rather than `main`'s behaviour — that is exactly why I am opening an issue instead of sending the local patch:

- tracked source files: **187,590** before ignore tuning, **107,852** after
- a full walk: **190s** → **59s** after the same tuning
- with a 45s tool timeout, **no symbolic call ever reached the language server**; clangd received zero `didOpen` requests across an entire session

The tuning that produced the second column was ours, in `.serena/project.yml`. The point is that even a well-tuned large repository leaves ~100k files to stat before every call.

### Possible directions

Not sure which you would prefer, and I do not want to guess in a PR:

1. **Debounce.** Skip the poll if one completed within the last N milliseconds. A handful of symbolic calls in quick succession is the common pattern, and they cannot each need a fresh walk.
2. **Watch instead of poll.** `watchdog` is already a dependency elsewhere in the ecosystem; an OS-level watcher turns the per-call cost into a background one. Bigger change, best outcome.
[...]

PR #2078, opened 2026-09-18
perf: let a directory traversal tell is_ignored_path what it already knows (5.8x faster walk)

Volksie:
Addresses the root cause found while investigating #2077. **5.8× faster tree walk, byte-identical output.**

### What

`Project.gather_source_files` walks the tree with `os.walk`, which hands back directories and files *separately*, and then calls `is_ignored_path` on each one. That method re-derived file-ness from the filesystem:

- `os.path.exists` and `os.path.isfile` in `_is_ignored_relative_path`
- `os.path.isdir` in `match_path`

Three syscalls per path, for an answer the caller already had.

`is_ignored_path`, `_is_ignored_relative_path` and `match_path` now take an optional hint, and `gather_source_files` supplies it from the `os.walk` split.

### Result

Measured on a repository with **97,549 tracked source files out of 707,889 total** (Unreal Engine source plus three game projects; Windows 11, Python 3.13, five language servers configured):

| | |
|---|---|
| `gather_source_files()` before | **67.8s** |
| `gather_source_files()` after | **11.6s** |

The returned list is **byte-identical** — sha256 over the sorted relative paths, 10,390,077 bytes, equal before and after.

For context on why it was so expensive: a bare `os.walk` of the whole 708k-file tree takes **10.7s**, so this was never I/O-bound. `cProfile` over 15,000 real `is_ignored_path` calls attributed **3.41s** to `nt._path_exists`, `nt._path_isfile` and `nt._path_isdir` — about **82%** of the per-call cost. Full numbers are in #2077.

### Compatibility

The parameter defaults to `None`, which determines file-ness from the filesystem exactly as before. **No existing caller changes behaviour**, and the one caller that passes it is the one that already knew the answer.

### Test

`test/serena/config/test_ignored_path_is_file_hint.py` asserts *equivalence* rather than specific verdicts: for every path in a fixture tree, across five ignore configurations and both `ignore_non_source_files` settings, the hinted call must agree with the unhinted one. A hint that could disagree with the filesystem would turn a performance change into a correctness one, so that is the property worth pinning.

It also covers the two cases where guessing file-ness from the *name* would go wrong — a directory with a suffix (`assets.bundle`) and an extensionless file (`Makefile`) — and I checked it against a deliberate inversion of the hint, which it catches (2 failures).

### Checks

`ruff check`, `ruff format --check` and `ty check src/serena src/solidlsp` pass. The existing tests touching this code pass: `test/serena/util/test_file_system.py`, `test/serena/config/test_global_ignored_paths.py`, `test/solidlsp/test_is_ignored_path_missing.py`, `test/serena/test_ls_file_sync.py` — 70 passed, 6 skipped.

### Note on scope

This does not by itself make the freshness poll cheap enough to run per symbolic tool call, which is what #2077 is really about — 11.6s is a large improvement on 67.8s but still not free. I have deliberately kept this to the redundant-syscall fix, since it stands on its own and benefits every caller of `gather_source_files`. Happy to follow up with `os.scandir` (which makes the classification free rather than merely cheap) and with the debounce discussed in #2077, as separate PRs if you would like them.


opcode81:
Thanks for your contribution!

PR #2080, opened 2026-09-18
fix(editing): account for the exclusive match end in occurrence line numbers and diff preview

feiiiiii5:
## Checklist

## Summary

`MultiFileContentReplacer`, which backs `replace_in_files`, reads the end of a match as if `match.end()` were inclusive. `search_text` in the same module already handles this correctly — see the comment at `text_utils.py:169-172`, added in #1708 (*"do not mark the following line as matched when a match ends at a line break"*, merged 2026-07-17); the newer occurrence code did not get the same rule.

Two observable effects, both on any match whose pattern consumes a line break (e.g. `replace_in_files("beta\n", "BETA\n")`):

* `ReplacementOccurrence.end_line` — documented at `:508-509` as the 0-based line number of the match end — names the *following* line, so the diff header reads `lines 1-2` for a one-line match.
* `render_occurrence_diff` — documented as *"the full lines spanned by the match"* — runs its window to the next line break after the exclusive end, so the line after the match is rendered as both removed and added:

```
input:      "alpha\nbeta\ngamma\n"    needle: "beta\n"    replacement: "BETA\n"

before:                                   after this PR:
  [f.txt:0@6c007a] lines 1-2                [f.txt:0@6c007a] line 1
    - beta                                     - beta
    - gamma                                    + BETA
    + BETA
    + gamma
```

Since this header and body are what an agent sees before approving a multi-file replacement, an unchanged neighbour line shown as removed-and-re-added is misleading about the effect of the edit. `apply_to_content` is not affected (it uses the character offsets), which the third test below pins as a control.

## Changes

`src/serena/util/text_utils.py`, `MultiFileContentReplacer.find_occurrences`:

* `end_line` is decremented when the match spans more than one line *and* consumed its line break — the same condition and reasoning as `search_text`. `start_line` is unchanged.

`src/serena/util/text_utils.py`, `MultiFileContentReplacer.render_occurrence_diff`:

* trailing lines that are identical on both sides of the block are dropped before formatting. This is deliberately expressed that way rather than by shortening the window: a line that is equal before and after the replacement carries no information about the change, and trimming it cannot hide one. It also keeps the two cases where the window is genuinely needed:
  - a replacement that *adds* lines still shows all of them (`test_render_occurrence_diff_shows_lines_added_by_the_replacement`);
  - a replacement that *removes* a line break still shows the resulting merge, because the sides then differ in line count and nothing is trimmed (`test_render_occurrence_diff_shows_lines_merged_by_a_removed_line_break`).

`test/serena/test_text_utils.py`, four tests in `TestMultiFileContentReplacer`: the corrected line span, the phantom line gone, and the two controls just above.

No other line-number arithmetic of this kind exists in the class family: `ContentReplacer` (the single-file counterpart) computes no line numbers, and `search_text` already applies the rule.

## Testing

`cwd` is the repo root in all commands; `PYTHONPATH=<tree>/src` with `<tree>` being either a pristine `c6fbd1c5` export or this branch, one interpreter (`.venv/bin/python`, `serena-agent 2.0.0.dev0`). Each run first prints the module it imported, so the two trees cannot be confused:

```
base  LOADED /private/tmp/qoder-r6-serena-base/src/serena/util/text_utils.py
fixed LOADED /private/tmp/qoder-r6-serena/src/serena/util/text_utils.py
```

Same test file in both trees, so old and new behaviour are compared rather than assumed:

```
$ python -m pytest test/serena/test_text_utils.py -q          # base source + these tests
4 failed, 103 passed
  FAILED ...::test_match_consuming_a_line_break_ends_on_the_matched_line
  FAILED ...::test_render_occurrence_diff_omits_the_line_after_a_line_break_match
  FAILED ...::test_render_occurrence_diff_shows_lines_added_by_the_replacement
  FAILED ...::test_render_occurrence_diff_shows_lines_merged_by_a_removed_line_break

$ python -m pytest test/serena/test_text_utils.py -q          # this branch
107 passed
```

Each hunk is isolated by mutation, so no test is vacuous: reverting only the `end_line` correction gives `3 failed, 1 passed`; reverting only the trailing-line trim gives `2 failed, 2 passed`; restoring gives `4 passed`.

Wider regression check over the three files that exercise this API
(`test/serena/test_text_utils.py`, `test/serena/test_edit_api.py`, `test/serena/test_serena_agent.py`), identical command and test file on both trees:

```
base:  18 failed, 166 passed, 7 skipped
fixed: 14 failed, 170 passed, 7 skipped
newly passing: exactly the four tests added here
failing only at fixed: none
failing in both: 14, all of them language-server dependent
   (test_serena_agent.py::...[go_*], [csharp_*], [kotlin_*], [powershell_*]; the corresponding
    language servers are not installed in this environment)
```

Project checks:

```
$ python -m ruff check src/serena/util/text_utils.py test/serena/test_text_utils.py
All checks passed!                                    (ruff 0.12.5, the pinned version)
$ python -m ruff format --check <same two files>
2 files already formatted
$ python -m ty check src/serena src/solidlsp
Found 57 diagnostics                                  (ty 0.0.24)
```

The `ty` count is **57 on `c6fbd1c5` as well**, with the only line mentioning `text_utils.py` being `:14:6` in both trees, i.e. an existing unresolved import in this environment rather than something introduced here; CI's `poe lint` runs in a fully synced environment. Not run: the language-server-marked suites (no servers installed), and the REPL — this change is reachable through `replace_in_files`, whose tool-level tests are in the comparison above.


src/serena/util/text_utils.py
```diff
@@ -633,6 +641,18 @@ def render_occurrence_diff(
             line_end = len(content)
         old_block = content[line_start:line_end]
         new_block = content[line_start : occ.start] + occ.replacement + content[occ.end : line_end]
+
+        # the window runs to the next line break after the exclusive match end, so when the match
+        # itself consumed a line break the following line is included on both sides; drop such
+        # trailing lines, as a line the replacement leaves unchanged is not part of the change
+        old_lines = old_block.split("\n")
+        new_lines = new_block.split("\n")
+        while len(old_lines) > 1 and len(new_lines) > 1 and old_lines[-1] == new_lines[-1]:
```
AmirF194, line 675: I ran this against `cd63b31e` in a clean container.

`render_occurrence_diff`'s new trailing-line trim (text_utils.py:645-654) does not cover the case where the match consumes the file's own final newline. `old_block` there is a literal slice of the original content, so at true EOF it still ends in `\n` even though there is no following line; splitting it on `\n` produces a trailing empty string that has nothing to compare against when the replacement itself has no trailing newline, so the pop loop never removes it. The rendered diff then reports a line being deleted that never existed in the file.

```
content = "alpha\nbeta\n"
occ = replacer.find_occurrences([("f.txt", content)], "beta\n", "BETA")[0]
render_occurrence_diff(occ, content)
->
  [f.txt:0@6c007a] line 1
    - beta
    - 
    + BETA
```

The file has one line ("beta"), but the diff shows two lines removed.

A fix could strip a single trailing "\n" from `old_block` before splitting when `line_end == len(content)`, matching how a POSIX final newline is normally treated as a file property rather than a line of its own, rather than relying only on the pairwise trailing-line comparison.

feiiiiii5: Reproduced at `cd63b31e`, and the mechanism is as you described: `line_end` falls back to `len(content)`, so at true EOF the window ends on the file's own terminator, splitting materialises the empty remainder behind it as a line, and the pairwise pop loop has nothing on the other side to pair it against.

Fixed in 20a47004. Two deltas from the snippet in your comment:

* the terminator is stripped from `new_block` as well as `old_block`. With `old_block` only, the phantom moves to the added side for a replacement that itself ends in a line break at EOF — `content="alpha\nbeta\n"`, `"beta\n" -> "BETA\n"` renders `- beta / + BETA / + ` (empty remainder on the new side, no counterpart to pop). Measured by applying your snippet verbatim on top of `cd63b31e`: that one case goes red while the other five stay green. The shape you posted renders correctly either way only because the two phantoms cancel, so the guard has to be symmetric to be a fix rather than a reshuffle.
* the guard itself is `line_end == len(content)` exactly as you wrote it, and that is what keeps a genuine blank line before the terminator visible: `content="alpha\nbeta\n\n"` with `"beta\n\n" -> "BETA"` still reports `- beta / - / + BETA`, because there the window ends on the excluded break rather than at the end of the content.

`test/serena/test_text_utils.py` gains six cases: the single-line EOF shape, a replacement ending in a line break at EOF, a multi-line match at EOF, a CRLF file, the blank final line that must stay shown, and a whole-file match. Against pristine `cd63b31e` five of the six are red, yours included:

```
assert '[f.txt:0@6c007a] line 1' in '  [f.txt:0@6c007a] lines 1-2\n    - beta\n    - \n    + BETA'
```

With the fix that file is 113 passed, and `ruff check` / `ruff format --check` are clean on both files (`ty` reports one `unresolved-import` for `sensai.util.string` at `text_utils.py:14`, identical before and after — an environment artifact, not attributable to the diff).

One correction to the framing, because it changes what the commit is: the phantom is not produced by this PR. At the merge base `c6fbd1c5` the same call renders `lines 1-2 / - beta / - / + BETA`, so the empty line predates the change and this PR's trailing-line trim merely never reached it; the `lines 1-2` header is the part this PR is about. Scope is display-only either way — `render_occurrence_diff` has one production caller (`src/serena/util/text_utils.py:867`, inside `render_listing`, reached from `src/serena/repl/api/edit_api.py:68`) and nothing parses the rendered lines back, so `apply_to_content` is unaffected.

Not checked: the LSP-gated `test_replace_in_files_*` cases and the rest of the matrix, which need the language servers; I ran the pure-python tests only.

<sub>Written with an AI assistant under my direction; every run quoted above was re-executed here before posting.</sub>

AmirF194: Confirmed, the symmetric strip in 20a47004 covers the case I ran and the one you added on top of it. Thanks for the fix.

PR #2089, opened 2026-09-21
Bridge external language server adapters through the registry

riverolfe:
## What

The plugin mechanism (`ExternalLanguageServerId` + `LanguageServerRegistry` via `solidlsp.language_server_registration` entry points) was added but **three places that consume language IDs still only iterate the built-in `LanguageServerId` enum**, making externally-registered adapters unreachable from `project.yml` and the CLI.

This change bridges all three to also consult `LanguageServerRegistry.get_instance()`.

## Changes

- `_determine_project_language_servers` — auto-detection now scans externally-registered LSes for matching projects; priorities for externals come from `ls.get_priority()` (no user-override mechanism exists for external LSes yet).
- `serena project create --ls` — accepts registry keys as a fallback when the lookup against the `LanguageServerId` enum fails.
- No behavior change for built-in language servers. Existing tests in `test/serena/config/test_serena_config.py` continue to pass (they exercise the enum path only).

## Why

This unblocks first-class support for language servers like `vhdl_ls` (VHDL-LS/rust_hdl) that ship as separate companion packages rather than being vendored into the main repo. Without this change, the plugin mechanism exists but is unreachable from project config.

## Backwards compatibility

- All changes are additive fallbacks.
- Built-in enum paths are unchanged.
- Unknown IDs produce an error message that now also lists externally-registered IDs as valid (when the registry has any).

## Testing

- Manual: registered a `vhdl_ls` companion package via a Python entry point, observed `project.yml` previously rejecting `language_servers: [vhdl]`; after this patch, the entry is accepted and `find_symbol` returns VHDL symbols via the LSP.
- Existing test suite in `test/serena/config/test_serena_config.py` covers the enum path; new test cases for the external-adapter path would require a test-only companion package, which I'll add if reviewers prefer.

Diff: 2 files changed, 25 insertions(+), 6 deletions(-).


riverolfe:
Pushed a follow-up commit (`c761581e`) addressing the `ty` type-check failures the CI flagged:

- `compute_language_server_support_composition`: accepts/returns `LanguageServerIdLike` instead of `LanguageServerId` (since externally-registered adapters are also `LanguageServerIdLike`).
- `ProjectConfig.autogenerate`: same widening for the `languages` parameter.
- Two string-formatting sites in `_determine_project_language_servers` use `.get_key()` instead of `.value` so they work for either type.

Verified locally with `uv run ty check src/serena src/solidlsp` — the two errors from CI are resolved. Remaining diagnostics in the codebase are pre-existing and unrelated to this PR.

src/serena/config/serena_config.py
```diff
@@ -369,11 +369,24 @@ def _determine_project_language_servers(
...
         for language in LanguageServerId:
             priority = serena_config.get_ls_priority(language)
             if priority > 0:
                 ls_priorities[language] = priority
+        # externally-registered language servers (via solidlsp.language_server_registration entry points)
+        # — always include so auto-detection picks them up for matching projects; priority is fixed
+        # (no per-user override mechanism exists for external LSes).
+        registry = LanguageServerRegistry.get_instance()
+        builtin_keys = {l.value for l in LanguageServerId}
+        for ls_key in registry.get_keys():
```
opcode81: There is no reason for the separation; we can iterate over all registered language server ids (which will include the built-in ones) by adding a method `iter_registered_ls_ids() -> Iterator[LanguageServerIdLike]` to the registry.
Similarly, there is no reason why there shouldn't be per-user overrides for external language servers; we can change the interface of `get_ls_priority` to accept `LanguageServerIdLike`.

src/serena/cli.py
```diff
@@ -720,14 +720,20 @@ def _create_project(project_path: str, name: str | None, language: tuple[str, ..
...
+                    languages.append(LanguageServerId(ls_key))
                 except ValueError:
-                    all_langs = [l.value for l in LanguageServerId]
-                    raise ValueError(f"Unknown language '{lang}'. Supported: {all_langs}")
+                    # fall back to the registry for externally-registered adapters
+                    if ls_key in registry.get_keys():
+                        languages.append(registry.resolve(ls_key))
+                    else:
+                        all_langs = [l.value for l in LanguageServerId]
+                        raise ValueError(f"Unknown language '{lang}'. Supported: {all_langs}")
```
opcode81: Here, too, we don't need to handle this in a fallback. We can use the registry only.

opcode81 (changes requested):

opcode81:
Also, please accept the CLA. Otherwise we cannot proceed with this PR.

riverolfe:
Pushed a revision (`e794f116`) addressing your review feedback:

**`LanguageServerRegistry.iter_registered_ls_ids()`** added — yields all registered LSes (built-in enum members + externally-registered adapters) in a single iterator, alphabetical order matching `get_keys()`.

**`ProjectConfig._determine_project_language_servers`** now uses the new iterator instead of the split "enum + dedup-externals" loop. ~9 lines collapse to 4.

**`SerenaConfig.get_ls_priority`** signature widened to accept `LanguageServerIdLike` (from `LanguageServerId`). Implementation uses `ls_id.get_key()` (defined on both enum members and `ExternalLanguageServerId`) so user-configured priorities via `serena_config.ls_priorities` now apply to external LSes too — matches the behavior you'd expect.

**`serena project create --ls`** drops the try/except fallback. Resolves via the registry directly; unknown keys now report the full registry key list in the error message (so users see "vhdl" alongside built-ins like "python" / "rust").

Net diff vs the previous attempt: `+21/-27` (simpler than the `+33/-14` before).

`uv run ty check` is clean for the modified lines; pre-existing diagnostics elsewhere are unchanged.

riverolfe:
> Also, please accept the CLA. Otherwise we cannot proceed with this PR.

Hi, I signed it the first time. It's in pending status "Waiting for status to be reported".
I can't re-sign it (almost empty page), and the recheck link just reloads this page.

3 jobs failed due to infrastructure issues, I'll try to rerun the full pipeline.

opcode81:
> Hi, I signed it the first time. It's in pending status "Waiting for status to be reported".

I believe the issue is that the commits are not actually authored by you. Please rebase and properly assign the author.

opcode81:
Thanks for your contribution.

PR #2065, opened 2026-09-18
fix: assert-based sandbox checks, project list cache, find_file param, prompt/docs contracts

Yu-0312:
## Summary

Code-review findings that are small bug fixes / documentation improvements per CONTRIBUTING.md. Behaviour changes that need discussion first are filed as issues instead:

- #2063 TaskExecutor timeout starts the next tool while the previous worker is still running
- #2064 SolidLSP sends Python code-point columns as LSP Position.character (UTF-16 mismatch)

### Security (assert stripped under `python -O`)

- `create_text_file` always validates via `Project.validate_relative_path` + explicit `ValueError` (new files no longer rely on `assert` for path containment)
- `QueryProjectTool` read-only / registered-project checks raise `ValueError` instead of `assert`

### Agent-visible contracts

- `find_file`: implement documented `skip_ignored_files` (default `False` = previous hardcoded behaviour)
- `jetbrains_rename` docstring: `rename_in_comments` / `rename_in_text_occurrences` default **False**, matching code
- Claude Code `system_prompt.yml` override: map tasks to real tool names (`rename_symbol`, `safe_delete_symbol`, JetBrains-only names for inline/type hierarchy)

### Config / packaging / docs

- Invalidate `SerenaConfig.project_names` / `project_paths` after project add/remove
- Declare `click==8.3.1` as a direct dependency (all three console scripts import it); refresh `uv.lock`
- `server.json`: version 1.7.0, include `grok`/`codebuddy`, drop template-only `context.template`, say YAML not markdown
- C++ guide: `language_servers` is a list; `compile_commands_dir` lives under `ls_specific_settings.cpp`
- Ansible language note: repair parentheses, use `language_servers`
- OmniSharp spelling; “the effect may be insufficient”; “you can you pass”
- Dashboard fallback warning f-string

## Test plan

- `python -m py_compile` on edited modules
- `server.json` / `pyproject.toml` / `system_prompt.yml` parse cleanly
- `uv lock` resolves with `click` as a direct `serena-agent` dependency
- Manual review of diffs against the issues/PR corpus (no overlap with open #1993/#2048/#1362 beyond intentional coordination notes)

## Checklist

- [x] Well-defined scope (small bug fixes + docs/metadata from one review pass)
- [x] CHANGELOG updated for non-docs changes
- [x] CLA will be accepted via CLA assistant on first PR interaction

PR #2092, opened 2026-09-21
fix(config): derive project lists on access instead of caching them

Yu-0312:
SerenaConfig.project_names / project_paths were cached_property values never invalidated after adding/removing projects mid-session, so user-facing project lists and error messages stayed stale. Invalidate the caches on add/remove.\n\nSplit out of #2065.


opcode81:
I think we should just remove the cache annotations instead.

Yu-0312:
Agreed — switched `project_names`/`project_paths` from `cached_property` to plain `property` and removed the `_invalidate_project_list_caches` helper and its call sites, so the lists are derived from `self.projects` on each access.

PR #2094, opened 2026-09-21
fix(dashboard): correct unsupported-mode fallback warning f-string

Yu-0312:
DashboardManager's unsupported-mode fallback warning logged the literal text {fallback_mode.value} because only the first string fragment was an f-string. Fix the f-string.\n\nSplit out of #2065.


CHANGELOG.md
```diff
@@ -30,6 +30,8 @@ Status of the `main` branch. Changes prior to the next official version change w
   - Fix: process-tree cleanup signaled descendant language-server processes without waiting for them,
     which could leave grandchildren as zombies; cleanup now waits for the discovered descendants (#1464)
   - Fix: `read_only` restriction in project definition was not applied to base tool set when in single-project context (#1938)
+  - Fix: DashboardManager's unsupported-mode fallback warning logged the literal text
+    `{fallback_mode.value}` because only the first string fragment was an f-string
```
opcode81: This should be moved to section "Dashboard".

Yu-0312: Moved — the entry now sits under the Dashboard section.

PR #2096, opened 2026-09-21
docs: correct configuration key names and typos

Yu-0312:
Correct configuration key names (languages -> language_servers) and typos across several docs.\n\nSplit out of #2065.


opcode81:
Thanks!

PR #2098, opened 2026-09-22
Bump Expert to v0.1.10 and drop the fixed-bug xfails in the Elixir suites

lukegalea:
Bumps the pinned Expert version for the Elixir language server from `v0.1.0-rc.6` to `v0.1.10` and removes the xfails that documented bugs fixed in the meantime.

## Why

`v0.1.0-rc.6` has two crashes that our Elixir suites were carrying xfails for:

- `textDocument/documentSymbol` for a file that was never `didOpen`-ed (or was evicted from the open-document store) raised a `FunctionClauseError`, surfacing here as a `nil` `document_symbols` result for some files — this is exactly what the `document_symbols` xfail documented.
- A related definition-request crash covered by the three `definition` xfails.

Both are fixed upstream by `v0.1.10` (for context, see the discussion on expert-lsp/expert#903), so this PR:

- Updates `EXPERT_VERSION` and the five per-platform SHA256 pins. New hashes come from the official `expert_checksums.txt` of the `v0.1.10` release; as a sanity check on the method, the old pins were confirmed to match the `rc.6` checksums file exactly, and the `expert_linux_amd64` v0.1.10 binary hash was verified against a fresh download.
- Removes the 4 version-bug xfails (`test_elixir_symbol_retrieval.py` ×1, `test_elixir_ignored_dirs.py` ×3... actually: symbol retrieval ×1 + definition ×3, and one in ignored dirs), each annotated with a comment pointing at the fixed version.
- Updates one stale containing-symbol assertion in `test_request_containing_symbol_function` to Expert's normalized name (`create_user`) — the old expectation was a Next LS-era full signature.
- Documents the new default `expert_version` in `docs/02-usage/050_configuration.md`.

The unversioned `xfail(reason="Flaky test, ...")` on `test_symbol_tree_structure` is intentionally left alone — it tracks an intermittent Expert-internal error, not this version bump.

## Verification

`pytest test/solidlsp/elixir/test_elixir_symbol_retrieval.py test/solidlsp/elixir/test_elixir_ignored_dirs.py` → **17 passed, 1 xpassed** (the xpass is the pre-existing flaky-marked test; `xfail_strict` is off). The run also exercised the new pins end to end: the v0.1.10 release binaries were downloaded and hash-verified during test setup.

Happy to run the remaining Elixir test files (`test_elixir_basic.py`, `test_elixir_integration.py`) before merge if useful.


lukegalea:
Follow-up context for reviewers: we also verified the v0.1.10 `documentSymbol` contract at the wire level — official release binary, `expert --stdio`, matching the `ProcessLaunchInfo` in `elixir_tools.py`. For any URI not in the open-document store — never opened or nonexistent — the server returns a success response with `result: null` (no error response, no crash, no temporary open in this path). That's the behavior the de-xfailed symbol tests now run against. Full probe details in our comment on expert-lsp/expert#903.

src/solidlsp/language_servers/elixir_tools/elixir_tools.py
MischaPanch, line 102: Here we should now detect the problematic old version and delete it - we can check the version string or the sha. This way users who already have used Serena and expert will automatically get the corrected version. Should be appropriately logged

MischaPanch (changes requested):
Thank you for the PR! Could you please sign the CLA and address the minor comment, then this can be merged

